PR #2, opened 2018-05-24
Add initial hello worlds and Github templates

conrad-vanl:
This adds in PR and Issue templates from NewSpring/Apollos and hello world examples straight from the [Apollo Server 2.0 tutorial](https://dev-blog.apollodata.com/apollo-server-2-0-30c9bbb4ab5e) and CRNA.

For CRNA, we are still having a discussion around to expo or not to expo, but creating the initial boilerplate with CRNA doesn't limit that decision and starts us off with best practices for everything else.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
+Fixes [issue-number]
+
+## Creator
+
+- [ ] Build relevant tests if any apply
+- [ ] Ensure there are no new warnings
+- [ ] Upload an animated GIF showcasing the solution (if it applies)
+- [ ] Set a relevant reviewer
+
```
burkeshartsis: Should we add a step for "build relevant stories if any apply?"

Or at the minimum, "ensure there are no new storybook errors." I found previously, that we would build something and then go to find storybook broken a lot.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
...
+## Reviewer
+
+- [ ] Run `test` and make sure all tests pass
+- [ ] Review function and form on iOS
+- [ ] Review function and form on Android
+- [ ] Review function and form on Web
```
burkeshartsis: Web? Pretty sure we can remove this 🙌

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
...
+- [ ] Review new test logic
+
```
burkeshartsis: Should we add a step to "review new stories if any apply."

api/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "apollos-church-api",
```
burkeshartsis: should we still call this `apollos-data`?

mobile/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "apollos-church-app",
```
burkeshartsis: wouldn't this be named `apollos-prototype`?

api/package.json
conrad-vanl: I'm trying to create clearer names. Initially, this `api` project is the `api` for the "Apollos Church", not a `data` library for `apollos`

mobile/package.json
conrad-vanl: As above, just trying to come up with consistent/clear names. `apollos-church-api` and `apollos-church-app`

I would be fine with `apollos-prototype-api` and `apollos-prototype-app` as well though

conrad-vanl:
> Personally, I feel like we have established some vocabulary around this project already.

You're referring to the vocabulary in NewSpring Apollos?

> If we think /api will eventually be an apollos-data repo I would suggest we call this folder data. As for the mobile folder. 

`/api` is _just_ the graphql backend for the Apollos Church app. We don't know _yet_ what will become an `apollos-data` repo, and making decisions around that potential eventuality is tricky. That isn't to say I don't mind the name `data` 😄 

> I'm wondering if we couldn't just move it's contents to root. What benefit do we have form moving it into a subfolder?

To keep the `package.json` files in separate directories and to prevent `api` from being even evaluated by the react-native packager. They're separate runtimes so I think it makes sense to keep them isolated.

If we have any code that's shared between the two runtimes (used in both `/api` and `/mobile`), we can put that at the root level of the project.

PR #23, opened 2018-05-30
Fix to travis linter builds

IsaacHardy:
Fixes linter builds where scripts cannot grab node_modules (cause they were never downloaded)

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


conrad-vanl:
@IsaacHardy do you for-see continuing down this path that generates a bunch of different shell commands inside of the script folder? If so, should we make the naming of these files more detailed?

IsaacHardy:
I don't for-see the need for more bash files for CI, but `install-yarn-dependecies.sh` is much better than just `yarn.sh`. Will make that change.

PR #26, opened 2018-06-01
Content Schema

### LINKED ISSUE #6: Schema Design: Content
## TODO

- [ ] This issue needs to have acceptance criteria flushed out

## Acceptance Criteria

- [ ] Can query for a single content item
- [ ] Can query for a list of content items by content channel
- [ ] Designed spec for initial content data points (such as title, content, etc)

conrad-vanl:
This is _finally_ ready. Spent longer then I was expecting to get test coverage up. But in the end, I think I came up with a pretty simple, clever, and easy-to-write way to test resolvers (mocking the Rock api and running real queries against our real schema).

We had several discussions on the merits of exposing the different types of content (Sermons, Articles, Devos, etc) as different graphql `types`, or exposing a single `ContentItem` type to handle all. In the end - I decided to write a generic `ContentItem` `interface` that we can use to create specific `ContentItem` types. For now, I started us off with a `UniversalContentItem` type that is the default and we can use to start building the app against. In the future, if we need additional content types - like a special `Sermon` type, it will be easy to add, and the resulting API would still be backwards compatible. I think that's the best win-win I could come up with.

Here are some example queries you can try:

```graphql
# Get the user/home feed (WC?)
query {
  userFeed {
    edges {
      node {
        id
        __typename
        title
        coverImage {
          sources { uri }
        }
      }
    }
  }
}
```

```graphql
# Get a list of the content channels and their child content channels
query {
  contentChannels {
    id
    name
    description
    childContentChannels {
      id
      name
    }
  }
}
```

```graphql
# Get content items by channel
# (should return a list of Series)
query {
  node(id: "7693f78b08331486861935e0a5855860efa97db6e7019519549ee7d717adc5ba") {
    ...on ContentChannel {
      id
      name
      childContentItemsConnection {
        edges {
          node {
            id
            title
            images {
              name
              sources {
                uri
              }
            }
          }
        }
      }
    }
  }
}
```

```graphql
# Get child content items of a content item
# This example returns the sermons in a series
query {
  node(id: "61a8c2fafd3ef77cca6226aec3df9059ad2929dbbf2e0494e63e09e46af012dc") {
    ...on UniversalContentItem {
      id
      title
      childContentItemsConnection {
        edges {
          node {
            title
          }
        }
      }
    }
  }
}
```

```graphql
# Example pagination
# Uses relay-style connection/edges/node model.
# The `first` argument specifies page size (get the first N items)
# The `after` argument you supply a cursor and the api will fetch items after that cursor
query {
  node(id: "7693f78b08331486861935e0a5855860efa97db6e7019519549ee7d717adc5ba") {
    ...on ContentChannel {
      id
      name
      childContentItemsConnection(first: 2, after: "4affc1122ad80d4edcf6c5bc9d88ae99") {
        edges {
          node {
            id
            title
          }
          cursor
        }
      }
    }
  }
}

# Paginating the user feed
query {
  userFeed(after: "7c613b3d35342b1bb0c53c1089ac11c6") {
    edges {
      cursor
      node {
        id
        __typename
        title
        coverImage {
          sources { uri }
        }
      }
    }
  }
}
```

Fixes #6 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [x] Review CI build results
- [x] Review new test logic


api/src/utils/cursor.js
```diff
@@ -0,0 +1,25 @@
+import Crypto from 'crypto';
+
+const secret = process.env.SECRET || 'SEfjsvoSDFnvblaE';
```
IsaacHardy: Is this displayed secret safe to display? I am not sure where it comes from.

api/src/utils/sanitize-html.js
```diff
@@ -0,0 +1,36 @@
+import sanitizeHtml from 'sanitize-html';
```
IsaacHardy: What is the purpose to sanitize the html? Just curious.

api/src/data/content-items/index.js
```diff
@@ -0,0 +1,196 @@
...
+    htmlContent: String
+    childContentItemsConnection(
+      first: Int
+      after: String
+    ): ContentItemsConnection
+    parentChannel: ContentChannel
+    # TODO theme: Theme
+  }
+
+  type UniversalContentItem implements ContentItem & Node {
```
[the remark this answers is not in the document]
conrad-vanl: Yeah 😇 

I didn't realize it made it into GraphQL. Last time I looked it wasn't possible 🙌

api/src/utils/cursor.js
conrad-vanl: Yeah, it's fine. Heighliner did. It's just used to encrypt the cursors.

api/src/utils/sanitize-html.js
conrad-vanl: Two reasons:

1. The HTML comes from a WYSIWYG editor. They're notorious for producing junky HTML. React-Native is brittle when rendering HTML,
2. Allows us to specify a set of supported tags, and strip out the rest.

api/src/connectors/rest.js
```diff
@@ -63,6 +63,7 @@ export default class RestConnector {
 
     return Promise.all(
       urls.map(async (url) => {
+        console.log('Fetching', url);
```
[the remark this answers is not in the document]
conrad-vanl: yeah, actually. Logging requests so we can see the traffic in the server log.

PR #30, opened 2018-06-01
Add React Navigation

richarddubay:
This is just an initial two-page navigation so that we have something.

Fixes [issue-number]

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


richarddubay:
@conrad-vanl This is just a simple two-page navigation setup that I know will need to be changed. FWIW, I just went through the entire "Fundamentals" section of their docs to get familiar with what we would be doing. What would you like to see changed or added here? Tab navigation might be one thing. Actually using real files and not just a details page I created for testing? Other things...

conrad-vanl:
@richarddubay great start. I think you could probably start skeleton'ing the app: 

We will probably start off with these tabs, following the NewSpring app closely (except for the connect screen):

1. Home Screen
2. Sections Screen
3. "Next steps" / Connect Screen
4. Search Screen
5. Profile Screen

Once you have that, I think you could "demo" how we could both do stack navigation with the tab bar still visible, and stack navigation without the tab bar visible. Ex: tapping on a card on the home screen probably pushes a route that goes on top of the tab bar, but tapping on a section in the sections screen pushes a route that keeps the tab bar visible.

You can go as far as making dummy components that are named the different screens we need: `HomeFeed`, `SectionsMenu`, `Connect`, `Search` (or `Discover`), `Profile`, etc. Right now they don't need to render much...but gives us a skeleton to start building the app inside of.

If you can - break this task down to a few <=16 point (2 day) issues, add them to github and assign estimates against them. And it'd be great if you can have  _something_ ready to be merged before end of sprint (next tuesday). It it's just a simple `Tabs` component at that point - that's fine!

Thanks!!

richarddubay:
@conrad-vanl A couple things:
1. I haven't added tasks yet ... but I will do that before EOD.
2. I'm not a huge fan of how I had to do the styling per stack navigator, but it didn't care when I had those settings on the tab navigator. Is there a good way to not have to specify that everywhere? Maybe one stack navigator that holds all the other stack navigators.
3. Is it realistic to think that the routes that need to be applied "on top of" the tab navigation would all live in the root stack navigator? I guess so ... but just thought I'd ask.
4. I'm getting a bunch of `Warning: isMounted(...) is deprecated...` warnings. I'm not specifically calling out `isMounted` anywhere. Is that React Navigation, do you think?

Lastly ... any feedback on what I have so far?

mobile/src/article.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
```
conrad-vanl: I believe with there also being an `articles/index.js` file this file will never be imported?

mobile/src/tabs/sections/index.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import { Button, StyleSheet, Text, View } from 'react-native';
+import { createStackNavigator } from 'react-navigation';
+import PropTypes from 'prop-types';
+import Articles from '../../../src/articles';
+import Devotionals from '../../../src/devotionals';
+import News from '../../../src/news';
```
conrad-vanl: I believe you can do absolute imports (@burkeshartsis ?)

mobile/App.js
```diff
@@ -10,16 +22,70 @@ const styles = StyleSheet.create({
...
+    Article,
+  },
+  {
+    initialRouteName: 'Tab',
+    headerMode: 'none',
+  }
+);
+
+/* eslint-disable */
+export default class App extends React.Component {
```
conrad-vanl: This file feels like there's too many different "things" in it. I think you could likely create the `TabStack` inside of `tabs/index.js`. Also, `HomeScreen` and `HomeStack` feel like they might belong in their own sub-folder as well.

mobile/src/article.js
conrad-vanl: ah, nvm. this is singular, other is plural. On `Newspring/Apollos` our content folders usually had 2 or 3 components inside of it, like inside of `articles` there was: `articles/Single` and `articles/List`. I'm not sure if that was eaiser to read/conceptualize then having a root-level `articles` and `article`. Thoughts?

mobile/App.js
richarddubay: I agree with this. I think I was getting an error at some point when I moved Home to its own thing, but it definitely needed to move to the tabs section.

mobile/src/article.js
richarddubay: Yep. This definitely makes sense. I did the singular as a test for the home screen before I made the plural for the sections test. I went ahead and made a "single" and moved it inside the `articles` folder.

richarddubay:
@conrad-vanl I've got everything updated that you suggested. Would love your feedback on the `tabs/index.js` now. Importing and exporting the same things feels kind of weird. Can I make this better?

mobile/src/tabs/index.js
```diff
@@ -0,0 +1,20 @@
...
+import { HomeStack } from './home';
+import { ProfileStack } from './profile';
+import { SearchStack } from './search';
+import { SectionsStack } from './sections';
+
+export { default as ConnectStack } from './connect';
+export { default as HomeStack } from './home';
+export { default as ProfileStack } from './profile';
+export { default as SearchStack } from './search';
+export { default as SectionsStack } from './sections';
```
conrad-vanl: I think this looks good, actually. Also, you don't _have_ to export these stacks at all if you think it's weird

mobile/src/tabs/home/index.js
```diff
@@ -0,0 +1,56 @@
...
+import PropTypes from 'prop-types';
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    backgroundColor: '#fff',
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+});
```
conrad-vanl: At one point I believe you made a comment thinking these `StyleSheet.create`'s were weird. While I don't disagree, I think they're fine for now. Likely these root-level `View`s will be replaced with `FlexedView` down the road.

mobile/src/tabs/index.js
conrad-vanl: If you wanted to clean this up any, you could just do

```
export {
  HomeStack,
  ProfileStack,
  SearchStack,
  SectionsStack,
}
```

Since they're already imported. but again, I don't think you need to export them at all.

richarddubay: If you're good with it, I'll leave it. I'm sure this will all change anyway.

mobile/src/tabs/home/index.js
richarddubay: That sounds good. I'm sure we'll move a lot of this stuff around later.

PR #52, opened 2018-06-07
add codecov

conrad-vanl:
Adds ~coveralls~ codecov (didn't realize codecov was also free for o/s) for code test coverage reporting. Also combines the previous build stages `lint` and `test` to a single `test` build stage with commands that run asynchronously - which should make the builds go faster.

Fixes [issue-number]

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


conrad-vanl:
@IsaacHardy let me know what you think about doing this. It adds test coverage reporting to PRs, and integration with Codecov which is a pretty slick service. IT's tied to your github login so all of us should be able to login and see coverage on codecov as well. 

I'm not making the codecov status checks required, but it will allow us to keep up with test coverage on PRs easier.

IsaacHardy:
@conrad-vanl Is the only to to implement Codecov is to create a `yarn.lock` and `package.json` in the root directory? I was under the impression we didn't want to do this.

Also, I'd recommend pulling in the master branch. There's a couple of changes that will probably interfere with this PR.

PR #57, opened 2018-06-08
updated ios provisioning profile

IsaacHardy:
Should allow for AppCenter builds and deploys

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


IsaacHardy:
@conrad-vanl Might need to add more to this PR.

IsaacHardy:
@conrad-vanl This should be good to go. I've made some provisioning changes to this and would like to make sure it doesn't break the dev process on Xcode (even though we should need it all too often with App Center doing most of the work).

Also, I added some default icon images to the build so we could push the app to iTunes Connect.

Issue #19, opened 2018-05-29
Move remaining Apollos UI primitives


burkeshartsis:
## User Story

As an engineer I want to use [Newspring-Apollos UI primitives](https://github.com/NewSpring/Apollos/tree/alpha/src/%40ui) so that I can more rapidly develop visually consistent interfaces in the Apollos Church prototype app.

#### Problem Details _(Optional)_

Initial plan is to move just the core UI primitives and to keep the same folder structure.

PR #59, opened 2018-06-08
Home Feed (#39)

### LINKED ISSUE #39: Create initial home feed of cards
## User Story

As a user, I want to read content important to me on the home screen

#### Acceptance Criteria

* [ ]  Home feed renders list of cards across content channels, like the newspring app

richarddubay:
Fixes #39 
Partially addresses #19 

## Creator

- [ ] Ensure there are no new warnings (in-app AND in storybook)
- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


mobile/src/ui/ChannelLabel/index.js
```diff
@@ -29,29 +29,31 @@ const StyledH7 = styled(({ theme }) => ({
...
-const CategoryLabel = enhance(({ label, icon, withFlex, isLoading, theme }) => (
+const ChannelLabel = enhance(({ label, icon, withFlex, isLoading, theme }) => (
   <Wrapper flexed={withFlex}>
-    <Icon
-      name={icon || 'audio'}
-      size={theme.helpers.rem(1.2)}
-      fill={theme.colors.text.secondary}
-      isLoading={isLoading}
-    />
+    {!icon ? (
```
richarddubay: @burkeshartsis I see that this would display a default icon if someone doesn't pass one, but what if they do? No icon will display. I'm pretty sure that's not what you intended. Should this always display an icon? Did we want to allow for the option that someone might not want to display an icon?

burkeshartsis: Fixed !

mobile/src/ui/FeedView/index.js
```diff
@@ -0,0 +1,136 @@
...
+      }
+    />
+  )
+);
+
+FeedView.defaultProps = {
+  isLoading: false,
+  renderItem: defaultFeedItemRenderer,
+  onEndReachedThreshold: 0.7,
+  keyExtractor: (item) => item && (item.id || item.entryId),
```
conrad-vanl: Any reference to `entryId` can be removed as that was a holtzman thing

mobile/src/tabs/home/index.js
```diff
@@ -1,40 +1,31 @@
...
-      <View style={styles.container}>
-        <Text>This is the home screen</Text>
-        <Button
-          title="Go to Article"
-          onPress={() => this.props.navigation.navigate('ArticleSingle')}
-        />
-      </View>
+      <Query query={GET_USER_FEED}>
+        {({ loading, error, data }) => {
+          if (loading) return null;
```
conrad-vanl: Why return `null` if `loading`? `FeedView` should accept a loading state (and is passed the prop below)

mobile/src/tabs/home/query.js
```diff
@@ -0,0 +1,28 @@
+import gql from 'graphql-tag';
+
+const GET_USER_FEED = gql`
```
conrad-vanl: Nice!

mobile/src/ui/ChannelLabel/index.js
```diff
@@ -0,0 +1,61 @@
...
+  ...(withIcon
+    ? { paddingHorizontal: theme.sizing.baseUnit / 2 }
+    : { paddingRight: theme.sizing.baseUnit / 2 }),
+}))(View);
+
+const StyledH7 = styled(({ theme }) => ({
+  color: theme.colors.text.secondary,
+}))(H7);
+
+// TODO: Get icon name from content channel
```
conrad-vanl: I believe this TODO can get removed

mobile/src/ui/ChannelLabel/index.js
```diff
@@ -0,0 +1,61 @@
...
+const ChannelLabel = enhance(({ label, icon, withFlex, isLoading, theme }) => (
```
conrad-vanl: @burkeshartsis is it worth generalizing the name of this component? All it does is display a provided `icon` next to some text. Before it did some mapping of the label to pick an icon, which warranted calling it a `ChannelLabel`. Not so sure now.

mobile/src/ui/FeedItemCard/index.js
```diff
@@ -0,0 +1,95 @@
...
+          />
+        </LikeButton> */}
+      </CardActions>
+    </Card>
+  )
+);
+
+const sourcePropType = PropTypes.oneOfType([
+  PropTypes.arrayOf(
+    PropTypes.shape({
```
conrad-vanl: This proptype shape seems off. They're sources for images, which I believe the shape is more like:

```
PropTypes.arrayOf(
  PropTypes.shape({
    uri,
  }),
)
```

mobile/src/tabs/home/index.js
richarddubay: So @IsaacHardy and I discovered this morning that if I just pull `data.userFeed`, the initial load returns "undefined" but then the data pulls again and I get data. If, on the other hand, I specify `data.userFeed.edges` (the data that I _actually_ need), then the data never gets re-fetched and I get an error saying that it couldn't find "edges of undefined". By adding the check for `loading` I was able to make sure that the `edges` get loaded and passed correctly. By all means, if there is a better way to handle this, let's do it.

mobile/src/tabs/home/query.js
richarddubay: 👍 Thanks!

mobile/src/ui/ChannelLabel/index.js
richarddubay: Good catch! I shall remove.

mobile/src/ui/ChannelLabel/index.js
richarddubay: I guess we could pass any kind of icon and any kind of label to something like this and it would work. I like this idea.

mobile/src/ui/FeedItemCard/index.js
richarddubay: This is what was copied over from NS Apollos ... but I guess the shapes are different now. Another good catch! I will change!

mobile/src/ui/ChannelLabel/index.js
burkeshartsis: @conrad-vanl, yeah I had talked with @richarddubay about this a little. I agree that it could effectively be even more generalized to something like `HorizontalTextWithIcon` or maybe better `HorizontalIconLabel` or... all of these names are rather verbose. I suppose we could beef up the component to handle a vertical orientation (icon on top label on bottom) and accept any size typography component (sizes are currently fixed) and then could have a name as generic as `Label` ¯\_(ツ)_/¯

PR #60, opened 2018-06-12
7 people schema

### LINKED ISSUE #7: Schema Design: People
## Acceptance Criteria

- [ ] Model Rock's core people schema in graphql. Lets start with the basics - first name, last name, nick name, brith date, email. Associations (like campus or groups) can come later
- [ ] Can get a person by ID using the Node query
- [ ] Can query for users by email

IsaacHardy:
Fixes #7.

This is just the initial setup for the people schema.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic

### DIFF
--- a/api/src/data/content-channels/index.js
+++ b/api/src/data/content-channels/index.js
@@ -23,8 +23,9 @@ export const resolver = {
       context.models.ContentChannel.all(),
   },
   ContentChannel: {
-    id: ({ id }, _, $, { parentType }) => createGlobalId(id, parentType.name),
-    childContentItemsConnection: ({ id }, input, { models }) =>
+    id: ({ id }, args, context, { parentType }) =>
+      createGlobalId(id, parentType.name),
+    childContentItemsConnection: ({ id }, args, { models }) =>
       models.ContentItem.paginate({
         cursor: models.ContentItem.byContentChannelId(id),
         input,
--- a/api/src/data/content-channels/index.js
+++ b/api/src/data/content-channels/index.js
@@ -28,7 +28,7 @@ export const resolver = {
     childContentItemsConnection: ({ id }, args, { models }) =>
       models.ContentItem.paginate({
         cursor: models.ContentItem.byContentChannelId(id),
-        input,
+        args,
       }),
   },
 };
--- a/api/src/data/content-items/index.js
+++ b/api/src/data/content-items/index.js
@@ -92,7 +92,7 @@ export const defaultContentItemResolvers = {
   childContentItemsConnection: async ({ id }, args, { models }) =>
     models.ContentItem.paginate({
       cursor: await models.ContentItem.getCursorByParentContentItemId(id),
-      input,
+      args,
     }),
 
   parentChannel: ({ contentChannelId }, args, { models }) =>
@@ -145,13 +145,13 @@ export const defaultContentItemResolvers = {
   },
 
   coverImage: async (root, args, { models }) => {
-    const defaultImages = defaultContentItemResolvers.images(node);
+    const defaultImages = defaultContentItemResolvers.images(root);
     // return top image by defalt. TODO: probably better logic to default to.
     if (defaultImages.length) return defaultImages[0];
 
     // If no image, check parent for image:
     const parentItems = await (await models.ContentItem.getCursorByChildContentItemId(
-      node.id
+      root.id
     )).get();
 
     if (parentItems.length) {
@@ -171,7 +171,7 @@ export const resolver = {
     userFeed: (root, args, { models }) =>
       models.ContentItem.paginate({
         cursor: models.ContentItem.byUserFeed(),
-        input,
+        args,
       }),
   },
   UniversalContentItem: {
--- a/api/src/data/content-items/index.js
+++ b/api/src/data/content-items/index.js
@@ -86,15 +86,16 @@ const isAudio = ({ key, attributeValues, attributes }) =>
     attributeValues[key].value.startsWith('http')); // looks like an audio url
 
 export const defaultContentItemResolvers = {
-  id: ({ id }, _, $, { parentType }) => createGlobalId(id, parentType.name),
+  id: ({ id }, args, context, { parentType }) =>
+    createGlobalId(id, parentType.name),
   htmlContent: ({ content }) => sanitizeHtml(content),
-  childContentItemsConnection: async ({ id }, input, { models }) =>
+  childContentItemsConnection: async ({ id }, args, { models }) =>
     models.ContentItem.paginate({
       cursor: await models.ContentItem.getCursorByParentContentItemId(id),
       input,
     }),
 
-  parentChannel: ({ contentChannelId }, input, { models }) =>
+  parentChannel: ({ contentChannelId }, args, { models }) =>
     models.ContentChannel.getFromId(contentChannelId),
 
   images: ({ attributeValues, attributes }) => {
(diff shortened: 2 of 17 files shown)


IsaacHardy:
@conrad-vanl I wanted to go ahead and set this PR to review. If you'd like more to be added, please let me know but i thought it a good idea to get this done and merged so you and I can begin work on the Authorization.

api/src/data/content-channels/index.js
```diff
@@ -23,11 +23,12 @@ export const resolver = {
...
   ContentChannel: {
-    id: ({ id }, _, $, { parentType }) => createGlobalId(id, parentType.name),
-    childContentItemsConnection: ({ id }, input, { models }) =>
+    id: ({ id }, args, context, { parentType }) =>
+      createGlobalId(id, parentType.name),
+    childContentItemsConnection: ({ id }, args, { models }) =>
       models.ContentItem.paginate({
         cursor: models.ContentItem.byContentChannelId(id),
-        input,
+        args,
```
conrad-vanl: I don't think this should've been changed from `args` to `input`

api/src/data/content-items/index.js
```diff
@@ -86,15 +86,16 @@ const isAudio = ({ key, attributeValues, attributes }) =>
...
-  id: ({ id }, _, $, { parentType }) => createGlobalId(id, parentType.name),
+  id: ({ id }, args, context, { parentType }) =>
+    createGlobalId(id, parentType.name),
   htmlContent: ({ content }) => sanitizeHtml(content),
-  childContentItemsConnection: async ({ id }, input, { models }) =>
+  childContentItemsConnection: async ({ id }, args, { models }) =>
     models.ContentItem.paginate({
       cursor: await models.ContentItem.getCursorByParentContentItemId(id),
-      input,
+      args,
```
conrad-vanl: I don't think this should've been changed from `args` to `input`

api/src/data/content-items/index.js
```diff
@@ -167,10 +168,10 @@ export const defaultContentItemResolvers = {
 
 export const resolver = {
   Query: {
-    userFeed: (_, input, { models }) =>
+    userFeed: (root, args, { models }) =>
       models.ContentItem.paginate({
         cursor: models.ContentItem.byUserFeed(),
-        input,
+        args,
```
conrad-vanl: I don't think this should've been changed from `args` to `input`

api/src/data/people/schema.js
```diff
@@ -0,0 +1,15 @@
...
+export default gql`
+  type Person implements Node {
+    id: ID!
+    firstName: String!
+    lastName: String!
+    nickName: String
+    birthDate: String
+    birthDay: Int
+    birthMonth: Int
+    birthYear: Int
```
conrad-vanl: exposing both `birthDate` and day/month/year separately is weird to me. Lets either remove birthday from schema altogether until we need it, or just expose a single `birthDate: String`

api/src/data/people/schema.js
```diff
@@ -0,0 +1,15 @@
...
+    email: String
```
conrad-vanl: Access-control for `Person` seems sticky. We are essentially exposing a church's entire person database - first name, last name, birthday, email, over graphql w/ no access control 🤔 that seems dangerous. What were your thoughts here?

api/src/data/people/schema.js
IsaacHardy: So I agree. In NS/Apollos, we used `moment` on the `birthDate` to determine the `age` of a person, however, `age` is a field we can already get from the Rock API. 

I'm not thinking of any use-cases besides that for `birthDate` so I think we can remove that one for now. I do think it is good to have `birthDay`, `birthMonth`, and `birthYear` separate for future use-cases.

api/src/data/content-channels/index.js
IsaacHardy: I used `args` instead because of this [tutorial](https://www.howtographql.com/graphql-js/2-a-simple-query/) and this [article](https://blog.graph.cool/graphql-server-basics-the-schema-ac5e2950214e).

Seems like the naming used for the four parameters are `root`, `args`, `context`, and `info`.

api/src/data/people/schema.js
IsaacHardy: We definitely need some access control on Person data. I'm not 100% sure how we or if we are implementing it within NS/Apollos but my assumption is that authentication will need to be completed before implementing AC.

api/src/data/people/schema.js
conrad-vanl: The problem with adding properties to a graphql schema that we don't know if we need is we can then never remove it - the nature of graphql is a versionless api that must maintains backwards compatibility. So if we don't have a use-case for an attribute, we shouldn't include it.

conrad-vanl: https://graphql.org/learn/best-practices/#versioning

conrad-vanl: Also - I think it would be wise to always express dates the same way in the schema, so we don't have different patterns at different parts in the schema. Do we want to express an Article's posted date as `postedDay` `postedMonth` `postedYear`? or provide a single `postedAtDate` of some sort?

Since we're making this question now, my advice would be to keep it out of this PR altogether and tackle dates as a separate task.

conrad-vanl: Maybe we eventually implement something like this to provide a consistent date format: https://github.com/excitement-engineer/graphql-iso-date

api/src/data/content-channels/index.js
conrad-vanl: Sounds like scope creep to the underlying task, but that's fine, I see your reasoning. But I believe this change breaks the `paginate` method, as it relies on the argument being named `input`:

https://github.com/ApollosProject/apollos-prototype/blob/master/api/src/connectors/rock/Model.js#L56

You'll have to update these functions too, which I don't believe you did.

api/src/data/people/schema.js
IsaacHardy: Wow, I didn't know Graphql was so inclusive with versioning. Thanks for the heads up.

I'll remove dates completely.

api/src/data/content-channels/index.js
IsaacHardy: Ah. Good catch. I thought I caught everything that was needed.

IsaacHardy:
@conrad-vanl Changes have been made and this is ready for review again.

api/src/data/people/resolver.js
```diff
@@ -0,0 +1,15 @@
...
+  Query: {
+    people: (root, { email }, { models }) => models.Person.getFromEmail(email),
+  },
+  Person: {
+    id: ({ id }, args, context, { parentType }) =>
+      createGlobalId(id, parentType.name),
+    firstName: ({ firstName }) => firstName,
+    lastName: ({ lastName }) => lastName,
+    nickName: ({ nickName }) => nickName,
+    email: ({ email }) => email,
```
conrad-vanl: since these attributes (`firstName`, `lastName`, `nickName`, `email`) share the same name in both the Rock API and the graph schema, and their values are the same,  you don't need these lines at all. Apollo-Server is smart enough to return the attributes as-is if the naming is the same. So you can completely remove lines 10-13 here and everything would still work.

conrad-vanl: You only need resolver functions at all if the `object` needs any mapping of keys or transposing of values. Like how we do the `createGlobalId` bit above instead of returning the raw Rock ID.

IsaacHardy: Awesome. FYI, I mimicked NS/Apollos here, but that definitely makes this easier which I encourage!

IsaacHardy: Done.

PR #61, opened 2018-06-12
Added Live now Button and Modal implementation

### LINKED ISSUE #43: A user can watch a static live stream in app
## TODO:

- [ ] Capture a live now mockup from newspring to use as example

--------

## User Story

A user can watch a static ("fake") live stream in the app

* @conrad-vanl created a Church Online Platform account for Apollos. He can give you credentials
* We may not be able to show a real church online platform live stream???                       

* [ ] The app will be live all the time (for now) in the app
* [ ] The app will feature a "Live Now" banner just like NewSpring
* [ ] Tapping on the "Live Now" banner will open a modal webview with the ChOP player
* [ ] This issue is only for the UI side - essentially displaying a ChOP player in the app with static/hard-coded endpoints
* [ ] MVP is to use a webview like NewSpring

cdanwards:
Fixes #43 

Initial implementation of the Live Now Bar and Live Feed Modal. 

TODO:
- [ ] Add LiveNowButton to all tabs
- [ ] Style LiveNowButton

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


conrad-vanl:
@cdanwards I think this is looking good as a great start! Do you want insight on next steps, or what were you planning on doing from here?

cdanwards:
@conrad-vanl some insight on next steps would be great!

mobile/src/live/liveModal.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { StyleSheet, WebView } from 'react-native';
+
+const styles = StyleSheet.create({
+  iframe: {
```
IsaacHardy: Have you thought about using React Natives `Dimensions` module?

`
import { Dimensions } from 'react-native';

let {width, height} = Dimensions.get('window')
`

mobile/App.js
```diff
@@ -2,11 +2,18 @@ import React from 'react';
...
 import ArticleSingle from 'articles/Single';
 import { TabStack } from 'tabs';
+import ModalScreen from 'live/liveModal';
 
 export const RootStack = createStackNavigator(
   {
     Tab: TabStack,
     ArticleSingle,
+    MyModal: {
+      screen: ModalScreen,
```
conrad-vanl: This should be renamed to something that makes it clear it's the live now modal (and rename the import from `ModalScreen`

mobile/src/live/liveModal.js
conrad-vanl: I would not advise using `Dimensions`. Relative units are better (and responsive to screen size changes)

mobile/src/live/index.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { StyleSheet, Text, View, TouchableOpacity } from 'react-native';
+import PropTypes from 'prop-types';
+
+const styles = StyleSheet.create({
```
conrad-vanl: On NewSpring/Apollos we expressed all components as styled components, using styled to do styling, so the container would become something like:

```
import styled from 'ui/styled';
const Container = styled({
  flex: 1,
  justifyContent: 'center',
  paddingHorizontal: 10,
})(View);
```

This is so that we can access theme values inside of styles. Things like colors and common spacing is exposed via theme. You can access theme inside of styled:

```
const Button = styled(({ theme }) => ({
   backgroundColor: theme.colors.primary,
})(TouchableOpacity);
```

Revisiting the container example above, all spacing values should refer back to our `baseUnit` that's defined in theme. That way the visual eb and flow of the UI is referred from a core set of values and things like visual spacing remain consistent. The `baseUnit` is `20` and I think our general rule is that you can use the `baseUnit` in multiples of 0.5, so we can do this:

```
import styled from 'ui/styled';
const Container = styled(({ theme }) => ({
  flex: 1,
  justifyContent: 'center',
  paddingHorizontal: theme.sizing.baseUnit / 2,
}))(View);
```

(@burkeshartsis maybe it's time we get some documentation together on styled and theme shape. Lets sync up on this sometime today)

mobile/src/live/index.js
```diff
@@ -0,0 +1,35 @@
...
+  },
+});
+
+const LiveNowButton = (props) => (
+  <View style={styles.container}>
+    <TouchableOpacity
+      style={styles.button}
+      onPress={() => props.navigation.navigate('MyModal')}
+    >
+      <Text> Live Now! </Text>
```
conrad-vanl: Should probably use `import { UIText } from 'ui/typography'` instead of react-native's default `Text`

mobile/src/live/liveModal.js
```diff
@@ -0,0 +1,22 @@
...
+  },
+  noScroll: {
+    overflow: 'hidden',
+  },
+});
+
+const ModalScreen = () => (
+  <WebView
+    source={{ uri: 'https://apollos.churchonline.org/' }}
+    style={{ styles }}
```
conrad-vanl: Hmmm....I don't think you're passing styles correctly here. You're essentially passing an object with two style definitions (`styles.iframe` and `styles.noScroll`). React-Native won't know what to do with it. I'd consider something like:

```
import styled from 'ui/styled';
const FullScreenWebView = styled({
  width: '100%',
  height: '100%',
  flex: 1,
})(WebView);
```

mobile/src/live/index.js
conrad-vanl: Didn't think about this - but we'll need to pull this change into this branch as well for theme to work: https://github.com/ApollosProject/apollos-prototype/pull/59/commits/790a9d4dcaa2dce53478b898ad3c07942c73ac29

you can just do `get fetch origin` then `git cherry-pick 790a9d4dcaa2dce53478b898ad3c07942c73ac29` from the command line to pull this change in (or you can do it by hand)

mobile/src/live/liveModal.js
cdanwards: @conrad-vanl would you recommend still using an iframe? or is there a better 'container'?

PR #66, opened 2018-06-14
Copied over NewSpring TableView

### LINKED ISSUE #64: Port `<TableView />` from NewSpring/Apollos
Lets go ahead and bring over `TableView` from NewSpring/Apollos, so we can move forward on building the connect menu:

https://github.com/NewSpring/Apollos/tree/alpha/src/%40ui/TableView

#### Acceptance Criteria

* [ ] Can see the TableView stories in storybook
* [ ] Snapshot test coverage for these components are added

cdanwards:
Fixes #64 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic

### DIFF
--- a/mobile/src/ui/TableView/index.js
+++ b/mobile/src/ui/TableView/index.js
@@ -1,6 +1,6 @@
 import { StyleSheet, View } from 'react-native';
 import PropTypes from 'prop-types';
-import { compose, renderComponent, setPropTypes } from 'recompose';
+import { compose, renderComponent } from 'recompose';
 import styled from 'ui/styled';
 import { enhancer as mediaQuery } from 'ui/MediaQuery';
 import Card from 'ui/Card';
@@ -12,9 +12,6 @@ export { default as Divider } from './Divider';
 export { default as FormFields } from './FormFields';
 
 const TableView = compose(
-  setPropTypes({
-    responsive: PropTypes.bool,
-  }),
   styled(
     ({ theme }) => ({
       marginBottom: theme.sizing.baseUnit,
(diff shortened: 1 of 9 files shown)


mobile/src/ui/TableView/index.js
```diff
@@ -0,0 +1,36 @@
...
+import Card from 'ui/Card';
+
+export { default as Cell } from './Cell';
+export { default as CellText } from './CellText';
+export { default as CellIcon } from './CellIcon';
+export { default as Divider } from './Divider';
+export { default as FormFields } from './FormFields';
+
+const TableView = compose(
+  setPropTypes({
```
burkeshartsis: This can be defined above the export using the default react syntax as `TableView.propTypes = { //.. }`

mobile/src/ui/TableView/index.js
```diff
@@ -0,0 +1,36 @@
...
+    responsive: PropTypes.bool,
+  }),
+  styled(
```
cdanwards: @burkeshartsis can this styled be defined above?

burkeshartsis: You would leave `styled` where it is.

cdanwards: Do you mind explaining why that is?

conrad-vanl: It's only because we're trying to get away from using HOCs for everything. For now, in the components we're porting over, we're just trying to get rid of HOCs that can be replaced by vanilla react syntax.

You can always set `propTypes` on a component like:

```
Component.propTypes = {
   someProp: PropTypes.string,
}
```

In this case, we're just choosing to get rid of using the `setPropTypes` HOC and going back to just using the vanilla React syntax for proptypes.

`styled` and `mediaQuery` we don't have other replacements for, so we will still use them as is.

PR #67, opened 2018-06-14
(44) Display Content Feed Item Details

### LINKED ISSUE #44: Tapping on a content item in the feed takes you to a content single
TODO:

- [ ] Flush out issue criteria

----

## User Story

* [ ] It should [SPEC_DETAIL]

richarddubay:
Fixes #44 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic

### DIFF
--- a/mobile/src/tabs/home/index.js
+++ b/mobile/src/tabs/home/index.js
@@ -19,6 +19,8 @@ export class HomeScreen extends React.Component {
     }),
   };
 
+  onPress = () => this.props.navigation.navigate('ArticleSingle');
+
   render() {
     return (
       <FlexedView>
@@ -29,6 +31,7 @@ export class HomeScreen extends React.Component {
               content={get(data, 'userFeed.edges', [])}
               isLoading={loading}
               error={error}
+              onPressItem={this.onPress}
             />
           )}
         </Query>
--- a/mobile/src/ui/FeedView/index.js
+++ b/mobile/src/ui/FeedView/index.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { FlatList } from 'react-native';
+import { FlatList, TouchableWithoutFeedback } from 'react-native';
 import PropTypes from 'prop-types';
 import { pure, compose, branch, withProps, defaultProps } from 'recompose';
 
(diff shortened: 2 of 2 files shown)


IsaacHardy:
@conrad-vanl Has the Codecov tests failed here, because of a lack of written tests?

conrad-vanl:
@isaachardy correct. Saying that overall average test coverage decreased with this PR

mobile/src/content/Single.js
```diff
@@ -0,0 +1,46 @@
...
+
+  static propTypes = {
+    navigation: PropTypes.shape({
+      getParam: PropTypes.func,
+    }),
+  };
+
+  render() {
+    const { navigation } = this.props;
+    const item = navigation.getParam('item', []);
```
conrad-vanl: instead of getting the item details from the navigation parameter, I think we should probably query for it...that way we don't end up with a gigantic query for the feed view querying for every attribute the single view needs. And in the future, when we link to content from other places, we don't have to worry about setting all of the items' info in the navigation action

mobile/src/content/index.js
```diff
@@ -0,0 +1,26 @@
...
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    backgroundColor: '#fff',
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+});
+
+class Content extends React.Component {
```
conrad-vanl: I don't think this file is used...?

mobile/src/tabs/home/query.js
```diff
@@ -5,20 +5,21 @@ const GET_USER_FEED = gql`
...
           __typename
-          title
+          id
           coverImage {
             name
             sources {
               uri
             }
           }
+          htmlContent
```
conrad-vanl: lets remove `htmlContent` from this query once my other commend is addressed

mobile/src/content/Single.js
```diff
@@ -0,0 +1,46 @@
...
+import PaddedView from 'ui/PaddedView';
+import { H3 } from 'ui/typography';
+
+class ContentSingle extends React.Component {
+  static navigationOptions = ({ navigation }) => {
+    const item = navigation.getParam('item', []);
+    const itemTitle = item.node.parentChannel.name;
+    return {
+      title: itemTitle || 'Content Channel Title Here',
+      headerStyle: {
```
conrad-vanl: lets get rid of these custom colors/fontweight and such for now

mobile/src/content/Single.js
richarddubay: I'm not exactly sure what you mean here. So instead of passing `item` as a param of the `onPress` function and getting that info in this file from that param, I should write a GraphQL query specifically for this file that returns the data for just this single content item? Does our API support getting one item like that? Am I missing what you're wanting to do here?

mobile/src/content/index.js
conrad-vanl: > could be used
> down the road
=== remove

mobile/src/content/Single.js
conrad-vanl: Correct, essentially. You'll still pass something to the navigation handler - but we shouldn't rely on being able to pass _all_ the data that's needed into the navigation handler. The minimum you should be required to pass to the navigation handler is just the content's ID. (think of using a web url to link to content - you'll only put the ID of the content or some permalink in the URL, not the entire data object of the content in the URL)

You can query for an individual item using the `node` query, just like on Heighliner:

```
query {
  node(id: "ssome-id") {
    ... on ContentItem {
      id
      htmlContent
      # and so on
    }
  }
}
```

mobile/src/client/cache.js
```diff
@@ -0,0 +1,26 @@
+import {
+  InMemoryCache,
+  IntrospectionFragmentMatcher,
+  defaultDataIdFromObject,
+} from 'apollo-cache-inmemory';
+
+const cache = new InMemoryCache({
+  dataIdFromObject(obj) {
+    // eslint-disable-next-line no-underscore-dangle
+    if (obj.entryId) return `${obj.__typename}:${obj.entryId}`;
```
conrad-vanl: `Apollos-data` won't ever return `entryId` so we can get rid of this line, which means we can get rid of this entire function

mobile/src/client/cache.js
```diff
@@ -0,0 +1,26 @@
...
+  },
+  fragmentMatcher: new IntrospectionFragmentMatcher({
+    introspectionQueryResultData: JSON.parse(
+      '{"__schema":{"types":[{"kind":"INTERFACE","name":"Node","possibleTypes":[{"name":"ContentChannel"},{"name":"UniversalContentItem"},{"name":"Person"}]},{"kind":"INTERFACE","name":"ContentItem","possibleTypes":[{"name":"UniversalContentItem"}]}]}}'
+    ),
+  }),
+  cacheRedirects: {
+    Query: {
+      node: (root, { id }, { getCacheKey }) =>
+        getCacheKey({ __typename: 'Content', id }),
```
conrad-vanl: lets also not include this `cacheRedirect` for now - this won't work

conrad-vanl:
Looks like we still have a code coverage outage on this branch, what's the plan there?

mobile/src/content/Single.js
```diff
@@ -0,0 +1,52 @@
...
+    }),
+  };
+
+  render() {
+    const { navigation } = this.props;
+    const itemId = navigation.getParam('itemId', []);
+    return (
+      <Query query={GET_CONTENT} variables={{ itemId }}>
+        {({ loading, error, data }) => {
+          if (loading) return null;
```
conrad-vanl: We should display placeholder views for loading. @burkeshartsis how do you think we should treat this? I know the typography components can support an `isLoading` prop, maybe we can pass `isLoading={isLoading}` to the `GradientOverlayImage`, `H3`, and `HTMLView` below, and then use lodash's `get` to keep errors from being thrown if `data.node === null` while loading?

mobile/src/content/Single.js
```diff
@@ -0,0 +1,52 @@
...
+import HTMLView from 'ui/HTMLView';
+import PaddedView from 'ui/PaddedView';
+import { H3 } from 'ui/typography';
+import GET_CONTENT from './query';
+
+class ContentSingle extends React.Component {
+  static navigationOptions = ({ navigation }) => {
+    const itemTitle = navigation.getParam(
+      'itemTitle',
+      'Content Channel Title Here'
```
conrad-vanl: Maybe we should have a better default...either just blank or 'Content' or something?

richarddubay: Yeah ... agreed. I'll just make it "Content".

richarddubay:
@conrad-vanl I need to write probably one test, maybe two to cover the `renderItem` and `onPressItem` pieces. Trying to decide if I need to figure out how to actually "press" an item? I don't know how to make that happen right now.

conrad-vanl:
@richarddubay you won't be able to "press" an item with our snapshot tests, unfortunately

richarddubay:
@conrad-vanl I added a test to pass an empty function to `onPressItem` then ... but I don't think it's going to see it as having been tested. Is there something else that I can do?

conrad-vanl:
Good question. Our options:

1. live with the test coverage decrease
2. Use something like enzyme to simulate taps
3. Unit test the function in question...treat the component as a normal Javascript class, instantiate it, and call the function directly. Like:

```
a = new FeedView(someProps);
a.renderItem();
a.onPress(item);
// .... etc
```

PR #68, opened 2018-06-14
User Authentication

### LINKED ISSUE #42: Design GraphQL schema for user authentication
Determine graphql schema for user authentication that handles signing in with Rock credentials, and storing a rock token. Also allows for requests made to the Apollos-church api to include an auth token.

1. user signin mutation ==> returns a user token
2. on subsequent graphql queries, you can include that user token to authenticate that query

----

## User Story

* [ ] It should include schema
* [ ] Allow for an `Authorization` header on graphql requests
* [ ] Create additional issues to hook up resolvers to actually authenticate with rock

conrad-vanl:
Implementation of an `authenticate` mutation that authenticates a user with Rock. Example mutation:

Example:

```gql
mutation {
  authenticate(identity: "myuser", password: "mypassword") {
    token
    # can also query for the user here if you'd like:
    user {
      id
    }
  }
}
```

The token it returns is a JWT token with an expiration date of 60days. On Heighliner we explicitly used `email` in the login mutation. I've simplified the process to take any identity that rock recognizes. Which removes some special-case handling.

I'm also opting to use Rock's cookie-based authentication flow, even on the server. It appears to be the best way to do user-auth with the Rock api. So the token that we get back from this authenticate mutation actually has Rock's cookie encoded in it.

To authenticate against the API in future requests, use the `token` provided in a standard `authorization` header, in graphiql, it looks like this:

image

This also adds a `currentUser` query:

```gql
query {
  currentUser {
    id
    profile {
      id
      email
      firstName
      lastName
    }
  }
}
```

`currentUser` returns an `AuthenticatedUser` type with two attributes: `id` and `profile`. `profile` points to the `Person` object for that account. I'm returning `AuthenticatedUser` instead of `Person` directly in this query so that we can add more functionality to this later (like if we need to support connections to other data sets - like their profile on church online platform - for example.

Fixes #42 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [x] Review CI build results
- [x] Review new test logic

### DIFF
--- a/api/src/data/people/__tests__/__snapshots__/model.tests.js.snap
+++ b/api/src/data/people/__tests__/__snapshots__/model.tests.js.snap
@@ -15,13 +15,13 @@ Array [
 ]
 `;
 
-exports[`Person gets person from email 2`] = `
+exports[`Person gets person from id 1`] = `
 Object {
   "email": "isaac.hardy@newspring.cc",
 }
 `;
 
-exports[`Person gets person from id 1`] = `
+exports[`Person gets person from id 2`] = `
 Array [
   Array [
     "https://apollosrock.newspring.cc/api/People/51",
@@ -36,7 +36,7 @@ Array [
 ]
 `;
 
-exports[`Person gets person from id 2`] = `
+exports[`Person gets person from id 3`] = `
 Object {
   "id": 51,
 }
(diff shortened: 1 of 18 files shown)


api/src/data/people/__tests__/__snapshots__/model.tests.js.snap
```diff
@@ -15,13 +15,13 @@ Array [
 ]
 `;
 
-exports[`Person gets person from email 2`] = `
+exports[`Person gets person from id 1`] = `
```
IsaacHardy: I've noticed that this text is changed for every snapshot. Curious as to how?

IsaacHardy:
@conrad-vanl Is it an issue that some of the string keys on a People object are returning as objects?
screen shot 2018-06-15 at 11 11 48 am

PR #71, opened 2018-06-15
Design tweaks

conrad-vanl:
A few quick design tweaks:
- Clean up tab headings to be more consistent
- Tweak styles on live now button
- add pull-to-refresh support on the home screen
- added some icons and implemented a nicer tab bar
- created a `<BackgroundView/>` component

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


mobile/src/tabs/home/index.js
```diff
@@ -2,23 +2,39 @@ import React from 'react';
...
-    title: 'Apollos Church',
-    // TODO: for the life of me, I can't figure out how to get these colors
-    // to be dynamic from theme or props.
-    headerStyle: {
-      backgroundColor: '#17B582',
-    },
-    headerTintColor: '#ffffff',
+    header: ({ state }) => ({
+      title: 'Apollos Church',
+      // TODO: for the life of me, I can't figure out how to get these colors
```
burkeshartsis: @conrad-vanl I left you a 🎁 here. I recommend cleaning this up before merging.

mobile/src/tabs/home/index.js
```diff
@@ -2,23 +2,39 @@ import React from 'react';
...
+      },
+      headerTintColor: state.params.tintColor,
+    }),
   };
 
+  constructor(props) {
+    super(props);
+
+    // eslint-disable-next-line
+    props.navigation.setParams({
```
burkeshartsis: Setting navigation params here causes `HomeScreen` to render twice. Maybe look into moving this to `getDerivedStateFromProps` to reduce unnecessary renders.

conrad-vanl: `getDerivedStateFromProps` is a React lifecycle method, right? How does it interact with `react-navigation`'s navigation state? You're setting navigation state here, not component state....right???

conrad-vanl: image

Also, not sure if it's working?

conrad-vanl: I'm not following how that plays into navigation state. Navigation state !== component state. `getDerivedStateFromProps` gets component state

conrad-vanl: i think that'd be like saying we could use `getDerivedStateFromProps` to change state in redux

mobile/src/ui/BackgroundView/__snapshots__/BackgroundView.tests.js.snap
```diff
@@ -0,0 +1,18 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`The BackgroundView component should render 1`] = `
+<BVLinearGradient
+  colors={
+    Array [
+      4294967295,
```
burkeshartsis: what values are these? 🙃

mobile/src/ui/ChannelLabel/index.js
```diff
@@ -47,7 +47,7 @@ const ChannelLabel = enhance(({ label, icon, withFlex, isLoading, theme }) => (
 ));
 
 ChannelLabel.propTypes = {
-  label: PropTypes.string.isRequired,
+  label: PropTypes.node.isRequired,
```
burkeshartsis: why would the label accept an object?

mobile/src/ui/theme/types/light.js
```diff
@@ -9,7 +9,7 @@ const light = ({ colors, alpha }) => ({
       link: colors.primary,
     },
     background: {
-      default: colors.lightSecondary,
+      default: colors.lightPrimary,
```
burkeshartsis: are we going to use `paper`. If so this feels like it should be the paper value. Personally I'm not sure we need a `default` and `paper` value.

mobile/src/tabs/home/index.js
```diff
@@ -2,23 +2,50 @@ import React from 'react';
...
+
 import GET_USER_FEED from './query';
+import tabBarIcon from '../tabBarIcon';
 import LiveNowButton from '../../live';
 
-export class HomeScreen extends React.Component {
-  static navigationOptions = {
-    title: 'Home',
-  };
+// TODO: what are our thoughts around using this @-syntax for HOCs?
```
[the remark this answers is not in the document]
burkeshartsis: Can we remove this `TODO` if we are going to keep decorators?

mobile/src/live/index.js
```diff
@@ -1,28 +1,32 @@
...
+  backgroundColor: theme.colors.lightSecondary,
+}))(Card);
 
 const LiveNowButton = (props) => (
-  <Container>
-    <Button onPress={() => props.navigation.navigate('LiveNowModal')}>
-      <Title> Live Now! </Title>
-    </Button>
-  </Container>
+  <Touchable onPress={() => props.navigation.navigate('LiveNowModal')}>
```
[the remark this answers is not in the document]
conrad-vanl: out of scope for this PR

mobile/src/tabs/connect/index.js
```diff
@@ -1,71 +1,61 @@
...
-            <CellText>I would like to give</CellText>
-          </Cell>
-          <Divider />
-          <Cell>
-            <CellIcon name="building" />
-            <CellText>Find Service Times & Locations</CellText>
-          </Cell>
-        </TableView>
-      </ScrollView>
+      <BackgroundView>
```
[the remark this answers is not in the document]
conrad-vanl: out of scope for this PR

mobile/src/tabs/home/index.js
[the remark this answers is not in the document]
conrad-vanl: I'm fine with leaving the `TODO` for now in case others come along to see this code and have thoughts they know it's an open discussion. I still want to research a little more on current state of decorators

mobile/src/tabs/tabBarIcon.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Icon from 'ui/Icon';
+
+const tabBarIcon = (name) => {
```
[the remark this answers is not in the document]
conrad-vanl: this isn't a component per say, it allows you to create an "instance" of an icon component that renders a certain icon: `tabBarIcon('profile')` will return a component that always renders a profile icon. Kind of like using `withProps` or something...but in this case it also sets some other prop defaults specific to tab bars. This comes in handy for our different Tab screens:

```
ConnectStack.navigationOptions = {
  tabBarIcon: tabBarIcon('profile'),
};
```

Without this, you'd have to use an inline function, which I believe you also dislike @burkeshartsis:

```
ConnectStack.navigationOptions = {
  tabBarIcon: ({ tintColor }) => <Icon name="profile" size={24} fill={tintColor} />,
};
```

> If we HAVE to have this can we at least create some tests for it make sure it doesn't needlessly rerender ala pure from recompose maybe?

How would you write a test for that? Snapshots test a single render state, not re-renders.

mobile/src/tabs/index.js
```diff
@@ -1,23 +1,25 @@
...
+export const TabStack = createBottomTabNavigator(
+  {
+    Home: HomeStack,
+    Sections: SectionsStack,
+    Connect: ConnectStack,
+  },
+  {
+    tabBarOptions: {
+      showLabel: false,
+      activeTintColor: '#17B582', // TODO: get from theme
```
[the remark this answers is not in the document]
conrad-vanl: Can we make this change in a later PR? I spent more time then I care to admit trying to get `screenProps` to work. We already had some static colors before this PR...I'm afraid if I tried to come up with a solution for this I'd come up with something that's not much better then this.

mobile/src/ui/BackgroundView/__snapshots__/BackgroundView.tests.js.snap
conrad-vanl: I assume they map to react-native color values under the hood maybe?

mobile/src/ui/ChannelLabel/index.js
conrad-vanl: it doesn't accept an object. It accepts a renderable node. That way you can have text styles inside the label:

```
label={<BodyText bold>Some bold text!</BodyText>}
```

I do it here: https://github.com/ApollosProject/apollos-prototype/pull/71/files/3fa0256486dec3a9597a6e022878a71c1f9607a7#diff-06253f7bd94dfc5cd710806d1fa0ce1dR20

mobile/src/ui/theme/types/light.js
conrad-vanl: naming is probably not right. But in line with what I've seen with material design spec, there are two background colors. A really light off-white, which is the primary or default background, and a pure-white "paper" color. React Native Paper does it too, if you look at their theme definition: https://callstack.github.io/react-native-paper/theming.html

conrad-vanl: image

The gray color behind the cards is `background.default`, and the card and tabbar is `background.paper`

mobile/src/tabs/tabBarIcon.js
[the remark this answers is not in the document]
burkeshartsis: Personally, I don't see the purpose for this function. I'm fine with the arrow function as it doesn't add any unnecessary layers of abstraction. Maybe `renderComponent()` from recompose might be cleaner but I don't necessary want to add recompose where we don't have to.

mobile/src/ui/ChannelLabel/index.js
burkeshartsis: Would this better represent what you are trying to do? My fear is that an `array` is a valid `node` but would not render properly and an `int` would just be cast as a `string`.

```
label: PropTypes.oneOfType([
    Proptypes.string,
    Proptypes.element
]),
```

mobile/src/tabs/tabBarIcon.js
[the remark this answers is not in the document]
conrad-vanl: even though we are basically duplicating props for every single tab, and an inline function would result in an impure component? I basically took this `tabBarIcon` component from react-navigation's example app that they have, let me see if I can track it down....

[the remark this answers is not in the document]
conrad-vanl: https://github.com/react-navigation/react-navigation-material-bottom-tabs/blob/master/example/src/shared/tabBarIcon.js

mobile/src/ui/ChannelLabel/index.js
conrad-vanl: Done.

PR #92, opened 2018-06-18
fix now deployment scaling

IsaacHardy:
Fixes scaling issues for now deployment.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ jobs:
     - stage: Deploy Api
       script: skip
       before_deploy: npm install -g now
+      after_deploy: now scale https://apollos-church-api.now.sh 0 --token $NOW_TOKEN && now scale https://apollos-church-api.now.sh 1 --token $NOW_TOKEN
       deploy:
         - provider: script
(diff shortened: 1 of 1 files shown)


conrad-vanl:
I think it's better then what we have, unless you think we should take scaling out entirely from our automated process and do it manually?

IsaacHardy:
@conrad-vanl @richarddubay Found a command in the documentation that allows us to remove urls that are not aliased to our main domain. We've tested this and this seems to be the best solution. It builds, alias, then remove unused instances. Also, it makes every new build scaled to 1.

conrad-vanl:
@IsaacHardy nice!! are you going to implement that change to this PR?

.travis.yml
```diff
@@ -25,6 +25,6 @@ jobs:
       before_deploy: npm install -g now
       deploy:
         - provider: script
-          script: cd api && now --public --token $NOW_TOKEN && now alias --token $NOW_TOKEN
```
conrad-vanl: should the `now rm` be done _before_ the deploy? it seems like doing it after like this is just going to remove the instance you just deployed?

PR #107, opened 2018-06-21
#101 Files structure / skeleton clean up

### LINKED ISSUE #101: File structure / skeleton next set of screens to be built
## User Story

As a developer I want a consistent code base and component API so that I can write faster and more efficiently with less errors.

- [x] Load graphql queries with a babel loader, allows for `.graphql` files
- [x] cleans up import/exports
- [x] simplify naming so there's fewer competing naming structures for things
- Enforce eslint rules for: 
  - [x] import/order
  - [x] import/no-named-as-default-member
  - [x] import/prefer-default-export
  - [x] react-native/split-platform-components
  - [x] react-native/no-unused-styles
  - [x] react-native/no-inline-styles
  - [x] react-native/no-color-literals
  - [x] react/jsx-handler-names
- [x] Separate navigator and screen components to separate files

conrad-vanl:
Closes #101 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


burkeshartsis:
@conrad-vanl what do you think about enforcing the following eslint rules

- [react/jsx-no-bind](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md)
- [react/no-multi-comp](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-multi-comp.md)
- [react-native/no-inline-styles](https://github.com/Intellicode/eslint-plugin-react-native/blob/master/docs/rules/no-inline-styles.md)
- [react-native/no-color-literals](https://github.com/Intellicode/eslint-plugin-react-native/blob/master/docs/rules/no-color-literals.md)

conrad-vanl:
> - [react/jsx-no-bind](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md)

Like the idea of it except we don't have a problem with using `.bind` and as far as arrow functions go, I don't see us getting away from 'em. For example, the new `Query` component we're using from Apollo-Client makes heavy use of using arrow-functions in component props, because it excepts you [to pass a function to its children](https://www.apollographql.com/docs/react/essentials/queries.html#basic).

> - [react/no-multi-comp](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-multi-comp.md)

Yes, as long as it doesn't barf at having multiple styled component defs in one file (maybe we enable `ignoreStateless`?).

> - [react-native/no-inline-styles](https://github.com/Intellicode/eslint-plugin-react-native/blob/master/docs/rules/no-inline-styles.md)

💯 

> - [react-native/no-color-literals](https://github.com/Intellicode/eslint-plugin-react-native/blob/master/docs/rules/no-color-literals.md)

💯

burkeshartsis:
There are a few more things this PR could address as sort of a catch all PR for things that needed clean up. That said, I didn't enable the this eslint rule (mentioned above). 

```'react/jsx-no-bind': [2, { allowFunctions: false }],```

The primary reason for wanting this rule being to eliminate potential spots where an arrow function at render time would cause an otherwise pure component to always render. There are a lot of places this rule throws an error in our current code base and would require refactoring a lot of places just to pass lint with no performance difference. You might be thinking, "oh PTL, I love arrow functions and hate doing unnecessary work." To that I would like to add that while there are many places we shouldn't be concerned with this (eg `<Button onPress(() => (` that pushes a navigation change or something that requires a render anyways) the point of these types of rules are to enforce good habits that pay off when it does matter and to make you not have to worry if you are looking at otherwise inefficient code when you see an arrow function passed as a prop. All that said, making this change is likely it's own PR either way. 

@conrad-vanl Not sure who should review this. We both did a lot of work on it.

**Comment by burkeshartsis** [comment]
there do appear to be some `key` warnings from the various feeds. I believe those were here before this PR. Not sure if we want to fix them here or not.

**Comment by burkeshartsis** [comment]
@conrad-vanl what was your thinking on the naming with `/content-feed` and `/content-single`? Our current Components = PascalCase, functions = camelCase.

also, I'm not sure what `tabBarIcon` does? Why the need for a function that returns a component? It's not clear how the `tintColor` prop is used either? Should we add documentation to this?

**Comment by conrad-vanl** [comment]
Man, you really don't like `tabBarIcon`, do you? You brought this up in the PR it was added and I don't think this PR introduces any changes to it. It's just simply for being able to statically define a tab bar icon...it was taken from one of React-Navigation's examples. It's just a function that returns a component with the correct icon and the other defaults needed for tab bar icons.

Using dasherized names in `content-feed` and `content-single` was continuing the naming system already used....where modules use dasherized names, components use PascalCase. 

I'm all for changing all of these folder names to PascalCase, fwiw

**Comment by richarddubay** [comment]
@conrad-vanl @burkeshartsis 
From a purely functional point-of-view:

- I notice that the "series" in the current app are missing the horizontal scroller that was previously merged into `master`. It looks like `HorizontalTileFeed` is still there in the content single, so maybe it's broken?
- There are some `key` warnings like Burke mentioned. I have no problem moving them into a different PR, but they should probably be taken care of.
- Editorial items on the Discover tab are missing images ... but that _might_ just be the data in Rock?

File: mobile/src/content/query.js [hunk 1]
```diff
@@ -1,41 +0,0 @@
...
-        id
-        title
-        htmlContent
-        coverImage {
-          name
-          sources {
-            uri
-          }
-        }
-        childContentItemsConnection {
```
richarddubay: Removing this could be why the children items aren't showing on series.

mobile/.babelrc
```diff
@@ -7,10 +7,11 @@
     [
       "module-resolver",
       {
-        "root": ["./src"],
-        "extensions": [".android.js", ".ios.js", ".js"]
+        "root": ["./src", "./"],
+        "extensions": [".android.js", ".ios.js", ".js", ".graphql"]
```
richarddubay: I'm not sure how this helps us. There has to be an advantage other than being able to use `.graphql` files, right?

mobile/src/live/index.js
```diff
@@ -1,38 +1,2 @@
...
-
-LiveNowButton.propTypes = {
-  navigation: PropTypes.shape({
-    navigate: PropTypes.func,
-  }),
-};
-
-export default withNavigation(LiveNowButton);
+export default from './LiveModal';
+export LiveButton from './LiveButton';
```
richarddubay: I guess I'm okay with this. This seems like really nit-picky separation though. It makes the `index` file almost useless. It's just another file to hold all the exports. But I get it.

mobile/storybook/decorators/CenterView/index.js
```diff
@@ -1,11 +1,16 @@
-import React from 'react';
 import PropTypes from 'prop-types';
 import { View } from 'react-native';
-import style from './style';
 
-export default function CenterView(props) {
-  return <View style={style.main}>{props.children}</View>;
-}
+import styled from '../../../src/ui/styled';
```
richarddubay: Relative vs. Absolute paths?

mobile/src/live/index.js
burkeshartsis: The aim here is to make it easier/cleaner to write tests and stories for our composed screens. It should help with #110.

mobile/storybook/decorators/CenterView/index.js
burkeshartsis: because storybook lives outside the `/src` folder I originally thought I would need to navigate out and into there for this to work. Looks like babel actually doesn't have a problem with it all. Pushing a commit with an absolute path now.

mobile/.babelrc
burkeshartsis: this change allows all queries to distinguishable by their file type

mobile/src/content/query.js
burkeshartsis: @conrad-vanl was this change in the query on purpose?

conrad-vanl:
Aight, I went in and fixed child content items displaying and loading states / warnings that were popping up....going to look over the rest of the changes

conrad-vanl:
> The primary reason for wanting this rule being to eliminate potential spots where an arrow function at render time would cause an otherwise pure component to always render. There are a lot of places this rule throws an error in our current code base and would require refactoring a lot of places just to pass lint with no performance difference. You might be thinking, "oh PTL, I love arrow functions and hate doing unnecessary work." To that I would like to add that while there are many places we shouldn't be concerned with this (eg <Button onPress(() => ( that pushes a navigation change or something that requires a render anyways) the point of these types of rules are to enforce good habits that pay off when it does matter and to make you not have to worry if you are looking at otherwise inefficient code when you see an arrow function passed as a prop. All that said, making this change is likely it's own PR either way.

@burkeshartsis I brought this up before, but it appears that using arrow functions in component props is the new way with React, and so I hope performance improvements (or proof that there's not really performance impact here) are on the roadmap. For examples, checkout their new context API: 

image

Which is identical to the recommended `<Query>` component on Apollo:

image

So, until we can provide _proof_ of a felt impact to performance on this, I'd like to err on the side of the community. If the parent component is pure, we will be preventing unnecessary re-renders anyways, which is what this pattern should help us do. Higher order functions would allow us to keep from using so many arrow functions, but the problem is that tracking purity with HOCs become much more difficult.

conrad-vanl:
What are our thoughts on folder naming inside of the root-level `src` folder? I've been treating these folders as "modules" or "packages" of code, and thus been using dasherized names according to node-standards. Inside, individual files are PascalCase or camelCase depending on if that file exports a Component or a function. However, I would be 👍 treating all of these folders in `/src` as Component exports, and naming them PascalCase instead.

We used the dasherized names on NS/Apollos as we could only root-import folders that started with an "@" symbol, and the dasherized names helped make the non-root-import folders look different. In this case, we don't need that separation any more.

mobile/src/content/query.js
conrad-vanl: Good catch. Fixed!!

mobile/.babelrc
conrad-vanl: If we're putting queries in separate files, it's a mild nicety to be able to use `.graphql` files versus javascript. We definitely don't have to, and if y'all don't like this change, I can back it up. GQL is a different language then JS, and before we were just putting GQL inside of a string, inside of a JS file. I'm good either way though.

mobile/src/live/index.js
conrad-vanl: Trying to keep any one file to be all about just one component. We had a lot of files that defined two components - the navigator component and the screen component. I think it makes sense to separate these into separate files across the board to help illustrate that these are in fact separate.

HOWEVER, in this case, I could side both ways. The live screen i'd be fine with living in `live/index.js`, since that's still just one component per file. I'm not sure if having them separate makes testing or readability any easier. But on the flip, I'm fine with leaving this as is 👍

mobile/.babelrc
richarddubay: I'm fine with it. I honestly like the `.graphql` designation. But my concern was more around having to use Babel to process the files and if that would slow things down because more processing usually means more time. If it's not going to slow things down once we get 100's of queries being processed, then I'm fine with it.

mobile/src/live/index.js
richarddubay: Let's just leave it. I do think that if we want to go this way then we should make it a pattern everywhere. Not just one-offs. But this can stay the way it is.

richarddubay:
@conrad-vanl I don't think I have a strong opinion about the folder naming. Sticking with the standards seems to be the best way to go about it. But I don't have to have it that way. My biggest priority would be that we do whatever makes it simple for the community to understand and participate. That being true, I don't think that casing of folders would be as important as how those folders are structured (what goes in what folder and how easy it is to find it).

mobile/src/tabs/discover/Discover.js
```diff
@@ -0,0 +1,63 @@
+import React, { PureComponent } from 'react';
+import { Query } from 'react-apollo';
+import { get } from 'lodash';
+
+import FeedView from 'ui/FeedView';
+import BackgroundView from 'ui/BackgroundView';
+
+import TileContentFeed from './tileContentFeed';
```
richarddubay: I think `./tileContentFeed` needs to be `./TileContentFeed` now.

mobile/src/tabs/discover/TileContentFeed.js
```diff
@@ -1,11 +1,13 @@
...
 import { withNavigation } from 'react-navigation';
 import PropTypes from 'prop-types';
+
 import PaddedView from 'ui/PaddedView';
 import { H4 } from 'ui/typography';
 import HorizontalTileFeed from 'ui/HorizontalTileFeed';
 import styled from 'ui/styled';
 import { ButtonLink } from 'ui/Button';
+
 import TileImageItem from './tileImageItem';
```
richarddubay: The same capitalization issue here.

mobile/.babelrc
burkeshartsis: babel transpiles the js when the app is bundled. In production mode there would be no difference at all.

richarddubay:
screenshot 2018-07-11 13 53 25

It looks like when there are no children to show it still renders the space for them. Not sure if it needs to happen on this PR, but this shouldn't render that space.

conrad-vanl:
> It looks like when there are no children to show it still renders the space for them. Not sure if it needs to happen on this PR, but this shouldn't render that space.

Yeah, agreed, doesn't really belong in this PR. Perhaps in #93

mobile/src/tabs/discover/Discover.js
conrad-vanl: fixed

mobile/src/tabs/discover/TileContentFeed.js
conrad-vanl: and fixed

richarddubay:
Yeah, #93 works for me for fixing that spacing issue up.

richarddubay:
@conrad-vanl and @burkeshartsis I'm feeling pretty good about this. Any last minute things you guys want to add?

PR #111, opened 2018-06-22
Created Mutation to Register

### LINKED ISSUE #82: Mutation to create user logins
This is a backend only task (no user interface will result from this issue)

## User Story

Add ability to create a user login with the graphql API (aka...signup a new user)

#### Problem Details

There are two paths forward on this:

1. We kick user out to webview for creating an account on a Rock-powered website. The Pro to this is that sign up will be easily customized to that church and support other Rock features, like merges and any other customizations the church might've already built.

2. Create a signup mutation in our graphql and we handle the UI natively

If 1 can be done easily, I might like to start there as it would remove complexity on our end and allow us to move towards SSO implementations quicker. If not possible, no big deal though.

* [x] You can create a new profile with atleast an identity (email) and password
* [x] Creates or links profile on Rock
* [x] After sign up, a user can sign in
* [ ] Create new issues in Github for handling the UI side of this - whether it's creating the necessary forms or the webview flow with rock

IsaacHardy:
Fixes #82 

To be able to register, you'll need to pass an email and password to the following mutation:

``` js
mutation {
  registerPerson (email: "enter@email.here", password: "password") {
    token
    user {
      id
      profile {
        id
        email
      }
    }
  }
}
```

The scope of this issue only requires an email and password, but in the future we will add more fields to be required when registering a new user. 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [x] Review CI build results
- [x] Review new test logic
- [x] Review the changelog entry


api/src/data/auth/model.js
```diff
@@ -57,4 +58,75 @@ export default class AuthModel extends RockModel {
       throw e;
     }
   };
+
+  makeNewGuid = () => {
```
conrad-vanl: Is this function still needed?

IsaacHardy: In the current scope, no. Completely forgot to remove it after refactoring the tests.

api/src/data/auth/model.js
```diff
@@ -57,4 +58,61 @@ export default class AuthModel extends RockModel {
...
+    }
+  };
+
+  createUserLogin = async (props = {}) => {
+    try {
+      const { email, password, personId } = props;
+
+      return await this.context.connectors.Rock.post('/UserLogins', {
+        PersonId: personId,
+        EntityTypeId: 27, // A default setting we use in Rock-person-creation-flow
```
conrad-vanl: @IsaacHardy who's the "We" in this context - NewSpring? Or is this a Rock standard.

IsaacHardy: I actually don't know the answer to this question as of now. @richarddubay So you know is this standard is a NewSpring or Rock? (i.e. We create a personId then create a userLogin)

richarddubay: It looks like it's in Rock in general. `Rock.Security.Authentication.Database` is Id 27 in NewSpring's Rock instance AND the Apollos Rock instance. It probably gets set up that way in the original migrations.

Issue #91, opened 2018-06-18
Create horizontal scrollers on engage screen


conrad-vanl:
## User Story

As a user I want to browse a few of the top content items in each channel

#### Problem Details _(Optional)_

![image.png](https://images.zenhubusercontent.com/55bcf69f4f7a08fbb7f4086a/dfa7323b-d8ed-4b90-b855-6c3b3b02dc98)

#### Acceptance Criteria _(Optional)_

* [ ] It should create a reusable horizontal feed/scroller type component
* [ ] MVP can utilize the same type of card for all content channels (the horizontal card with the photo on the left and title on the right). This works similarly to new App Store app on iOS

IsaacHardy:
@conrad-vanl @burkeshartsis Will this reusable component behave like a component in our `UI` directory?

IsaacHardy:
@conrad-vanl I'm also a bit confused about the second point of the acceptance criteria. In the mock, I see a square image that is centered with text, but in the AC you are suggesting **(the horizontal card with the photo on the left and title on the right)**

burkeshartsis:
Hey @IsaacHardy the horizontal feed/scroller would likely be a reusable component in `/ui` much like `FeedView` is now.

I agree these is some ambiguity on which card lives in the scroller for what content. It seems like @conrad-vanl is just suggesting to simplify things for now and use the card that shows a title as it would work for all content types. I'm happy to meet with you tomorrow and talk to you about this more.

Issue #89, opened 2018-06-18
Show content children on content screens


conrad-vanl:
## User Story

As a user I want to browse a series' sermons, or a studies' devotionals

#### Problem Details _(Optional)_

* Implement the horizontal scroller from NS/Apollos to navigate to a content's child items

![image](https://user-images.githubusercontent.com/103927/41607852-a0d56100-73ac-11e8-8067-d33829ee629d.png)

#### Acceptance Criteria _(Optional)_

* [ ] It should show a horizontal scroller that links to individual child items
* [ ] Should only show a horizontal scroller if that item has children

richarddubay:
Do you feel like this task and #91 would use different horizontal scrollers? Same scroller with different props? They feel like two different scrollers, but I thought I'd ask.

PR #116, opened 2018-06-27
Add tiles and horizontal scrollers

richarddubay:
This is a part of both #89 and #91.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [x] Review CI build results
- [x] Review new test logic
- [x] Review the changelog entry


richarddubay:
@burkeshartsis and @IsaacHardy This is really fast and dirty. I'm expecting changes. I cannot guarantee that the props being passed are the props that are needed or that they are even called the correct thing. I imagine these will have to be modified as we go. 

Also Burke, I believe that we could probably do something with the `CardTile` to get rid of the `OverflowFix` and `WebAspectRatioFix`, but when I tried to remove them originally the tile looked pretty janky in Storybook. I'm not _quite_ sure what I need to do there. 

So ... what needs to change right now?

mobile/src/ui/TileImage/TileImage.tests.js
```diff
@@ -30,13 +30,12 @@ describe('the SquareTile component', () => {
...
   });
   it('passes an onPressItem prop', () => {
-    const onPress = () => {};
     const tree = renderer.create(
       <Providers>
         <SquareTile
           image={'https://picsum.photos/600/400/?random'}
           link={'https://github.com'}
-          onPressItem={onPress}
+          onPressItem={jest.fn}
```
richarddubay: Of course! This is so smart.

PR #117, opened 2018-06-28
91 Create horizontal scrollers on engage screen

### LINKED ISSUE #91: Create horizontal scrollers on engage screen

IsaacHardy:
Fixes #91 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


conrad-vanl:
@IsaacHardy what are your thoughts on code coverage here? The patch (~88%) is a bit below our current standard (~93%). Generally anything above 90% I'm Ok with but if it's close I'm open to interpretation.

IsaacHardy:
@conrad-vanl I haven't been able to figure out why the patch is such a low percentage.

conrad-vanl:
@IsaacHardy if you want to work on the conflicts I will work on the review/merge-readiness

conrad-vanl:
nvm, i'll fix theconflicts :)

conrad-vanl:
@richarddubay since Isaac is out and I ended up doing a lot of work on this PR, could you do a review on this?

conrad-vanl:
This has been simplified a bit since @IsaacHardy's initial integration, we'll follow up with more cleanup.

mobile/src/tabs/discover/index.js
```diff
@@ -1,93 +1,40 @@
...
-    } = this.props;
-    return (
-      <BackgroundView>
-        <Query query={GET_DISCOVER_ITEMS}>
-          {({ loading, error, data, refetch }) => {
-            if (loading) return 'Loading...';
+const DiscoverScreen = () => (
+  <BackgroundView>
+    <Query query={GET_DISCOVER_ITEMS}>
+      {({ loading, data: { contentChannels = [] } = {}, refetch }) => (
```
richarddubay: No error checking?

mobile/src/tabs/discover/tileContentFeed.js
```diff
@@ -0,0 +1,66 @@
...
+        />
+      )}
+      loadingStateObject={{
+        id: 'fake_id',
+        title: '',
+        coverImage: [],
+      }}
+      isLoading={isLoading}
+      refreshing={isLoading}
+    />
```
richarddubay: Seems redundant to have to pass `isLoading` twice. Is there a way to not have to?

mobile/src/ui/FeedView/index.js
```diff
@@ -59,12 +60,12 @@ export class FeedView extends React.Component {
...
-          channelTypeIcon={item.node.parentChannel.iconName}
-          images={item.node.coverImage.sources}
-          isLoading={item.node.isLoading}
+          id={get(item, 'node.id')}
+          title={get(item, 'node.title') || get(item, 'node.name') || ' '}
+          channelType={get(item, 'node.parentChannel.name')}
+          channelTypeIcon={get(item, 'node.parentChannel.iconName')}
+          images={get(item, 'node.coverImage.sources')}
+          isLoading={get(item, 'node.isLoading')}
         />
```
richarddubay: More curious than anything here ... is there benefit to having to go through `lodash` to do the `get` instead of just naming the data directly? Seems like an extra call to get data that we already have.

mobile/src/ui/HorizontalTileFeed/index.js
```diff
@@ -68,20 +68,22 @@ const generateLoadingStateData = (loadingStateObject, numberOfItems) => {
 
   while (loadingStateData.length < numberOfItems) {
     const newData = itemData();
-    newData.node.id = `fakeId${loadingStateData.length}`;
+    newData.id = `fakeId${loadingStateData.length}`;
     loadingStateData.push(newData);
```
richarddubay: I'm not sure how this works. Isn't the id on the `node` from the data?

mobile/src/ui/SideBySideView/Left.js
```diff
@@ -0,0 +1,20 @@
...
+export default compose(
+  mediaQuery(
+    ({ md }) => ({ minWidth: md }),
+    styled({
+      width: `${100 - 41.6666666}%`,
+      height: '100%',
+      overflow: 'hidden',
+      ...Platform.select({
+        web: {
+          maxHeight: '100vh',
```
richarddubay: I think you can remove anything related to "web" here.

mobile/src/ui/SideBySideView/Right.js
```diff
@@ -0,0 +1,23 @@
...
+  mediaQuery(
+    ({ md }) => ({ minWidth: md }),
+    styled(
+      {
+        width: '41.6666666%',
+        height: '100%',
+        overflow: 'hidden',
+        ...Platform.select({
+          web: {
+            maxHeight: '100vh',
```
richarddubay: The same thing with "web" here.

mobile/src/ui/ThumbnailCard/index.js
```diff
@@ -0,0 +1,93 @@
...
+import FlexedView from 'ui/FlexedView';
+import { H5, BodyText } from 'ui/typography';
+import Paragraph from 'ui/Paragraph';
+import { enhancer as mediaQuery } from 'ui/MediaQuery';
+import ChannelLabel from 'ui/ChannelLabel';
+
+import Thumbnail from './Thumbnail';
+
+const enhance = compose(
+  setPropTypes({
```
richarddubay: This could probably be moved out into its own thing.

richarddubay:
@conrad-vanl I left some comments above. Most of it probably falls in the "we'll follow up with more cleanup." category, but I wanted to mention them. If you feel good about all of this, I can approve.

richarddubay:
@conrad-vanl Also, the test coverage dropped ... are you okay with that?

mobile/src/ui/FeedView/index.js
conrad-vanl: `get` is a "safe" way to access nested items. For example, if `parentChannel` or `coverImage` comes back `null` from graphql (which according to the current schema, _can_ happen). I was running into issues where `coverImage === null` and this was throwing error. `get` is really nice to use in mapping component props as we can build our components to accept `null` values, and then map the props using `get` like this and not build a lot of special cases, like:

```
images={item.node && item.node.coverImage && item.node.coverImage.sources}
```

mobile/src/tabs/discover/tileContentFeed.js
conrad-vanl: 👍  I'll look

mobile/src/tabs/discover/index.js
conrad-vanl: 😇 I'll add

mobile/src/ui/HorizontalTileFeed/index.js
conrad-vanl: Yeah....so the whole `edges`/`node` schema has fubar'd our existing feed components a bit. I'd like to clean this up, bu one things this PR does is `map` our `node` and `edges` and clean up the data shape before passing down into the feed components, so our feed components don't have to mess with the shape of `edges` and `nodes`

mobile/src/ui/SideBySideView/Left.js
conrad-vanl: 👍

mobile/src/ui/FeedView/index.js
richarddubay: Well, I'd definitely rather not have that. :)

conrad-vanl:
I'd actually like to address all of your comments @richarddubay  - i'll follow up with changes.

Very minor test coverage drop, and on close inspection, it appears to only hit lines of code that we are only able to easily test with enzyme (such as onClick handlers). Since it's so close, I'm okay with the drop (-0.1%)

mobile/src/ui/HorizontalTileFeed/index.js
richarddubay: Any chance that we can get rid of the whole `node/edges` thing? I get the premise of it coming from Relay, but it kind of muddies the water.

conrad-vanl:
@richarddubay this should be G2G, let me know

PR #127, opened 2018-07-12
93 Content Polish

### LINKED ISSUE #93: Placeholder content item design issue
## TODO

this is a placeholder issue for doing a 💄 pass on content item screens. This can encompass things like typography, placeholders, colors, etc.

----

## User Story

* [ ] It should [SPEC_DETAIL]

burkeshartsis:
closes #93 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


api/src/data/node/model.js
```diff
@@ -5,20 +5,21 @@ const secret = process.env.SECRET || 'LZEVhlgzFZKClu1r';
 export function createGlobalId(id, type) {
   const cipher = Crypto.createCipher('aes192', secret);
 
-  let encrypted = cipher.update(`${type}:${id}`, 'utf8', 'hex');
+  let encrypted = cipher.update(`${id}`, 'utf8', 'hex');
   encrypted += cipher.final('hex');
 
-  return encodeURI(encrypted);
+  return `${type}:${encrypted}`;
```
richarddubay: This isn't an issue, I'm just not sure I understand the point of this. Why is adding the type of the data relevant?

mobile/src/content-single/index.js
```diff
@@ -43,7 +52,8 @@ class ContentSingle extends PureComponent {
   }
 
   handleOnPressItem(item) {
-    this.props.navigation.push('ContentSingle', {
+    const { navigation } = this.props;
+    navigation.push('ContentSingle', {
```
richarddubay: This is two lines to replicate the effect of the one line that was removed. Is this just a readability thing? A React Native pattern we'd like to follow? Just asking questions...

mobile/src/content-single/index.js
```diff
@@ -67,34 +77,56 @@ class ContentSingle extends PureComponent {
...
+        fetchPolicy="cache-only"
+      >
+        {({ data: cachedData }) => (
+          <Query query={getContentItem} variables={this.itemId}>
+            {({ loading, error, data }) => {
+              const content = {
+                ...(cachedData.node || {}),
+                ...(data.node || {}),
+              };
+
```
richarddubay: I have no idea how adding the second query _outside_ the first query helps at all with making the query load data (and thereby load something to the screen) faster. I get the idea that chances are it will get some cached content. But we're not setting that content until we're inside the function of the second query. I can trust you that it works ... I just don't understand how.

mobile/src/live/LiveButton.js
```diff
@@ -19,31 +19,30 @@ const LiveNowButton = () => (
   <Query query={getLiveStream}>
     {({ loading, data }) => {
       const isLive = get(data, 'liveStream.isLiveNow', false);
-      return (
+
+      return isLive ? (
```
richarddubay: Moving this out here is a good move! 💯

api/src/data/node/model.js
burkeshartsis: cc: @conrad-vanl

conrad-vanl: It helps us in cache lookups on the client: https://github.com/ApollosProject/apollos-prototype/blob/93-content-polish/mobile/src/client/cache.js#L17

mobile/src/content-single/index.js
conrad-vanl: I believe the linter caught this for some reason. It's a lint rule that I'd be all in favor of removing. Something about forcing object destructuring.

mobile/src/content-single/index.js
conrad-vanl: Yup! Apollo changed their semantics on this and won't return any cached data unless the entire query matches what's in the cache. So if we're querying for any attribute that we don't have cached, Apollo won't return any data. So we used two queries instead - a very minimal query that we will likely have cached data for from the feed view, and a second query that more likely will have to hit the server for data.

Unfortunately, there's no way to do this in Apollo with only one query

mobile/src/content-single/index.js
burkeshartsis: nah I just removed the destructuring and it didn't complain. Pushed a commit clean that up.

mobile/src/content-single/index.js
conrad-vanl: Oh, and the second query will still render immediately - just with no data and `loading` set to true. Which allows for the first query to bring in its cached data

richarddubay: Two queries _seems_ more inefficient ... but I get what you're saying. I would think that would be something that Apollo wouldn't want to have be that way. Hmmm...

mobile/src/live/LiveButton.js
burkeshartsis: 💥

richarddubay:
television-justin-bieber-celebs-yC7vLP2q7kUHC

PR #128, opened 2018-07-12
78 Live Stream

### LINKED ISSUE #78: Determine ability to track live stream schedule or status
High-level goal of this issue is to determine how we're going to integrate with ChOP live status, and create a graphql schema for it.

## User Story

We only want to display the "We're live" banner when the church is actually broadcasting a live event

#### Problem Details

* There might be a Church Online api that allows us to get live events or streaming status
    * If not, lets sync with the Life Church people on next steps here, as it was suggested this is something they'd be willing to build. At this point, we can explore other potential alternative points of integration as well (can we integrate directly with Rock to get a worship schedule, for example)
    * If so, lets understand what the API provides and how it structures information
* Lets consider how we want to integrate with such an API...we can either ping the API ever so often to see if the church is currently live, or if the API exposes a schedule, we can use that schedule to statically determine if the church _should be_ live or not

#### Acceptance Criteria

* [ ] Create a graphql schema for querying live status and the streaming URL. We do **NOT** need to hook this query up to the ChOP API though, this issue is just for understanding their API and spec'ing out what schema we want to use
* [ ] Write a resolver for above query that returns fake data
* [ ] Hook our "Live Now" button up to this data
* [ ] Document and create issues for next steps on ChOP integration

richarddubay:
Fixes #78 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


richarddubay:
@conrad-vanl So this just sets up the query to return live is true, but it doesn't match the return values from the https://newspring.churchonline.org/api/v1/upcoming_event_times endpoint. I can set that up if that's a better use case. 

Also, I'm not confident in what I've got in the resolver. I feel like it's not right. I'm returning `Query` and `LiveFeed`, but I'm not sure I need to?

If you get a minute, can you take a look at this PR and let me know where I can make it better?

FWIW, if I change the resolver to `false`, the live bar goes away. So at least that's hooked up correctly. :)

api/src/data/live/schema.js
```diff
@@ -0,0 +1,7 @@
+import { gql } from 'apollo-server';
+
+export default gql`
+  type LiveStream {
+    isLiveNow: Boolean!
```
conrad-vanl: 💯

mobile/src/live/LiveButton.js
```diff
@@ -1,36 +1,51 @@
...
-              }
-            />
-          </CardContent>
-        </LiveCard>
-      </Touchable>
-    )}
-  </WebBrowserConsumer>
+  <Query query={getLiveStream}>
+    {({ loading, error, data }) => {
+      if (error) return <ErrorCard error={error} />;
```
conrad-vanl: I wonder...in real world if we really want to display an error here? Most of the time the church won't be live, and showing a user an error message for UI that normally doesn't exist, and they didn't ask for, is a good user experience. @burkeshartsis ??

richarddubay: Yeah, I don't think we do honestly. If it's broken, it should probably just not show anything.

richarddubay:
@conrad-vanl So I started off here with the boolean `isLiveNow` because I thought it'd be the easiest to build out and get working in the app. 

Now, I'm not opposed to seeing if we can get the time of the next event. I'm not sure at all how we would make that work though. I just don't have any experience with that. 

If we pull that data when the app loads and get a date/time, where does that live while we're waiting on that time to roll around, and how does the `liveButton` know when to display? Maybe that's where the subscription comes in.

conrad-vanl:
@richarddubay I agree, lets keep it with `isLiveNow: bool` _and_ in the future we can add a subscription on that if we don't want to poll.

The timestamp idea was just to get us thinking a little bit...but looking at the data that the ChOP api returns, it returns events and the times they start. So we could return the start time of the next event, and the client would just start a clock and render the "Live Now" button once the clock is >= the startTime. But I think `isLiveNow` is more explicit, keeps the client simpler, and we have an improvement path in the future (subscriptions).

api/src/data/live/__tests__/model.tests.js
```diff
@@ -0,0 +1,23 @@
+import fetch from 'isomorphic-fetch';
+import LiveStream from '../model';
+// import RockConnector from '../../../connectors/rock';
+
+describe('LiveStream', () => {
```
conrad-vanl: I don't see that you added a `live/model.js` anywhere, or am I missing that? So....what is this testing?

richarddubay: There is a `live/model.js`, but it literally just returns a resource. I wasn't sure what else to do with this file since it _technically_ doesn't even need to connect to the `RockModel`. Maybe I should do something else with it?

richarddubay: I removed the tests though. Because they weren't really testing much ... just that the correct `resource` was returned. I was just trying to keep the test coverage up.

conrad-vanl: lets remove the model for now too then

api/src/data/live/model.js
```diff
@@ -0,0 +1,5 @@
+import { RockModel } from '../../connectors/rock';
+
+export default class LiveStream extends RockModel {
```
conrad-vanl: Yeah lets just remove this for now. It won't even be a `RockModel` when we hook it up, since it's connecting to ChOP and not Rock

PR #147, opened 2018-07-18
Improved splash screens

burkeshartsis:
This added a package to handle the transition from the native launch/splash screen to the point at which JS is actually loaded and ready to render something. Without this a white screen will flash after the launch screen as React Native is mounted. It's really just [another bridge in a series of bridges](https://imgflip.com/i/2e8ljv). 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic

### DIFF
--- a/mobile/App.js
+++ b/mobile/App.js
@@ -1,7 +1,7 @@
-import React from 'react';
+import React, { Component } from 'react';
 import { createStackNavigator } from 'react-navigation';
 import { Sentry } from 'react-native-sentry';
-
+import SplashScreen from 'react-native-splash-screen';
 import Providers from 'Providers';
 import ContentFeed from 'content-feed';
 import ContentSingle from 'content-single';
@@ -24,10 +24,18 @@ const AppNavigator = createStackNavigator(
   }
 );
 
-const App = () => (
-  <Providers>
-    <AppNavigator />
-  </Providers>
-);
+class App extends Component {
+  componentDidMount() {
+    SplashScreen.hide();
+  }
+
+  render() {
+    return (
+      <Providers>
+        <AppNavigator />
+      </Providers>
+    );
+  }
+}
 
 export default App;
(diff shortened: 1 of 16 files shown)


conrad-vanl:
🤔

IsaacHardy:
@conrad-vanl What concerns do you have about the package?

conrad-vanl:
@burkeshartsis might be better suited to detail them. It has a lot of open issues - https://github.com/crazycodeboy/react-native-splash-screen/issues, and it appears that it could be a source of headache when we want to upgrade our react-native version.

burkeshartsis:
So just looking at this package on github shows it's a little unstable but if I'm honest I'm not really sure there is a better solution. In this PR this package is pinned to an older version of of itself because the latest has typos in iOS and bugs in Android. There are work arounds but they aren't pretty. The good news is that it appears to be actively maintained so stuff looks to at least eventually get addressed, albeit maybe slower than we might like. Another issue with this package is that it's kind of entirely configured outside of JS which means future projects will need to set it up themselves or we'll need to have steps in a build tool. It's really not much more than what is required for adding assets to release to the app store though.

TLDR; I'd probably just add it. It works right now and I believe fixes will be merged soon that will allow it to get bumped back to the most recent version. I also think there is a better solution. Kind of a not fun pill to take but hey... splashy splash screens

burkeshartsis:
You are correct @conrad-vanl. It shows the splash screen and then adds a JS hook to hide it that is executed when the app has mounted. I believe a lot of the instability with this package is recent. The number of stars on Github and the fact that there isn't really a better that does this the right way suggest to me that we may end up revisiting this package anyways. FWIW, Android is really the one that is affected most noticeably here. If you feel sketchy about merging it I understand. I see two options,
1. Don't merge and close the PR.
2. Merge the PR because it works right now and when we build create-apollos-app we can make this an experimental feature.

Whichever way you choose to go, I'm willing to support your decision.

conrad-vanl:
@burkeshartsis is there anything you want to add to the user-value side of things?

PR #149, opened 2018-07-18
109 Add Enzyme

### LINKED ISSUE #109: Add Enzyme
## User Story

As a developer, I want to make sure that our app is tested as well as possible so that I can be confident that what we send to our people works.

In this case, implement Enzyme (http://airbnb.io/enzyme/docs/guides/react-native.html) to test our components. This would be super helpful for testing events.

* [ ] Research whether or not we can use Enzyme for testing our components.
* [ ] It should test the `onPressItem` function in `ui/FeedView`.

richarddubay:
Fixes #109 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


richarddubay:
One thing I should probably do is split out these one big test into multiple tests ... but that seems kind of redundant with the fact that it is required to dig down so deep into the stack in order to reach the props we need to run the test. Maybe that's not worth splitting up right now?

conrad-vanl:
@richarddubay I see you made some changes and might've gotten the test coverage to improve. What from your comment above would you still like help with? Was @IsaacHardy syncing with you on this?

conrad-vanl:
Also, @richarddubay, instead of doing a bunch of `.dive()`, can we make use of the `testId` prop? Maybe: https://github.com/wix/enzyme-drivers/blob/master/README.md

richarddubay:
@conrad-vanl Yep ... I found a way to get it so that we got the `onPress` function covered. Which is great. I just don't feel super great about what I had to do in order to make that happen. No, @IsaacHardy and I haven't talked about it. I'll take a look at that link you sent ... maybe there's something there I can use.

Also, thinking about heading this direction: https://www.joinroot.com/blog/mounting-react-native-components-with-enzyme-and-jsdom/

But I've pretty much exhausted my 8 points on this. So I don't want to spend too much more time on it.

richarddubay:
I think a lot of the problem with being able to get to certain pieces of the rendered stack and do stuff is because everything is so wrapped up in other HOC's that it's almost impossible to untangle them. All of the docs I've been reading just show a snapshot that looks like: 

```
<View>
   <Text>
      Here's so text
   </Text>
</View>
```

What we end up with in our snapshots start out like:

```
<branch(getContext(mapProps(WithWindow)))
```

and then it only gets worse from there. So many of the examples show someone being able to do something like `wrapper.find(MyComponent)` and then do something with it. We can't do that ... or at least, I haven't been able to figure out how to make it work. Too many things wrapping too many things. 

@conrad-vanl or @IsaacHardy If one of y'all know how to make this work better, I'm all ears.

mobile/src/ui/FeedView/FeedView.tests.js
```diff
@@ -144,4 +146,81 @@ describe('The FeedView component', () => {
...
+
+    // This is a snapshot of the component at it's highest level.
+    expect(wrapper).toMatchSnapshot();
+
+    // This digs down into the component to get to the level where the
+    // 'onPressItem' function can be reached.
+    // FWIW, I feel like this is NOT maintainable long term in
+    // it's current state.
+    // Maybe this would get better after https://github.com/airbnb/enzyme/issues/1436
+    const render = wrapper
```
IsaacHardy: Have you looked into grabing the element where the `onPressItem` lives by an enzyme selector? http://airbnb.io/enzyme/docs/api/selector.html

IsaacHardy: You could then use something the `find()` method to select that specific element. http://airbnb.io/enzyme/docs/api/ReactWrapper/find.html

richarddubay: Yeah, I've tried a lot of this. I can't seem to get them to work. I just get `undefined` most of the time.

richarddubay: Per this comment (https://github.com/airbnb/enzyme/issues/947#issuecomment-302311398) it looks like to `shallow` render, you have to do one dive per HOC.

IsaacHardy:
@richarddubay I have a meeting at 2pm, but let's 🍐 on this afterwards.

IsaacHardy:
Hey @conrad-vanl, @richarddubay and I took a look at this and found this useful tool https://github.com/Stupidism/enzyme-shallow-until. However, we were unable to get it working. 

Because the way in which `shallow` works in `enzyme`, testing HOCs can be a huge chore. It seems using the `.dive()` method multiple times is an option but not very good one. We should not merge this in until we can determine there are no other possible solutions to testing HOCs.

richarddubay:
@conrad-vanl I just wanted to chime in real quick. @IsaacHardy and I looked at a lot of things, found some things we thought might be helpful but, like he said, we were never able to get them to work. 

The current solution _does_ get our test coverage up and it _does_ prove that the `onPress` function works. That being said, I think there are probably a couple of these tests that could be removed (redundancy or just not needed here). Also, once again I'll say that I'm not super happy with having to `.dive()` 9 times. But according to [this](https://github.com/airbnb/enzyme/issues/947#issuecomment-302311398), that is exactly what needs to happen. 

We haven't found a good way to do what should be simple. Things like `.find()` just won't work with our HOC setup. There is also a feeling I've been getting that this won't truly be solved for React Native until we get an adapter (https://github.com/airbnb/enzyme/issues/1436).

conrad-vanl:
So what's the consensus here...do we still want to get this PR merged as is?

IsaacHardy:
@conrad-vanl That will have to be your call. I personally think having to `.dive()` that many times to connect to a component will be quite annoying, yet possible.

IsaacHardy:
Another thought, if there is a way to make this process easier with Enzyme while using HOCs, then it will take some substantial time figuring out how.

IsaacHardy:
@conrad-vanl @richarddubay After looking into this a bit more, I think we should go ahead and close this issue for the time being.

I think the amount of work needed to implement a successful test is tedious, but doable. However, the main reason I am closing this issue is the potential for future issues.

Enzyme doesn't seem to behave as fluidly with React-Native and requires a couple extra packages for it to come close to the configuration and functionality of using Enzyme with React. Because of the potential for issues and the complexity, I think it wise to close this issue until we can come up with a better solution.

Other news, Detox is supporting Android now even though only in React Native v51 and earlier which means we might be able to use that in the future.

PR #155, opened 2018-07-19
User can see their name and profile picture on the connect screen once logged in

### LINKED ISSUE #88: Show user profile on connect screen
## TODO

- [ ] Needs new/updated design mockup

-----

## User Story

As a user I want to know the app knows who I am, and begin to associate space within the app as "my own".

* [ ] It should show the user's profile photo
* [ ] It should show the user's name
* [ ] Create issues for adding the ability to edit your profile and change your profile photo

IsaacHardy:
Fixes #88 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


IsaacHardy:
simulator screen shot - iphone x - 2018-07-19 at 16 30 15

IsaacHardy:
simulator screen shot - iphone x - 2018-07-19 at 16 30 32

api/src/data/people/__tests__/resolvers.test.js
```diff
@@ -26,6 +26,9 @@ describe('Person', () => {
           lastName
           nickName
           email
+          photo {
```
conrad-vanl: Thoughts on having this return an `ImageMediaSource` instead? https://github.com/ApollosProject/apollos-prototype/blob/master/api/src/data/media/index.js#L44

My thought would be to make all assets (image, video, audio) use the same types, so as we support more features they are easier to add. Also, the idea with the `ImageMediaSource` type is it matches the object structure exactly that React Native's `Image` component expects (ie...using "uri" instead of "url", hopefully eventually supporting `width` and `height` attributes).

mobile/src/ui/UserAvatarView/index.js
```diff
@@ -0,0 +1,106 @@
...
+    isLoading,
+    refetch,
+    onPhotoPress,
+    setIsUploadingFile,
+    isUploadingFile,
+    ...viewProps
+  }) => {
+    const ImageContainer = Touchable;
+    // todo: handle file select stuff
+    /* eslint-disable react-native/no-inline-styles */
```
conrad-vanl: Booooo lets get rid of this eslint-disable

mobile/src/ui/UserAvatarView/index.js
```diff
@@ -0,0 +1,106 @@
...
+    return (
+      <Container {...viewProps}>
+        <Content>
+          <ImageContainer
+            webWrapperStyle={{ alignItems: 'center' }}
+            onUploadStarted={() => setIsUploadingFile(true)}
+            onUploadEnded={() => setIsUploadingFile(false)}
```
conrad-vanl: Maybe all of these props should be removed for now?

mobile/src/ui/UserAvatarView/index.js
```diff
@@ -0,0 +1,106 @@
...
+        paddingTop: '100%',
+        borderRadius: '50%',
+      },
+    }),
+  },
+}))(Avatar);
+
+const enhance = compose(
+  pure,
+  setPropTypes({
```
conrad-vanl: We should use native `propType` api now instead of HOCs

mobile/src/ui/UserAvatarView/index.js
```diff
@@ -0,0 +1,106 @@
...
+    isLoading: PropTypes.bool,
+    refetch: PropTypes.func,
+    onPhotoPress: PropTypes.func,
+    blurIntensity: PropTypes.number,
+    allowProfileImageChange: PropTypes.bool,
+    ...View.propTypes,
+  }),
+  withThemeMixin({ type: 'dark' }),
+  withState('isUploadingFile', 'setIsUploadingFile', false)
+);
```
conrad-vanl: I'm almost thinking we can remove _all_ of these HOCs. the `setPropTypes` should be moved down to the end like `UserAvatarView.propTypes = {}`, the `withThemeMixin` can be removed, which will make this component light instead of dark, and the `withState` was specific to our uploading file implementation, which can be removed for now.

mobile/src/ui/UserAvatarView/index.js
```diff
@@ -0,0 +1,106 @@
...
+});
+const Name = copyStyles(H4);
+const City = copyStyles(BodyText);
+
+const StyledAvatar = withTheme(({ theme }) => ({
+  containerStyle: {
+    marginRight: 0,
+    marginBottom: theme.sizing.baseUnit / 2,
+    ...Platform.select({
+      web: {
```
conrad-vanl: Web styles not needed.

mobile/src/ui/Avatar/index.js
```diff
@@ -0,0 +1,73 @@
...
+import ConnectedImage from 'ui/ConnectedImage';
+import styled from 'ui/styled';
+import { withTheme } from 'ui/theme';
+import ActivityIndicator from 'ui/ActivityIndicator';
+
+export { default as AvatarList } from './List';
+
+const enhance = compose(
+  pure,
+  setPropTypes({
```
conrad-vanl: Use native react api for setting proptypes

mobile/src/tabs/connect/Connect.js
```diff
@@ -15,59 +20,75 @@ class Connect extends PureComponent {
...
+          {({ data: { isLoggedIn = null } }) => {
+            if (isLoggedIn)
+              return (
+                <Query query={getUserProfile}>
+                  {({ data: { currentUser: { profile } = {} } = {} }) => (
+                    <UserAvatarView user={profile} />
+                  )}
+                </Query>
+              );
+            return '';
```
conrad-vanl: What would you think about if the user isn't logged in, displaying an empty `UserAvatarView`?

mobile/src/ui/UserAvatarView/index.js
IsaacHardy: So what it sounds like to me, is that you'd like me to revamp the components I've moved over from NS/AP and remove all HOCs?

mobile/src/ui/UserAvatarView/index.js
IsaacHardy: Good catch.

mobile/src/tabs/connect/Connect.js
IsaacHardy: I actually debated that with myself. I felt that to display nothing was a better experience for the user.

api/src/data/people/__tests__/resolvers.test.js
IsaacHardy: I do like that. Let me look into implementing this and I'll get back at you.

mobile/src/ui/UserAvatarView/index.js
IsaacHardy: Good point.

mobile/src/ui/UserAvatarView/index.js
conrad-vanl: Correct. that's generally what we've been doing when moving components over

mobile/src/tabs/connect/Connect.js
conrad-vanl: I'm fine with rendering nothing for now, as we are still working on final design directions for this screen, but I think the pro to displaying an empty state here is that it gives the use a preview for what the UI should look like, and promote them to do what we "want" them to do (login). We could even make tapping on the empty state trigger the login flow.

IsaacHardy:
@conrad-vanl Changes have been made!

mobile/src/ui/Avatar/Avatar.stories.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react-native';
+
+import FlexedView from 'ui/FlexedView';
+
+import Avatar from '.';
+
+const source = {
+  url:
```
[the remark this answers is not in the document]
IsaacHardy: Hmm, I'm surprised it ever works while using `url`.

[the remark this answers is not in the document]
conrad-vanl: our custom image component (`ConnectedImage`) was built to support either `url` or `uri` to make working with Heighliner easier.

IsaacHardy:
@conrad-vanl Updated with `uri` changes to storybook and snapshot.

mobile/src/ui/UserAvatarView/UserAvatarView.stories.js
```diff
@@ -0,0 +1,22 @@
...
+const photo = {
+  url:
+    'https://dg0ddngxdz549.cloudfront.net/images/cached/images/remote/http_s3.amazonaws.com/ns.images/all/member_images/members.nophoto_1000_1000_90_c1.jpg',
+};
+
+storiesOf('UserAvatarView', module).add('default', () => (
+  <FlexedView>
+    <UserAvatarView
+      photo={photo}
+      firstName="John"
```
conrad-vanl: Did you test this story? The `propTypes` for `UserAvatarView` shows you passing it a `user` object as a prop, with `user.photo`, `user.firstName`, etc. Here they're blown out.

IsaacHardy: The story has been tested. This is how the Component is setup in NS/AP. [I think you wrote it](https://github.com/NewSpring/Apollos/commit/f9e5fa14bb43379f1e3d4886488f37a229f2d6cc#diff-d69ff039d05236de9a7492c7ecd77c92). 

I don't mind changing it if we think it is a better representation of the Component, but if I had to guess, I'd say that this story is merely attempting to test that the Component can receive the correct data and generate the require view.

conrad-vanl: Fair enough. I'll reiterate though. There's an inconsistency in how this component is being used...

The source code for the `UserAvatarView` shows you passing a `user` prop as an object: https://github.com/ApollosProject/apollos-prototype/pull/155/files/c8d3612e059d9fbdc412540a6918214a1ea0a1e4#diff-4864a053a8001abf9871a6da58e5af39R68

This story is not passing that prop. These separate `photo`, `firstName`, etc props are not even going to be evaluated. Which means the story shows up blank:

image

So this should be corrected in this PR 😄

conrad-vanl: Alternatively, generally the separate props are better design from a prop-design perspective (passing objects as props should be avoiding as much as possible), so I'd be 👍 on changing the prop types for this component to match the current story

IsaacHardy: @conrad-vanl I've made changes to the UserAvatarView component. Check those out to see if that is what you were trying to explain.

IsaacHardy:
@conrad-vanl I have a fix. Not having a profile pic is def the issue. There are also unwanted behavior for not having a firstName and lastName either. Gonna fix those and make a push.

PR #174, opened 2018-07-25
add enableAapt2 to gradle.props

IsaacHardy:
## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic

### DIFF
--- a/mobile/android/app/release/output.json
+++ b/mobile/android/app/release/output.json
@@ -1,1 +1,0 @@
-[{"outputType":{"type":"APK"},"apkInfo":{"type":"MAIN","splits":[],"versionCode":6,"versionName":"0.3.0","enabled":true,"outputFile":"app-release.apk","fullName":"release","baseName":"release"},"path":"app-release.apk","properties":{}}]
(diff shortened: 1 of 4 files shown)


mobile/android/app/release/output.json
```diff
@@ -0,0 +1 @@
+[{"outputType":{"type":"APK"},"apkInfo":{"type":"MAIN","splits":[],"versionCode":6,"versionName":"0.3.0","enabled":true,"outputFile":"app-release.apk","fullName":"release","baseName":"release"},"path":"app-release.apk","properties":{}}]
```
burkeshartsis: I think this file and the entire `/release` directory can be removed.

mobile/android/gradle.properties
```diff
@@ -18,3 +18,4 @@
 # org.gradle.parallel=true
 
 android.useDeprecatedNdk=true
+org.gradle.configureondemand=true
```
burkeshartsis: I do not believe we need this anymore.

mobile/appcenter-pre-build.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
burkeshartsis: can this file live in `/scripts`

IsaacHardy: No it cannot.

mobile/android/gradle.properties
IsaacHardy: You might be right.

mobile/android/app/release/output.json
IsaacHardy: No idea where this came from 😬

burkeshartsis: It's generated when you create a release build.

Issue #142, opened 2018-07-16
Hook up live stream API to live ChOP data


conrad-vanl:
## User Story

A user can watch the live stream during a church's live stream service time

#### Acceptance Criteria _(Optional)_

* [ ] Gets data from ChOP API
* [ ] Determine if there's a way we can mock this or artificially trigger live or not (cuz the apollos church won't really ever be "live"). Maybe we can upload a fake schedule to ChOP?

richarddubay:
@conrad-vanl In addition to https://newspring.churchonline.org/api/v1/upcoming_event_times, I also found out about this end point: https://newspring.churchonline.org/api/v1/events/current

It seems to show the next event and whether or not it's live. This might be a good place to start?

PR #175, opened 2018-07-25
#142 - Live Bar connection to ChOP API (WIP)

cdanwards:
…eam api

Co-Authored-By: Isaac Hardy <isaac.hardy@newspring.cc>

Fixes 142

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


conrad-vanl:
After looking into this issue at length, I have decided to punt test coverage for the livestream resolvers onto next sprint. I'm okay merging this temporarily without tests, and heres why:

Up to this point, our testing strategy for API has been to create a `graphql` instance (using facebook's reference `graphql` package) with our schema and our resolver functions independent from Apollo Server, and use that `graphql` instance in our tests. This works great as it allows us to run full integration tests on our schema and resolvers without setting up a server instance to test against.

However, as Apollo Server 2 has matured our current testing strategy has become fundamentally broken. Apollo Server 2 includes more and more functionality baked-in that our tests don't cover. For example, accessing `dataSources` is impossible with our current testing strategy, as well as any other convenience features that Apollo Server includes that we might eventually rely on - such as File Uploads.

Therefore, I believe our best path forward is to re-address our testing strategy to better integrate with Apollo Server 2. This is entirely out of scope from this current issue, and I believe would make a great first task for our new resource that is starting this week (more on that later today 😉). So, my vote is to punt test coverage and this scope of work to the next sprint, and for now, LGTM!

PR #176, opened 2018-07-30
Fix infinite splashscreen in storybook

burkeshartsis:
Fixes infinite splashscreen in storybook 😇😬

## Creator

- [x] Ensure there are no new warnings (in app AND in storybook) 🙈🙉🙊
- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic

### DIFF
--- a/mobile/storybook/storybook.js
+++ b/mobile/storybook/storybook.js
@@ -5,7 +5,6 @@ import {
   configure,
   addDecorator,
 } from '@storybook/react-native';
-import SplashScreen from 'react-native-splash-screen';
 
 import Providers from 'Providers';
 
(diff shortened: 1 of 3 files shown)


mobile/storybook/storybook.js
```diff
@@ -26,6 +27,10 @@ const StorybookUIRoot = getStorybookUI({ port: 7007, onDeviceUI: true });
 // https://github.com/storybooks/storybook/issues/2081
 // eslint-disable-next-line react/prefer-stateless-function
 class StorybookUIHMRRoot extends Component {
+  componentDidMount() {
```
conrad-vanl: Hmmm....should this somehow be in our `Providers` component located in `src/Providers.js` ?

burkeshartsis: Not a bad idea! What is the clear benefit? One solution/implementation in one spot?

conrad-vanl:
@burkeshartsis check out 72b30c5 and let me know what you think. This "hides" the `splashscreen.hide()` function a little bit and keeps the splash screen visible for an undetectable amount of longer time, but it keeps it visible until after our caches have been hydrated, which will prevent any "double rendering" visibility to the user with the app becoming visible before cache has hydrated.

PR #159, opened 2018-07-20
Issue 110 - Add Test Coverage to Query Components

cdanwards:
Fixes 110

## Creator

- [x] Build relevant tests if any apply
- [x] Ensure there are no new warnings (in app AND in storybook)
- [x] Upload an animated GIF showcasing the solution (if it applies)
- [x] Set a relevant reviewer
- [ ] Add a line to the changelog documenting the high-level change this PR introduces

## Reviewer

- [ ] Review CI build results
- [ ] Review function and form on iOS
- [ ] Review function and form on Android
- [ ] Review new stories if any apply
- [ ] Review test coverage: Are all new or changed states tested?
- [ ] Review new test logic
- [ ] Review the changelog entry

Issue #136, opened 2018-07-16
Clean up and flush out discover feed categories


conrad-vanl:
#### Problem Details _(Optional)_

Right now the discover feed only contains 3 categories - Editorial, Sermon Series, Devotion Series. 

Lets flush this out to include:

- Articles (change name from Editorial)
- Stories
- News
- Sermon Series
- Devotion Series

And we should port over boilerplate/example content from NewSpring (or other sources) in doing so.

Issue #135, opened 2018-07-16
A user can create an account from within the app


conrad-vanl:
## TODO

- [ ] Generate design asset to reflect UI

Rough mockup:

![image](https://user-images.githubusercontent.com/103927/42836608-1eedb64e-89c1-11e8-85dd-408bde4b658c.png)

spacing / padding probably isn't right...this is rough. But I think this could be nice.

-----

## User Story

A user wants to easily sign up for a new account from within the app.

#### Problem Details _(Optional)_

* Screenshots, GIFs, or a video
* Any additional information (version, device, etc.)

#### Acceptance Criteria _(Optional)_

* [ ] A user can create an account from within the app
* [ ] Turn the login screen to a two-tab view (similar to NS/Apollos) with Login on one tag, Signup on the other

conrad-vanl:
@IsaacHardy I plan to start building out the UI on this today, hoping I can get to it :-D

My current plan, let me know thoughts:

1. The [auth/index](https://github.com/ApollosProject/apollos-prototype/blob/master/mobile/src/auth/index.js) component will render a tab view with the Sign up and Login forms
1. I'll add an `auth/signup` form similar to [`auth/login`](https://github.com/ApollosProject/apollos-prototype/tree/master/mobile/src/auth/login)
1. [This bit](https://github.com/ApollosProject/apollos-prototype/blob/master/mobile/src/auth/login/index.js#L15-L19) should be taken out of the login form and moved into a client-side graphql mutation in our client-side graphql [resolver](https://github.com/ApollosProject/apollos-prototype/blob/master/mobile/src/store/index.js), so that it can be easily called from both the login and signup components

IsaacHardy:
I think that plan sounds pretty great!

PR #177, opened 2018-07-31
updated to v0.4.0

IsaacHardy:
Updating app and api versions to v0.4.0.


conrad-vanl:
@IsaacHardy starting with 0.4.0, you want to remove the changelog.md file in favor of just keeping up with https://github.com/ApollosProject/apollos-prototype/releases ?

This will require a little more diligence when a release is created - essentially going back through the commits included in the release and copy/pasting the commit title and the PR #.

Requires a little more time at the end, but we don't need to worry about it per PR, and then our releases page can link to the PRs/Commits for each "change" in that release, which is arguably a better changelog. Thoughts?

IsaacHardy:
I really like using that package. Is the tag it generates customizable?

conrad-vanl:
@IsaacHardy good question! We'll have to try it out :)

conrad-vanl:
Looks like we have to follow these commit conventions: https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md

🤔

conrad-vanl:
Here's the actual commit spec website: https://conventionalcommits.org/

I'm not against the idea of adhering to this....but wondering if we have to

IsaacHardy:
To run the script, you will `cd` into the `scripts` directory on root, and run 

`sh release.sh <version_num>`

Where `version_num` equals your next incremented version number. (e.g. 0.4.0).

The script will update both iOS and Android application binary with the correct version, and also update both API and MOBILE directories `package.json` version number.

Finally, the script updates the change log based on specific commit messages found in the git log.

IsaacHardy:
@vinnyjth @richarddubay @burkeshartsis @cdanwards Would like some input from y'all.

I'm updating this PR to contain a new change-log script and release script. The release script merely updates the Android and iOS versions while the change-log will grab our squashed merge commits and generate/update a change-log.

@conrad-vanl and I are wanting feedback on how we think we could organize these messages in a clear and meaningful way.

conrad-vanl:
For more clarity to the above ^, right now what we 've talked about doing is we'll just take the commit messages generated by each PR. Right now, they'd look like this:

```

- 161 Adding individual devos to the home feed (#195) (ef3ba9b)
- 136 Discover Feed (#197) (a7b73db)
- [API] Enable Absolute Local Imports (#194) (fdf08f3)
- Improved splashscreen dismissal (8adfb52)
```

We want to get better at writing these messages. How do we communicate that in our PR template?

IsaacHardy:
@vinnyjth @burkeshartsis @richarddubay @cdanwards Just wanted to tag you all again for feedback.

vinnyjth:
@IsaacHardy I like this approach! This gives me more incentive to have more descriptive names for my PRs.

IsaacHardy:
@vinnyjth and @burkeshartsis Specifically, we are "wanting feedback on how you think we could organize these messages in a clear and meaningful way."

cdanwards:
I think that if we made it so that each PR title is a high level overview of what is going on in the PR it would help a lot. PR's like #155 , #159 , or #150 are clear on what's going on in the PR. One's like #135 or #174 are less clear.

IsaacHardy:
@cdanwards That is actually the topic of discussion. How should we structure those PR titles?

conrad-vanl:
I'll go. My vote would be to structure PR titles similar to user stories, specifying what the result of that PR does. 

> User can see their name and profile picture on the connect screen once logged in

Is a 💯 A++ 👍  PR title.

Something like:

> 182 - Design GraphQL schema for ESV scripture

I would write to be something more like:

> Ability to query for ESV scripture in GraphQL

richarddubay:
Okay, my turn. I don't think this needs to be terribly convention heavy. I think these need to be simple one sentence structures. Start with words that say what you did. Things like `Updated`, `Added`, `Removed`, `Changed`. Then add what you changed.

> Added individual devotionals to the home feed.

I think that is as hard as this needs to be.

I did notice that some PR's tend to wrap more than one `feature` or major change in them. Is it feasible to think that we could break up big changes into multiple PR's in order to get more specific about what is changing? Perhaps we break them up into features? For example, #197 actually has to important things in it. It (1) changed the sorting order of individual feeds to be in newest first order, and it (2) added topics to the discover feed. But using that as a PR title is kind of long. That PR should probably have been two PR's. 

So instead of:
> #136 Discover Feed

or even:
> Changed the sorting order of individual feeds to be in newest first order and added News and Stories topics to the discover feed.

Maybe this becomes two PR's:

- > Changed the sorting order of individual feeds to be in newest first order.
- > Added News and Stories topics to the Discover feed.

I guess maybe that just means that we need to make sure that our issues are broken down into individual features and do not encompass more than one major change? FWIW, I think I added the sorting order change to the PR mentioned above because I noticed it was not the correct order and just thought I'd fix it while I was already in that code. So that one is one me ... maybe this isn't really an issue?

conrad-vanl:
@richarddubay for your #136 Discover Feed example, I think it could also just be:

> Improved content feeds by adding new sections and adjusting order of items

If we need more details (like specifying _what_ the new sections are, or what the _new_ order is) we can do that in the commit body (when we merge the PR we can tweak the PR commit message body as well)

IsaacHardy:
@conrad-vanl updated PR template

Issue #167, opened 2018-07-20
Sibling content items


conrad-vanl:
## User Story

A user can browse a content item's siblings from the content-single view.

For example, you're viewing a sermon and you can see the other sermons in the same series

#### Problem Details _(Optional)_

* Screenshots, GIFs, or a video
* Any additional information (version, device, etc.)

#### Acceptance Criteria _(Optional)_

* [ ] It should [SPEC_DETAIL]

PR #199, opened 2018-08-03
Sibling Content api

vinnyjth:
References #167 

Prior to this PR, the only way to get sibling items from a `ContentItem` was to fetch the parent and then go back down the tree to the children. This suffered from a few problems, one of them being that the source item would returned in the list "siblings" (requiring additional filtering on the client) 

This PR exposes a "siblingContentItemsConnection` endpoint on `ContentItems` that fetchs all `ContentItems` with the same direct parent as the current item. This includes items with an `order` both before and after the current item. This is a fairly expensive query, requiring three different requests to rock, but no N+1 requests. I would be happy to hear if anyone has any ideas to optimize this query further :) 

I included tests over the `resolver` and `model` which following the existing testing setup. 

# Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


api/src/data/content-items/model.js
```diff
@@ -43,6 +43,36 @@ export default class ContentItem extends RockModel {
     return request.orderBy('Order');
   };
 
+  getCursorBySiblingContentItemId = async (id) => {
```
IsaacHardy: I can't find any improvements as of now on this method, however, I'd like us to speak with Jon E from ROCK to see if there is anyway to implement this functionality within rock itself. Maybe even as a plugin, because I feel that this query might be to burdensome as it currently stands.

@conrad-vanl What are your thoughts?

conrad-vanl: I need to spend some time looking at the code in this file, but from my perspective, the resulting schema is exactly what I had in mind, I believe, when this issue was started. 

I also don't see how this could be implemented in the Rock API directly in an unobtrusive fashion

IsaacHardy: Aye. I'm gonna go with your understanding of this functionality and approve this PR then. I was hoping for a better way.

IsaacHardy:
@vinnyjth  Before we merge, do you have an example query I can check?

api/src/data/content-items/model.js
conrad-vanl: I wasn't disagreeing with your assessment of the complexity of the integration with Rock. Just more looking from the perspective of the api schema we end up with

vinnyjth:
Yep! @IsaacHardy 
```
query getContent {
  node(id:"UniversalContentItem:f1184b518af2c2607b4e580e777d0f57") {
    ... on ContentItem {
      id
      title
      siblingContentItemsConnection {
        edges {
          node {
            id
            title
          }
        }
      }
    }
  }
}
```

conrad-vanl:
@IsaacHardy if the graph schema is what you're concerned with, this "connection" paradigm does follow the same paradigm we're using for child items, and I believe follows Relay's spec for pagination, and Graphql's recommendation for more advanced pagination models: https://graphql.org/learn/pagination/

IsaacHardy:
@conrad-vanl Thanks for the link!

IsaacHardy:
@vinnyjth LGTM!

conrad-vanl:
@IsaacHardy I threw the don't merge on here on purpose :)

I had some questions on a gotcha I see with this, which I'll add to this comment.

IsaacHardy:
@conrad-vanl Ah! My bad!

conrad-vanl:
@IsaacHardy no worries 😄 

I updated my comment above 😉

vinnyjth:
@conrad-vanl Makes sense :D I think if we relied on the `parentChannel` attribute, we could also ditch one of the queries in this PR (or at least simplify it). I'll test it out and see if I can identify any noticeable changes in my test queries. Still would love to hear from @IsaacHardy  or @richarddubay about their expectations around content with multiple parents.

conrad-vanl:
@vinnyjth what we want to be careful with though is we _don't_ want to query for siblings that don't have matching parent items. For example, we don't want to get sermons from a _different_ series...

My decision to _not_ include a `parentItem` connection to the schema thus far falls under this same reason...I'm not sure how to handle multiple parent scenarios, as I don't know when an item might have multiple parents.

vinnyjth:
@conrad-vanl after digging into the weeds for a bit, I don't `parentChannel` is different than the `parent` I find when grabbing `contentChanels` via `ChildContentChannelItemId`. Beyond that, if you go  `parentChannel` -> ` `childContentItems` you return an array of all items in the __parents__ parent channel. 

Thus - I think grabbing parents via  `ChildContentChannelItemId` is more specific than `parentChannel`, from what I can see right now. Would love to hear from @issachardy or @richarddubay if they have ideas :D

vinnyjth:
Talked with @conrad-vanl, safe to merge :) This PR contains the intentional functionality given the data we know about.

PR #203, opened 2018-08-07
Added videoUrl to Video Content type

vinnyjth:
Fixes [issue-number]

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


api/src/data/media/index.js
```diff
@@ -26,6 +32,7 @@ export const schema = gql`
     sources: [VideoMediaSource]
     # duration: Float
     embedHtml: String
+    videoUri(format: VIDEO_FORMATS): String
```
conrad-vanl: I'm fine with going this direction, but we might want to consider back tracking on the `sources` property above, or adding this `format` argument to that line instead....thoughts?

Originally i was thinking we would like have multiple `sources` for a Video....and that `sources` prop could take care of presenting those sources somehow

Having two props that exposes urls is a miss in my book...and I don't care about backwards compatibility with the schema at this point in time.

conrad-vanl: the point behind using `sources: Array` _originally_ came from a desire to maintain a consistent format used for Video, Audio, Images, etc. For images, it's common to deliver multiple resolutions to the client as React-Native's `<Image/>` component can take an array of sources, like:

`<Image source={[{ uri: 'source-1' }, { uri: 'source-2' }]} />` and RN works out the best image to use based on the device's screen size and what not.

Granted, I agree a single `uri` is likely much simpler then an array. But I think if we go that route, I'd want to go that route for all media type, and not just video.

Furthermore, right now, these high-level resolvers `images`, `video`, `audio`, are fairly abstract, and return _all_ of the images, video, or audio available for a content item. If we turned these into only returning a single resource, I'd want to do something simliar to what we did with the `coverImage` property - and describe _what_ that media is specifically and how it should be used....as it is very likely that content will always have multiple media of the same type. In retrospective, this `embedHTML` attribute is probably also not good.

conrad-vanl: I could see a world where this attribute returns an array by default, but if you specify the format to use, it only returns a single item...or something like that....  🤔

conrad-vanl: A sermon might have both a `sermonTrailer` and a `sermonVideo`... or they might have a `seriesImage`, `sermonThumbnail`, `speakerThumbnail`, etc....

vinnyjth: Yah - it makes perfect sense to have some consistency between each of the  different `Contents` that can be returned. So do you think it makes sense to have a singular `video` or `mainVideo` attribute that would function similar to the `coverImage`?

burkeshartsis: ### TL;DR
I vote for a filterable array of sources for videos. This logic, at least at present time, does not need to extend to any other media types (i.e. images).

#### Tell me more!
Having thought through this a bit in a few different contexts I believe you both have emphasized important aspects. First and foremost, I believe media types while similar do not need to be identical. What is required to render an image in the best possible experience is different than that of a video. I also believe we shouldn't reinvent any logic that already exists for us in one place somewhere else. For example, React Native's `Image` accepts and array of sources and does the complex task of figuring out which image is best. At this time there is no need to move that logic to the server. Video, however, is a different story. Each platform supports different formats and encodings with very little overlap and different preferred/optimized formats. In this case, I believe we are forced to build this logic into the server with a schema that supports it. To that end, I believe all media related queries should return an array of sources. It makes senes for consistency and flexibility. Video assets demand the ability to filter that array. Down the road, we may benefit from being able to filter image assets as well in order to return platform optimized formats like webM to Android for a better user experience and to lower bandwidth costs. But that is certainly not a priority at this juncture. 

Hope this somewhat coherent thoughts are relevant enough to be useful.

conrad-vanl: So....it sounds like based on @burkeshartsis thoughts that we'll keep the `sources` property that returns an array, but add an argument to it (for video only) that allows us to filter it.

If I properly distilled his thoughts above, I like that.

vinnyjth:
@conrad-vanl @burkeshartsis 

Awesome discussion here :D I'm going to hold off writing any code, as there's really not a need right now :D The URLs we have from Wistia support both platforms. In the future we can revisit the schema ideas in the PR, and implement the "`source` URL filtering via argument" strategy

PR #214, opened 2018-08-10
Convert Rock Data Source to ApolloDataSourceREST

vinnyjth:
Fixes #200 

This PR removes our custom REST data source, and replaces it with the offically provided ApolloRESTDataSource. Some advantages include... 

- Better caching support
- Less dependency on "boilerplate"  code that we need to maintain. 
- More in line with Apollo's best case scenario setup. 

Some caveats 

- Testing is a little more difficult.
- Needed to do some hacky things to enable stuff like returning a `Response` object (rather than returning `response.json()` 

**Some notes about this PR**
To ensure that the functionality of the app didn't change, I confirmed that **no resolver snapshot tests were changed**. 
Almost all of the model snapshot tests changed, but only because the way I captured the results changed. If you compare the old snapshots to the new snapshots, you should note that they are functionally equivalent (but slightly different data is captured slightly differently). 

API coverage percentage went down. This is because I removed several large files that had 100% coverage. The cover is slightly better compared to master, on a line by line basis. 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [x] Review CI build results
- [x] Review new test logic
- [x] Review the change-log entry
- [x] Review that the PR title adheres to our patch-notes/change-log style


api/src/connectors/rock/data-source.js
```diff
@@ -0,0 +1,79 @@
...
+  didReceiveResponse(response, request) {
+    // Can't use await b/c of `super` keyword
+    if (request.method === 'GET') {
+      return super
+        .didReceiveResponse(response, request)
+        .then((parsedResponse) => this.normalize(parsedResponse));
+    }
+    // Shortcut the parsing, we need the headers.
+    // Should be a better, cleaner way to do this...
+    return response;
```
conrad-vanl: 😕

Would it be better to ditch the parsing? So the return type is consistent? Or in the `GET` instance normalize the response, but still return the full response object?

🤔 This seems like a smell just because it can create a gotcha later

api/src/data/auth/data-source.js
```diff
@@ -78,11 +62,12 @@ export default class AuthModel extends RockModel {
     try {
       const { email } = props;
 
-      return await this.context.connectors.Rock.post('/People', {
+      const response = await this.post('/People', {
         Email: email,
         IsSystem: false, // Required by Rock
         Gender: 0, // Required by Rock
       });
+      return response.json();
```
conrad-vanl: And the result of the gotcha as I bet this object type is `PascalCase` data vs `camelCase`

api/src/connectors/rock/data-source.js
vinnyjth: @conrad-vanl I spent a LONG time trying to figure out the right way to do this. I like the idea of ditching the parsing, but then we need to run the `normalizer` EVERYWHERE which is very obnoxious. I also thought about normalizing, but including the original response, but then we need to deconstruct everywhere we call `get` and that's annoying as well!

For reference, the ApolloRESTDataSource doesn't return a `Response` object by default, and there's no config to change that behavior. It always returns the parsed JSON. I think best case scenario is we find a way to configure if a Response should be sent back via call to `post`, `get` etc. So you could do `get({withResponse: true})` or even `getWithResponse()`. 

All of those approaches are difficult to implement, mostly because the `Request` object created prior to the request isn't the exact same `Request` object that comes back in the function we need to short circuit.

All that said, I'll look into it further and find something better before this PR is merged :D

conrad-vanl: Remind me - we only need the response object for reading headers in the event of grabbing the cookie, right? Could we somehow just.....do that here instead? and still return the parsed json? Might be a little icky but less icky then this?

vinnyjth: Yah - I'm thinking that's a great option :D So far here is the only place where we actually need access to the Headers

vinnyjth:
@conrad-vanl Changes implemented!

PR #227, opened 2018-08-17
Complete flow for Liking Content

vinnyjth:
Fixes [issue-number]

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic

### DIFF
--- a/api/src/connectors/rock/constants.js
+++ b/api/src/connectors/rock/constants.js
@@ -7,6 +7,9 @@ export const {
 
 export const GET_IMAGE = url.resolve(ROCK_API, 'GetImage.ashx');
 
+export const APP_INTERACTION_CHANNEL = 9;
+export const CONTENT_ITEM_TYPE_ID = 208;
+
 export const FIELD_TYPES = {
   TEXT: 1,
   MULTI_SELECT: 2,
--- a/api/src/connectors/rock/rock-constants.js
+++ b/api/src/connectors/rock/rock-constants.js
@@ -1,0 +1,96 @@
+/* eslint-disable no-console */
+import RockApolloDataSource from '/api/connectors/rock/data-source';
+
+const mapApollosNameToRockName = (name) => {
+  switch (name) {
+    case 'ContentItem':
+    case 'UniversalContentItem':
+      return 'ContentChannelItem';
+    default:
+      throw new Error(`${name} has not been mapped into a Rock type!`);
+  }
+};
+
+const RockConstantCache = new Map();
+export default class RockConstants extends RockApolloDataSource {
+  async findOrCreate({ model, objectAttributes }) {
+    // Turns {ChannelId: 7, Name: 'Something'} into '(ChannelId eq 7) and (Name eq 'Something')'
+    const filter = Object.keys(objectAttributes)
+      .map((key) => {
+        if (typeof objectAttributes[key] === 'string') {
+          return `(${key} eq '${objectAttributes[key]}')`;
+        }
+        return `(${key} eq ${objectAttributes[key]})`;
+      })
+      .join(' and ');
+
+    const objects = await this.request(model)
+      .filter(filter)
+      .get();
+    if (objects.length) {
+      console.log('Found', objects[0]);
+      return objects[0];
+    }
+    const objectId = await this.post(`/${model}`, objectAttributes);
+    const ret = await this.get(`/${model}/${objectId}`);
+    console.log('Created', ret);
+    return ret;
+  }
+
+  async createOrFindInteractionComponent({ componentName, channelId }) {
+    return this.findOrCreate({
+      model: 'InteractionComponents',
+      objectAttributes: { Name: componentName, ChannelId: channelId },
+    });
+  }
+
+  async createOrFindInteractionChannel({ channelName }) {
+    return this.findOrCreate({
+      model: 'InteractionChannels',
+      objectAttributes: { Name: channelName, UsesSession: true },
+    });
+  }
+
+  async interactionChannel() {
+    if (RockConstantCache.has('interactionChannel')) {
+      return RockConstantCache.get('interactionChannel');
+    }
+    const channel = await this.createOrFindInteractionChannel({
+      channelName: 'Apollos App',
+    });
+    RockConstantCache.set('InteractionChannel', channel);
+    return channel;
+  }
+
+  async interactionComponent() {
+    if (RockConstantCache.has('InteractionComponent')) {
+      return RockConstantCache.get('InteractionComponent');
+    }
+    const channel = await this.interactionChannel();
+    const component = await this.createOrFindInteractionComponent({
+      componentName: 'Apollos Content Item',
+      channelId: channel.id,
+    });
+    RockConstantCache.set('InteractionComponent', component);
+    return component;
+  }
+
+  async modelTypeId(nameInput) {
+    const name = mapApollosNameToRockName(nameInput);
+    const cacheKey = `type:${name}`;
+
+    if (RockConstantCache.has(cacheKey)) {
+      return RockConstantCache.get(cacheKey);
+    }
+    const types = await this.request('EntityTypes')
+      .filter(`Name eq 'Rock.Model.${name}'`)
+      .get();
+    if (types.length) {
+      // eslint-disable-next-line prefer-destructuring
+      RockConstantCache.set(cacheKey, types[0]);
+      return types[0];
+    }
+
+    return null;
+  }
+}
--- a/api/src/data/content-items/index.js
+++ b/api/src/data/content-items/index.js
@@ -34,6 +34,7 @@ export const schema = gql`
     parentChannel: ContentChannel
 
     theme: Theme
+    interactions: [Interaction]
   }
 
   type UniversalContentItem implements ContentItem & Node {
@@ -56,6 +57,7 @@ export const schema = gql`
     terms(match: String): [Term]
 
     theme: Theme
+    interactions: [Interaction]
   }
 
   type Term {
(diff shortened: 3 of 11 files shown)


api/src/data/interactions/schema.js
```diff
@@ -0,0 +1,28 @@
...
+    operation: INTERACTION_OPERATION!
+    interactionDateTime: String!
+  }
+
+  enum INTERACTION_OPERATION {
+    Like
+  }
+
+  input CreateInteractionInput {
+    contentId: String!
```
conrad-vanl: Two comments....

1. Should this be of type `ID` and not `String`?
2. It looks like you tried to make the `InteractionInput` somewhat abstract by using the `INTERACTION_OPERATION ` enum, but then include a `contentId` that's required, which says to me that all future interaction operations must have a `contentId`. Would it make more sense to either a) not make this required, or b) generalize this? perhaps `nodeId`? (For future reference: users will be able to like other things besides content in the future. For example, groups)

api/src/data/interactions/schema.js
```diff
@@ -0,0 +1,28 @@
...
+    sessionId: String!
+    operation: INTERACTION_OPERATION!
+  }
+
+  extend type Mutation {
```
conrad-vanl: is this `extend` syntax new? Never seen it before...this is awesome!

api/src/data/interactions/schema.js
```diff
@@ -0,0 +1,28 @@
+import { gql } from 'apollo-server';
+
+export default gql`
+  type Session implements Node {
```
conrad-vanl: How does the frontend need to interact with `Session`? Are you expecting it to call `createSession` every time there's a "new" session - as in - the user leaves the app and comes back?

Just making sure we shouldn't offload more logic to the backend, and use like a `getOrCreateCurrentSession` type mutation.....I think the frontend can better handle this, just wondering.

api/src/data/interactions/data-source.js
```diff
@@ -0,0 +1,53 @@
+import { AuthenticationError } from 'apollo-server';
+import RockApolloDataSource from '/api/connectors/rock/data-source';
+import { parseGlobalId } from '/api/data/node';
+
+export default class Interactions extends RockApolloDataSource {
```
conrad-vanl: We might consider documenting this flow out into something we can eventually send over to the Rock guys to confirm we're using this correctly

api/src/connectors/rock/constants.js
```diff
@@ -7,6 +7,9 @@ export const {
 
 export const GET_IMAGE = url.resolve(ROCK_API, 'GetImage.ashx');
 
+export const APP_INTERACTION_CHANNEL = 9;
+export const CONTENT_ITEM_TYPE_ID = 208;
```
conrad-vanl: I couldn't find these constants anywhere else in the changelog...am I missing them?

api/src/data/index.js
```diff
@@ -21,10 +23,11 @@ const data = {
   LiveStream,
   Theme,
   ESVScripture,
+  Interactions,
+  RockConstants: { dataSource: RockConstants },
```
conrad-vanl: Should this be exposed as a `dataSource` or as a global singleton? Ie....cached per-request or cached once globally?

api/src/data/content-items/index.js
```diff
@@ -34,6 +34,7 @@ export const schema = gql`
     parentChannel: ContentChannel
 
     theme: Theme
+    interactions: [Interaction]
```
conrad-vanl: For the current user story around `liking`, I wonder if this is the right schema to use. Can't help but feel something more basic like `isLiked: Bool` or something wouldn't better fit the story.

If we kept it like this, I think I'd atleast want a way to filter the results by an `INTERACTION_OPERATION`?

`interactions(type: INTERACTION_OPERATION): [Interaction]`

conrad-vanl: While I'm here...the other thing I'm thinking about...is a like is a toggle-state interaction. You either like something, or you don't. You don't add a like interaction something you already liked, but you can unlike it.

conrad-vanl: So I don't know if that means we need to have both `Like` and `Unlike` interaction types, or we need to make interactions more hidden in the API, and create a more basic `like` mutation?????

FWIW, I like the flexibility the interaction schema provides....

conrad-vanl: But either way, an `isLiked: Bool` attribute on the schema maybe the nicest thing for the client to work with

api/src/connectors/rock/rock-constants.js
```diff
@@ -0,0 +1,96 @@
...
+  switch (name) {
+    case 'ContentItem':
+    case 'UniversalContentItem':
+      return 'ContentChannelItem';
+    default:
+      throw new Error(`${name} has not been mapped into a Rock type!`);
+  }
+};
+
+const RockConstantCache = new Map();
```
conrad-vanl: We haven't really talked about a caching strategy yet... I like this, but I wonder what edge-cases it might create. For instance, we will need to create some sort of caching / optimization strategy for the other DataSources as well, and Apollo does have a bit of english on the subject ---

https://blog.apollographql.com/automatic-persisted-queries-and-cdn-caching-with-apollo-server-2-0-bf42b3a313de

https://blog.apollographql.com/apollo-server-2-0-performance-and-error-reporting-built-in-baa6158f3c09

https://blog.apollographql.com/easy-and-performant-graphql-over-rest-e02796993b2b

api/src/connectors/rock/constants.js
vinnyjth: Need to remove these, thanks!

api/src/connectors/rock/rock-constants.js
vinnyjth: @conrad-vanl Totally agree that we should rely more on Apollo's built in caching :D I think this is something we can explore further, and may require some collaboration with the Rock team.

From what I discovered, there's two ways of caching that apply to ApolloServer. 

1. Schema level caching (cache whole parts of the GQL schema). This uses Apollo Engine, and doesn't apply to this context (we're not using the schema to get at this data). 

2. The DataSource's internal cache. This is the cache we **should** be using. It's not being used in this PR because (as far as I can tell) the response we are getting back from Rock is not cacheable because it doesn't come over with a TTL header. However, I think that's something we can bypass by setting the `ttl` option in request options. **Setting TTL by hand is totally undocumented** but I see no reason not to use it :) 
https://github.com/apollographql/apollo-server/blob/master/packages/apollo-datasource-rest/src/RESTDataSource.ts#L23

api/src/data/content-items/index.js
vinnyjth: I think having an `isLiked: Bool` on the schema is a great option :D 

As for figuring out how to tell if something is liked or unliked, I can count the number of like/dislike operations and determine which there is more of. Kinda hacky, but I think that's the best solution.

api/src/data/index.js
vinnyjth: All `DataSources` are instantiated per request, but the cache is persisted across multiple requests.  So I think we're doing it properly here. I experimented with setting it up as a singleton, but writing it as a singleton broke a lot of the advantages provided by ApolloRESTDataSource

api/src/data/interactions/data-source.js
vinnyjth: 👍

api/src/data/interactions/schema.js
vinnyjth: I was hoping a client would be responsible for creating sessions, and then keeping track of it's own `sessionId`. The client would also be responsible for "forgetting" a sessionId and creating a new session when it needed too. 

I think the mutations we need depend on what count as a "session" though. Is it based on length of time? Opening/closing the app? Login/logout?

api/src/data/interactions/schema.js
vinnyjth: 1. 👍 
2. I'll swap `contentId` for `nodeId`. :)

api/src/data/interactions/schema.js
vinnyjth: Yah! One of my `todos` is to move all our root fields to use the `extend` syntax :)

api/src/data/interactions/schema.js
conrad-vanl: > I think the mutations we need depend on what count as a "session" though. Is it based on length of time? Opening/closing the app? Login/logout?

The obscurity in this and potential complexity is why I was thinking it could be suitable as a backend responsibility....but still not sure

api/src/connectors/rock/rock-constants.js
conrad-vanl: ah, interesting. Gotcha. Would it be better to set a `ttl` and go that route then? Less complexity on our end?

And opposed to `ttl`s on these Rock resources coming from Rock....I'm not really sure if that would make sense for them to add, as these aren't really things that have a defined ttl. Almost feel like `ETag` headers would be a better solution if we were going to ask for something from the Rock team....

api/src/data/interactions/schema.js
vinnyjth: Sounds good. What if right now it just returns a new session every time?

packages/apollos-church-api/src/connectors/rock/rock-constants.js
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable no-console */
-import RockApolloDataSource from '/api/connectors/rock/data-source';
```
conrad-vanl: this is probably an instance where it makes more sense to use a relative import

vinnyjth: 👍

api/src/connectors/rock/rock-constants.js
vinnyjth: I implemented a way to pass `ttl` to ApolloRESTDataSource :)

api/src/data/content-items/index.js
IsaacHardy: @vinnyjth Are you able to change the category name of an interaction? If so, is it possible to "unlike" something by simply changing that category?

PR #246, opened 2018-08-28
Fix now deploys

### LINKED ISSUE #241: [Deployment] Fix Deployments caused by Package Restructure
- `Now` builds are deploying using Nodemon. We should probably just use `node`
- Appcenter doesn't pick the right `package.json`
- Possibly another `now` error?

vinnyjth:
Fixes #241 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,9 @@
     ]
   },
   "scripts": {
-    "start": "./node_modules/.bin/lerna run start --parallel",
+    "start": "if test \"$NOW_ENV\" = \"true\" ; then yarn start:now ; else yarn start:dev ; fi",
+    "start:now": "./node_modules/.bin/lerna run start --scope apollos-church-api",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -8,7 +8,9 @@
     ]
   },
   "scripts": {
-    "start": "./node_modules/.bin/lerna run start --parallel",
+    "start": "if test \"$NOW_ENV\" = \"true\" ; then yarn start:now ; else yarn start:dev ; fi",
+    "start:now": "./node_modules/.bin/lerna run start --scope apollos-church-api",
```
IsaacHardy: This is more of a question about how lerna works, but when you run this, are the `yarn start` commands being ran from within all directories under the `packages` folder?

vinnyjth: Nope! But only because I include the `--filter`. Otherwise they are :)

PR #255, opened 2018-08-28
Added Devotional Content Type / Add ability to query scripture HTML from ContentItem

### LINKED ISSUE #218: Add GraphQL endpoint for Scripture attribute from Rock.
Make sure our GraphQL server can bring over the Scripture attribute from Rock. May be more to this issue.

vinnyjth:
Fixes #218 

Adds a new Content Type, `DevotionalContentItem` which contains a scripture field (which pulls from the scripture field in Rock. On the front end, you will be able to check if a content item is a `UniversalContentItem` or a `DevotionalContentItem` by checking the `__typename` field. 

You can query for features of the `DevotionalContentItem` on the front end like so...
```
{
   __typename
  _id
  name
  (etc)
  body
  ... on DevotionalContentItem {
    scriptures
  }
}
```

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


conrad-vanl:
I think this looks good to me. Any possibility on the code cov?

vinnyjth:
@conrad-vanl It's the darn `__typename` resolver :/ I don't really understand why it's not getting called when I expect, so there's no easy way to fix the coverage that I see right now.

richarddubay:
@vinnyjth & @conrad-vanl So ... not sure if it matters or not, but the URL for the API.Bible `GET` is different than the one that we were using for the ESV. Specifically, where before we were passing something like `Song+1:1-4` or `Song+of+Solomon+1:1-4` as part of the URL to the ESV, the API.Bible structure wants something like `SNG.1.1-SNG.1.4`. Not sure if that plays into any of this that you're doing here, but I thought I should mention that. In particular, if we want people to be able to type in "Song of Solomon 1:1-4" in their CMS ... how do we translate that to "SNG.1.1-SNG.1.4"?

vinnyjth:
Great question @cdanwards. I'm digging through the api.bible to see if they have an endpoint where you can resolve a scripture query term into the format they expect.

If not, this is a function I wrote a while back to parse a reference into a JS object with the data split out. We would then need to map the Song of ... into SNG. Maybe via a lookup table of some sort?
```
// 'John 3:16-17'
// -> [{ book: 'John', chapter: 3, verse: 16}, { book: 'John', chapter: 3, verse: 16}, ]
// eslint-disable-next-line
export function getVersesFromStringReference(reference) {
  const matches = reference.match(/(^[\d+\s]*[\w+\s]+)\s(\d+):(\d+)(?:-(\d+))?(?:\s(\w{3}))?/);

  if (!matches || !matches[1] || !matches[2] || !matches[3]) {
    console.error(`${reference} is not a valid verse reference.`);
  }

  const book = matches[1];
  const chapter = parseInt(matches[2], 10);
  const verse = parseInt(matches[3], 10);
  const translation = matches[5] || 'NIV';
  let verseRange = [verse];

  if (matches[4] && parseInt(matches[4], 10) > verse) {
    const endVerse = parseInt(matches[4], 10);
    const rangeLength = endVerse - verse;
    // build a range of verses between the start verse and the end verse.
    // eg Mathew 2:10-15 builds [10, 11, 12, 13, 14, 15];
    // I chose not to use lodash.range to reduce dependencies.
    verseRange = [...Array(rangeLength + 1).keys()].map(i => i + verse);
  }
  return verseRange.map(v => ({ book, chapter, verse: v, translation }));
}
```

conrad-vanl:
@vinnyjth you're tagging @cdanwards but I think you mean to be tagging @richarddubay

vinnyjth:
Fixing this to work with new API.

vinnyjth:
not sure what's going on here - might need to trigger a rebuild

conrad-vanl:
@vinnyjth try merging it with origin master and pushing up. I think I've ran into this issue before and recall that fixing it

vinnyjth:
@conrad-vanl Already upto date :/ I think we'll need to merge something else in first. Rebuilding in Jenkins did not fix our issue.

conrad-vanl:
> Build config did not create any jobs.
https://travis-ci.org/ApollosProject/apollos-prototype/requests

trying to remember how i fixed this

richarddubay:
What README is complete without a 🍕 emoji? ❤️ this. 😉

PR #178, opened 2018-07-31
create a protected action handler (#85)

conrad-vanl:
Fixes #85 

## Creator

- [x] Build relevant tests if any apply
- [x] Ensure there are no new warnings (in app AND in storybook)
- [x] Upload an animated GIF showcasing the solution (if it applies)
- [x] Set a relevant reviewer
- [x] Add a line to the changelog documenting the high-level change this PR introduces

## Reviewer

- [x] Review CI build results
- [x] Review function and form on iOS
- [x] Review function and form on Android
- [x] Review new stories if any apply
- [x] Review test coverage: Are all new or changed states tested?
- [x] Review new test logic
- [x] Review the changelog entry

PR #257, opened 2018-08-30
Implemented ability to like a piece of content through an interaction Visually

### LINKED ISSUE #244: [Development] Inject ability to like a piece of content within the already existing image cards

* [ ] A user can like a piece of content within the app.
* [ ] A user can unlike a piece of content within the app.

IsaacHardy:
Fixes #244 

This PR gives basic UI functionality to like and unlike a piece of content.

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


IsaacHardy:
simulator screen shot - iphone 8 plus - 2018-08-30 at 11 36 27

IsaacHardy:
simulator screen shot - iphone 8 plus - 2018-08-30 at 11 40 39

IsaacHardy:
simulator screen shot - iphone 8 plus - 2018-08-30 at 11 41 46

packages/apolloschurchapp/src/ui/Like/Like.tests.js
```diff
@@ -0,0 +1,38 @@
...
+import renderer from 'react-test-renderer';
+
+import { ThemeProvider } from 'apolloschurchapp/src/ui/theme';
+
+import Like from '.';
+
+describe('the Share component', () => {
+  it('should render a Like', () => {
+    const tree = renderer.create(
+      <ThemeProvider>
```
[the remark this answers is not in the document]
IsaacHardy: Got it!

packages/apolloschurchapp/src/auth/login/index.js
```diff
@@ -5,18 +5,31 @@ import * as Yup from 'yup';
...
+import getSessionId from '../getSessionId';
 
 import authenticateMutation from './authenticate';
 import LoginForm from './Form';
 
 const Login = ({ onLogin }) => (
   <Mutation
     mutation={authenticateMutation}
-    update={(cache, { data: { authenticate } }) => {
+    update={(
```
conrad-vanl: Question here: would like feedback from both @vinnyjth and @IsaacHardy on this:

So essentially this component is our `Login` component. We also have a `Signup` component that doesn't appear to be covered under this functionality (a user signs up, but no session is created).

This leads me to believe that we should move this state maintenance (creating the session) out of our view-logic component, and into our state store (src/store/index). Potentially creating a new mutation - like `handleLogin` or something that both stores the authentication token, and fires a `createSession` mutation and stores the result of that as well....

But wondering what else we should be considering here....

packages/apolloschurchapp/src/content-single/ActionContainer.js
```diff
@@ -0,0 +1,73 @@
...
+                    },
+                  }
+                ) => {
+                  cache.writeQuery({
+                    query: getContentItem,
+                    data: { isLiked: operation === 'Like' },
+                  });
+                }}
+              >
+                {(createSession) => (
```
conrad-vanl: so should this function `createSession` really be named like `createInteraction` or `like` or something? It's not actually creating a session, yeah? It's what's triggering the like

packages/apolloschurchapp/src/content-single/ActionContainer.js
```diff
@@ -0,0 +1,73 @@
...
+                  <Like
+                    itemId={itemId}
+                    sessionId={sessionId}
+                    isLiked={isLiked}
+                    operation={isLiked ? 'Unlike' : 'Like'}
+                    toggleLike={async (variables) => {
+                      try {
+                        await createSession({ variables });
+                        await refetch();
```
conrad-vanl: Is this line required? That `cache.writeQuery` above setting `isLiked` should handle this, without another network request

packages/apolloschurchapp/src/content-single/ActionContainer.js
```diff
@@ -0,0 +1,73 @@
...
+
+import createInteraction from './createInteraction';
+import getContentItem from './getContentItem';
+
+const ActionContainer = ({ content, itemId }) => (
+  <SideBySideView>
+    <Query query={getSessionId} fetchPolicy="cache-only">
+      {({ data: { sessionId } }) =>
+        sessionId ? (
+          <Query query={getContentItem} variables={{ itemId }}>
```
conrad-vanl: This is querying for too much data, I'd vote for this component to use a different query that just gets `isLiked`

packages/apolloschurchapp/src/content-single/index.js
```diff
@@ -28,9 +28,10 @@ const ContentContainer = styled({ paddingVertical: 0 })(PaddedView);
 class ContentSingle extends PureComponent {
   static navigationOptions = ({ navigation }) => {
     const shareObject = navigation.getParam('sharing', 'Content');
+    const itemId = navigation.getParam('itemId', []);
     return {
       title: shareObject.title,
-      headerRight: <Share content={shareObject} />,
+      headerRight: <ActionContainer itemId={itemId} content={shareObject} />,
```
conrad-vanl: is it possible to get this `shareObject` from a query as well?

packages/apolloschurchapp/src/ui/FeedItemCard/index.js
```diff
@@ -52,14 +52,12 @@ const FeedItemCard = enhance(
...
         />
-        {/* <LikeButton id={id}>
-          <Icon
-            name={isLiked ? 'like-solid' : 'like'}
-            size={theme.helpers.rem(1.2)}
-            fill={theme.colors.text.primary}
-            isLoading={isLoading}
-          />
-        </LikeButton> */}
+        <Icon
```
conrad-vanl: So how hard would it be to abstract the `Like` button inside of `ActionContainer` into it's own `Like` component that could be reused here?

packages/apolloschurchapp/src/ui/Like/index.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { compose, pure } from 'recompose';
+import Icon from 'apolloschurchapp/src/ui/Icon';
+import Touchable from 'apolloschurchapp/src/ui/Touchable';
+import { withTheme } from 'apolloschurchapp/src/ui/theme';
+
+const enhance = compose(
```
conrad-vanl: You could bring the queries/mutations for liking into this component instead of keeping them inside of the `ActionContainer` component (probably could do the same thing with the `Share` component)

packages/apolloschurchapp/src/content-single/ActionContainer.js
IsaacHardy: Good catch.

packages/apolloschurchapp/src/content-single/ActionContainer.js
IsaacHardy: That is a good suggestion.

packages/apolloschurchapp/src/content-single/ActionContainer.js
IsaacHardy: Without this `refetch()`, the cache does not update the the view when 'liking' a piece of content.

packages/apolloschurchapp/src/content-single/index.js
IsaacHardy: I actually think it is and didn't think about this before. Let me look into seeing if i can implement that.

packages/apolloschurchapp/src/ui/FeedItemCard/index.js
IsaacHardy: If that is the user experience we want, then I can def look to see what the implications are for doing this.

packages/apolloschurchapp/src/ui/Like/index.js
IsaacHardy: We def could do this, but wouldn't it go against our current `ui` folder structure? I don't think we are placing the queries/mutations in any other component.

packages/apolloschurchapp/src/auth/login/index.js
IsaacHardy: I like this

conrad-vanl:
@IsaacHardy something else I just thought about on this, I think we might need to merge #178 for this? Essentially that PR gives you the ability to let a user tab on the like button when they're not logged in, it'll open the auth modal, they login, then it goes back and triggers the like. Thoughts?

IsaacHardy:
If that is the user experience we want, then I think you are correct that that PR is needed.

packages/apolloschurchapp/src/auth/login/index.js
IsaacHardy: It is quite possible that this functionality will require additional API work. I'll need @vinnyjth to que in on this but it seems the only way to implement this type of functionality would be to create a mutation that exists as both a "signup" and "login" mutation.

If that is the case, I recommend that we added the same writeQuery to the signup form as we did for the login form until we are able to add that api work.

packages/apolloschurchapp/src/content-single/index.js
IsaacHardy: @conrad-vanl Was there a specific benefit that you were thinking of by querying the data in rather than grabbing it from the nav params?

packages/apolloschurchapp/src/auth/login/index.js
conrad-vanl: no @IsaacHardy i don't think I'm saying that. Just that we need to have a client-side mutation that handles both storing the auth token and creating the session. This can be added to our state store `src/store/index`, like a `handleLogin` mutation that takes an authToken as input, and does exactly what this cache updater is doing now.

packages/apolloschurchapp/src/content-single/index.js
conrad-vanl: We don't have to remember to add it to `nav params` any time we route to content, which is an easy to create but not catch edge-case bug.

packages/apolloschurchapp/src/ui/FeedItemCard/index.js
conrad-vanl: It seems like `Like` can standalone and be re-used easily this way

packages/apolloschurchapp/src/ui/Like/index.js
conrad-vanl: Got it. Good call.

conrad-vanl:
@IsaacHardy maybe I missed it, but how are you handling tapping on the `Like` button when there's no user logged in?

IsaacHardy:
@conrad-vanl The current UX is that there is no like button if you are not logged in. I recall you mentioning that the protected login function PR you have would be a requirement to add functionality for liking content while not logged in. (e.g. We could move a user to a login screen if they were to like a piece of content)

packages/apolloschurchapp/src/ui/FeedItemCard/index.js
IsaacHardy: I agree with the `Like` component being re-usable, what I mean is if we want a user to have the ability to like a piece of content via the feed view, rather then liking it within the content view.

packages/apolloschurchapp/src/content-single/index.js
IsaacHardy: FYI, it seems that there might be some limitations with attempting this functionality.

packages/apolloschurchapp/src/auth/login/index.js
IsaacHardy: I spoke with @vinnyjth earlier and he help clarified what you were suggestion earlier. That makes more sense!

IsaacHardy:
@conrad-vanl @vinnyjth Updated the caching and session creation to the store component. Check it out and let me know if you have any suggestions.

PS: there's an internal server error when i try to register, not sure if that is in the current app or not.

PR #262, opened 2018-09-04
Move Root Queries to be inline 

vinnyjth:
I found that the schema is easier to work with when the root queries are inline with the folders/files where their resolvers are implemented. 

This is now possible thanks to the `extend type` syntax from Graphql :D 

I didn't move the auth related queries, because I moved those in another PR. I want to avoid creating merge conflicts in `git`. 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


conrad-vanl:
I fully support this PR, but I'll let @IsaacHardy review it as I think he will be 🤩over this!

packages/apollos-church-api/src/data/bible/schema.js
```diff
@@ -5,4 +5,8 @@ export default gql`
     reference: String
     content: String
   }
+
+  extend type Query {
+    scripture(query: String!): ESVScripture
```
IsaacHardy: Running into an issue with `ESVScripture`. I think this is supposed to be just `Scripture` based on the type above.

IsaacHardy:
There was a side effect to your API changes. A snapshot on the client end was updated. Not sure if this was related in anyway, but we should make a note to run `yarn test` from the root of the project rather than in each package to make sure no unexpected changes were made.

I've updated everything that was blocking this PR so I'm marking, MERGE.

Issue #228, opened 2018-08-21
Integrate Appcenter analytics and diagnostics


conrad-vanl:
#226 removed sentry tracking due to problematic tendencies with the react-native-sentry package. Originally, I was planning on adding a javascript-only sentry tracker (like we had on NewSpring Apollos), but go to thinking: Why don't we try out Appcenter? It has both an analytics and diagnostics solution built-in, we are already integrated with it, and it might be a good "single-tool" solution for most churches. 

I feel like our analytics approach is going to be strongly focused on Rock interactions in the short term anyways, and so really we just need bug tracking and some basic analytics coupled together to support the data we're sending to Rock, which appcenter appears to provide.

https://docs.microsoft.com/en-us/appcenter/sdk/getting-started/react-native

PR #263, opened 2018-09-05
Adds Server Side Analytics Support

### LINKED ISSUE #240: [Analytics] Settle on an Analytics Aproach
## User Story

As a Admin I want to view App Analytics so that I can make informed decisions about content and the product.

We should spec out the options for Analytics. Current ideas include - 
- Using a generic Analytics lib [(Analytics.js?)](https://segment.com/docs/sources/website/analytics.js/) and allowing clients to configure which SaaS they would like to use.
- Talk to a company like [Astronomer](https://www.astronomer.io/) to enable "all in one" approach to collecting analytics.
- Use Microsoft Appcenter #228 
- Explore using Rock Interactions to provide Analytics
   - Probably should determine if you can do things like create Charts using Rock Interactions.
   - I'm worried that Rock Interactions are great at collecting Data, and bad at visualizing data.

Once we have brainstormed ideas, we should equip ourselves to pick a route forward and begin integrating into the application.

vinnyjth:
Fixes #240 
References #228

Video of feature in action :D 
https://www.dropbox.com/s/xe6gc9pmbpqgy8i/Sep-05-2018%2017-00-01.mp4?dl=0

Adds a GraphQL interface and two implementations of Analytics Tracking. Right now it's configured to send all data points to segment, and 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


conrad-vanl:
I _think_ I'm in support of this PR....

However, I really do think that it's a miss that `interactions` and `analytics` are separate APIs. From a client-perspective, they feel like the same thing - "record that user did something". I feel like it could be very reasonable for a church to eventually want ScreenViews to be tracked in Rock, for example. I'd vote to consider merging the schemas, if at all possible.

If not...I'd almost favor a client-side analytics implementation.

vinnyjth:
@conrad-vanl That's a really good point. I've thought through this a bit, and I believe I need to think a bit more before I come up with a great solution to integrate Interactions into this PR. The biggest issues, as I see it, is that you need to setup an `InteractionChannel` and an `InteractionComponent` before you can create an `Interaction`. I need to make some changes in the current API to reflect the configuration required before you can fire off an arbitrary event :)

vinnyjth:
Ready for approve/re-review @conrad-vanl !

packages/apollos-church-api/src/data/analytics/interfaces/ga.js
```diff
@@ -0,0 +1,16 @@
+import ua from 'universal-analytics';
+
+export default class GoogleAnalytics {
```
conrad-vanl: Thoughts on having these classes extend a base `Analytics` class with noop functions set for `track` and `identify`? That way we can remove the `shouldIdentify` and `shouldTrack` vars and conditional logic you have elsewhere?

vinnyjth: Love it :D Originally I built this using a `Base Analytics` but ended up ditching it because I couldn't see a clear use. However, I think it would be helpful to have it as a base to work off of for future implementations.

vinnyjth:
@conrad-vanl I was totally wrong about `createInteraction`. I had said on our call that the mutation was `likeContent`, but it's definitely `createInteraction`. I propose we rename the `createInteraction` mutation to `likeComment` and modify the args to fit that paradigm.

vinnyjth:
> Is there possibility for the "default" to be something other then segment? segment is too much of a cost premium for most of these churches to consider it

We could. Our other option that uses `analytics-node` is astronomer. I know there are other, but I'm having trouble finding them....

conrad-vanl:
What about google analytics? It looks like you had that set up

PR #264, opened 2018-09-06
fixed up the tests that my PR broke.

richarddubay:
Fixes [issue-number]

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic

### DIFF
--- a/packages/apollos-church-api/src/data/content-items/__tests__/resolvers.tests.js
+++ b/packages/apollos-church-api/src/data/content-items/__tests__/resolvers.tests.js
@@ -125,7 +125,7 @@ describe('UniversalContentItem', () => {
             id
             title
             scriptures {
-              html
+              content
             }
           }
         }
--- a/packages/apollos-church-api/src/data/content-items/__tests__/resolvers.tests.js
+++ b/packages/apollos-church-api/src/data/content-items/__tests__/resolvers.tests.js
@@ -125,7 +125,7 @@ describe('UniversalContentItem', () => {
             id
             title
             scriptures {
-              content
+              html
             }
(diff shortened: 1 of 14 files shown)


IsaacHardy:
@richarddubay Let out the PR that just got merged to master. If might have updated this PR a bit.

richarddubay:
@IsaacHardy This still passed. So we should be good?

packages/apollos-church-api/src/data/content-items/__tests__/resolvers.tests.js
```diff
@@ -125,7 +125,7 @@ describe('UniversalContentItem', () => {
             id
             title
             scriptures {
-              html
```
conrad-vanl: hmmm... I guess I messed as I didn't notice that this attribute changed. Is the change to `content` only because the new api uses the key `content` for this? I personally prefer `html` (more explicit to what it is), but could be swayed otherwise

IsaacHardy:
@richarddubay I'm assuming it passed because you're comparing it with a non-master branch.

packages/apollos-church-api/src/data/content-items/__tests__/resolvers.tests.js
vinnyjth: I agree with @conrad-vanl, I like using HTML here.

richarddubay: I don't have a problem using `html`, but yeah, the new API uses `content` so I just went with the default.

PR #292, opened 2018-09-13
Adds Rock as Analytics Interface

vinnyjth:
Fixes #292 

This PR accomplishes two tasks.

#1 Renames the `createInteraction` mutation to `updateLikeEntity`. The purpose of this changes is to reduce confusion about when to use `createInteraction` by removing it (now that we have an analytics API) 

#2. Adds Rock Interaction analytics interface. We can now send specific Analytics events to Rock Interactions. Right now a single analytics event is implemented and whitelisted, but in the future we can implement further events.  

## Creator

- [x] ~~Upload an animated GIF showcasing the solution (if it applies)~~
- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


packages/apollos-church-api/src/data/analytics/data-source.js
```diff
@@ -90,14 +93,19 @@ export default class Analytics extends DataSource {
...
-      iface.track({
-        userId: currentUser && currentUser.id,
-        anonymousId,
-        properties: parsedProps,
-        event: eventName,
-        context: deviceInfo,
-      });
+      if (
+        iface.eventWhitelist === null ||
+        iface.eventWhitelist.includes(eventName)
```
IsaacHardy: Could you explain the `eventWhiteList` property?

packages/apollos-church-api/src/data/analytics/interfaces/rock_interactions.js
```diff
@@ -0,0 +1,34 @@
...
+        return null;
+    }
+  }
+
+  trackViewContent({ contentId, sessionId }) {
+    if (!contentId || !sessionId) {
+      console.log('No ContentId or SessionId included in `track` call.');
+      return null;
+    }
+    return this.context.dataSources.Interactions.createInteraction({
```
IsaacHardy: Wouldn't this make the Interactions table EXTREMELY large? How would we handle this data at scale?

PR #296, opened 2018-09-18
Added Devotional Component

### LINKED ISSUE #267: [Scripture] Create Devotional Component
## User Story

As a consumer of in-app devotional content, I want to read my devotional along with its associated scripture so that I can grow closer to God.

#### Acceptance Criteria

* [ ] A devotional component is created that takes advantage of the tabbed view, displaying the devotional in one tab and the associated scripture in another.

richarddubay:
Fixes #267 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [x] Review CI build results
- [x] Review new test logic
- [x] Review that the PR title adheres to our patch-notes/change-log style


packages/apolloschurchapp/src/ui/Devotional/index.js
```diff
@@ -0,0 +1,57 @@
...
+  const hasScripture = isLoading || scripture.length;
+  const tabRoutes = [{ title: 'Devotional', key: 'devotional' }];
+  if (hasScripture) tabRoutes.push({ title: 'Scripture', key: 'scripture' });
+
+  return (
+    <BackgroundView>
+      <TabView
+        routes={tabRoutes}
+        renderScene={SceneMap({
+          devotional: withProps({
```
burkeshartsis: As a whole we have been trying to reduce our dependency on recompose where there are more common patterns available. In this case it might even prove to be more performant to remove `withProps` in favor of something like `renderTabName()`. This would allow you to pass Props into your Tab components in a more nature way that is not dependent on recompose.

packages/apolloschurchapp/src/ui/Devotional/index.js
```diff
@@ -0,0 +1,57 @@
...
+            scripture,
+          })(ScriptureTab),
+        })}
+      />
+    </BackgroundView>
+  );
+};
+
+Devotional.propTypes = {
+  content: PropTypes.shape({
```
burkeshartsis: I like your naming of `content` here. It's semantic and pairs well with your use of scripture in naming. Unfortunately, this is the only place I think it's used in your component. My suggestion would be to change your references to `devotional` and DevotionalTab` to `Content`. Another option would be to update this to use the `devotional` naming pattern. It's just less clear but more consistent if you were to go that route.

packages/apolloschurchapp/src/ui/Devotional/index.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import { withProps } from 'recompose';
+import PropTypes from 'prop-types';
+
+import TabView, { SceneMap } from 'apolloschurchapp/src/ui/TabView';
+import BackgroundView from 'apolloschurchapp/src/ui/BackgroundView';
+import DevotionalTab from 'apolloschurchapp/src/ui/Devotional/DevotionalTab';
+import ScriptureTab from 'apolloschurchapp/src/ui/Devotional/ScriptureTab';
+
+const Devotional = ({
```
burkeshartsis: I'm not super in love with how generically named this is as `Devotional`. Devotional what? Suppose another church wanted to present devotions in multiple different ways. Tabs, stories, or something else. I think we would run into naming conflicts. Naming will always be hard and you can always import it as a different name but we might be able to help a little from the onset.

packages/apolloschurchapp/src/ui/Devotional/index.js
```diff
@@ -0,0 +1,57 @@
...
+  content: { body, title, ...otherContentProps },
+  isLoading,
+  scripture,
+}) => {
+  const hasScripture = isLoading || scripture.length;
```
burkeshartsis: So I think an `isLoading` prop for this component is good I'm just not sure we're implementing it in a way that leverages our placeholders and how they do a lot of the work for you.

packages/apolloschurchapp/src/ui/Devotional/index.js
burkeshartsis: Additionally, I think some portion of these components could benefit from being a `PureComponent` class. My guess would be this parent component as it manages all the props for both tabs. You could argue though that if the tabs were `PureComponents` they could manage if they render independently of themselves.

packages/apolloschurchapp/src/ui/Devotional/Devotional.tests.js
```diff
@@ -0,0 +1,84 @@
...
+
+// Needed to add our own ScriptureTab component here. This is because the real
+// one uses the Scripture component, which calls a query and we can't handle
```
burkeshartsis: Can we not mock the `Query`?

packages/apolloschurchapp/src/ui/Devotional/Devotional.tests.js
```diff
@@ -0,0 +1,84 @@
...
+  scriptures: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.string,
+      reference: PropTypes.string,
+      html: PropTypes.string,
+    })
+  ),
+};
+
+describe('the Devotional component', () => {
```
burkeshartsis: You `index.js` exports a `Devotional` component but it's not used here. Instead it looks like you're rebuilding it inside the test which leaves the component outside of test coverage 😢

packages/apolloschurchapp/src/ui/Devotional/Devotional.stories.js
```diff
@@ -0,0 +1,78 @@
...
+      id: PropTypes.string,
+      reference: PropTypes.string,
+      html: PropTypes.string,
+    })
+  ),
+};
+
+const stories = storiesOf('Devotional', module);
+
+stories.add('God is Our Banner', () => (
```
burkeshartsis: You're story seems to be rebuilding the `Devotional` component which means it's not really a story of the code you use.

packages/apolloschurchapp/src/ui/Devotional/Devotional.tests.js
burkeshartsis: Seems like the underlying problem was you needing to get around the `Query` in `ScriptureTab`. I think mocking the `Query` is probably the best route but if you would rather not (because mocking can be a pain) I recommend exposing tab props to pass components into.

packages/apolloschurchapp/src/ui/Devotional/Devotional.tests.js
burkeshartsis: Another option, you could update `ScriptureTab` to check if you are getting a ref (needs the `Query`) or an object and then swap the `Scripture` component for a scripture `Item`. Better yet, read my comments about proptypes and implement this change to Scripture itself thus bypassing/not rendering the `Query` if it's not needed. Thats sort of the long way around the mock the query problem though but could be a userful feature is someone wanted to display scripture without using a ref.

packages/apolloschurchapp/src/utils/scriptures/list.js
```diff
@@ -0,0 +1,11 @@
+function scriptureList({ scripture, commas = true } = {}) {
```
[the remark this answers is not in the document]
richarddubay: This is smart. I always just bring everything over as it is first and never think if it's the best way to handle it. You probably commented about it, but I usually just end up using `recompose` as well. Consider this moved.

packages/apolloschurchapp/src/ui/Devotional/ScriptureTab.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { ScrollView } from 'react-native';
+import PropTypes from 'prop-types';
+import PaddedView from 'apolloschurchapp/src/ui/PaddedView';
+import Scripture from 'apolloschurchapp/src/ui/Scripture';
+
+const ScriptureTab = ({ isLoading, scripture }) => (
+  <ScrollView>
+    {!isLoading ? (
```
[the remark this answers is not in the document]
richarddubay: Well that was just silly. Fixed.

packages/apolloschurchapp/src/ui/Devotional/index.js
conrad-vanl: Also, should `Devotional` live inside of `content-single` and not `ui`? just a thought 🤔

packages/apolloschurchapp/src/ui/Devotional/ScriptureTab.js
```diff
@@ -0,0 +1,28 @@
...
+      <PaddedView>
+        <Scripture references={scripture} />
+      </PaddedView>
+    ) : null}
+  </ScrollView>
+);
+
+ScriptureTab.propTypes = {
+  isLoading: PropTypes.bool,
+  scripture: PropTypes.arrayOf(
```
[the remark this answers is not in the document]
richarddubay: @burkeshartsis Do you feel like it's beneficial to have both? I'd be okay with just changing it to have the correct propTypes that the Scripture component is expecting.

packages/apolloschurchapp/src/ui/Devotional/DevotionalTab.js
```diff
@@ -0,0 +1,57 @@
...
+  body,
+  isLoading,
+  route: { jumpTo },
+}) => (
+  <ScrollView>
+    {!isLoading ? (
+      <ContentContainer>
+        <H2 padded>{title}</H2>
+        {scripture && scripture.length ? (
+          <ScriptureLink padded onPress={() => jumpTo('scripture')}>
```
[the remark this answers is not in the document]
richarddubay: screenshot 2018-09-20 16 24 52

So by changing this out to use the `ButtonLink` and leaving it's only real style (text color) alone, I've now got a situation where the tab underline color no longer matches the link color. In this case, should I then change the tab underline color? Or switch the order from `<H4><ButtonLink>` to `<ButtonLink><H4>` so that I can override the link color to match the tab underline? I'm guessing from your comment that it should be the former, but I wanted to check.

packages/apolloschurchapp/src/ui/Devotional/DevotionalTab.js
```diff
@@ -0,0 +1,57 @@
...
+import PropTypes from 'prop-types';
+import PaddedView from 'apolloschurchapp/src/ui/PaddedView';
+import { H2, H4 } from 'apolloschurchapp/src/ui/typography';
+import styled from 'apolloschurchapp/src/ui/styled';
+import HTMLView from 'apolloschurchapp/src/ui/HTMLView';
+import scriptures from 'apolloschurchapp/src/utils/scriptures';
+
+const ContentContainer = styled({ paddingVertical: 0 })(PaddedView);
+
+const ScriptureLink = styled(({ theme }) => ({
```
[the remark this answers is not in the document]
richarddubay: I went ahead and moved this into the `ScriptureList` component. It makes sense that they go together.

packages/apolloschurchapp/src/ui/Devotional/index.js
richarddubay: Actually, this might not be true.

richarddubay: It didn't like that setup at all. So I just made the index a pure component and set up the routes like you mentioned and that I saw in other components. That should work.

packages/apolloschurchapp/src/ui/Devotional/index.js
richarddubay: Changed everything to be "content" based. Love it.

packages/apolloschurchapp/src/ui/Devotional/index.js
richarddubay: @burkeshartsis I'd love to know your thoughts on how we could make this better. Should we add a `Placeholder` component?

richarddubay: I went ahead and added a placeholder to the `ContentTab` so that both tabs would show placeholders while data was loading.

packages/apolloschurchapp/src/ui/Devotional/Devotional.tests.js
richarddubay: There are now two scripture components: one connected and the other flat. The connected component is mocking the query and return data.

packages/apolloschurchapp/src/ui/Devotional/Devotional.tests.js
richarddubay: This is taken care of with the split of the scripture into connected and flat components. We're now using the actual component and not a rebuilding of the pieces of the component inside the tests.

packages/apolloschurchapp/src/ui/Devotional/Devotional.stories.js
richarddubay: Again, with the split out of the scripture component into connected and flat components I am able to use the actual `Devotional` component in Storybook.

packages/apolloschurchapp/src/ui/Devotional/index.js
richarddubay: @conrad-vanl and @burkeshartsis - Point of discussion per the comments above: 

1. Currently, the `Devotional` component is still named `Devotional` and still lives in the `ui` folder. I'm not sure what else I would name it. Are there suggestions?
2. `Devotional` is now a flat component having no connection to any data. It is, however, a component made up of other components in the `ui` folder. Do you feel like this should still be moved out to the `content-single` folder? I can see where the answer could go both ways. No, it should stay in the `ui` folder because it is a component that doesn't deal with data. Yes, it should move out of the `ui` folder because if the `ui` folder will one day become an NPM package, it probably shouldn't include the `Devotional` component. So ... thoughts?

I feel like this is pretty close to being ready for review again. Whatever we decide here should take it over the top.

richarddubay:
@conrad-vanl Moved the component to the `content-single` folder! I feel great about that move.

packages/apolloschurchapp/src/content-single/Devotional/index.js
```diff
@@ -0,0 +1,82 @@
...
+   *   body: body text of the devotional.
+   *   title: title of the devotional.
+   * isLoading: Boolean prop that shows placeholders while the data is loading.
+   * scripture: An array of scripture verses containing:
+   *   id: The ID of the verse (i.e. '1CO.15.57')
+   *   reference: The scripture location (i.e. '1 Corinthians 15:57')
+   *   html: The HTML of the verses to render.
+   */
+  static propTypes = {
+    content: PropTypes.shape({
```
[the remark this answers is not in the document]
richarddubay: I think the original idea was that yeah, it would be nested to keep it separated from scripture. But I checked the return data from the devotional and it doesn't have it nested, so there is no sense in doing it here. Removed the nesting.

packages/apolloschurchapp/src/ui/Scripture/ScriptureList.js
```diff
@@ -0,0 +1,45 @@
...
+    <StyledH4>
+      <ButtonLink padded onPress={() => jumpTo('scripture')}>
+        {combo}
+      </ButtonLink>
+    </StyledH4>
+  );
+};
+
+ScriptureList.propTypes = {
+  scripture: PropTypes.arrayOf(
```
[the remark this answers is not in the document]
richarddubay: Updated the Devotional and it's content components to only pass down and use the necessary parts of the scripture. In this case, that's just the scripture references. No need to pass down :allthethings: if we don't need to.

PR #297, opened 2018-09-18
Extract Like component to make it reusable

### LINKED ISSUE #280: [Liking Content] Extract code from <ActionContainer /> to form own <Like /> component
We should create a `<Like />` component that is reusable which contains the code/logic from the `<ActionContainer />` component. 

* [ ] We should have a reusable `<Like />` component

IsaacHardy:
Fixes #280 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


packages/apolloschurchapp/src/ui/LikeButton/index.js
```diff
@@ -0,0 +1,79 @@
...
+              {(createNewInteraction) => (
+                <Like
+                  itemId={itemId}
+                  sessionId={sessionId}
+                  isLiked={isLiked}
+                  operation={isLiked ? 'Unlike' : 'Like'}
+                  toggleLike={async (variables) => {
+                    try {
+                      await createNewInteraction({ variables });
+                      await refetch();
```
vinnyjth: Do we still need the refetch? I think we can get rid of this now that we have the `cache.writeQuery` :)

vinnyjth:
Reviewing 👀

packages/apolloschurchapp/src/content-single/ActionContainer.js
```diff
@@ -1,79 +1,22 @@
...
-                )}
-              </Mutation>
-            )}
-          </Query>
-        ) : null
-      }
-    </Query>
+    <LikeButton
+      itemId={itemId}
+      updateLikeEntity={updateLikeEntity}
```
vinnyjth: Nit-pick. The Share button can be put anywhere and passed an object to share. The same isn't true with the like button. You need to inject the `updateLikeEntity` function and the `getLikedContentItem` function. What do you think about moving those functions inside the like button component so they don't need to be passed?

packages/apolloschurchapp/src/ui/LikeButton/LikeButton.tests.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
```
vinnyjth: TESTS :D 

nice work :)

packages/apolloschurchapp/src/ui/LikeButton/index.js
```diff
@@ -0,0 +1,79 @@
...
+          )}
+        </Query>
+      ) : null
+    }
+  </Query>
+);
+
+LikeButton.propTypes = {
+  itemId: PropTypes.string,
+  getLikedContentItem: PropTypes.shape({}),
```
vinnyjth: Nit pik - it's annoying to fill these out, but it always helps future readers of this component. We probably won't need them if we move the functions inside this component, which would be nice :D

packages/apolloschurchapp/src/ui/LikeButton/index.js
IsaacHardy: I swore I removed this... Maybe I'm losing my mind.

IsaacHardy:
@vinnyjth Made updates to comments

packages/apolloschurchapp/src/content-single/ActionContainer.js
vinnyjth: You're very right, good catch

PR #303, opened 2018-09-21
Fix release script so CHANGLOG.md users proper version

vinnyjth:
Fixes [issue-number]

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,5 @@
 {
   "name": "apollos",
-  "version": "0.5.0",
   "private": true,
   "workspaces": {
     "packages": [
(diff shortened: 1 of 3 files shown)


vinnyjth:
@conrad-vanl I figured out that the `package.json` version was not being set for our root package.json, and hence the changelog was borked. Fixed that (`lerna` won't do it :/), and also update the README so it's a tidy bit more accurate.

package.json
```diff
@@ -14,8 +14,9 @@
     "lint": "./node_modules/.bin/lerna run lint",
     "test": "./node_modules/.bin/lerna run test",
     "ci-test": "./node_modules/.bin/lerna run test -- --maxWorkers=2",
-    "release": "./node_modules/.bin/lerna version --no-push && echo \"Release has been tagged!\nPlease look over the CHANGELOG.md file, and use 'git commit -amend' to amend any edits.\nWhen done, be sure to 'git push --tags' and create a release PR\"",
-    "postversion": "./scripts/update-change-log.sh && ./node_modules/.bin/lerna run postversion"
+    "release": "./node_modules/.bin/lerna version --no-push && yarn set-from-api-version && echo \"Release has been tagged!\nPlease look over the CHANGELOG.md file, and use 'git commit -amend' to amend any edits.\nWhen done, be sure to 'git push --tags' and create a release PR\"",
```
conrad-vanl: We shouldn't _need_ this `set-from-api-version` script with lerna. Other monorepo projects that utilize lerna don't depend on other processes to set `package.json` versions, and so I'm confused why we need to here.

Looking at a few examples, such as the babel repo, I don't think we need a root-level `version` in our package.json. Why do you think we need one? 

Ahhh....we use a custom changelog script, that's why. Instead of adding a version to root `package.json`, which I don't think we need, maybe we can change [this line](https://github.com/ApollosProject/apollos-prototype/blob/master/scripts/change-log/generate-git-log.sh#L4) to point at `lerna.json`'s version instead...thoughts?

vinnyjth: Yah, that's we need the root version, for that script. Pointing at `lerna.json` makes much more sense, good catch. Didn't realize that the `lerna.json` file existed :)

PR #321, opened 2018-10-02
Adds Push Notification Deep Linking

### LINKED ISSUE #290: [Push Notifications] [Exploratory] Set up deep linking from Push Notifications
As a developer / administrator, I want my Push Notifications to be able to dynamically link to a specific view or piece of content within the app.

This needs to be flushed out a little more. This may be able to be accomplished through OneSignal in combination with some functionality that can be added through React Navigation which we're already using.

cdanwards:
Fixes #290 

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


packages/apolloschurchapp/src/Notifications.js
```diff
@@ -1,8 +1,11 @@
 import { Component } from 'react';
+import { Linking, Platform } from 'react-native';
```
conrad-vanl: I'm glad you decided to use `Linking` for this, it was one of the things I was going to suggest trying.

I think in the future, we can have our deep linking (a user tried to open a link on the church's website) and this push notification linking handler user the same set of code.

Good start :)

packages/apolloschurchapp/src/Notifications.js
```diff
@@ -12,7 +15,18 @@ export default class NotificationsInit extends Component {
     OneSignal.addEventListener('ids', this.onIds);
   }
 
+  componentDidMount() {
+    if (Platform.OS === 'android') {
```
conrad-vanl: Why the check for just android here?

packages/apolloschurchapp/src/Notifications.js
```diff
@@ -33,6 +47,23 @@ export default class NotificationsInit extends Component {
     console.log('Device info: ', device);
   };
 
+  navigate = (url) => {
```
conrad-vanl: Good start for this function. One of the things I'm going to ask the Rock team is: 

Can Rock help us "get" a permalink for a content item - say - `ContentChannelItem` with ID of `3` translates to `https://newspring.cc/articles/my-article-name`

And on the flip-side, given some permalink, say, `https://newspring.cc/articles/my-article-name`, can we have Rock figure out what content that link is pointing at?

My hypothesis is that Rock should be able to do this for any church that uses Rock for their website. But we'll see what they say, I'm sure the functionality doesn't exist today but would be super helpful for this, because then a church could just copy-n-paste the website URL in for the push notification instead of trying to construct some app url.

cdanwards: @conrad-vanl Does our graphQL server give a randomized ID to each article like it does in the NS app? I know there it shows up as `/studies/e8e958963bc68592e57b1dad29ad81e9/entry/4712ab25bbaa2701e411398f3a7052f5` or something like that. Then it links it to the parent `entityId` or something like that.

packages/apolloschurchapp/src/Notifications.js
cdanwards: The guide I was following along with had this reasoning:
`If we are on Android, we immediately call the navigate method passing in the url. If we are on iOS, We add an event listener to call handleOpenUrl when an incoming link is detected.`

cdanwards: And with that, I think that the `handleOpenUrl` isn't working as intended. I'm not sure what's off about it though.

packages/apolloschurchapp/src/Notifications.js
cdanwards: Also, @conrad-vanl would this be categorized with the other open [issue](https://github.com/ApollosProject/apollos-prototype/issues/308) when talking to Spark?

conrad-vanl: Yes, and yes. The ID the app uses is different then the Rock IDs.

packages/apolloschurchapp/src/index.js
```diff
@@ -40,7 +41,11 @@ const AppModalNavigator = createStackNavigator(
 const App = () => (
   <Providers>
     <BackgroundView>
-      <AppModalNavigator />
+      <AppModalNavigator
```
vinnyjth: Nice :D

packages/apolloschurchapp/src/NavigationService.js
```diff
@@ -0,0 +1,23 @@
+import { NavigationActions } from 'react-navigation';
```
vinnyjth: Well done :D

packages/apolloschurchapp/ios/apolloschurchapp/AppDelegate.m
```diff
@@ -10,9 +10,26 @@
...
 
+- (BOOL)application:(UIApplication *)application openURL:(NSURL *)url
+  sourceApplication:(NSString *)sourceApplication annotation:(id)annotation
+{
+  return [RCTLinkingManager application:application openURL:url
+                      sourceApplication:sourceApplication annotation:annotation];
+}
+
+// Only if your app is using [Universal Links](https://developer.apple.com/library/prerelease/ios/documentation/General/Conceptual/AppSearch/UniversalLinks.html).
+- (BOOL)application:(UIApplication *)application continueUserActivity:(NSUserActivity *)userActivity
```
vinnyjth: This snippet is perfect :D

packages/apolloschurchapp/src/Notifications.js
```diff
@@ -12,12 +16,44 @@ export default class NotificationsInit extends Component {
...
     OneSignal.removeEventListener('received', this.onReceived);
     OneSignal.removeEventListener('opened', this.onOpened);
     OneSignal.removeEventListener('ids', this.onIds);
   }
 
+  handleOpenURL = (event) => {
+    NavigationService.navigate(event.url);
+  };
+
+  navigate = (url) => {
```
vinnyjth: Really like how you're handling this here :)

packages/apolloschurchapp/src/index.js
```diff
@@ -40,7 +41,11 @@ const AppModalNavigator = createStackNavigator(
...
+        ref={(navigatorRef) => {
+          NavigationService.setTopLevelNavigator(navigatorRef);
```
conrad-vanl: Didn't even know this was a thing. Good find.

packages/apolloschurchapp/src/Notifications.js
conrad-vanl: How is it behaving? This is a little different because it uses a different router, but here's how we handled it on NS Apollos: https://github.com/NewSpring/Apollos/blob/alpha/src/%40ui/NativeWebRouter/DeepLinking.js#L25-L28

vinnyjth:
@cdanwards want to give you props, because this is a great example of a well scoped PR :D

packages/apolloschurchapp/src/Notifications.js
```diff
@@ -17,41 +19,25 @@ export default class NotificationsInit extends Component {
...
-    if (routeName === 'AppStackNavigator') {
-      navigate('AppStackNavigator', { id });
-    } else if (routeName === 'Tabs') {
-      navigate('Tabs');
-    } else if (routeName === 'ContentSingle') {
-      navigate('ContentSingle', { id });
-    } else if (routeName === 'Connect') {
-      navigate('Connect');
-    }
+  navigate = (rawUrl) => {
```
cdanwards: @vinnyjth This is awesome! I knew we'd need to implement something like this.

packages/apolloschurchapp/ios/apolloschurchapp/AppDelegate.m
```diff
@@ -10,9 +10,26 @@
 #import <React/RCTBundleURLProvider.h>
 #import <React/RCTRootView.h>
 #import "RNSplashScreen.h"
+#import <React/RCTLinkingManager.h>
 
 @implementation AppDelegate
 
+- (BOOL)application:(UIApplication *)application openURL:(NSURL *)url
```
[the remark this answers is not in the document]
conrad-vanl: Is this the only change still needed to mark this PR ready for review again?

Issue #324, opened 2018-10-10
[Toolkit] Add Developer / Tester "Toolkit"


vinnyjth:
## User Story

As a developer  I want to have access to a an in-app control panel so that I can test the livestream.
As an app tester  I want to have access to a an in-app control panel so that I can test push notifications.

#### Problem Details _(Optional)_

* Screenshots, GIFs, or a video
* Any additional information (version, device, etc.)

#### Acceptance Criteria _(Optional)_

* [ ] It should [SPEC_DETAIL]

IsaacHardy:
@vinnyjth Is this needing to be closed?

vinnyjth:
@IsaacHardy Not yet, as I haven't accomplished all the objectives within :)

PR #326, opened 2018-10-11
Adds Initial Testing control panel

vinnyjth:
References #324 

Adds a control panel interface accessed through the Connect tab, where we can house various testing tools. Right now there's a button that toggles the livestream on and off. Once this merged, I will add more features :) #keepprsshort

Video: https://www.dropbox.com/s/g4xa49nk3wpx81o/Screen%20Recording%202018-10-11%20at%2012.21.03%20PM.mov?dl=0

## Creator

- [x] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


packages/apolloschurchapp/src/testing-control-panel/changeLivestream.js
```diff
@@ -0,0 +1,15 @@
+import { client } from 'apolloschurchapp/src/client';
```
conrad-vanl: Are there any scenarios when getting the `client` singleton won't work? (I'm thinking - versus using the `withApollo` HOC)

vinnyjth: Yah, I think if you are swapping out clients in the provider dynamically (not sure why you would want to do that....) you would benefit from the `withApollo` HOC. Or if you want to pass in a fake client to test the component. Besides that, I haven't run into any reasons why the singleton model doesn't work.

conrad-vanl: Is there a simple way to use the context-provided client instead of the singleton?

vinnyjth: At one point, I was passing the client into the function, and getting the function from the `withApollo` HOC, but that pattern felt really dirty, so I imported it directly. I don't anticipate using the `client` as a singleton ever being an issue :)

IsaacHardy:
@vinnyjth This is really cool and useful! Great job!

packages/apolloschurchapp/src/testing-control-panel/index.js
```diff
@@ -0,0 +1,38 @@
...
+import changeLivestream from './changeLivestream';
+
+const ChangeLivestream = () => (
+  <Query query={getLiveStream}>
+    {({ data }) => {
+      const isLive = get(data, 'liveStream.isLive', false);
+      return (
+        <TouchableCell
+          handlePress={() => changeLivestream({ isLive: !isLive })}
+          iconName={isLive ? 'pause' : 'play'}
```
IsaacHardy: Would `CircleOutlineCHeckMark` and `CircleOutlineXMark` work better here?

vinnyjth:
Done @IsaacHardy  :)

PR #355, opened 2018-10-23
Move apollos rock data source

### LINKED ISSUE #352: [Server Cleanup] Split apollos-church-api into several different packages
## User Story

* [ ] It should [SPEC_DETAIL]

vinnyjth:
How this works

1. Uses the `babel-preset-apollos`
2. The other packages only see what's inside the `./lib` folder. 
3. The `lib` folder only exists when `yarn` or `yarn install` is run from the root. (yarn install builds all dependancies
4. If making changes to `apollos-rock-apollo-data-source` you will have to run `yarn` or `lerna build` every time you want to include changes. The way to avoid this is to practice TDD and write tests in the `src` directory of `packages/apollos-rock-apollo-data-source`. There should probably be a better option though

My next target is the Constants file. 

Fixes #352 

## Creator

- [ ] Set a relevant reviewer

## Reviewer

- [ ] Review CI build results
- [ ] Review new test logic


vinnyjth:
Coverage went down because the RockApolloDataSource was being tested implicitly by other modules. Now that the main API doesn't track coverage over the `RockApolloDataSource`, we will need to write some explicit tests.

conrad-vanl:
What were your thoughts behind adding a file-watcher and triggering build off of a `start` command?

vinnyjth:
@conrad-vanl I think I could make it happen! Having a ton of watchers will get unwieldy though, the more packages we add :/

conrad-vanl:
Hmmm...if the watcher is scoped to just that package's `/src` dir, you think it will get pretty bad?

It's just right now development on this project centers around running the packager/node server locally. Losing that might be confusing/disruptive.

vinnyjth:
@conrad-vanl Honestly not sure. I think at this point though, it's safe to throw caution to the wind and monitor by default. We can rethink if people's machines get bogged down :)

packages/apollos-church-api/package.json
```diff
@@ -31,6 +31,7 @@
     "analytics-node": "^3.3.0",
     "apollo-datasource-rest": "0.1.1",
     "apollo-server": "^2.0.5",
+    "apollos-rock-apollo-data-source": "file:../apollos-rock-apollo-data-source",
```
conrad-vanl: Once we start publishing packages on NPM, this can turn into a version number, and locally `yarn` will still use the local version of the package, but deploys I think will use the published version

conrad-vanl: We can actually make this a version number number, but it breaks travis

vinnyjth: It looks like `apollo` uses the `file:../` syntax, while `babel` uses the version number approach. I am fine keeping the `file:../` for now.

vinnyjth:
@IsaacHardy Once you either review this or opt out of reviewing this I can merge it :)

PR #357, opened 2018-10-24
Adds Apollos Config

### LINKED ISSUE #351: [Rock Configuration] Add "All in one" Configuration Source for apollos-church-api
## User Story

* [ ] It should [SPEC_DETAIL]

vinnyjth:
Adds an ApollosConfig package that will automatically load and manage all the configuration variables for the apollos project and all dependencies, once imported and called. 

This package can be used by any of the server side applications that need access to ENV variables. As of right now, we will keep the documentation for variables inside the `config.yml` file, but in the future we should probably annotate the `config.yml` file more like a piece of documentation itself and keep our values out of `git`.  

Also includes a bunch of changes to support testing.

### What design trade-offs/decisions were made
As of right now, we are keeping the `config.yml` inside version control. In the future, this should be removed from version control and replaced with a more lightweight `config.example.yml`

### How do I test this PR?
1. Clone.
2. `yarn && yarn start` and click around the app to make sure everything works.

n/a

## TODO:

- [x] Closes #351 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-church-api/config.yml
```diff
@@ -0,0 +1,166 @@
+# This file, in conjuction with your .env file,
```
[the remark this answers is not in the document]
vinnyjth: Thinking about two options here (yml `extends` syntax is 💩 )

- Have two yml files, you can edit the advanced one for complex setups but overall you shouldn't need to touch it. This is nice because it might be tricky to debug what's going on when merging changes. For example, if I add `ROCK.APP_URL`, should it wipe out or keep `ROCK.APP_KEY`? I feel like different people might expect different behavior. 

- Have a yml file that's pretty much empy that a user could copy paste in, and that would provideoverridess. Default values would be kept in `src/config.js`. There would be some values you would need to define in `config.yml` (like your rock settings) to get the app to boot. There would be a config.sample.yml you could copy/paste when you clone.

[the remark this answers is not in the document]
conrad-vanl: Lame. Before I answer, have we thought much about what the public api for apollos api/server is going to look like? In other words, as a church, when I create an apollos api application, what files do I have, what do I need?

I imagine we'll need some sort of `@apollosproject/server` dependency inside of a node project, and you'd start an `apollos` server similarly to `apollo-server`:

```
new ApollosServer();
```

Why not pass the `config` as an argument to that function?

```
new ApollosServer({
  config,
});
```

And at that point, the `config` can just be a javascript object in the same file? Maybe we're starting at the wrong spot by starting with the config file...maybe we need to figure out what the "base" thing that a person needs to do in order to start their apollos api / server application.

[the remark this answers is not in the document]
vinnyjth: I can see us going two routes right now.

The first is what I'll call the `middleware` route where you can `yarn add @apollosproject/server` and then ```
new ApollosServer({
  config,
});
``` and boom you're done. 

I think the second is more like a create react app route, where you would run `npx create-apollos-church-server` and it would scaffold out a whole project, and we would compartmentalize each of the app building blocks to the point where you would really be passing different elements that we build into an `new ApolloServer` (not **APOLLOS**) instance. 

I've been building as if we're aiming for the second. Simple because I think it will be easier to create an mvp version of that product. I think we'll build something that's hard to maintain if we try to create a monolithic `@apollo/server` that does everything but is configurable vs if we build a really nice scaffolding system with all of the **real** logic that's difficult to maintain hidden in our own `@apolloschurch` packages. 

That said, I think the first approach will be easier if we have a lot of churches that need very little configuration, and I think the second approach will be easier if each church desires a slightly different setup. 

In the first case, updating a fleet of churches would be as easy as `npm upgrade @apolloschurch/server` and then updating their config / overridden resolvers / howeverelse the church has customized their instance. In the second case, we may have to update a @apolloschurch/server-parts package and then make sure we don't need to update any of the client's `actual` resolvers / schema / custom data sources. I anticipate this case being easier **if** each church is doing a high level of configuration. 

That's kinda all of the top of my head - and what I've been thinking about as I build things. 

##TLDR
I think the most important thing to clarify is **I've been working recently with the concept that apollos-church-api is really an implementation of an example churches API** and not a package that will eventually be `@apolloschurch/server`. My goal is to gut `apollos-church-api` into separate packages to the point where it could become a mere scaffold. Hopefully that makes sense and we are on the same page here.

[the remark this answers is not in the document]
Comment (conrad-vanl) [reply packages/apollos-church-api/config.yml]: > I think the most important thing to clarify is I've been working recently with the concept that apollos-church-api is really an implementation of an example churches API and not a package that will eventually be @apolloschurch/server. My goal is to gut apollos-church-api into separate packages to the point where it could become a mere scaffold. Hopefully that makes sense and we are on the same page here.

Correct. I think `apollos-church-api` continues to be gutted and reduced until it's really just a template implementation.

File: packages/apollos-church-api/src/data/bible/data-source.js [hunk 2]
```diff
@@ -1,18 +1,21 @@
 import { RESTDataSource } from 'apollo-datasource-rest';
+import ApollosConfig from '@apolloschurch/config';
+
+const { BIBLE_API } = ApollosConfig;
```
conrad-vanl: 🤔

Q: say that `BIBLE_API` was a string instead of an object, or, I did this on this line:

```
const { BIBLE_API } = ApollosConfig;
const { KEY } = BIBLE_API;
```

Then used this string/key later on, like in a function. Would this string point to the default value, or to the value that you set after calling `loadYaml`?

conrad-vanl: What I'm getting at is execution order and such....

conrad-vanl: and the potential for obscure bugs to happen

vinnyjth: I've accounted for both w/ how I merge the objects in when the file is loaded using two strategies. 

1. requring the package as early as possible and keep the package init synchronous 
2. Using non-functional friendly object assignment inside the package to attach config. `Object.assign(obj, newObj)`instead of `Object.assign({}, obj, newObj)` In the above case, that means that KEY should still be correct, because BIBLE_API will be pointing the the same object. I do this merge deeply so the same objects, even the ones a layer deep in the config, are reused. 

That said, I'm open for suggestions on how to make this better because it feels very brittle :D

conrad-vanl: The only way I can see making this less brittle is by forcing the usage of a getter function, ie...

`ApollosConfig.get('BIBLE_API')` as then, you can't easily use destructure assignment at the top of a file, which I'd imagine is a no-no when it comes to the order-execution above.

vinnyjth: @conrad-vanl I like that idea a lot. I'll keep it front of mind as we move forward, and maybe put together a test suite to try and break the current setup.

PR #358, opened 2018-10-24
Clean up / polish navigation on the connect tab

conrad-vanl:

- Cleans up navigation on user settings forms by adding a navigation header and tweaking where in the stack the navigation happens
- Tweaks/simplifies the component structure on the personal details + change password form. Primary output is that this decreases code debt
- Fixes typo on "Recently Liked" header on connect tab
- Adds navigation header to Recently Liked feed
- Fixes data loading on recently liked feed. Previously, not all information was showing up (the query was incomplete)

The main trade-offs made had to do with the design on the forms. Mainly with the way the cancel and Save buttons worked. Currently on master, these forms have a single "Cancel" or "Done" button in the upper right hand. However, it's not clear what happens when you swipe down or if you want to cancel your changes after you've made some. We need to have two clear actions: a cancel action, and a save/complete/done action. 

In order to accomplish this with the least code debt, I opted to add a `Save` button at the bottom of the screen,  and a `Cancel` button inside of the navigation header. Our `<Button/>` component cleanly handles the different loading/validation states of our forms, so there is a usability win in using `<Button />`. I tried to add a "Save" text button at the top right hand corner inside of the navigation header, but there's not a clean way to do that with `react-navigation` (due to component scope). 

I also tried several different variations of using no navigation header, making this component a modal with a close button in the top right, etc. I believe this ended up being the best solution.

image

### How do I test this PR?

Sign in, test out the forms. Make sure the software keyboard is visible. Make sure layout still looks okay on smaller devices, especially android.

Refactored existing components. Existing components don't have tests, but there were snapshot changes.

## TODO:

- [x] ~Closes [tag-issues-here]~
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apolloschurchapp/src/tabs/connect/getLikedContent.js
```diff
@@ -1,22 +1,29 @@
 import gql from 'graphql-tag';
 
+import { contentItemFragment } from 'apolloschurchapp/src/content-single/getContentItem';
+
 export default gql`
   query getAllLikedContent {
     getAllLikedContent {
       ... on ContentItem {
+        ...contentItemFragment
```
IsaacHardy: I really need to look into fragments to better understand their usecase.

packages/apolloschurchapp/src/user-settings/ChangePassword.js
```diff
@@ -1,38 +1,28 @@
...
-
-const DoneButton = styled(() => ({
-  fontWeight: '800',
-}))(ButtonLink);
-
 class ChangePassword extends PureComponent {
   static navigationOptions = () => ({
-    title: 'Personal Details',
+    title: 'Change Password',
+    headerBackImage: (
```
IsaacHardy: We might need to adjust some of this styling on Android:

screen shot 2018-10-25 at 9 34 22 am

packages/apolloschurchapp/src/user-settings/PersonalDetails.js
```diff
@@ -1,39 +1,28 @@
...
-
-const DoneButton = styled(() => ({
-  fontWeight: '800',
-}))(ButtonLink);
-
 class PersonalDetails extends PureComponent {
   static navigationOptions = () => ({
     title: 'Personal Details',
+    headerBackImage: (
+      <PaddedView vertical={false}>
```
IsaacHardy: We might need to adjust some of this styling on Android:

screen shot 2018-10-25 at 9 28 52 am

packages/apolloschurchapp/src/user-settings/ChangePassword.js
IsaacHardy: Also, on iOS these "Headers" are center in comparison to Android where they are flexed to the left. What design would be appropriate for these nav headers?

packages/apolloschurchapp/src/tabs/connect/getLikedContent.js
conrad-vanl: So we're not really using them super effectively yet. This example, we're really just using it to make sure we're selecting the minimal data that the `content-single` view needs so that when it opens, it opens without placeholders for the big things.

The main use case for fragments is to have pre-defined selections for similar things that you can reuse. Like the `sharing` object - we could have a `sharingFragment` that selects everything you need to display a `share` component. Similar with media like an `image` or something - contains the `src`, `width`, and `height`.

packages/apolloschurchapp/src/user-settings/ChangePassword.js
conrad-vanl: Hmmm....I'm tempted to move the "cancel" button to the right side on all devices. Thoughts?

IsaacHardy: Yea, I think that is an easy fix. Go for that.

conrad-vanl: Done.

PR #360, opened 2018-10-25
add `PaginationInfo` and `pageInfo` to our schema

conrad-vanl:
[ ] TODO: flush out this PR description

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-church-api/src/data/content-items/index.js
```diff
@@ -364,4 +364,7 @@ export const resolver = {
     title: ({ title }) => title,
     message: () => '',
   },
+  ContentItemsConnection: {
```
conrad-vanl: Would a better way to do this is to make `ContentItemsConnection` itself a function that just returns `root`? that might make more sense....

conrad-vanl:
```suggestion
  ContentItemsConnection: (root) => root,
```

vinnyjth: I'm fine with the way we are doing this right now.

packages/apollos-church-api/src/data/content-items/index.js
```diff
@@ -364,4 +364,7 @@ export const resolver = {
...
+    pageInfo: (root) => root,
```
vinnyjth: I don't think you need this line :) 
```suggestion
```

conrad-vanl: I thought that at first two, but without this line, the query always return `null`. That's why I was thinking that perhaps moving this function up to `ContentItemsConnection` might be better (my other comment). I believe `ContentItemsConnection` needs a resolver?

vinnyjth: OH okay your other comment makes more sense now. I think moving the logic that creates the `pageInfo` cursor would be better off at the `ContentItemsConnection` level, only because it will let other items that don't return edges use the `startCursor` and `endCursor` logic. I can make the changes myself if you're fine with that.

conrad-vanl: sounds good!

packages/apollos-church-api/src/data/content-items/index.js
```diff
@@ -364,4 +364,16 @@ export const resolver = {
     title: ({ title }) => title,
     message: () => '',
   },
+  ContentItemsConnection: {
+    pageInfo: ({ edges }) => ({
+      startCursor: async () => {
```
conrad-vanl: Why is this better then the way we had it? This seems like we don't really have any point for the `Pagination` resolver now...

vinnyjth: It's better because now we can support different implementations of `pageInfo`. For example, if we add a connection that keeps the cursor on a different field or something. Trying to decouple `PaginationInfo` from the specific implementation of `ContentItems`.

vinnyjth: that's just my $0.02. If you want to revert until we need it, I'm fine with that :)

conrad-vanl: I'm just wondering at this point why have the `PaginationInfo` resolver at all? It's not really doing anything....this resolver still has to do "all the work"

conrad-vanl: and, wonder if there's a less but better path here. Before, the resolver was based on having `edges` and working off of that. That seems pretty repeatable (`PaginationInfo` will always be next to an `edges` attribute). It seems like we'll probably end up with duplicated logic looking at the beginning and end of that `edges` array now whenever we add pagination to a resource. Is that correct?

vinnyjth: Yah  - good point. In `crossroads` we got in the habit of creating HOCs for commonly used resolvers that worked really well. I'll try that pattern here and push it up.

vinnyjth:
@conrad-vanl This is probably the best version of the code thus far. Give it a shot :)

PR #361, opened 2018-10-25
Fixes Dockerfile, `yarn link`, and package naming

vinnyjth:

- Removes `link:/` and replaces it with `file:/`. This removes the warnings when running `yarn install`, - Rename `apollos-rock-apollo-data-source`  to `@apolloschurch/rock-apollo-data-source`
- Fixes the dockerfile so it will pick up new packages automatically in the future.
- Removes the need to specify `port` in `.env`

The code I use to optionally specify PORT is ugly. If you try and do `port: null` or just pass an arg of undefined, it will choose a random port. If you default to `4000` here, then the tests break because they need a random port specified. Try it out if you don't trust me ;) 

### How do I test this PR?

Boot the API and make sure the app still works as expected.

N/a

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


Dockerfile
```diff
@@ -1,12 +1,7 @@
 FROM node:8-alpine
-RUN mkdir -p /usr/src/app
-RUN mkdir -p /usr/src/babel-preset-apollos
-RUN mkdir -p /usr/src/apollos-rock-apollo-data-source
+COPY ./packages /usr/src/
+RUN mv /usr/src/apollos-church-api /usr/src/app
```
conrad-vanl: Under "Edge cases: what assumptions are made about input?", this `Dockerfile` assumes that we're _always_ wanting to run/build the `-api`, and ignores the mobile side.

That's fine for now, but in the future, HMW better package this?

packages/apollos-church-api/package.json
```diff
@@ -31,7 +31,7 @@
     "analytics-node": "^3.3.0",
     "apollo-datasource-rest": "0.1.1",
     "apollo-server": "^2.0.5",
-    "apollos-rock-apollo-data-source": "file:../apollos-rock-apollo-data-source",
+    "@apolloschurch/rock-apollo-data-source": "file:../apollos-rock-apollo-data-source",
```
conrad-vanl: Only thing I noticed, but I think this happens on master, is that if you make a change to `-data-source` while `-api` is running, `-api` doesn't restart on it's own.

vinnyjth: Yah that's true :/ I don't have a great solution for this yet but I'm actively trying to come up with something.

PR #363, opened 2018-10-29
Fixes Recently Liked

### LINKED ISSUE #347: Fixes to recently liked query
#### Problem Details

It appears the recently liked query is broken in the following ways:

- The items in my recently liked feed- when I tap on them, it doesn't show that I've liked them
- This creates a bug that if I open that content, re-like it, then un-like it again, it doesn't remove the item from the feed. It's forever locked in my feed
- When I like new content, it gets added to the end of the feed, instead of the beginning of the feed

vinnyjth:

- The "Recently Liked" screen previously took a while to update when liking / unliking items. This PR fixes that. 
- The "Recently Liked" screen previously did not have a consistent order. This PR also fixes that and sorts items by most recently liked.

- I saw some performance improvements that could be made on the API side that I did not make at this time b/c I want to keep the size of this PR down.

### How do I test this PR?

1. Launch the app.
2. Unlike all items from your "Recently Liked" feed. They should disappear from the feed as soon as you unlike them (no need to refresh or wait).
3. Like a bunch of items on the home feed. They should appear in the "Recently Liked" feed at once, without needing to refresh. They should also be sorted with the most recently liked first.

## TODO:

- [x] Closes #347 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


IsaacHardy:
@vinnyjth I am having local API issues so I'm currently unable to test this locally. Still trying to debug this network issue.

packages/apollos-church-api/src/data/content-items/index.js
```diff
@@ -280,46 +280,67 @@ export const resolver = {
     getAllLikedContent: async (root, args, { dataSources }) => {
       // Get All Interactions from current user
       const interactions = await dataSources.Interactions.getForContentItems();
```
conrad-vanl: did you consider constructing any of the below logic into a better ODATA query?

packages/apolloschurchapp/src/ui/LikeButton/updateLikedContent.js
```diff
@@ -0,0 +1,74 @@
...
+import { contentItemFragment } from 'apolloschurchapp/src/content-single/getContentItem';
+
+const addItemToLikedContentList = ({ cache, item }) => {
+  try {
+    const data = cache.readQuery({ query: getAllLikedContent });
+    const fullItem = cache.readFragment({
+      id: `${item.__typename}:${item.id}`,
+      fragmentName: 'LikedContentItemParts',
+      fragment: gql`
+        fragment LikedContentItemParts on ContentItem {
```
conrad-vanl: Do we need this full fragment? And if so, since this is touching the `getAllLikedContent` query, can we add this fragment to live alongside that query, and import it here? That way if the query changes in the future, this doesn't break?

packages/apollos-church-api/src/data/content-items/index.js
vinnyjth: I did, came up with two ideas, neither of which I implemented because I implemented them in the `interactions` PR.

First one was changing `resolveUserContentFromInteractions` to fetch all at once rather than doing N fetches. I implemented this in a branch I'm working on called 348-interactions-cleanup
Second was preloading the Entity with `getForContentItems`. 

I'm hesitant to make these changes here b/c this file will change as soon as we reimplement Like/Unlike to use `Following` (something I think we should do next sprint)

packages/apolloschurchapp/src/ui/LikeButton/updateLikedContent.js
vinnyjth: We do need this full fragment, otherwise the `getAllLikedContent` query will be missing attributes. I'm an advocate of changing `getAllLikedComment`to use the `... contentItemFragment`, which I will implement :D Good call.

PR #365, opened 2018-10-30
Android Like Button Fix

redreceipt:
## DESCRIPTION

kapture 2018-10-30 at 15 21 24

The like button on Android devices is unclickable, this fixes that.

The `TouchableNativeFeedback` component [can only accept](https://facebook.github.io/react-native/docs/touchablenativefeedback) a single `View` component as a child which was not how our Like button was implemented ... *I think*. Refactoring the `compose` HOC fixed it.

### How do I test this PR?

Click the Like button and watch the magic happen.

None...yet :)

## TODO:

- [x] Closes [#331]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apolloschurchapp/src/ui/Touchable/index.js
```diff
@@ -1,16 +1,11 @@
...
 
 const IOSTouchable = TouchableOpacity;
 
 IOSTouchable.defaultProps = {
   activeOpacity: 0.5,
 };
 
-const Touchable =
-  Platform.OS === 'android' ? TouchableNativeFeedback : IOSTouchable;
+const Touchable = Platform.OS === 'android' ? TouchableHighlight : IOSTouchable;
```
conrad-vanl: This incurs a rather significant change - as the `<Touchable/>` component is used _all over_ the app. Does not seem reasonable to change the touchable here since it impacts things application-wide.

Could we  just refactor the `Like` button to have a single child? Perhaps wrap the current children in a single `<View>` ?

redreceipt:
@conrad-vanl @IsaacHardy any idea as to why the refactoring fixed it given the knowledge that `TouchableNativeFeedback` only can accept a single `<View>` component? Also, why the two props were added to the snap?

packages/apolloschurchapp/src/ui/Like/index.js
```diff
@@ -1,19 +1,17 @@
...
 
-const enhance = compose(
-  pure,
-  withTheme()
-);
-
-export const LikeIcon = enhance(({ isLiked, theme }) => (
-  <Icon name={isLiked ? 'like-solid' : 'like'} fill={theme.colors.secondary} />
-));
+const LikeIcon = compose(
```
vinnyjth: If using a single `HOC`, there's not a good reason to use `compose`. You can probably eliminate `compose` if you have decided you don't need the `pure` `HOC`.

burkeshartsis: @vinnyjth is right. This slipped by me while we were pairing!

redreceipt: done.

conrad-vanl:
@redreceipt the two props were added to the snap because `styled` passes all props down into it's child component. So those props, although used by `styled`, now get passed into the view it's rendering.

As for your other question - this change shouldn't affect the bug that this PR is aiming to fix, so I'm not sure. Are you positive the issue no longer exists??

burkeshartsis:
@conrad-vanl I can confirm it fixed the issue.

Also these changes don't use `styled` just `withTheme`. Does `withTheme`, which is basically a `withProps` under the hood, also pass its props down to its children?

conrad-vanl:
@burkeshartsis that is correct. `withProps` passes all existing props down, just adds new props.

conrad-vanl:
For `react-native`, that practice is fine.

Just a no-no on `react` for the web

redreceipt:
@vinnyjth wanna green check me so we can merge?

PR #380, opened 2018-11-01
Add Content Channels Support to Apollos Config

vinnyjth:

Allows the user to configure content channels used in the config.yml file.

n/a

### How do I test this PR?

- Boot the API and the App
- Remove an id from the `config.yml` under `ROCK_MAPPINGS -> DISCOVER_CONTENT_CHANNEL_IDS` and make sure the associated disappears from the home feed.
- Remove an id from the `config.yml` under `ROCK_MAPPINGS -> FEED_CONTENT_CHANNEL_IDS` and make sure the associated disappears from the category feed.

Updated existing tests to use new functionality 

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-church-api/src/config.js
+++ b/packages/apollos-church-api/src/config.js
@@ -1,4 +1,5 @@
 import path from 'path';
+import dotenv from 'dotenv/config'; // eslint-disable-line
 import ApollosConfig from '@apolloschurch/config';
 
 ApollosConfig.loadYaml({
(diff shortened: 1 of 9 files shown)


vinnyjth:
> I was getting the same issue on the cards branch i'm working on, but I think we need to move the dotenv import to be above the apollos config import...

Not getting this problem, but your change makes perfect sense.

>Otherwise...I think this looks good. Locally I can't get tests to pass on apolloschurchapp from within this branch, and I think it's some sort of compile cache issue...is there a way to reset it? I'm seeing errors that definitely aren't in the code on this branch (but were on another branch I was working on) and I've tried lerna run build from the root

Try `yarn install --force` :) I've noticed that `yarn install` doesn't update local packages in all cases.

conrad-vanl:
Weird...wonder why you're not getting that problem....

PR #414, opened 2018-11-06
Adds `@apolloschurch/core-api` package

vinnyjth:
The goal of this logic is store logic/schema parts that can be shared across all integrations of the apollos project. Right now it's just a bunch of helper functions to generate a schema/resolvers and the `node` type, but in the future we might be able to include more.  

n/a

### How do I test this PR?

0. `yarn install`
1. Boot up API
2. Ensure app functionality hasn't changed

n/a, moved tests 

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
@vinnyjth I haven't looked at this yet (😇) ... but any thoughts around creating like a core apollos "data" package that contains core data code that's isomorphic? Something that exports the schema and utility functions we use on both the backend and frontend? Those seem to be highly dependent on each other, but I guess the dependencies that are required between front-end and back-end might not let it make sense?

Been trying to figure out what separation of packages we want or need...

On the front-end, there seems to be three big buckets:

- Core components - things like Buttons
- Native components - things like media players and content feeds. Some of these could be data-bound somehow
- Data - everything from `react-apollo` to client-side state

vinnyjth:
@conrad-vanl I am all for that - as it stands right now, I'm not sure what in this PR will be of any value to the API. On the API, I see two big buckets

- Utility functions and classes (stuff like the `createGlobalId` and `ApolloDataSourceRest`) 
- All-in-one data models (resolvers/schema/datasource) for a specific part of the API (like `Auth` or `Live`).

packages/apollos-church-api/package.json
```diff
@@ -28,6 +28,7 @@
     "testEnvironment": "node"
   },
   "dependencies": {
+    "@apollosproject/rock-apollo-data-source": "file:../apollos-rock-apollo-data-source",
```
conrad-vanl: Is this a duplicate line now?

packages/apollos-church-api/src/data/index.js
```diff
@@ -33,31 +40,36 @@ const data = {
   Analytics,
   Family,
   Pagination,
+  UniversalContentItem: {
+    model: ContentItem.model,
```
conrad-vanl: Are we still using `model` for anything? I thought it was all just `dataSource` now?

packages/apollos-church-api/src/data/index.js
```diff
@@ -33,31 +40,36 @@ const data = {
...
 
-export const schema = [
-  gql`
-    type Query {
-      _placeholder: Boolean # needed, empty schema defs aren't supported
-    }
-
-    type Mutation {
-      _placeholder: Boolean # needed, empty schema defs aren't supported
+export const dataSources = createDataSources(data);
```
conrad-vanl: What's the value of pushing these `createDataSources`, `createResolvers`, etc functions into our lib code? All they do is create an object mapping right? Would it make more sense to leave that code here since there's not any magic that's going on?

conrad-vanl: Thinking through this some more....

I think the only "weird" thing to me from an api perspective with these functions, is that the `createResolvers` function, for example, expects an object that looks like:

```
{
   ResolverName: {
      resolver: { ... }
   },
}
```

and in reality, the object actualy looks like:

```
{
   ResolverName: {
      resolver,
      dataSource,
      schema,
      // .... etc
   }
}
```

which is kind of an "obscure" api from a implementation perspective - like - someone comes along and reads this code and has a little bit of a "wtf is going on?" reaction (I would, and I wrote most of this logic originally 🤭).

The `createResolvers` function actually doesn't even _need_ the `dataSources` or `schema` fields. Which is why this is a code smell, in my mind.

Anyways, would it make more sense from an api/implementation perspective if we ditched the key mapping for these functions, and have the intended input for the `createResolvers` function (and the others) to look like:

```
{
   ResolverName: resolver
}
```

The other option...bundle these different functions `createDataSources`, `createResolvers`, etc into one function that does it all....

```
export createApolloServerConfig(data);
```

packages/apollos-church-api/src/data/index.js
```diff
@@ -33,31 +40,36 @@ const data = {
...
+  UniversalContentItem: {
+    model: ContentItem.model,
+    dataSource: ContentItem.dataSource,
+  }, // alias
+  DevotionalContentItem: {
+    model: ContentItem.model,
+    dataSource: ContentItem.dataSource,
+  }, // alias
 };
+// UniversalContentItem: ContentItem.model, // alias
```
conrad-vanl: Can this comment line be removed?

vinnyjth: yep

packages/apollos-church-api/src/data/index.js
vinnyjth: You know, I was under the assumption that we did, but after looking through all the model files, you are correct :) I will remove this logic and the files.

packages/apollos-church-api/src/data/index.js
vinnyjth: @conrad-vanl Which of these options do you like more? I really like the idea of exporting a single function which, given data, returns `{ context, dataSources, schema }` 

I can see how passing in the same object to each function is confusing, especially because each function only uses a part of that objec.t

conrad-vanl: I feel like the single function might result in less code overall...

and gives us the opportunity to tie in relationships/hooks between these functions in the future....

For example - `context` currently sets up some auth-bits. But the auth-bits are dependent on which auth-data-module you're using....so having one function with access to the entire data stack might make some that interoperability in the future???

vinnyjth:
@conrad-vanl Made your changes, merging.

Issue #382, opened 2018-11-02
Media player control tapability issues


burkeshartsis:
## User Story

As a `<type of user>` _(who)_ I want to `<perform an action>` _(what)_ so that I can `<achieve a goal>` _(why)_.

#### Problem Details

Investigate sizing, padding, and hitSlop (Android) to improve tapability of all buttons.

#### Acceptance Criteria _(Optional)_

* [ ] It should [SPEC_DETAIL]

PR #416, opened 2018-11-09
Fixes native button ripple on Android

burkeshartsis:
This fixes the native button ripple effect on Android when a `Touchable` has rounded corners (e.g. every button in the app!).

This fix just wraps the `TouchableNativeFeedback` that is rendered on Android with a `View` and exposes a `borderRadius` prop on `Touchable` to round the corners of the wrapping `View`. This fix doesn't just fix our buttons though. Any `Icon` that is wrapped in a `Touchable` can now elegantly render Android's ripple effect 🕺

This also sets the Android ripple color based on the Theme `type` (light/dark) so that the color is always appropriate to background 🕺🕺 

#### After
nov-09-2018 11-41-21

#### Theme driven responsive ripple color!
nov-09-2018 12-45-23

NONE! 💪

### How do I test this PR?
Storybook has examples of buttons. You can also in app for places known to use `Touchable` or `Button`

Coverage apparently didn't drop so I didn't add any but there are totally new untested props.

## TODO:

- [x] Partial fix for #382 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apolloschurchapp/src/ui/Touchable/index.js
```diff
@@ -1,16 +1,45 @@
...
+
+const Touchable = ({ borderRadius, children, ...otherProps }) => (
+  <AndroidWrapper borderRadius={borderRadius}>
+    <PlatformTouchable {...otherProps}>{children}</PlatformTouchable>
+  </AndroidWrapper>
+);
+
+Touchable.propTypes = {
+  borderRadius: PropTypes.number,
+  children: PropTypes.node,
```
conrad-vanl: This is a good opportunity to make sure we throw a prop type warning if `children` contains multiple elements, since that is invalid on android. I believe `PropTypes.node` will...maybe we should leave a comment here about how on android, `children` most be a single element?

burkeshartsis: Good idea. https://reactjs.org/docs/typechecking-with-proptypes.html#requiring-single-child

burkeshartsis: fixed in 663f3f6

burkeshartsis:
@conrad-vanl the media player buttons may not look/function like the gif above. I have another PR coming to make that work correctly. That just shows that it's possible 😉

Issue #250, opened 2018-08-28
Investigate possible media casting functionality


burkeshartsis:
## User Story

As a `<type of user>` _(who)_ I want to `<perform an action>` _(what)_ so that I can `<achieve a goal>` _(why)_.

#### Problem Details _(Optional)_

- platform specific casting (ios = airplay, android = google cast)
- bonus points for google on both 💥

#### Acceptance Criteria _(Optional)_

* [ ] It should [SPEC_DETAIL]

burkeshartsis:
possibly the most straight forward ios solution. https://github.com/disjfa/react-native-airplay-btn

burkeshartsis:
related way to possibly improve the airplay UX a bit in app.

https://github.com/react-native-community/react-native-video/blob/639f2bbd4cf7da89d303adc5f29af1a2c167a87e/Video.js#L377

redreceipt:
https://github.com/react-native-google-cast/react-native-google-cast

conrad-vanl:
@redreceipt open a PR for your branch and then we can close this issue :)

redreceipt:
#422

PR #422, opened 2018-11-14
Adds Airplay Implementation

redreceipt:
DESCRIPTION
-----------

kapture 2018-12-10 at 16 16 45

This gives the user the ability to stream via Apple Airplay and Google Cast.

TODO:
-----

- [ ] Closes [#250]
- [ ] Set a relevant reviewer

REVIEW:
-------

**Manual QA**


conrad-vanl:
image

😲 

Looks like something needs to be added to the gitignore?

redreceipt:
Maaaaaybe

redreceipt:
A lot of work to be done, but it works!

packages/apolloschurchapp/src/ui/MediaPlayer/FullscreenControls.js
```diff
@@ -135,11 +157,21 @@ class FullscreenControls extends PureComponent {
...
   componentWillUnmount() {
     this.backHandler.remove();
     if (this.closeTimeout) clearTimeout(this.closeTimeout);
+
+    // remove airplay handlers
+    this.airPlayConnected.remove();
+    this.airPlayAvailable.remove();
+
+    // TODO: when should we disconnect?
+    // AirPlay.disconnect();
```
redreceipt: @conrad-vanl when do you think is the right time to disconnect Airplay? If ever?

packages/apolloschurchapp/src/ui/MediaPlayer/FullscreenControls.js
```diff
@@ -244,7 +276,8 @@ class FullscreenControls extends PureComponent {
                     <Title>{get(mediaPlayer, 'currentTrack.title')}</Title>
                     <Artist>{get(mediaPlayer, 'currentTrack.artist')}</Artist>
                   </Titles>
-                  <IconSm name="empty" />
+                  <AirPlayButton />
+                  {/* <IconSm name="empty" /> */}
```
redreceipt: @burkeshartsis what's the best way to approach this UI to make it right? The icon needs a little margin on the right but it looks pretty good now (see the image in PR description). Can the button be wrapped in `IconSm`?

packages/apolloschurchapp/src/ui/MediaPlayer/FullscreenControls.js
burkeshartsis: You'll probably want to move all things airplay related to a separate component like we did with `MusicControls` and then you'll probably "render" it [here](https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apolloschurchapp/src/ui/MediaPlayer/FullscreenPlayer.js#L200). Then pass what hooks you need as props to the `FullScreenControls` so you can selectively render the AirPlay icon when airplay is available. 

Based on how `MusicControls` is setup you want your component to be mounted and unmounted similarly. So `Airplay.disconnect()` would live in the airplay component I recommended above inside that components `componentWillUnmount`.

packages/apolloschurchapp/src/ui/MediaPlayer/FullscreenControls.js
burkeshartsis: Update your branch with master and `IconSm` will use `ButtonIcon` and should look good.

packages/apolloschurchapp/src/ui/MediaPlayer/FullscreenControls.js
redreceipt: @burkeshartsis help me understand why we would use `FullScreenPlayer` at all here?

Seems confusing to put the `AirplayStuff` component in `FullscreenPlayer` and the `AirplayButton` in `FullscreenControls` when we could just put both of them in `...Controls`.

burkeshartsis: I think we'll want to schedule a talk to dive into how the media player works. The short is, `FullScreenPlayer` is what renders `FullScreenControls`. Depending on our approach we may want to move the Airplay stuff higher up the stack or likely just into it's own component like we do with `MusicConrols` in which case we'll need to manage it's visibility in state higher up the chain.

packages/apolloschurchapp/src/ui/MediaPlayer/FullscreenControls.js
redreceipt: When I wrap it in `IconSm` it doesn't load, is there something I need to do to use it?

redreceipt:
This is going to need some QA testing

packages/apolloschurchapp/ios/apolloschurchapp.xcodeproj/project.pbxproj
```diff
@@ -1575,6 +1600,13 @@
...
 			sourceTree = BUILT_PRODUCTS_DIR;
 		};
+		DC7D3D0521A32C8500626550 /* libRNAirplay.a */ = {
+			isa = PBXReferenceProxy;
+			fileType = archive.ar;
+			path = libRNAirplay.a;
+			remoteRef = DC7D3D0421A32C8500626550 /* PBXContainerItemProxy */;
+			sourceTree = BUILT_PRODUCTS_DIR;
+		};
 /* End PBXReferenceProxy section */
```
redreceipt: @burkeshartsis after you build on your machine, commit the changes so this part goes back to normal :)

packages/apolloschurchapp/src/ui/MediaPlayer/FullscreenControls.js
```diff
@@ -91,6 +94,9 @@ const IconLg = withTheme(({ theme }) => ({
   iconPadding: theme.sizing.baseUnit * 0.3125,
 }))(ButtonIcon);
 
+const AirPlayContainer = styled({
+  padding: 40,
```
conrad-vanl: This should use a theme value, and ideally should be an even derivative of `baseUnit`. `baseUnit` is 16px I believe, so the closest would be one of these:

```javascript
const AirPlayContainer = styled(({ theme }) => ({
  padding: theme.sizing.baseUnit * 2,
}))(View);

// ... or

const AirPlayContainer = styled(({ theme }) => ({
  padding: theme.sizing.baseUnit * 3,
}))(View);
```

packages/apolloschurchapp/src/ui/MediaPlayer/FullscreenControls.js
```diff
@@ -262,7 +268,10 @@ class FullscreenControls extends PureComponent {
                     <Title>{get(mediaPlayer, 'currentTrack.title')}</Title>
                     <Artist>{get(mediaPlayer, 'currentTrack.artist')}</Artist>
                   </Titles>
-                  <IconSm name="empty" disabled />
+                  {/* TODO: this is terrible */}
+                  <AirPlayContainer>
```
conrad-vanl: What's keeping this from showing up on android?

Issue #409, opened 2018-11-06
Empty Scripture Tab


redreceipt:
## User Story

As a `<type of user>` _(who)_ I want to `<perform an action>` _(what)_ so that I can `<achieve a goal>` _(why)_.

#### Problem Details _(Optional)_

When there's no scripture in a devotional, there should be no tab. Right now it's just a blank empty one.

![20181106_093050](https://user-images.githubusercontent.com/2659478/48070931-fa99bc00-e1a6-11e8-9441-2cede02f2d4f.gif)

* Screenshots, GIFs, or a video
* Any additional information (version, device, etc.)

#### Acceptance Criteria _(Optional)_

* [ ] It should [SPEC_DETAIL]

PR #427, opened 2018-11-15
Fix Empty Scripture Tab

redreceipt:

When there's no scripture in a devotional, it was showing a blank tab. This fixes that. See #409.

kapture 2018-11-15 at 11 21 32

I don't think anything...but I did see one time that it tried to preload the scripture tab and then it disappeared. Not sure if this will happen every time or not. Might have been cached on my system or something.

### How do I test this PR?

Go to a devotional with no scripture. 

None.

## TODO:

- [x] Closes [#409]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
I don't remember what I'm supposed to do with `jest`. Also, not sure why Travis is failing, didn't change much.

packages/apolloschurchapp/src/content-single/DevotionalContentItem/index.js
```diff
@@ -79,7 +78,11 @@ class DevotionalContentItem extends PureComponent {
     loading,
   }) => {
     if (error) return <ErrorCard error={error} />;
-    const hasScripture = loading || scriptures.length;
+    // only include scriptures where the references are not null
+    const validScriptures = scriptures.filter(
+      (scripture) => scripture.reference != null
+    );
+    const hasScripture = loading || validScriptures.length;
```
redreceipt: @burkeshartsis do we need the `loading` here? I think this is why I've seen it pre-load the scripture tab and then disappear.

burkeshartsis: The approach we take with loading and its goal is to pre-render the UI so that when content is eventually available the user already has context for what they are looking at and how they will interact. For the time being, we should probably error on the side of the majority here, in that most devos will have scripture. I'm not imposing it as a requirement moving forward as I could see a devotion based on a different book but we are really engineered for that at the moment.

I think the best solution is to render the tab ui in a loading state along with the screen content. Then when the content is ready render it all and hide the scripture tab if we don’t have scripture. This minimizes transitions in an already supported way.

See #434 to track this moving forward.

burkeshartsis:
@redreceipt it's not your fault. I believe @vinnyjth just merged a PR to fix it.

burkeshartsis:
@redreceipt so your tests aren't passing because your change affected the snapshots. `yarn test` will show you which tests are failing and what you need to address.

redreceipt:
@burkeshartsis the same two tests are failing in both of my open PRs on files that I didn't touch, not sure where to go from here...

```
Summary of all failing tests
FAIL src/FeedView/FeedView.tests.js
  ● Test suite failed to run

    Cannot find module 'numeral' from 'Metrics.js'



      at Resolver.resolveModule (../../node_modules/jest-resolve/build/index.js:221:17)
      at Object.<anonymous> (src/ContentCard/Metrics.js:1396:89)

FAIL src/ContentCard/ContentCard.tests.js
  ● Test suite failed to run

    Cannot find module 'numeral' from 'Metrics.js'



      at Resolver.resolveModule (../../node_modules/jest-resolve/build/index.js:221:17)
      at Object.<anonymous> (src/ContentCard/Metrics.js:1396:89)
```

packages/apolloschurchapp/src/content-single/DevotionalContentItem/index.js
```diff
@@ -40,9 +40,8 @@ class DevotionalContentItem extends PureComponent {
    * Returns: an array of scripture references.
    */
   getScriptureReferences = (scripture) => {
-    if (scripture && scripture.length) {
-      return scripture.map((ref) => ref.reference);
-    }
+    // only add refs to the array if not null
+    if (scripture) return scripture.filter((ref) => ref.reference != null);
```
burkeshartsis: Here is the reason your tests keep failing. `getScriptureRefrences()` used to return a new array of ref strings (`['1 John 3: 16']`). Your use of filter returns the entire scripture object (`[{id: '1 John 3: 16', html: '<htmlbits ../>}]`). Now while you didn't make any changes to any components, down the render chain `ScriptureList` expects an array of strings. Which resulted in the app rendering `[object Object]` on the screen and snapshots failing because they weren't rendering the scripture references. Funny enough, what `getScriptureRefrences()` takes in and expects to return is even documented in the doc block above in lines 37-41. Suffice to say, this is EXACTLY why we write snapshot tests. I went ahead and fixed this in 51b0442

PR #432, opened 2018-11-16
Adds Rock content data connector

vinnyjth:
## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-data-connector-people/.npmignore
```diff
@@ -0,0 +1,2 @@
+*
+!lib/**
```
conrad-vanl: ❓ on this. on `ui-kit`, I had to add a `files` key to the `package.json` and explicitly include `src/**` (since on react-native packages there's no reason to-precompile them). Doing so also effectively ignores all other files in the package. Would that be better/worse then using `.npmignore`?

packages/apollos-data-connector-rock-content/src/content-channels/index.js
```diff
@@ -1,28 +1,8 @@
...
-    ): ContentItemsConnection
-
-    iconName: String
-  }
-
-  extend type Query {
-    contentChannels: [ContentChannel]
-  }
-`;
+export { contentChannelSchema as schema } from '@apollosproject/data-schema';
```
conrad-vanl: Nice!

packages/apollos-data-connector-people/.npmignore
vinnyjth: I found that creating this file was easier to do across all projects than creating a `files` key. I think either approach is great, and I think I should have made the file differently for `ui-kit` because everything is kept in `src`.

PR #465, opened 2018-12-11
Filters by active in content items data source

vinnyjth:
In the past, items would show up in the app before they went live or after they expire. Many churches use those dates to force content to go live or to remove content from their feed, so we need to respect the values of those properties. This PR does that. 

I also added some methods to the `request-builder.js` so that request filter chaining should be more concise and read better.

n/a

### How do I test this PR?
1. Open the app w/ local API.
2. Examine the content item feed and pick an item.
3. In Rock, edit the Active At or Expires At dates to make the item inactive.
4. Refresh the content feed in the app, the item should disappear.  

Updated existing tests.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
@vinnyjth what's the instance of Rock this is pulling from? apollosrock.newspring.cc doesn't seem like the right content?
screen shot 2018-12-13 at 8 46 00 pm

vinnyjth:
@redreceipt the different content is due to all of the content you normally see being expired :P w/ we need to login to the rock instance and change the dates on all content that _should_ show up in the app now.

conrad-vanl:
@vinnyjth I'd say that content _really_ looks like the content from NewSpring's rock instance, not Apollo's?

vinnyjth:
@conrad-vanl True. @redreceipt, I'm assuming you have the wrong values in your .env file. Compare what you have against one of your teammates values, or the values in basecamp.

packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -2,6 +2,8 @@ import RockApolloDataSource from '@apollosproject/rock-apollo-data-source';
 import ApollosConfig from '@apollosproject/config';
 
 const { ROCK_MAPPINGS } = ApollosConfig;
+const LIVE_CONTENT = (date = new Date()) =>
```
conrad-vanl: I'm fairly certain that `new Date()` gets evaluated once at initial app boot, so once the date changes this default value will be wrong...I think you _have_ to do something like:

```
const LIVE_CONTENT = (date) => {
   if (!date) date = new Date();
```

vinnyjth: @conrad-vanl ooooh good catch, but I think in Javascript we are okay
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Evaluated_at_call_time

conrad-vanl:
Actually, it might appear that I'm incorrect, as I think this is doing the same thing: 

image

packages/apollos-data-connector-rock/src/content-items/data-source.js
conrad-vanl: yeah! I'm noticing that now, okay, back to 👀  i'll have this approved shortly me thinks

vinnyjth:
@conrad-vanl yep - MDN says that args defaults are evaluated at runtime
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Evaluated_at_call_time

Which honestly I didn't even consider, so I"m glad you brought it up.

PR #488, opened 2019-01-17
Fixes timezone of dates sent to rock when filtering content items

vinnyjth:

This PR does two things

1. Rock expects all dates passed to it to be in the timezone of the rock instance (which was Eastern, for the ChristFellowship folks, and is the same for ApollosRock). If you pass it ISO8601 strings (which we do when filtering) it ignores the timezone of the string. So if you send it UTC 5:00PM, it interprets it as EST 5:00PM. This was leading to content items not appearing or disappearing in the app at the correct time.

This PR fixes that behavior, and allows users to set a timezone in the app, and all dates will be offset by the appropriate offset so as to match the timezone of Rock.

2. This app sets up the `CONTENT_LIVE` filter as a property of the ContentItem data source, so it can be overridden by implementations as needed. That way churches that need to change the logic won't be dependent on a PR to core. 

n/a
### How do I test this PR?
n/a
Existing test still work. 

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-data-connector-rock/src/content-items/data-source.js
+++ b/packages/apollos-data-connector-rock/src/content-items/data-source.js
@@ -6,12 +6,21 @@ const { ROCK_MAPPINGS, ROCK } = ApollosConfig;
 
 export default class ContentItem extends RockApolloDataSource {
   resource = 'ContentChannelItems';
+
   LIVE_CONTENT = () => {
     // get a date in the local timezone of the rock instance.
     // Rock doesn't respect timezone information, hence the need to warp the date
-    const date = moment().utc().add(moment().tz(ROCK.TIMEZONE).utcOffset(), 'minutes');
+    const date = moment()
+      .utc()
+      .add(
+        moment()
+          .tz(ROCK.TIMEZONE)
+          .utcOffset(),
+        'minutes'
+      );
     return `((StartDateTime lt datetime'${date.toISOString()}') or (StartDateTime eq null)) and ((ExpireDateTime gt datetime'${date.toISOString()}') or (ExpireDateTime eq null)) `;
-  }
+  };
+
   expanded = true;
 
   getCursorByParentContentItemId = async (id) => {
--- a/packages/apollos-data-connector-rock/src/content-items/data-source.js
+++ b/packages/apollos-data-connector-rock/src/content-items/data-source.js
@@ -1,13 +1,17 @@
 import RockApolloDataSource from '@apollosproject/rock-apollo-data-source';
 import ApollosConfig from '@apollosproject/config';
+import moment from 'moment-timezone';
 
-const { ROCK_MAPPINGS } = ApollosConfig;
-const LIVE_CONTENT = (date = new Date()) =>
-  `((StartDateTime lt datetime'${date.toISOString()}') or (StartDateTime eq null)) and ((ExpireDateTime gt datetime'${date.toISOString()}') or (ExpireDateTime eq null)) `;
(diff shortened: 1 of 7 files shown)


packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -1,13 +1,19 @@
...
 
-const { ROCK_MAPPINGS } = ApollosConfig;
-const LIVE_CONTENT = (date = new Date()) =>
-  `((StartDateTime lt datetime'${date.toISOString()}') or (StartDateTime eq null)) and ((ExpireDateTime gt datetime'${date.toISOString()}') or (ExpireDateTime eq null)) `;
+const { ROCK_MAPPINGS, ROCK } = ApollosConfig;
 
 export default class ContentItem extends RockApolloDataSource {
   resource = 'ContentChannelItems';
 
+  LIVE_CONTENT = () => {
```
conrad-vanl: Why the capitalization on this method name?

conrad-vanl: Oh, it was a const above before. I'm fine with it like this, just wondering if this fits our standards/coding style

packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -1,13 +1,19 @@
...
+    // get a date in the local timezone of the rock instance.
+    // Rock doesn't respect timezone information, hence the need to warp the date
+    const date = moment().utc().add(moment().tz(ROCK.TIMEZONE).utcOffset(), 'minutes');
+    return `((StartDateTime lt datetime'${date.toISOString()}') or (StartDateTime eq null)) and ((ExpireDateTime gt datetime'${date.toISOString()}') or (ExpireDateTime eq null)) `;
```
conrad-vanl: so the `toISOString` - what does it output now? Does it show the `America/New_York` tz? Does it use `UTC` time? in other words, say in the future Rock does start supporting timezones in the way we'd expect it to....would this theoretically still work? Said in a third way: are we sending the same time, just in another timezone, with the new timezone still included in the iso string?

vinnyjth: It outputs a string with no apparent timezone. So, if Rock started respecting timezones we should be okay as long as Rock doesn't start assuming all new Timezones are UTC instead of local time

conrad-vanl: Are you sure it doesn't contain a timezone? Pretty sure `ISOString` should have one. Even if it's just a `Z` at the end of the timestring

conrad-vanl: (`Z` would indicate UTC, fwiw)

vinnyjth: yah you're totally right actually. I forgot that Z === zulu === utc/gmt. 

Not sure the right way to handle this besides changing that string to remove the Z. Maybe we should circle up with John from Rock to get an idea about time zone parsing future capabilities and plans. 

Otherwise, Next course of action I would take is explicitly removing the Z from the string.

conrad-vanl: so with the `Z` included I'm imagining this makes the date technically incorrect

vinnyjth: yah totally. I need to confirm that rock won't throw an error if exclude the Z

Issue #494, opened 2019-01-30
Create a ready-to-go caching solution for the API


vinnyjth:
We can mimic work that we did in Christ Fellowship. This is a very nuanced and several step task. 

1. Add cache hints to the schema. You use syntax like `@cacheControl(maxAge: 200)`. We should start by setting a reasonable default (10 minutes) and then overiding that default for things we should cache longer, and things we shouldn't cache. You also need to configure `apollo-server` to send the cacheControl headers. 

2. Caching only works w/ `GET` requests. ApolloClient is configured to use `POST`, but you can configure HTTPLink to use `GET`. However, the `uploadHTTPLink` that we use doesn't respect the `useGetForQueries` option. You will need to use `switch` from ApolloLink to use a different link depending on if the request is a query or mutation. I've done all that work in ChristFellowship, and can provide an example. 

3. Setup Fastly. Make sure you default the cache time to 0. Fastly by default caches things without a CacheControl header, and apollo doesn't send a cacheControl header if the max age for the query is 0. This is bad. Setting the default cache time to 0 in the admin panel fixes it.

https://github.com/apollographql/apollo-cache-control

https://github.com/christfellowshipchurch/christ-fellowship-conference-apollos/commit/aaa8c76edc81cf1a47ccd2556927cbb661ed4b87

PR #502, opened 2019-01-31
Adds Fastly CDN and Data Caching

redreceipt:

This will add a cache layer to all our queries and a corresponding Fastly CDN route to handle them.

None.

### How do I test this PR?

- [ ] Make sure most queries are being cached, you can use the Network inspector from `react-native-debugger`

Run queries using the following types and make sure they aren't cached:

- [ ] `AuthenticatedUser`
- [ ] `Person`
- [ ] `userFeed`
- [ ] `isLiked`
- [ ] `likedCount`
- [ ] `getAllLikedContent`

None.

## TODO:

- [ ] Closes [#494]
- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-data-schema/src/index.js
+++ b/packages/apollos-data-schema/src/index.js
@@ -419,8 +419,8 @@ export const followingsSchema = gql`
     likedCount: Int
 `)}
 
-  extend type Query @cacheControl(maxAge: 0) {
-    getAllLikedContent: [ContentItem]
+  extend type Query {
+    getAllLikedContent: [ContentItem] @cacheControl(maxAge: 0)
   }
 `;
 
--- a/packages/apollos-data-schema/src/index.js
+++ b/packages/apollos-data-schema/src/index.js
@@ -415,7 +415,7 @@ export const followingsSchema = gql`
   }
 
   ${extendForEachContentItemType(`
-    isLiked: Boolean
+    isLiked: Boolean @cacheControl(maxAge: 0)
     likedCount: Int
 `)}
 
--- a/packages/apollos-data-schema/src/index.js
+++ b/packages/apollos-data-schema/src/index.js
@@ -6,7 +6,7 @@ export const testSchema = gql`
 `;
 
 export const authSchema = gql`
-  type AuthenticatedUser {
+  type AuthenticatedUser @cacheControl(maxAge: 0) {
     id: ID!
     profile: Person
   }
@@ -40,7 +40,7 @@ export const peopleSchema = gql`
     value: String!
   }
 
-  type Person implements Node {
+  type Person implements Node @cacheControl(maxAge: 0) {
     id: ID!
     firstName: String!
     lastName: String!
@@ -328,7 +328,7 @@ export const contentItemSchema = gql`
     cursor: String
   }
 
-  extend type Query {
+  extend type Query @cacheControl(maxAge: 0) {
     userFeed(first: Int, after: String): ContentItemsConnection
   }
 `;
@@ -419,7 +419,7 @@ export const followingsSchema = gql`
     likedCount: Int
 `)}
 
-  extend type Query {
+  extend type Query @cacheControl(maxAge: 0) {
     getAllLikedContent: [ContentItem]
   }
 `;
--- a/packages/apollos-data-schema/src/index.js
+++ b/packages/apollos-data-schema/src/index.js
@@ -417,7 +417,7 @@ export const followingsSchema = gql`
 
   ${extendForEachContentItemType(`
     isLiked: Boolean @cacheControl(maxAge: 0)
-    likedCount: Int
+    likedCount: Int @cacheControl(maxAge: 0)
 `)}
 
   extend type Query {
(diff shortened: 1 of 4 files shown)


packages/apollos-data-schema/src/index.js
```diff
@@ -330,6 +330,7 @@ export const contentItemSchema = gql`
 
   extend type Query {
     userFeed(first: Int, after: String): ContentItemsConnection
+      @cacheControl(maxAge: 0)
```
conrad-vanl: Is there a way to cache this by user?

vinnyjth:
> Does CDN caching handle partial query caching or just full queries? If just full, how would we implement partial query caching?

Full query caching only :/ @redreceipt and I talked about the need to comb the client for queries that use fields like "isLiked" and make sure that the fields that eliminate the ability to cache aren't included in cacheable queries 

For partial query caching, we need to either 
- add Max-Age headers to the data coming from Rock (so that the datasources will cache) 
- or add `.ttl` calls to all of our rock requests (and make sure that the `ttl` option is working  
- or Roll our own caching layer that is capable of handling partial query caching. There's no out of the box options at this time to do partial query caching with a cdn.

packages/apollos-data-schema/src/index.js
vinnyjth: @conrad-vanl Yes! Micheal and I talked about this yesterday. We need to change `scope: PRIVATE` and then write a cache policy in Fastly. I'm working with @redreceipt to figure out how to accomplish that in this PR (probably need to add some documentation about the Fastly side of things)

redreceipt:
@vinnyjth ok trying to test out what we've got, definitely getting some network errors.
screen shot 2019-02-04 at 9 37 07 am
screen shot 2019-02-04 at 9 37 12 am

packages/apollos-data-schema/src/index.js
```diff
@@ -415,12 +416,12 @@ export const followingsSchema = gql`
   }
 
   ${extendForEachContentItemType(`
-    isLiked: Boolean
-    likedCount: Int
+    isLiked: Boolean @cacheControl(maxAge: 0)
+    likedCount: Int @cacheControl(maxAge: 0)
```
vinnyjth: I don't think we need to cache `likedCount` because it should be the same for all users (and it's okay if it's a little slow to update)

conrad-vanl: Are we able to manually invalidate the cache for `likedCount` whenever a `liked` mutation is fired for that content item? I could see a problem where a user taps "like" and the like counter in app doesn't update...especially if it still shows "0 likes"

vinnyjth: Not without pulling in the Fastly API and manually flushing the cache when an item is liked :/

However, we could update the code here 
https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apolloschurchapp/src/ui/LikeButton/index.js
to increase the `likedCount` for the item on the client. It would fall back to an old value if the view was refetched, but hopefully by that time the cache has expired. 

Another option, we could dynamically add a `expires-at` header based on the number of likes using this API. We could set a max-age of 0 for content items with no likes, and start adding a cache at a likedCount of 10 or so.
```
const resolvers = {
  Query: {
    post: (_, { id }, _, { cacheControl }) => {
      cacheControl.setCacheHint({ maxAge: 60 });
      return find(posts, { id });
    }
  }
}
```

redreceipt:
@vinnyjth

I think figuring out `scope` takes care of 1 and 3. I'll make an issue for this.

I do have a documentation skeleton up and running, I'm going to make some finishing touches to the site, including instructions for Fastly, and get that up so we can all make edits as we go.

For 4, I could be thinking in the wrong direction completely but Fastly has an API that we could maybe use to flush certain queries?

vinnyjth:
Sounds good :D Once you create those issues reference this issues to those issues and I'll close out this PR.

Do you have a link to what you have right now in terms of documentation? 

I think that's a good direction to explore. A frustration with that approach is that it would increase our dependency on Fastly. Another problem will be that actually identifying a record in the cache will be very difficult. For example, the cache key for the isLive query is a hashed version of this. 

http://localhost:4000/?query=query%20getLiveStream%20%7B%0A%20%20liveStream%20%7B%0A%20%20%20%20isLive%0A%20%20%20%20__typename%0A%20%20%7D%0A%7D%0A&operationName=getLiveStream&variables=%7B%7D

if we ever change that query, the cache key will change. If any other queries include that sensitive field, we will need to somehow identify those queries in the cache. There might be a be a good way to do this, but we will be heavily dependent of the power of the Fastly tools for searching through data that's cached.

vinnyjth:
@redreceipt One more thing before we merge, can you change the value in `.env.production` to the fastly URL? Make sure we include `https://` in the url.

redreceipt:
@vinnyjth 

Issue #505 is linked.

The documentation is in the `ApollosProject.github.io` repo if you wanna poke around. We'll create new branches off of the `source` branch for making changes. We'll publish from the `source` branch which will push to `master` and go live. Haven't done that step yet, almost ready.

`.env.production` has been updated to Fastly URL.

packages/apollos-data-schema/src/index.js
vinnyjth: @conrad-vanl after thinking about this some more, we opted to keep the cache exception for this field for the time being.

PR #504, opened 2019-02-01
Add Passes GraphQL schema, template generation, and a REST endpoint to download pkpass files

### LINKED ISSUE #485: Create a Mobile Passes schema for users
## User Story

As a `user` I want to `access my mobile passes` so that I can `check-in` to physical installations hosted by my Church.

#### Acceptance Criteria

* [ ] It should provide a list of passes

conrad-vanl:

From a high-level, this PR adds the ability to query for passes on the graphql side, it exposes enough data over graphql to be able to render a pass in-app, and then provides a REST endpoint to download/generate passes on the fly.

Specifically, it:

1. Adds GraphQL schema for supporting Passes to deliver on the functionality [in this prototype](https://sketch.cloud/s/5PRpy/v8rLPQ8/play)
2. Utilizes [`passkit-generator`](https://github.com/alexandercerutti/passkit-generator) to read [PassKit package files](https://developer.apple.com/library/archive/documentation/UserExperience/Reference/PassKit_Bundle/Chapters/Introduction.html#//apple_ref/doc/uid/TP40012026-CH0-SW1) and generate `.pkpass` downloads
3. Adds `serverMiddleware` functionality that can be exposed by any `data-connector` to add middleware to the express app. This allows `data-connector`s the ability to expose REST endpoints, for example.
4. It adds a `Template` `data-source`, and a `Rock`-powered connector for rendering lava templates. The `pass.json` file is rendered through `Rock`, allowing churches to easily customize passes based on any features that Apple adds/allows

I don't feel like many trade-offs were made. The decision to treat `pass.json` as a lava template could be limiting if a church doesn't use Rock, as the only way that we customize passes out of the box currently is through a `Template` data-source. If the Church doesn't have some templating service that we can write a connector for, then they'd have to do a lot of custom work to get passes to generate with user information.

Another design decision is rendering `data-uri`s in the graph responses for images on passes. If a url is available, the graph schema will use the URL, else if an image is only available in the pass's template files, it encodes it as a `data-uri`. This makes the graph responses larger and potentially slower, but is a pretty lean approach.

### How do I test this PR?

Read the included `README.md` for instructions on querying for passes and download pkpass files. You'll need to ping @conrad-vanl for the proper `.env` vars to use (which we can add to Basecamp after this is merged).

I currently don't have automated tests for the rest endpoint. That is my biggest remaining todo. However, I have test coverage for the new resolvers and template data-source.

## TODO:

- [x] Closes #485 
- [ ] ~Upload GIF(s) of iOS and Android~
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
@vinnyjth I'll have to fill out the PR template at the top but I'd love your thoughts on where this ended up! There's a readme file with some really quick, rough, basic info.

But once you pull this in you should be able to query for user passes in GraphiQL and download pass files (that you can open on your phone by texting the file to yourself)

packages/apollos-data-connector-passes/src/data-source.js
```diff
@@ -0,0 +1,123 @@
...
+import util from 'util';
+import { DataSource } from 'apollo-datasource';
+import { UserInputError } from 'apollo-server';
+import { promise as DataURI } from 'datauri';
+import ApollosConfig from '@apollosproject/config';
+
+const readFile = util.promisify(fs.readFile);
+
+export default class Pass extends DataSource {
+  initialize(config) {
```
vinnyjth: Annoying that this isn't done automatically as part of the `DataSource` class....

packages/apollos-data-connector-passes/src/data-source.js
```diff
@@ -0,0 +1,123 @@
...
+    this.context = config.context;
+    this.cache = config.cache;
+  }
+
+  getPassPathFromTemplateName = (templateName) =>
+    ApollosConfig.PASS.TEMPLATES[templateName];
+
+  async getRawPassTemplate(templateFile) {
```
vinnyjth: I really like how you're using the cache here. In the future, if we use redis or memcache for the apollo cache, this idea will page huge dividends.

packages/apollos-data-connector-passes/src/index.js
```diff
@@ -0,0 +1,76 @@
...
+    const context = getContext({ req });
+    let passData;
+
+    try {
+      passData = await context.dataSources.Pass.generatePassData({
+        template: req.params.template,
+        includeImageDataURIs: false,
+      });
+    } catch (e) {
+      return res.send(e.message);
```
vinnyjth: Should we set a different status code here to make responding to an error easier on the client? Something like
```
res.status(500).send(e.message);
```

packages/apollos-data-connector-passes/src/data-source.js
```diff
@@ -0,0 +1,123 @@
+import fs from 'fs';
+import util from 'util';
+import { DataSource } from 'apollo-datasource';
+import { UserInputError } from 'apollo-server';
+import { promise as DataURI } from 'datauri';
+import ApollosConfig from '@apollosproject/config';
+
+const readFile = util.promisify(fs.readFile);
+
+export default class Pass extends DataSource {
```
vinnyjth: More of a generic question than a question specific to this schema ... 

Because this type `implements Node`, should we add a get `getFromId` function so a pass can be fetched via `node(id: ....)`? I don't think it's something that we will need to actually do, but in that case should this type really `implement Node`?

conrad-vanl: Yeah! Great idea. I completely forgot to go back and address IDs.

> but in that case should this type really `implement Node`?

yeah....good thought. I don't know? `Node` seems so...abstract. What other things that have `ids` don't implement `Node` right now?

For cache purposes, having everything implement `Node` that has an `ID` is actually helpful on the client, because we can set up a fairly [robust `cacheRedirect`](https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apolloschurchapp/src/client/cache.js#L17)

conrad-vanl:
I....don't see why tests are failing?

PR #524, opened 2019-02-13
Adds props support to BackgroundView

calebpanza:
There is no current way to override the BackgroundView gradient that is used in the app without having to rewrite the component locally.

Design decision was made to have 2 levels of overrides. First, the component checks the Component for a *colors* property, and then it uses the default value. *NOTE:* this component does require an array of colors, so if you want to test a solid color background, you will need to pass an array of it like `['#000000', '#000000']`

### How do I test this PR?
Place a background view somewhere in the app and pass in the parameter *colors* as an array. Additionally you can pass any prop available to a `View` or `LinearGradient`.

I added a snapshot that generates a red to blue gradient to test passing in *colors* as a property to the component. Additional tests where added for a solid color gradient and other supported props.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


burkeshartsis:
I'll review this later today.

packages/apollos-ui-kit/src/BackgroundView/index.js
```diff
@@ -5,8 +5,12 @@ import styled from '../styled';
 import { withTheme } from '../theme';
 
 const BackgroundView = compose(
-  withTheme(({ theme }) => ({
-    colors: [theme.colors.background.paper, theme.colors.background.screen],
+  withTheme(({ theme, colors }) => ({
+    colors: colors ||
+      theme.overrides.backgroundGradient || [
```
conrad-vanl: I don't _believe_ this is the proper usage of `overrides`. @burkeshartsis and I talked about this a little bit the other day. We don't have the intention for how to use `overrides` documented anywhere, unfortunately. But I'd really like to have a conversation about this before we merge anything.

conrad-vanl: And, I realize that comment isn't super helpful as I'm not disclosing the intended usage for `overrides` :) we can talk about that too

conrad-vanl: Well, I guess this was our intended usage for `overrides`: 

```javascript
overrides = {
   ComponentName: {
      // css styles
   }
}
```

So, for example:

```
overrides = {
   BackgroundView: {
      backgroundColor: 'red',
   }
}
```

However, `overrides` as we intended it would only override css styles, and not `prop` values, like you are doing here. I think allowing `overrides` to override props is a neat idea, but I'd want to think through the design of our api for doing so.

calebpanza: @conrad-vanl I see! Ok that makes a lot of sense. I'm pretty available all day so feel free to ping me at any point to touch base on this.

burkeshartsis: Just looked at this PR now and this

> However, `overrides` as we intended it would only override css styles, and not `prop` values, like you are doing here. I think allowing `overrides` to override props is a neat idea, but I'd want to think through the design of our api for doing so.

Appears to be the most pressing conversation.

burkeshartsis: if we do keep the overrides I might recommend moving this default array value to the theme file and default the component to white.

conrad-vanl: @burkeshartsis I'm tempted to say we should hold off on making changes to the `overrides` system until/unless we figure out whether we'd rather move over to something like `styled-components`. I don't think that refactor would be overly difficult and then we can just turn to what they're doing for things like this.

Which, to be fair, they don't have an override system :)

burkeshartsis:
@conrad-vanl this is ready for review.

PR #528, opened 2019-02-14
[Personalize Content] Created PersonaFeed query

IsaacHardy:
This PR is needed to add the query for getting unique content to a specific user based on their personas.

We had to create a brand new API endpoint to point this query toward to keep this feed efficient.

### How do I test this PR?
You can test the query like so:
```
query personaFeed {
   personaFeed {
    edges {
      node{
        id
        title
        likedCount
        htmlContent
        images {
          sources {
            uri
          }
        }
      }
    }
  }
}
```
Normal resolver and query tests.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


IsaacHardy:
@conrad-vanl @vinnyjth This is now ready for review.

packages/apollos-church-api/config.yml
```diff
@@ -67,6 +67,12 @@ ROCK_MAPPINGS:
       EntityType: ContentChannelItem
     ContentItem:
       EntityType: ContentChannelItem
+    Person:
```
vinnyjth: This was a nice fix to get the code working, but we should probably fix the original function so that it properly handles finding EntityTypes that aren't ContentItems.

packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -239,6 +239,36 @@ export default class ContentItem extends RockApolloDataSource {
     return request.orderBy('Order');
   };
 
+  // Generates feed based on persons dataview membership
+  byPersonaFeed = async () => {
+    try {
```
vinnyjth: Doing `try` and then `catch` and then throwing the `e` is pretty much the same as not catching the error in the first place :) I would recommend we do one or the other. (Handle the error by returning an empty array, or not handling the error)

packages/apollos-data-connector-rock/src/content-items/resolver.js
```diff
@@ -55,6 +55,13 @@ const resolver = {
         cursor: dataSources.ContentItem.byUserFeed(),
         args,
       }),
+    personaFeed: async (root, args, { dataSources }) => {
```
vinnyjth: Sweeeeeeeet

packages/apollos-data-connector-rock/src/people/data-source.js
```diff
@@ -16,6 +16,30 @@ export default class Person extends RockApolloDataSource {
...
+
+    try {
+      // Get current user
+      const { id } = await Auth.getCurrentPerson();
+
+      // Get the entity type ID of the Person model
+      const personEntityTypeId = await RockConstants.modelType('Person');
+
+      // Return a list of all dataviews by GUID a user is a memeber
+      return this.request('DataViews/GetPersistedDataViewsForEntity')
```
vinnyjth: Nice!

packages/apollos-data-schema/src/index.js
```diff
@@ -332,6 +332,8 @@ export const contentItemSchema = gql`
   extend type Query {
     userFeed(first: Int, after: String): ContentItemsConnection
       @cacheControl(maxAge: 0)
+    personaFeed(first: Int, after: String): ContentItemsConnection
+      @cacheControl(maxAge: 0)
```
vinnyjth: Good catch!

packages/apollos-data-connector-rock/src/people/data-source.js
vinnyjth: Will this require changes to the rock instance? or can anyone make the queries that you are making.

IsaacHardy: We have the controller that must be copied to the three Churches Rock instances. However, we also have a PR up on the main Rock repo: https://github.com/SparkDevNetwork/Rock/pull/3539

packages/apollos-data-connector-rock/src/content-items/data-source.js
IsaacHardy: Looks like I got lazy and forgot about it. Will fix

packages/apollos-church-api/config.yml
IsaacHardy: Could you elaborate a bit here?

conrad-vanl:
The `ready for review` label was never removed the first time around (after the first review), so I cna't tell if this is ready again or not, so I'm going to remove the label and add it back if it's ready again.

packages/apollos-church-api/config.yml
vinnyjth: This section of the `Config.yml` is used to define properties for ContentItems. Having `Person` here doesn't really jive with the intent of this config section. 

The reason we put this here, was that `getEntityType` throws an error if the value you pass doesn't exist in this array. That's was an oversight when I wrote that function, because we definitely wan't to be able to query for arbitrary EntityTypes. Rather than adding `Person` here, we should fix `getEntityType`

vinnyjth: @IsaacHardy Do you feel like you have what you need to make this change or should we pair on it?

vinnyjth:
👀

packages/apollos-data-connector-rock/src/content-items/__tests__/__snapshots__/resolvers.tests.js.snap
```diff
@@ -320,6 +320,17 @@ Object {
 }
 `;
 
+exports[`UniversalContentItem gets a persona feed 1`] = `
+Object {
+  "data": Object {
+    "personaFeed": null,
+  },
+  "errors": Array [
+    [GraphQLError: Cannot read property 'getPersonas' of undefined],
```
vinnyjth: 🤔 I'll see if I can fix this Snapshot :)

PR #537, opened 2019-02-21
Adds Rock Package Readme

### LINKED ISSUE #525: Create readme for `apollos-data-connector-rock`

vinnyjth:
## TODO:

- [ ] Closes #525 
- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-data-connector-rock/README.md
```diff
@@ -1,11 +1,223 @@
...
+- Interactions
+- People
+- PersonalDevices
+- RockConstants
+- Shareable
+- Template
+
+## Dependants / Dependencies
+
+None, right now. This module is 100% standalone. However, several modules internally require other modules internally. For example,
```
conrad-vanl:
```suggestion
None, right now. This module is 100% standalone. However, modules within the `data-connector` might depend on each other. For example,
```

packages/apollos-data-connector-rock/README.md
```diff
@@ -1,11 +1,223 @@
...
+  RockConstants,
+  Family,
+  Person,
+  ContentItem,
+  ContentChannel,
+  Sharable,
+  Auth,
+  PersonalDevice,
+  Template,
+  ...
```
conrad-vanl: Should we document the need to specify the `ContentItem` sub-types?

```
DevotionalContentItem: {
  dataSource: ContentItem,
}
```

I don't like that we have to do this..... fwiw this is required due to [this line](https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-server-core/src/node/model.js#L38). We expect there to be a `dataSource` present for each `type`. I've thought about a couple of times modifying the logic in `parseGlobalId` to look for a `dataSource` matching the `__type`'s `interface`... so for example if `DevotionalContentItem` didn't have it's own `dataSource` specified, it'd know to fall back to `ContentItem`. But I think that can get messy too :/

packages/apollos-data-connector-rock/README.md
```diff
@@ -1,11 +1,223 @@
...
+export { resolver };
+```
+
+#### 3. Import your new module.
+
+Naming this time isn't important because we don't have a data source.
+
 ```
-const apollosDataConnectorRockContent = require('apollos-data-connector-rock-content');
+// The import name is _very_ important. That will be the name of your dataSource.
```
conrad-vanl: This statement "the import name is _very_ important" seems to contradict with the line above "Naming this time isn't important". Am I reading this wrong?

packages/apollos-data-connector-rock/README.md
```diff
@@ -1,11 +1,223 @@
...
-// TODO: DEMONSTRATE API
+Adding a new ContentItemType is a common operation, and we have some configuration tools to make it easier. Once you define your new ContentItem type's schema by extending the `ContentItem` GraphQL interface, add a new line to your `config.yml` in the RockMapppings -> ContentItem section. For example, if we knew that all of our GroupContentItem's referenced above had a content channel id of 27, we could add this to our config.yml in the RockMappings section.
+
+```
+    GroupContentItem:
+      EntityType: ContentChannelItem
+      ContentChannelTypeId: [27]
 ```
+
+If determining if the ContentItem type is more complex than something like ContentChannelTypeId, you'll need to override the `__resolveType` function on the ContentItem resolver.
```
conrad-vanl:
```suggestion
If determining the `ContentItem` type is more complex than something like` ContentChannelTypeId`, you'll need to override the `__resolveType` function on the `ContentItem` resolver.
```

conrad-vanl: Also, same as my comment above about linking the additional `dataSource`s for the new content item type - 

```
GroupContentItem: {
  dataSource: ContentItem,
}
```

Would this be needed in this case as well?

packages/apollos-data-connector-rock/README.md
vinnyjth: I made a bad copy/paste here 😂 I can delete this comment

vinnyjth:
@conrad-vanl In regard to needing to copy over the data source multiple times, I would like to fix the logic at the `node` (it's always bothered me a tiny bit that we have to do that). Are you okay merging this if I create an in sprint issue to fix? I'm thinking we could go the route you suggested (best option) or we could rely on the array of ContentItemTypes inside `config.yml` (more brittle but would work if figuring out interface is hard)

packages/apollos-data-connector-rock/README.md
vinnyjth: Good catch 👌  I left a comment about ideas moving forward below.

PR #547, opened 2019-02-26
Update ButtonIcon to support style prop

burkeshartsis:
## DESCRIPTION
Update ButtonIcon to support additional styles passed somewhere we can render them appropriately.

We needed to be able to render a `backgroundColor` on `ButtonIcon`. This makes that possible among other things.

### How do I test this PR?
Add a style to a currently in use `ButtonIcon` somewhere in the app. Easiest place is in the media player.

🍪's 

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
1. Why can't we use `style`?
2. Do the other `Button` components support `backgroundColor`?

burkeshartsis:
Great questions @conrad-vanl! Answers below,

1. Originally I had avoided the using `style` because I would have to pass it to a child component and not the parent to get it to work correctly. Which would could lead to unexpected results for a future developer. I don't mind this approach this was just my only concern with it.

2. Other buttons do not have a similarly named prop. In fact, to you point in above I have always advocated for styles being passed as a `style` rather than extra api. 

The PR now reflects your suggestion.

packages/apollos-ui-kit/src/Button/ButtonIcon/index.js
```diff
@@ -52,6 +55,10 @@ class ButtonIcon extends PureComponent {
               ? this.props.iconPadding
               : this.props.size
           }
+          additionalStyles={this.props.style}
+          borderRadius={
+            this.props.iconPadding + this.props.size || this.props.size * 2
```
redreceipt: Why would want we want to double the size of the icon footprint implicitly if they don't pass in anything? Shouldn't we just let them decide that by letting the footprint be exactly the size they specified?

redreceipt:
@burkeshartsis not sure if I'm testing this right but shouldn't the tab icons turn colors here?
Screen Shot 2019-03-20 at 10 57 34 AM

packages/apollos-ui-kit/src/Button/ButtonIcon/index.js
burkeshartsis: One word, tappability. We want to provide a minimum level of usability. The available api allows for customizability in size while preserving this minimum level of ux sugar.

burkeshartsis:
@redreceipt your screenshot is using `Icon` not `ButtonIcon`

burkeshartsis:
@redreceipt also this just exposes the `style` prop for the `ButtonIcon` you don't pass a `backgroundColor` directly you have to pass it via a style object. Check the tests or stories for an example.

PR #567, opened 2019-02-27
Schema Modularization

conrad-vanl:

Alright @vinnyjth. I had an idea come to me last night on a refactor of schema to better modularize it. I took a first crack at it in this PR, but it's probably not good yet, and I need to explain my idea.

We are building a ton of connected ui components that rely on certain schema. A share button relies on certain share attributes, a theming component relies on a `theme` object, our card components relies on the existence of title, image, summary text, etc. This has created a lot of flexibility in exposing many different things as `ContentItem`s (like groups on the CF app) because then our components know how to render them. Is there a better way to express this in the schema?

Interfaces allow us to construct fragment queries on the interface, without having to know what the actual type is...For example, what if a connected share button component could take in the `ID` of any `Node`:

```jsx
<ShareButton id={props.id} />
```

```gql
query {
   node(id: ID) {
    ...on Shareable {
       sharing: {
          ...shareFragment
       }
    }
  }
}
```

Right now, components like this rely on the node being a `ContentItem` (ex: [`LikeButton`](https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apolloschurchapp/src/ui/LikeButton/getLikedContentItem.js#L6)). But it doesn't need to, it should only care if it's a node that supports the specific feature it needs. 

Same with our `ContentCard`, it looks for props like `title`, `summary`, `coverImage`. What if we just created some sort of `Card` interface in our schema, and could render a card for any node that implemented that interface?

```gql
type Cardable {
   title: String
   summary: String
   coverImage: ImageMedia
}
```

```jsx
<ContentCard id={props.id} />
```

```gql
query {
   node(id: ID) {
      ...on Cardable {
         ...cardableFragment
      }
   }
}
```

Our components are slowly moving towards this paradigm, albeit [with pretty complex queries](https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apolloschurchapp/src/ui/ContentCardConnected/query.js) because we have to support a lot of different types.

Lastly, when using interfaces, if the node we are querying for doesn't implement that interface, the query does not error out, and the component can cleanly handle the edge case. Say we have a default `ContentSingleView` that includes the ability to override theme. Our `<Theme />` component would query `...on Themeable`, if the node wasn't a `Themeable`, then the component can cleanly exit and not do anything to theme. 

As long as the graphql server defines these base-line interfaces, the church can choose what properties they want to support on their individual data types. They can create content types that don't implement certain interfaces, and our ui won't throw a fit. Right now if I try to create an apollos app that doesn't contain theme, there's no way to do that. Right now we have to define _all_ of the properties of our baseline `ContentItem` type, even if we have no plans to use them. Switching to interfaces, I believe we'd get best of both worlds.

THOUGHTS?

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


calebpanza:
@conrad-vanl where are we at with this idea? I really, _really_ like it. Would love to talk more about what this could look like and how it could be implemented.

conrad-vanl:
@calebpanza essentially if we can get this PR to a point where we're comfortable merging (ie...tests pass), I'd be happy to move on it. It's mainly additive to the schema, so it's mostly non-breaking, however I did start removing things I don't think we are using currently.

calebpanza:
@conrad-vanl what can I do to help get it to the point of merging?

vinnyjth:
I hit the "ready for review" button by accident 😬 
I'll be looking at this today, trying to get tests passing.

conrad-vanl:
@vinnyjth do you think this should be prioritized before we have any more production apps?

vinnyjth:
@conrad-vanl I want to say yes, but it will be hard because of the huge amount of work we will need to undertake to get this working. I don't honestly remember if there's still work to do to get the API working properly, but I know there's still significant changes that need to happen on the client and with the API tests to get this working. Not hard changes, just a lot of them....

I think we need to decide if this pattern is "really nice" or "essential for easier development with Apollos". In this upcoming sprint, I can take time to continue to hammer on making this PR work, but honestly my time is better spent right now making the UI more modular and improving update tools. 

So far (and it's hard to say because we haven't done any advanced customization) I can't say our current schema has significantly hampered our ability for our clients to customize, the app. At this point, I would rather focus on hiding away the connected components into packages, and that gives us more ability to upgrade our current pre-production apps, before they hit production. 

Summarizing my stream of thoughts above - If we are going to go live with this, we need to do it before any of our apps hit production. I find it hard to justify working on it right now, and hope that if . we focus on other modularity improvements we can sneak this change in for free.

conrad-vanl:
Yes, I totally don't want to take your focus away from other things. And agree with your thoughts.

I do see this as being in the realm of things that fit within our goal to make the framework more modular and easier to work with. I agree packaging UI needs to be done first as well.

PR #585, opened 2019-03-07
Adds support for Node to use Interface's DataModel instead of relying on Aliases

vinnyjth:

For a while, we have had to create `alias` references to DataSources for models that want to use `node` but are implementations of a root `interface`. With this PR, the `node` resolver will be smart enough to "work it's way up the hierarchy chain" and resolve without needing to alias every implementation. 

This PR is also smart enough that specific instances of an implementation that supply their own datasource will override default behavior. 

I also improved the error message for instances where a dataSource is present, but the dataSource does not implement a getFromId function.

n/a

### How do I test this PR?

Run the app, you shouldn't see any changes.

N/a. Fixed some existing tests that were pulling bad data prior to this PR.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


README.md
```diff
@@ -120,4 +120,4 @@ static propTypes = {
 };
 ```
 
-🍕🍕🍕
\ No newline at end of file
+🍕🍕🍕🍕
```
redreceipt: @vinnyjth please explain this

vinnyjth: I need to trigger another build on travis and the "rebuild" button wasn't working.

redreceipt: ha! oh

packages/apollos-server-core/src/node/__tests__/__snapshots__/model.js.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Node Node class should throw error if it can't find a matching model 1`] = `"No dataSource found using NoModel"`;
```
IsaacHardy: Is this the expected outcome?

packages/apollos-server-core/src/node/model.js
```diff
@@ -29,19 +30,36 @@ export function parseGlobalId(encodedId) {
   }
 }
 
+const getPossibleDataModels = ({ schema, __type }) => {
```
IsaacHardy: This whole method is hard to comprehend. Inline comments might be useful to help explain some of the work happenings.

packages/apollos-server-core/src/node/__tests__/__snapshots__/model.js.snap
vinnyjth: Yes

packages/apollos-server-core/src/node/model.js
vinnyjth: Can you expand on what you find hard to comprehend about it? I don't doubt you for a second, but I want to make sure that I leave the right comments. I tried to name the variables in such a way that the code was self documenting.

IsaacHardy: Line 34 seems to be grabbing the possibleType of a node, but I have no idea what `.getTypeMap()` is doing.

Line 35 has some very interesting properties to check: `possibleType.astNode.interfaces`

Line 38 as well ^

Line 42 has an array with possibleInterfaces, and I have no idea what that even looks like.

vinnyjth: 👍 adds more helpful comments

PR #591, opened 2019-03-12
Adds SMS Auth UI

### LINKED ISSUE #546: Build SMS-login UI
Modify the `/auth` package to default to an SMS-login flow.

This issue includes building out the 2 new SMS login screens: a phone number prompt and a code entry prompt. The name input screen will be a part of the onboarding flow, **not** auth.

As you start on this issue, feel free to break out into smaller chunks if it makes sense (multiple PRs). Otherwise, feel free to tackle at once.

Don't worry about hooking into the onboarding flow right now...just add to the existing `auth` package and `auth` routes that you can enter through the `Connect` tab currently.

**Stretch Goal**: Move `/auth` into a separate `apollos-ui-auth` package

conrad-vanl:
~Blocked by #576~

This PR adds sms authentication to our existing auth ui package.

None really :) 

### How do I test this PR?

Try to login, on a real device. @vinnyjth or someone will have to add you to the whitelist.

This PR still needs some snapshots for the forms

## TODO:

- [x] Closes #546 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
@redreceipt this PR was originally branched off of #576. Since #576 was merged into master, we just need to patch up the conflicts that merge generated. I briefly glanced through the merge conflicts...if you have any questions or run into issues please ping me. Once we get that merged up, I think you should do a run through and flag any other work that needs to be done, and we can divide and conquer the remaining work.

conrad-vanl:
@redreceipt after the merge conflicts are fixed, I believe the only thing left is the loading state UI on the next buttons. I talked to @burkeshartsis about this and he's able to assist while I'm out after dan's work is in a good place.

redreceipt:
ok @burkeshartsis I think I've got most of it 🤞 still a couple of conflicts I want your final say on. I don't understand enough of what is going on to make an informed decision. Three main files I've got questions about:

Both added `ui-auth/src/Providers.js`, which is the right one? Or which parts from it do I need from each?
[master](https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-ui-auth/src/Provider.js)
[this PR](https://github.com/ApollosProject/apollos-prototype/blob/ui-sms-auth/packages/apollos-ui-auth/src/Provider.js)

for `app/src/Providers.js`...

Screen Shot 2019-04-03 at 5 18 08 PM


for `app/src/client/index.js`...

Screen Shot 2019-04-03 at 5 19 15 PM

Screen Shot 2019-04-03 at 5 20 20 PM

redreceipt:
[Auth](https://trello.com/c/8X9fqLQ1/6-auth)

conrad-vanl:
I can't leave a review on this myself since I opened the PR, but I believe this needs just a bit more 💄 polish work. Mainly - the "loading" state on the next buttons are off (shouldn't change size, should show activity indicator). With the `loading` prop set to true, we should be showing an activity indicator as well - that exists in the button's source code here - https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-ui-kit/src/Button/Button.js#L104

@redreceipt do you feel like this is something you could look at?

burkeshartsis:
@conrad-vanl I can look at the loading state today

packages/apollos-ui-auth/src/AuthPassword/index.js
```diff
@@ -0,0 +1,76 @@
...
+            <BackgroundView>
+              <SafeAreaView style={StyleSheet.absoluteFill}>
+                <PaddedView>
+                  <PromptText>{this.flatProps.passwordPromptText}</PromptText>
+                </PaddedView>
+
+                <TabView
+                  routes={this.tabRoutes}
+                  renderScene={SceneMap({
+                    login: () => <LoginForm onLogin={closeAuth} />,
```
burkeshartsis: I'd suggest moving this function along with the line after to class properties or methods. Something like `this.loginFormRoute({ closeAuth })` maybe?

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
```diff
@@ -0,0 +1,165 @@
...
+    smsPasswordLoginPrompt: PropTypes.node,
+    screenProps: PropTypes.shape({}), // we'll funnel screenProps into props
+  };
+
+  static defaultProps = {
+    brand: <BrandIcon />,
+    authTitleText: 'Have we met before?',
+    smsPromptText:
+      "Lets get you signed in using your mobile number. We'll text you a code to make login super easy!",
+    smsPolicyInfo: (
```
burkeshartsis: I didn't know you could define a default prop to be a component like this. I mean it makes sense. I've just never seen it before.

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
```diff
@@ -0,0 +1,165 @@
...
+    }
+  }
+`;
+
+class PhoneEntry extends Component {
+  static propTypes = {
+    brand: PropTypes.node,
+    authTitleText: PropTypes.string,
+    smsPromptText: PropTypes.string,
+    smsPolicyInfo: PropTypes.node,
```
burkeshartsis: any particular reason we believe this shouldn't just be a `string` that get's passed into a component. That's how most of our component API's are defined almost everywhere else. Was there a particular reason for wanting to allow people to customize the style of this legal text? Would it be better to just have a sensible default and allow for style overrides via `theme`?

burkeshartsis: At the very least if we want to do it this way, I'd expect us to accept a `string` or `node` and conditionally type check whether to render a sensible default like the `H6` below with the provided string or render the passed in component if it's available.

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
```diff
@@ -0,0 +1,165 @@
...
+
+  validationSchema = Yup.object().shape({
+    phone: Yup.string().matches(/^\D?(\d{3})\D?\D?(\d{3})\D?(\d{4})$/),
+  });
+
+  get flatProps() {
+    return { ...this.props, ...(this.props.screenProps || {}) };
+  }
+
+  submitHandler = (mutate) => async (
```
burkeshartsis: I'm surprised our linter didn't pick this up but typically we try to enforce something like this.

```suggestion
  handleOnSubmit = (mutate) => async (
```

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
```diff
@@ -0,0 +1,165 @@
...
+
+    return (
+      <KeyboardAvoidingView style={StyleSheet.absoluteFill} behavior="padding">
+        <BackgroundView>
+          <Mutation mutation={requestPin}>
+            {(mutate) => (
+              <Formik
+                initialValues={{ phone: '' }}
+                validationSchema={this.validationSchema}
+                onSubmit={this.submitHandler(mutate)}
```
burkeshartsis:
```suggestion
                handleOnSubmit={this.submitHandler(mutate)}
```

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
```diff
@@ -0,0 +1,165 @@
...
+    } catch (e) {
+      setFieldError(
+        'phone',
+        'There was an error. Please double check your number and try again.'
+      );
+    }
+    setSubmitting(false);
+  };
+
+  handlePasswordLoginPress = () => {
```
burkeshartsis: Maybe?
```suggestion
  handleOnPressLogin = () => {
```

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
```diff
@@ -0,0 +1,165 @@
...
+                          onSubmitEditing={this.handleAdvance}
+                          enzblesReturnKeyAutomatically
+                          error={touched.phone && errors.phone}
+                          onChangeText={handleChange('phone')}
+                          value={values.phone}
+                        />
+                      </PaddedView>
+                      {allowPassword ? (
+                        <PaddedView>
+                          <ButtonLink onPress={this.handlePasswordLoginPress}>
```
burkeshartsis: If my suggestion above is accepted you'll want this to match 😉 
```suggestion
                          <ButtonLink onPress={this.handleOnPressLogin}>
```

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
```diff
@@ -0,0 +1,165 @@
...
+    ),
+    allowPassword: true,
+    smsPasswordLoginPrompt: "I'd rather use my email and a password",
+  };
+
+  validationSchema = Yup.object().shape({
+    phone: Yup.string().matches(/^\D?(\d{3})\D?\D?(\d{3})\D?(\d{4})$/),
+  });
+
+  get flatProps() {
```
burkeshartsis: I've never really seen props combined like this before. Nothing particularly wrong here but I was wondering what you were thinking or why the need?

Personally, I like to see where my props came from so I understand their usage in the context of where they came from.

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
```diff
@@ -0,0 +1,165 @@
...
+import gql from 'graphql-tag';
+import { Mutation } from 'react-apollo';
+
+import {
+  NextButtonRow,
+  NextButton,
+  TitleText,
+  PromptText,
+  BrandIcon,
+} from '../styles';
```
burkeshartsis: With so many shared components and given that they are used so similarly I'm wondering if the better solution would be to refactor `PhoneEntry.js` and `Verification.js` to use a shared layout component that consists of these shared styled components. I also wonder if `Slide` (from `/Onboarding`) would be a suitable solution.

packages/apollos-ui-auth/src/AuthSMS/Verification.js
```diff
@@ -0,0 +1,157 @@
...
+const verifyPinMutation = gql`
+  mutation verifyPin($phone: String!, $code: String!) {
+    authenticateWithSms(phoneNumber: $phone, pin: $code) {
+      token
+    }
+  }
+`;
+class Verification extends Component {
+  static propTypes = {
+    brand: PropTypes.node,
```
burkeshartsis: Was the thinking behind the generic prop name that this could be an icon or an image or something? Typically we have handled this with an icon and just accepted a string and defaulted to `brand-icon`.

packages/apollos-ui-auth/src/Header.js
```diff
@@ -0,0 +1,12 @@
...
+
+const Header = styled(({ theme }) => ({
+  padding: theme.sizing.baseUnit,
+  paddingBottom: theme.sizing.baseUnit * 1.5,
+  flexDirection: 'row',
+  alignItems: 'center',
+  backgroundColor: theme.colors.background.paper,
+}))(SafeAreaView);
+
+export default Header;
```
burkeshartsis: I can haz testz?

packages/apollos-ui-auth/src/AuthSMS/Verification.js
```diff
@@ -0,0 +1,157 @@
...
+
+  validationSchema = Yup.object().shape({
+    phone: Yup.string().matches(/^[6-9]\d{9}$/),
+  });
+
+  get flatProps() {
+    return { ...this.props, ...(this.props.screenProps || {}) };
+  }
+
+  submitHandler = ({ verifyPin, closeAuth }) => async (
```
burkeshartsis:
```suggestion
  handleOnSubmit = ({ verifyPin, closeAuth }) => async (
```

packages/apollos-ui-auth/src/AuthSMS/Verification.js
```diff
@@ -0,0 +1,157 @@
...
+                          authToken: authenticateWithSms.token,
+                        },
+                      });
+                    }}
+                  >
+                    {(verifyPin) => (
+                      <Formik
+                        initialValues={{ code: '' }}
+                        validationSchema={this.validationSchema}
+                        onSubmit={this.submitHandler({ verifyPin, closeAuth })}
```
burkeshartsis:
```suggestion
                        onSubmit={this.handleOnSubmit({ verifyPin, closeAuth })}
```

vinnyjth:
@conrad-vanl do you want to handle making the changes or, if you're cool with it, I can take over from here?

conrad-vanl:
@vinnyjth up to you, you're not supposed to be working this week 😉 

Let me at least respond to the review first.

vinnyjth:
@conrad-vanl Once you reply to the review, I don't mind making changes :D I'm ambivalent about Burke's changes above, so I'll just do everything unless you comment otherwise.

packages/apollos-ui-auth/src/AuthPassword/index.js
conrad-vanl: There's not any benefit in doing this, it would end up being:

```
login: () => this.loginFormRoute({ closeAuth }),
```

packages/apollos-ui-auth/src/index.js
```diff
@@ -1,22 +1,11 @@
...
-  H5,
-  styled,
-  Icon,
-  ButtonLink,
-  withTheme,
-} from '@apollosproject/ui-kit';
-
-import LoginForm from './login';
-import SignUpForm from './signup';
+  SMSPhoneEntry as AuthSMSPhoneEntry,
```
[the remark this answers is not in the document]
conrad-vanl: Yeah, the naming can be cleaned up some. `SMSPhoneEntry` is called `SMSPhoneEntry` within the context of the `Auth` package, but when it gets exported globally it gets named `AuthSMSPhoneEntry`. Taht's what I was trying to do...

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
conrad-vanl: The policy info has a link inside of it, so it can't be a string. `string`s might be more common, but we definitely use `node`s in a lot of places.

conrad-vanl: The way I look at it...values that come from the backend should be strings almost always. Values that get defined as part of your frontend code probably benefit being `node`s

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
conrad-vanl: `handleOnSubmit` is incorrect. This is a function that returns a function, so it is named differently (and should be named differently). I'm happy to consder a new name to make this more clear

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
conrad-vanl: See above comment

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
conrad-vanl: I think it's only because the child component these props get passed into has no knowledge of `screenProps` (nor should it, because it's not a screen component / doesn't have to be rendered inside a navigator), but within this screen component, there are some `screenProps` that need to get passed down. So we flatten it.

packages/apollos-ui-auth/src/AuthPassword/index.js
burkeshartsis: This is purely a code style. I think I think it's preferable to call a function even if that function just returns an arrow function that returns a component. Preference really but I like how we did it [here](https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apolloschurchapp/src/content-single/DevotionalContentItem/index.js#L101) with Devotional tabs

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
conrad-vanl: I'm open to this

packages/apollos-ui-auth/src/AuthSMS/Verification.js
conrad-vanl: yeah it probably needs to be more flexible then that

conrad-vanl: again, I think this goes back to my comments above where props that are part of the ui layout probably benefit from being `.node`, wheres props directly driven from backend/database benefit from being `string`. So in this case....`brand` is a layout component of this view. But something like an Article's title is not.

conrad-vanl:
@vinnyjth I made some comments

packages/apollos-ui-auth/src/AuthSMS/PhoneEntry.js
vinnyjth: @burkeshartsis Is that something you could do when you make the UI improvements you plan on making in another PR?

vinnyjth:
@conrad-vanl Sweet, I'll try and make the remaining changes before EoD.

packages/apollos-ui-auth/src/styles.js
```diff
@@ -0,0 +1,49 @@
...
+
+export const NextButtonRow = styled({
+  flexDirection: 'row',
+  justifyContent: 'flex-end',
+  alignItems: 'center',
+})(PaddedView);
+
+export const NextButton = (props) => (
+  <Button type="secondary" {...props}>
+    <>
```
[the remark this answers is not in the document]
vinnyjth: Can confirm it works on Android.

packages/apollos-ui-auth/src/Header.js
vinnyjth: We can hit this in the UI once-over.

PR #596, opened 2019-03-14
Creates new personalize component for home feed.

IsaacHardy:
## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apolloschurchapp/src/tabs/home/Home.js
```diff
@@ -56,6 +58,32 @@ class Home extends PureComponent {
                   <>
                     <LogoTitle source={require('./wordmark.png')} />
                     <LiveButton />
+                    <Query
+                      query={getPersonaFeed}
+                      fetchPolicy="cache-and-network"
+                    >
+                      {({ data: personaData, loading: actionLoading }) => (
+                        <ActionTable
+                          label={'FOR YOU'}
```
IsaacHardy: Discussion: Because MVP only involves this SPECIFIC component usage, I decided to hard code this label in. However, this should be updated to support more in the future.

packages/apolloschurchapp/src/tabs/home/Home.js
```diff
@@ -56,6 +58,32 @@ class Home extends PureComponent {
...
+                          isLoading={actionLoading}
+                          onPress={this.handleOnPress}
+                          DynamicHeader={
+                            <H3
+                              isLoading={actionLoading}
+                              numberOfLines={3}
+                              ellipsizeMode="tail"
+                            >
+                              Some random text that encourages you
```
IsaacHardy: Same as above ^

packages/apolloschurchapp/src/tabs/home/Home.js
```diff
@@ -56,6 +58,32 @@ class Home extends PureComponent {
                   <>
                     <LogoTitle source={require('./wordmark.png')} />
                     <LiveButton />
+                    <Query
+                      query={getPersonaFeed}
+                      fetchPolicy="cache-and-network"
+                    >
+                      {({ data: personaData, loading: actionLoading }) => (
+                        <ActionTable
```
conrad-vanl: CC @vinnyjth @burkeshartsis on this regarding naming. `ActionTable` isn't bad. I feel like `Card` should probably be in the name. `Table` I like. `Action` - i'm not sure about.

`ContentTableCard`
`ContentListCard`

are two other potential ideas.

packages/apolloschurchapp/src/ui/ActionTable/ActionItem.js
```diff
@@ -0,0 +1,78 @@
...
+  borderRadius: theme.sizing.baseUnit,
+  overflow: 'hidden',
+  marginRight: theme.sizing.baseUnit,
+}))(View);
+
+const StyledH6 = styled(({ theme }) => ({
+  color: theme.colors.text.tertiary,
+}))(H6);
+
+const TextContainer = styled(({ theme }) => ({
```
conrad-vanl: Just curious...did you try `CellContent` here instead? Also wondering if instead of using `marginTop`, `height` and such, if we should some how get this to vertically center using `flex`.

packages/apolloschurchapp/src/ui/ActionTable/index.js
```diff
@@ -0,0 +1,74 @@
...
+    content: PropTypes.array, // eslint-disable-line
+    label: PropTypes.string,
+    DynamicHeader: PropTypes.element,
+  };
+
+  render() {
+    const { onPress, isLoading, content, label, DynamicHeader } = this.props;
+
+    return (
+      <ContentCard
```
conrad-vanl: We talked about this a bit, but we should probably nix using `ContentCard` here. Think this can simply just become:

```
<Card>
  <Header>{...}</Header>
  <ContentTable>{...}</ContentTable>
</Card>
```

packages/apolloschurchapp/src/ui/ActionTable/index.js
```diff
@@ -0,0 +1,74 @@
...
+        header={
+          <Header>
+            <StyledH6 isLoading={isLoading}>{label}</StyledH6>
+            <ActionTitle>{DynamicHeader}</ActionTitle>
```
conrad-vanl: I feel like we should either:

- Use a `string` prop for both `label` and `header`
- Just use a `PropType.element` for all of `Header`'s children

So either:

```
<ActionTable label="some string" header="Some string" />
```

or

```
<ActionTable
  header={
     <>
          <H6>....</H6>
          <H4>....</H4>
     </>
  }
/>
```

conrad-vanl: I also _think_ you can use `<H4 padded />` and get rid of the `<ActionTitle />` wrapper. I'd rather let the typography elements dictate vertical rhythm vs us adding components with padding, even if that means this looks different then the sketch proto.

packages/apolloschurchapp/src/ui/ActionTable/ActionItem.js
IsaacHardy: I'm not using any TableView components here. They did not allow for some of the needed styles. @burkeshartsis can speak to that a bit.

packages/apolloschurchapp/src/ui/ActionTable/index.js
IsaacHardy: Aye.

packages/apolloschurchapp/src/ui/ActionTable/index.js
IsaacHardy: I would love a bit of elaboration on the first suggestion.

Regarding the second comment, the `<ActionTitle />` is needed for creating the wrapper for any element that might be passed in so that the loading state has correct dimensions for the rendered content. @burkeshartsis can also speak to this.

packages/apolloschurchapp/src/tabs/home/Home.js
IsaacHardy: Naming things is hard.

IsaacHardy: @vinnyjth @burkeshartsis PING

packages/apolloschurchapp/src/ui/ActionTable/index.js
conrad-vanl: Well you're using two different paradigms for these components - one is a `element` and the other is a `string`. Passing an `element` isn't bad, but I wonder if we do that if we shouldn't just allow them to pass an element for the entire `Header` instead of two separate props.

Keeping at two separate props isn't bad, but wonder why they can't just be `string`s at that point.

packages/apolloschurchapp/src/tabs/home/Home.js
vinnyjth: I like the language around `ContentTableCard` or `ContentListCard`. I don't understand the word `Action` in this context :)

packages/apolloschurchapp/src/ui/ActionTable/index.js
IsaacHardy: Ahhh. That makes a bit more sense. I'll update it to pass in an element for both label and header.

IsaacHardy:
@conrad-vanl @burkeshartsis @vinnyjth @richarddubay This is ready for review.

packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -257,6 +257,7 @@ export default class ContentItem extends RockApolloDataSource {
         .join()}`
     )
       .andFilter(this.LIVE_CONTENT())
+      .top(first)
```
conrad-vanl: What happens if you don't pass this argument at all, ie `first` is undefined? Is `first` a required argument to this function?

packages/apolloschurchapp/src/tabs/home/Home.js
```diff
@@ -19,6 +27,11 @@ const LogoTitle = styled(({ theme }) => ({
   resizeMode: 'contain',
 }))(Image);
 
+const StyledH6 = styled(({ theme }) => ({
+  color: theme.colors.text.tertiary,
+  marginBottom: theme.sizing.baseUnit / 2,
```
conrad-vanl: @burkeshartsis is he using this correctly? Should he have to add a margin to `H6` at all, should the vertical rhythm of our components just let this flow naturally?

Regarding color: it's becoming more common to change colors of ui kit text components (although, in this case, I'm not sure if we should pass in a different color). @burkeshartsis in the future, what would you think about passing in some prop to our `text` components that let you set a different theme color? Something simple like:

`<H6 primary />` `<H6 secondary />` `<H6 touchable />`

or maybe

`<H6 themeColor="primary" />`

packages/apolloschurchapp/src/tabs/home/Home.js
```diff
@@ -56,6 +69,36 @@ class Home extends PureComponent {
...
                     <LiveButton />
+                    <Query
+                      query={getPersonaFeed}
+                      fetchPolicy="cache-and-network"
+                    >
+                      {({ data: personaData, loading: actionLoading }) => (
+                        <ContentTableCard
+                          isLoading={actionLoading}
+                          onPress={this.handleOnPress}
+                          DynamicHeader={
```
conrad-vanl: Thoughts on naming this just `header` or something? Not sure why the need to `PascalCase` this prop either (I realize this prop changed just a bit from my last review so you just kept the same name). I'm just not sure we need to keep the name "Dynamic" in here (we haven't used this phrase in any other component ever, even though we have other props that use `PropTypes.element`)

packages/apolloschurchapp/src/tabs/home/Home.js
```diff
@@ -56,6 +69,36 @@ class Home extends PureComponent {
...
+                            <>
+                              <StyledH6 isLoading={actionLoading}>
+                                FOR YOU
+                              </StyledH6>
+                              <H3
+                                isLoading={actionLoading}
+                                numberOfLines={3}
+                                ellipsizeMode="tail"
+                              >
+                                Some random text that encourages you
```
conrad-vanl: Lets throw in some good default language here, even if it is temporary (since we know Michael likes to show off the prototype). What did you think about the language I used in the sketch mockup?

packages/apolloschurchapp/src/tabs/home/Home.js
```diff
@@ -56,6 +69,36 @@ class Home extends PureComponent {
...
+                              </H3>
+                            </>
+                          }
+                          content={get(
```
conrad-vanl: Nice! I like how you just pass in the query result into this component. Makes it pretty flexible.

packages/apolloschurchapp/src/ui/ContentTableCard/ContentTableCardItem.js
```diff
@@ -0,0 +1,78 @@
...
+}))(H6);
+
+const TextContainer = styled(({ theme }) => ({
+  marginTop: theme.sizing.baseUnit / 2.5,
+  borderBottomWidth: 0.5,
+  height: theme.sizing.baseUnit * 4.25,
+  borderColor: theme.colors.shadows.default,
+}))(FlexedView);
+
+const Cell = styled(({ theme }) => ({
```
conrad-vanl: We don't have to do this in this PR, but we have started using this version of the "TableView" elsewhere in concepts, and I think it's something we'll want to move into a standard UI Kit (the component that allows for a large-ish image thumbnail with two lines of text next to it).

I do wonder though what about `TableView` didn't allow you to do that. The [Willow Creek Prototype app does something similar here](https://github.com/Differential/willow-prototype/blob/master/packages/WillowCreekApp/src/tabs/my-willow/index.js#L67) and it uses the standard `TableView` components, other then a custom `CellImage` component. I'm curious to know what didn't work when you tried the `TableView` components, as I'd really like to be able to use these components elsewhere.

packages/apollos-data-connector-rock/src/content-items/data-source.js
IsaacHardy: This is required.

packages/apolloschurchapp/src/tabs/home/Home.js
IsaacHardy: it was PascalCased on @burkeshartsis 's suggestion. However, I believe this can change, now that the dynamic header returns more than just one element.

packages/apolloschurchapp/src/tabs/home/Home.js
IsaacHardy: I'm sure it was adequate.

packages/apolloschurchapp/src/tabs/home/Home.js
IsaacHardy: Updated H6 style.

burkeshartsis: The use of vertical rhythm here is debatable but it does appear to calculate to a similar value. 

I would love to talk about ui-kit constructs more especially as it relates to theme values and colors.

Issue #519, opened 2019-02-12
[Onboarding] Present on-boarding experience/routes when first opening app


conrad-vanl:
### User Story

As a user I should be presented the `Auth` route then the `Onboarding` route every time I log in.

### Description

Everytime a user isn't logged in, we should force the `Auth` route. After login, we should then direct the user to the `Onboarding` route. After `Onboarding`, we can take the user to the home screen.

Since we persist your login (we currently reset this each build but will be changing that) you should only have to login ~1 time per device. Every time you login on a new device, you need to be presented with the onboarding experience again to enable notifications and location services. For MVP, we are comfortable just re-showing onboarding every time you login. Again, we currently reset your auth state with every build, and we will change that to not function that way in a separate task.

### Acceptance Criteria

* [ ] It should show the `Auth` route if you're not logged in
* [ ] After logging in, push/switch to the `Onboard` route
* [ ] After onboarding, go back to the home tab

### Possible Implementation & Open Questions

* The yet-to-be-merge [`ui auth` package](https://github.com/ApollosProject/apollos-prototype/pull/591) adds an `afterAuth` style callback to make some of this easier, but I think we can go ahead and wire this up with the current `Auth` route in `apolloschurchapp` until that's merged
* We probably want to maintain separate `Auth` and `Onboarding` routes on the root navigator

### Other Resources

* Basecamp: https://3.basecamp.com/3926363/buckets/6947926/todolists/1506896432

conrad-vanl:
@richarddubay said (on Slack):

> the `Auth` component/route doesn't have any logic to tell whether or not someone is already signed in. It looks like that logic exists on the `Connect` component/route. Should we default the route to `Connect` then? Or update the `Auth` route to have the same `getLoginState` query that the `Connect` route does? Or maybe something else

conrad-vanl:
So I'm not 100% sure why you need the "login state" logic inside of `Auth`.

I think what you probably need to do is somewhere near the root-level of our app stack, check for user login state, and if user is not logged in, redirect to the `Auth` component. This root-level can use a `<Query />` component querying for login state. Something like:

```
const getLoginState = gql`
  query {
    isLoggedIn @client
  }
`;

<Query query={getLoginState}>
  {({ data: { isLoggedIn } = {}, loading: innerLoading }) => {
    if (isLoggedIn) showLoginSomehow();
  })}
</Query>
```

conrad-vanl:
[...]

PR #615, opened 2019-04-09
Adds `AuthLoadingSwitch` to Navigate different directions, depending on if a user is logged in.

vinnyjth:

Adds and implements an AuthLoadingSwitchComponent, which navigates different ways if a user is logged in. When this component is the default route of a navigator, it will wait for a query to return describing if a user is logged in (including waiting for the cache to load) and then navigate to a named route described by the component. 

This is used to handle navigating correctly when the app mounts. To the "home" if the user is logged in, and to the "auth screen" if the user isn't logged in.

Here's an example of usage

```
const AuthLoading = (props) => <AuthLoadingSwitch loggedInRouteName="Home" loggedOutRouteName="Auth" LoadingIndicator={LoadingComponent}

const AppNavigator = createStackNavigator(
  {
    AuthLoading,
    Home,
    Auth,
  },
  {
    initialRouteName: 'AuthLoading',
  }
);
```

Chose to create this component in the `ui-auth` package to reduce the amount of code that needed to be copy/pasted.

### How do I test this PR?

1. Open the app when logged out, refresh the app, you should end up on the auth screen.
2. Open the app when logged in, refresh the app, you should end up on the tab view.

Snapshot tests over new components.

## TODO:

- [x] References #519 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-auth/src/index.js
```diff
@@ -70,6 +71,7 @@ const CancelButton = styled(({ theme }) => ({
 class Auth extends PureComponent {
   static navigationOptions = {
     header: null,
+    gesturesEnabled: false,
```
burkeshartsis: I'm not sure we want to default this for everywhere auth is used in the app. I agree we need it for how we are using it in this PR though. At the same time, we are transitioning from an app that you auth when you need it to an app that forces auth up front. Either way, I feel like this component should be more flexible than restrictive. Also email login here is just temporary until we get SMS auth merged in.

packages/apollos-ui-auth/src/login/Form.js
```diff
@@ -81,7 +81,7 @@ class Form extends PureComponent {
             </View>
           </PaddedView>
         </KeyboardAwareScrollView>
-        <BottomSafeAreaView>
+        <BottomSafeAreaView forceInset={{ bottom: 'always' }}>
```
burkeshartsis: This is broken on Android. I'm sure it's broken in lots of places. We should probably handle this in a seperate PR but I wanted to mention it.

packages/apollos-ui-auth/src/index.js
burkeshartsis: This is the only real question to address before merge.

vinnyjth: @burkeshartsis Not worried about this at this point, we can still hit cancel to close. It's a good callout to make however. 

If `react-navigation` made configuring this option easier, I would be all for it, but as of right now we'll have to work some magic in order for this setting to be configurable via a prop.

Issue #617, opened 2019-04-09
Further abstract hardcoded Apollos names and styles


redreceipt:
for iOS, we need to use splash screen pngs, not the xib file. this is for the CLI later. 
https://medium.com/@hellosunschein/launchimage-s-explained-33b88c02d027
https://github.com/crazycodeboy/react-native-splash-screen/tree/master/examples/ios/examples/Images.xcassets

for Android, we need to name the color something more generic than "deep water". also need to abstract the LauncherActivity name to @string/app_name.

I can start work on this ASAP as I would consider it a prerequisite for anyone making their own apps

cc @burkeshartsis

vinnyjth:
@redreceipt Can you close this now?

PR #624, opened 2019-04-11
Adds Static iOS Launch Images

redreceipt:

We need to use static launch images for scalability and flexibility with our future CLI and allowing partner churches to use their own media without touching XCode.

Now we are using a background image and auto-scaling instead of using a background color.

### How do I test this PR?

Boot up iOS in a few different sims.

None.

## TODO:

- [x] Closes [#617]
- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
Before I say anything, I want to be clear: I don't have an opinion on this yet.

> We need to use static launch images instead of the LaunchScreen.xib for scalability with our future CLI and allowing partner churches to use their own media without touching XCode.

Why? Are you assuming that this "future CLI" will need to generate all of these images? Or are we asking churches to generate these manually? These images still need to be imported into the xcode project, right, requiring the opening of xcode?

redreceipt:
> Before I say anything, I want to be clear: I don't have an opinion on this yet.
> 
> > We need to use static launch images instead of the LaunchScreen.xib for scalability with our future CLI and allowing partner churches to use their own media without touching XCode.
> 
> Why? Are you assuming that this "future CLI" will need to generate all of these images? Or are we asking churches to generate these manually? These images still need to be imported into the xcode project, right, requiring the opening of xcode?

I believe as long as they match the name and size (our CLI could help with that) churches should never have to touch XCode theoretically. It'll just do a straight swap of the media files.

conrad-vanl:
@redreceipt doesn't the existing storyboard use two images as well? A separate background image and icon? How is this different/better?

vinnyjth:
@redreceipt Because the assets in the `Launchscreen.xib` are, at the end of the day, files stored on disk referenced in an asset catalog, I would love to see if we can continue to use the `xib`.

redreceipt:
@vinnyjth @conrad-vanl I definitely agree with both of you, I image won't be a problem. It's the [background color](https://github.com/ApollosProject/apollos-prototype/blob/5345ca62a7566e30f7d45d3e5449ce8e0db2a763/packages/apolloschurchapp/ios/apolloschurchapp/Base.lproj/LaunchScreen.xib#L23) that I think may be a problem. It's a custom color space (sRGB) with floats as the colors. May be able to use standard RGB instead and then it would be an easy swap for people. I'll play around with that. But it was generated by dropping the image in XCode I believe. 

Obviously, I'm assuming a lot so any other insight would be welcomed. I just want to make it easy for people. imagine this flow if you were a church IT guy: `apollos add-splash --img splash.png --color #AAA` 🙌

redreceipt:
Ok so after a little digging, the only thing I can come up with is we just use floats that we'll have to calculate (ex. 0/255, 85/255, 90/255) and drop the result into the `LaunchScreen.xib` file. A little hacky but it works. I think the `react-native-splash-screen` lib we're using is overriding any Objective-C I'm trying to write. (see the `AppDelegate.m` file).

What do y'all think?

conrad-vanl:
What if we made the background an image as well? For apollos-prototype, it's just an image of a solid color. That would give church's flexibility to use textures and stuff instead if they wanted.

redreceipt:
That just kinda defeats the purpose of using the `xib` in the first place since the image will have to scale for every new screen size. With `pngs` they recommend using so many images so the app doesn't have to do any heavy lifting on boot.

But it would be easier for the user to only have to make one image...

vinnyjth:
@redreceipt After pushing a few different apps to the app store, the `xib` is a lifesaver because I only have to make one image, and don't have to constantly make new images as the new iPhones and iPads with wonky screen sizes come out. I would love to figure out if we could parse that xib file and introduce a background color.

If we can't do that, the xib story board supports forcing images to "fill screen". it's a kinda weird hack, but you could have the CLI generate a single pixel of a color you prefer and assign it to that asset, and then the xib would blow that asset up to fill the whole viewport. Here's an example: 
https://www.slideshare.net/mystcolor/stretchable-images-in-uiimageview-using-ib-only

conrad-vanl:
You can also have it `tile` the image if you don't want to stretch it 😉

vinnyjth:
> You can also have it tile the image if you don't want to stretch it 😉

redreceipt:
@vinnyjth ok I think those would both work. 

1. The first option is we parse the `.xib` file. We can do this by replacing the floats that are generated from XCode with our own slightly less precise floats. (ex. red=0.484589459984 becomes red=0.4845). Probably unnoticeable and this would allow them to still use the storyboard UI fairly easily. It writes over itself in both directions.

2. This method would be more precise but maybe a little more complex for churches and partners to tinker with themselves. Our CLI could certainly generate the one-pixel image for them but if they ever opened up the XCode UI, it may be a little confusing as to what is going on.

What do you think?

conrad-vanl:
My vote would be to just instruct the church to provide an image, and if they want the background to be a solid color, to just provide an image with a solid color. That way there's flexibility that if the church wanted to provide a photo background or a texture they could.

redreceipt:
@conrad-vanl I don't think a photo or textured image would ever be the right option considering Xcode would have to scale it and it wouldn't be right. I think for the CLI we ask for a logo and a color and if they want to get fancy, they can go in Xcode and figure it out, it wouldn't be that difficult. They'd just have to use static images at different sizes. 

@vinnyjth out of the two options above, which do you like?

conrad-vanl:
@redreceipt i used a photo texture on the CLC / Watermark app and it looked great 😀
https://github.com/watermarkchurch/clc-conference-app/blob/master/packages/ConferenceApp/ios/ConferenceApp/Images.xcassets/SplashImage.imageset/CLC19%20SplashScreen%20RJ%20BG-1.png

redreceipt:
@conrad-vanl that one would look great when stretched in any direction. If the icon was included in that image it would be distorted. Photos would be distorted as well.

conrad-vanl:
@redreceipt there's different fill modes. `tile` for textures, `fill` (or `cover`?? can't remember) for photos, etc. I'm suggesting we keep 2 images - a separate background image and foreground image. The foreground image is the icon that sits in the center like we have. We'd just make the background an image as well (we could rip the scene right out of the CLC app I linked to). The instructions can be written based on a solid background, but just let that be an image. If we already support an image in our scene I just think we'll have more flexibility down the road.

redreceipt:
gotcha. so just imagining the CLI: 

```
apollos add-splash --color #FFF        // this would create a solid color image and use tile
apollos add-splash --texture file.jpg      // this would use tile
apollos add-splash --photo file.jpg        // this would use cover
```

cc @conrad-vanl

conrad-vanl:
i mean, i imagine V1 of the cli will just take an image and fill mode and won't create a solid color image on its own...but yeah something like that

```
--backgroundImage file.jpg --backgroundFillMode cover
```

🤷‍♂️

redreceipt:
yea I like that. ok I'll make those changes and push it.

redreceipt:
got it up, there's a little space around the edges that I'm trying to figure out.

redreceipt:
ok, I got it but it's a little weird how I did it. Made the image slightly larger than the frame and then set the auto-resize to full screen and the content mode to "scale to fill". 

cc @vinnyjth @conrad-vanl

PR #627, opened 2019-04-12
Updates the passkit package to latest version

### LINKED ISSUE #625: Update Passkit-generator version
### Details

> Dear Conrad,
> my name is Alexander and I'm the creator of Passkit-generator on Github and NPM.
> I'm contacting you because I found the package you contributed to, apollos-data-connector-passes, through NPM as dependent on mine.
> I was looking at package.json and found that you are relying on a version of passkit-generator that does not exist anymore on Github. Better, you are relying on a branch ("next") that does not exist anymore as it was merged into master.
> My fear is that if you try to do npm install on your package, you might not download correctly the dependency.
> I know, I might have had ignored this and let you discover this by yourself, but I preferred you alert you of this.
> 
> For the future, do not rely on branches outside master, as they will be deleted once ready to be merged. 😊
> 
> Hope you appreciated this and have a nice day,
> Alexander

vinnyjth:
This upgrades the PassKit package to the latest version. Thanks for the heads up @alexandercerutti !!!

We were overriding the `_patch` method, however the code we copy pasted from the original `_patch` method changed, so our override was no longer working. I refactored the code to use `super._patch` to improve the ease of updating the passkit package in the future. I had to make a change to the `_validateType` override as well to make this possible. 

n/a

### How do I test this PR?

1. Start the API
2. Login, and get your user's token
3. `curl "http://localhost:4000/pass/CHECKIN" \
     -H 'authorization: YOUR_TOKEN_HERE' --output pass.pkpass && open pass.pkpass`

n/a

## TODO:

- [x] Closes #625 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
Looks like my gif encoder doesn't like green 😂 

I promise the pass is still the right color.....
image

packages/apollos-data-connector-passes/src/PassGenerator.js
```diff
@@ -19,53 +19,23 @@ class ApollosPassGenerator extends Pass {
...
-      passFile[this.type].transitType = this.transitType;
-
-    return Buffer.from(JSON.stringify(passFile));
   }
 
   // Similar case here. We're just going to allow all types, because we can't reliably parse the pass template yet.
-  _validateType = () => true;
+  _validateType = () => {
+    // TODO: in the future, we might want to support more than the generic type.
+    this.type = 'generic';
```
conrad-vanl: Why do we need to set the type to generic? I think our previous code would support non-generic types, although we didn't explicitly test for it

vinnyjth: My comment is misleading :P The only place that `type` field is used, is to perform a validation in that `_patch` method. Setting this type doesn't restrict our ability to generate passes of other types (like my comment implies), but rather overrides a check that occurs around the transit type in `_patch`. 

I had to override that check, because the `_patch` method assumes a `type` property is set on the `PassGenerator` instance. If there's not `type` property (such is the case if I don't set it here) than the `_patch` method throws an error here: 

https://github.com/alexandercerutti/passkit-generator/blob/master/src/pass.js#L652

`Can't set property "transitType" of undefined`

vinnyjth: Now that I think about it though, I believe we might run into situations where if we set a type like `boardingPass`, we will run into the same issues (because we are assuming the type is `generic`). 

I see a few solutions:

1. Write a regex or similar that can parse the `type` out of the non-JSON parseable string we provide. 
2. Write a PR for Alexander that checks for the presence of `type` before checking for a property on type, therefore catching this error on his end (my favorite solution)
3. Add a warning in the README.md that, for now, you should stick to `generic` types.

conrad-vanl: Is it mergeable as-is, and we can open up an issue to tackle #2?

alexandercerutti: Hi @vinnyjth and @conrad-vanl,
I'm following a bit the topic.

Passkit generator was created with the concept of per-use models in mind. I mean, one model for one set of things. So each model, should have a type.
So, technically speaking, `this.type` has not been documented as passkit-generator expects to find it in the `pass.json`.
Now that I'm writing this, I might consider this behavior to be changed, looking at https://github.com/alexandercerutti/passkit-generator/projects/1#card-19901583, but I cannot assure you I'll do that (at least, soon).

vinnyjth: @conrad-vanl Totally mergeable as is :D I think option 2 is our best path forward, and I can work on that with Alexander! 

@alexandercerutti I might be getting some of the details wrong here, but the root of the problem is that we pass a JSON template to your library that isn't valid JSON. Down the line, we run that JSON through a renderer (kinda like `.erb` or `.haml`) renders some template tags into valid JSON. When we pass our JSON template to your library, without the extensions we make to your class, the class from your library fails (for good reason, we don't pass valid JSON 😂) 

I think we probably have some further thinking to do about how to handle case like this, but in the meantime it would be really helpful if you could add a check to this line
https://github.com/alexandercerutti/passkit-generator/blob/master/src/pass.js#L652
so that we check if `this.type` is defined, before using it to assign a value. Something like 
```
if(this.type){
  passFile[this.type]["transitType"] = this.transitType;
} else {
  console.warn('`type` is undefined. You shouldn't see this unless manually extending this class.')
}
```

alexandercerutti: @vinnyjth Excuse me, but I'm not getting how are you passing the rendered JSON to the library. So you create dynamically the model, then once rendered the JSON, you are calling `.generate()`?

About the check, it may have sense but sadly I don't release a new version for just one little change. I think I'll include it in v.1.6.0.

vinnyjth: Correct - it looks kinda like this

1. Create pass from `template`. The `template` may or may not be invalid JSON. 
2. Use our internal templating engine to render the invalid JSON into valid JSON, `passData`. 
3. Set the pass `_props = passData`
4. We then call `pass.generate`. 

No worries about the check! I have it on my todo to write you a PR, and then when you merge it I'll watch the releases.

alexandercerutti: I don't use Github releases. I usually notify personally who make PRs.
Anyway, I prefer to do this by myself because I have to write the debug message (I won't insert `console.warn`). Or do that, we'll discuss there. 😊

alexandercerutti:
I'm sorry for replying here: I know the PR is closed but I wanted to inform you that I'm going to change the name of `_validateType` on the next release (1.6.1). I'm also planning to change where the type is memorized (probably not anymore in that function, but in a new method with signature ` async _extractPassDefinition()`).

I'm informing you as you are the only public project I know that uses my package.
I'm available for anything you need.

Also, I'll try to find a way to let a type to be set manually (even if I won't document this as this shouldn't be done for me) as we had discussed earlier in the topic.

Keep an eye on the project. I'll try to inform you when I'll release the new version.
You can currently act in two ways: use the current version (1.6.0) statically - removing any other selector in `package.json` - or check another way to achieve what you are currently doing.

PR #680, opened 2019-05-06
Adds Storybook as an app

conrad-vanl:

This creates a new storybook app as a separate package and loads stories from all of our frontend packages.

This introduces a second react-native app in core. The positive to this is that we could transition core-team development over to working 100% on storybook - building out features, components and screens directly within storybook. We could go as far as loading the entire `apolloschurchapp` prototype app as a story. This gives us the ability to have a 100% "playground" environment, while we can then work towards transitioning `apolloschurchapp` over to 100% boilerplate/starterkit project.

The trade-off is `apolloschurchapp` now has no built-in storybook, and so churches that create their own apollos project won't have storybook built in.

@vinnyjth - @burkeshartsis and I talked quite a bit about this and I'm not sure that we are 100% sold on the direction of this PR, nor are we certain what would be best. 

### How do I test this PR?

You can run the storybook app by `cd`'ing into the `ApollosStorybook` folder and running `yarn start`. It's just a typical react-native project.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


package.json
```diff
@@ -6,8 +6,6 @@
       "packages/*"
     ],
     "nohoist": [
-      "**/react-native",
```
conrad-vanl: I think we can add these back...but interesting if the app works without them might keep this removed?

packages/ApollosStorybook/native-storybook.js
```diff
@@ -0,0 +1 @@
+export * from '@storybook/react-native';
```
conrad-vanl: This allows all of our other `ui-` packages to import the same instance of the storybook dependency, instead of having to require their own (which doesn't work anyways)

package.json
vinnyjth: Maybe. One thing I've noticed is that changes to `nohoist` don't always propagate to other's machines easily. Sometimes I'll push something that worked locally but find that `lerna` didn't apply the changes to my machine before I pushed, and so anyone who pulls and nukes will have a broken setup. 

tldr: We can know for sure after nuking :D

vinnyjth:
We might have less conflicts if you create a new branch and cherry pick your commits, I think this branch is based on an out of date `onboarding-fixes` and so merging will be a mess.

conrad-vanl:
> We might have less conflicts if you create a new branch and cherry pick your commits, I think this branch is based on an out of date onboarding-fixes and so merging will be a mess.

yeah i'm going to do this....

PR #681, opened 2019-05-06
Storybook app as package

conrad-vanl:
### What does this PR do, or why is it needed?

This creates a new storybook app as a separate package and loads stories from all of our frontend packages.

### What design trade-offs/decisions were made?

This introduces a second react-native app in core. The positive to this is that we could transition core-team development over to working 100% on storybook - building out features, components and screens directly within storybook. We could go as far as loading the entire `apolloschurchapp` prototype app as a story. This gives us the ability to have a 100% "playground" environment, while we can then work towards transitioning `apolloschurchapp` over to 100% boilerplate/starterkit project.

The trade-off is `apolloschurchapp` now has no built-in storybook, and so churches that create their own apollos project won't have storybook built in.

@vinnyjth - @burkeshartsis and I talked quite a bit about this and I'm not sure that we are 100% sold on the direction of this PR, nor are we certain what would be best. 

The other option is to just add storybook back to `apolloschurchapp`, but I think I'd want to make storybook the default thing (render it at the root, with one of the stories being the app), with a welcome screen that explains what's going on. And then a comment in the code inside of `App.js` on how to disable storybook being ran at the root, or something like that....not sure.

Basically, I'd love core development to transition inside of storybook, instead of something you just run optionally. But I still want to give church app developers an instance of storybook they can use for their app.

### How do I test this PR?

You can run the storybook app by `cd`'ing into the `ApollosStorybook` folder and running `yarn start`. It's just a typical react-native project.

## TODO:

- [x] PR has a relevant title that will be understandable in a public changelog (ie...non developers)
- [ ] Closes [tag-issues-here]
- [x] No new warnings in tests, in storybook, and in-app
- [ ] Upload GIF(s) of iOS and Android
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

- [ ] Manual QA on iOS and ensure it looks/behaves as expected
- [ ] Manual QA on Android and ensure it looks/behaves as expected

**Code Review: Questions to consider**

- [ ] Read through the "Files changed" tab _very carefully_
- [ ] Edge cases: what assumptions are made about input?
- [ ] What kind of tests could be written?
- [ ] How might we make this easier for someone else to understand?
- [ ] Could the code be simpler?
- [ ] Will the code be easy to modify in the future?
- [ ] What's one part of these changes that makes you excited to merge it?

> The purpose of PR Review is to _improve the quality of the software._

PR #683, opened 2019-05-07
Storybook as an NPM package

conrad-vanl:

A different take on #681. Instead of having a separate app within `apollos-prototype`, this creates a `@apollosproject/ui-storybook` package that can be implemented on any Apollos Church app. So we integrate it into Apollos Church app.

I do not currently like the way it is integrated. I THINK that I would like to make running storyboard the default in development, and the app the default in production, with the ability to turn off storyboard in development easy. But want advice on that.

You have to import storyboard stuff from `@apollosproject/ui-storybook` instead of `@storybook/react-native`

This doesn't implement anything for storybook-server yet. I think storybook-server becomes slightly more complex with storybook being implemented into the main app.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
I don't understand all the nuances, but I think I like this approach better :)

packages/apolloschurchapp/src/index.js
```diff
@@ -1,70 +1,16 @@
...
-    Onboarding,
-    LandingScreen,
-  },
-  {
-    initialRouteName: 'ProtectedRoute',
-    mode: 'modal',
-    headerMode: 'screen',
-  }
-);
+const App = Storybook;
```
vinnyjth:
```
let App = App;
if (__DEV__) {
  App = Storybook
}
```

I think the packager removes the `if` block entirely in prod

vinnyjth: I think we could also (using env variables) create a different `npm` command that would set an `env` variable that could drive this value. Make it easy to launch the app instead of storybook.

conrad-vanl:
Here's what I'm thinking this branch needs to be mergeable:

1. A new "Welcome" story that describes what storybook is, how to use it, and how to turn it off
2. An npm command to run w/o storybook enabled, and consideration on how to publish app w/o storybook for our church partners that adopt `apolloschurchapp`
3. A way to publish the storybook app to the app store. This is something we should discuss: I see immense value in having an Apollos Storybook app on the app store (probably more value then having a prototype app). But do we publish both the prototype app (just to the beta testflight) and the storybook app (all the way to production), or do we move to only publishing the storybook app?

vinnyjth:
@conrad-vanl That makes sense. One thing that just occurred to me is a lot of the changes we make here will be picked up by the upgrade tool, so partners will get this storybook setup by default when they upgrade.  That sounds like a double edged sword, great if they want it, but probably really confusing and added complexity if they don't?

conrad-vanl:
@vinnyjth that's why Number 1 in my list above is important IMO. I think storybook-driven development is something we can better advocate, and this allows us to advocate for it. But still provide an easy path to get rid of it if they don't want it :)

vinnyjth:
>An npm command to run w/o storybook enabled, and consideration on how to publish app w/o storybook for our church partners that adopt apolloschurchapp
>A way to publish the storybook app to the app store. This is something we should discuss: I see immense value in having an Apollos Storybook app on the app store (probably more value then having a prototype app). But do we publish both the prototype app (just to the beta testflight) and the storybook app (all the way to production), or do we move to only publishing the storybook app?

I think both of these we can solve by creating a `.env.storybook` and then driving the rendering / not rendering of storybook via `react-native-config`. If we do it like that, we could go as far as having the option of picking storybook/not storybook by setting a value in Appcenter.

vinnyjth:
https://github.com/luggit/react-native-config#different-environments

packages/apolloschurchapp/src/index.js
burkeshartsis: Try looking at how we currently pipe storybook into the packager without having to do any conditionals to `index.js`

conrad-vanl: I know, I saw that. That requires storybook server though yeah?

conrad-vanl: ah I see just with `--projectRoot`. Not a bad idea. the react-native-config stuff @vinnyjth showed is interesting too esp if it can be configured in appcenter

vinnyjth:
@conrad-vanl Give me the word and I'll implement the RN Config version of switching between Storybook / Main App. Maybe @burkeshartsis could write us a killer description?

conrad-vanl:
I can work on the welcome slides and such if @vinnyjth wants to take the architectural stuff

package.json
```diff
@@ -22,7 +22,9 @@
     "test": "./node_modules/.bin/lerna run test",
     "ci-test": "./node_modules/.bin/lerna run test  --stream -- --maxWorkers=2",
     "storybook": "./node_modules/.bin/lerna run storybook --stream --scope apolloschurchapp",
-    "ios": "./node_modules/.bin/lerna run ios --stream --scope apolloschurchapp",
+    "ios": "npm run ios:storybook",
```
conrad-vanl: so how does this work with just a plain `yarn start`? I believe the general instruction is to start apollos by running `yarn start` at the root (or atleast, that's how I always do it 😇 ). I'm not sure who actually runs the `yarn ios` cmd from the project folder itself

vinnyjth: The `react-native-config` values are set via building, and aren't updated when starting/restarting the packager :/

How do you normally put the app on the simulator / your device? One thing we can do (alongside this change) is create a xcode schema that defaults to storybook.

conrad-vanl: I usually just build with xcode

vinnyjth: Okay - I'll add a schema that picks defaults to the storybook.

conrad-vanl: Makes me wonder if @burkeshartsis's approach that's currently in master might be the better solution, I'm not sure...

vinnyjth: yah - having to run react-native run-ios is kinda a letdown. It's pretty fast one you run it once though, I tend to run it multiple times a day b/c it helps reset your app state when changing branches

conrad-vanl: Agree. Just having to create a new schema kinda sucks. His solution with starting the packager with a different `--projectRoot` arg is interesting to me

vinnyjth: Yah I do like that a lot. Downside of that approach is that I imagine it might be more difficult to build a production app that's not storybook?

conrad-vanl: why? doens't appcenter start a packager when it builds?

vinnyjth: It does, but I think how that packager is started and what it points at is not done via our package.json but is configured as a build step in xcode. I don't know the details 100%, and I'm sure we could solve it, but I'm not immediately familiar.

conrad-vanl: Gotcha. What if we dropped `.env.storybook` and just made the configuration point part of `.env`?

vinnyjth: Help me understand. You're suggesting

Same approach, remove the ability to switch "apps" via different run commands.
If people want different environments have change your `.env` files and rebuild. No complexity around building a new `.env` file. 

I'm totally on board with this. As a note, you can do this already by adding
```
STORYBOOK=true
```
To your `.env` file. If you do that right now, either run command will launch storybook 🤷‍♂

conrad-vanl: Correct. I don't dislike the multiple run commands, what I'm worried about is your suggestion to create a new scheme. I don't want to create a new scheme

conrad-vanl:
@vinnyjth how can I help get this merged today / by EoW? I'd like to get a testflight build up with storybook enabled for Michael R. I'll take care of the merge conflicts. I just want to make sure we have clear direction on how to run w/ or w/o storybook, and a path for us to publish the storybook version on testflight (and we probably need a discussion on whether we want to publish two separate builds to testflight - one w/ and one w/o storybook)

vinnyjth:
@conrad-vanl I'll take out all my magic with with the `npm start` commands, and we can drive the storybook purely through changing your .env variables. I'll have the production env setup so we can push the choice about doing what the builds down the road, and accomplish that purely through setting AppCenter environment vars.

conrad-vanl:
@vinnyjth might need your help with this one. Can't figure out why travis is failing....both of our scripts are exiting with a 0 exit code, yet at the end of the build it shows exiting with code 1.

PR #696, opened 2019-05-09
Adds Apollos.json to help improve upgrade tool.

vinnyjth:

This PR creates an `apollos.json` file. This file only exists so that the upgrade tool has something to work with when assuming the user's current version and current environment. With this file, we can simplify the code required to complete the process of minimizing the args that you need to pass to the upgrade tool. 

When you run the upgrade tool, it will look at this file to determine it's context (api or client) and then look at the version to find the "current version". When the upgrade tool is run, this file will automatically be updated to reference the new current version (the end user will never have to touch this file). 

I also added steps to the release process so that this file stays in sync with the current package version.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apolloschurchapp/apollos.json
+++ b/packages/apolloschurchapp/apollos.json
@@ -1,4 +1,4 @@
 {
-  "version": "0.8.5",
-  "environment": "client"
+    "version": "0.8.5",
+    "environment": "client"
 }
--- a/packages/apolloschurchapp/apollos.json
+++ b/packages/apolloschurchapp/apollos.json
@@ -1,0 +1,4 @@
+{
+  "version": "0.8.5",
+  "environment": "client"
+}
--- a/packages/apollos-church-api/apollos.json
+++ b/packages/apollos-church-api/apollos.json
@@ -1,0 +1,4 @@
+{
+  "version": "0.8.5",
+  "environment": "api"
+}
--- a/packages/apollos-church-api/apollos.json
+++ b/packages/apollos-church-api/apollos.json
@@ -1,4 +1,4 @@
 {
-  "version": "0.8.5",
-  "environment": "api"
+    "version": "0.8.5",
+    "environment": "api"
 }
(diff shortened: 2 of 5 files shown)


packages/apolloschurchapp/apollos.json
```diff
@@ -0,0 +1,4 @@
+{
+    "version": "0.8.5",
+    "environment": "client"
```
conrad-vanl: how is `environment` used on the upgrade tool? should this name be more explicit? (like `apolloschurchapp` or something that directly references what it is?) I'm just trying to think future proof....

vinnyjth: It's flag that we currently pass. It's either `api` or `client`, and so this is a direct translations of that flag. it drives whether the `diff` used to upgrade is the client diff or the api diff. 

 I think this is okay as it stands right now :)

vinnyjth: If you look at the upgrade tool branches, I use the `api`/`client` language very consistently. 
https://github.com/ApollosProject/apollos-upgrade/branches

And here's how we specify the "platform". Maybe I should rename this key to platform....
https://github.com/ApollosProject/apollos-upgrade/blob/master/src/index.js#L27

IsaacHardy:
Assuming this works as intended, I would run `apollos-upgrade upgrade --projectName newspringchurchapp --packageName com.newspringchurchapp` and we'd setup the script to fetch the apollos.json from CORE and read this local file to determine current version. 

Is this correct?

vinnyjth:
> Assuming this works as intended, 
You bet it works as intended ;) 

>I would run apollos-upgrade upgrade --projectName newspringchurchapp --packageName com.newspringchurchapp and we'd setup the script to fetch the apollos.json from CORE and read this local file to determine current version.

This is almost correct. We read the current version from the remote lerna file (which you already did) and then read the local file for the current version.

conrad-vanl:
My only worry is that will it matter if we have multiple "client" templates? like a react-native app, a conference app, a web site (down the road)

vinnyjth:
> My only worry is that will it matter if we have multiple "client" templates? like a react-native app, a conference app, a web site (down the road)

That's a good callout 🤔 In that case, we will have to reorg more than once place (also have to upgrade the diff repo to keep track of the different clients). For now I think it's safe to keep it as simple as `client/api`

PR #698, opened 2019-05-09
Fixes Content Items back redirecting to Onboarding

vinnyjth:

Under certain conditions, clicking "Back" on a ContentItem would redirect the user to Onboarding, rather than the appropriate tab screen. This is because "Onboarding" was still technically the top of the card stack.

### How do I test this PR?

1. Logout.
2. Complete onboarding
3. Open a content item
4. Close the content item
- You should be redirected to he screen you viewed the content item from, not "Onboarding".

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
@vinnyjth there's a similar bug that occurs but less frequently... when you open the sections menu and tap into a section, then tap on a content item in that section. Then tap back - some of the time it takes you back to the sections menu instead of the individual section. It's like it does a `history.back()` on both navigators instead of just the top most one. Is that related to this at all?

vinnyjth:
@conrad-vanl That's interesting, I think the underlying cause might be similar (the back button takes you back way too far). 

This fix is definitely needed (otherwise onboarding will live at the root) of the navigation stack rather than tabs. I believe this means that Android users could potentially hit their hardware back button to get to it. 

We can look at the back button in the card modal as well though, and figure out how to enforce only going back one level. Maybe we should be calling `pop` instead of `back`?

richarddubay:
I do think this is bigger than just the Onboarding. Like @conrad-vanl mentioned there are other places (Go to Discover > View All in a section > select a piece of content > then click "X" and it takes you back to Discover) where this happens. It's like sometimes whatever is last is just not being registered as being on the navigation stack. 

We see the same behavior in the prayer module where we open a list of prayers in a `<ModalView>` and we tap on a link that opens a piece of content in another modal and when you tap on the "X" on that piece on content it closes both modals.

Not sure what that means, it just feels like this is bigger than just Onboarding.

vinnyjth:
I agree that we have an issue to resolve with ContentItems back button going back too far. I don't think that excludes the need for the issue this PR solves, which might be better to phrase as 

"stops people from being able to revisit onboarding by going back too far" Tapping back on a ContentItem happens to be an easy way to reproduce this.

vinnyjth:
btw this is probably why we are going back too far.

https://github.com/ApollosProject/apollos-prototype/blob/c55811093bdb977f01af46730cb8cc13bfe9aee0/packages/apolloschurchapp/src/content-single/NavigationHeader.js#L10

vinnyjth:
Thinking on this - we should probably replace the code above with something like `goBack()` or event `goBack(ContentItem)`

vinnyjth:
@IsaacHardy made the change referenced above and tested, seems to be working well. Let me know how it works for you.

IsaacHardy:
@vinnyjth This implementation work for the NS app. I could not replicate the issue no matter how many times I tried.

PR #715, opened 2019-05-15
Include storybook files with NPM release

vinnyjth:

I took this chance to clean up the usage of `node_modules/.bin` in our root level package.json.

Before publishing our packages, we should bundle up the latest storybook files so that our handy storybook package can find and use them.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-kit/.npmignore
```diff
@@ -1,2 +0,0 @@
-*
```
vinnyjth: This file was unused, because we had the `files` array inside the package.json

package.json
```diff
@@ -16,24 +16,24 @@
...
-    "start:now": "./node_modules/.bin/lerna run start --scope apollos-church-api",
-    "start:dev": "./node_modules/.bin/lerna run start --parallel",
-    "lint": "./node_modules/.bin/lerna run lint",
-    "test": "./node_modules/.bin/lerna run test",
-    "ci-test": "./node_modules/.bin/lerna run test  --stream -- --maxWorkers=2",
-    "ios": "./node_modules/.bin/lerna run ios --stream --scope apolloschurchapp",
-    "android": "./node_modules/.bin/lerna run android --stream --scope apolloschurchapp",
-    "release": "./node_modules/.bin/lerna publish --no-push && echo \"Release has been tagged!\nPlease look over the CHANGELOG.md file, and use 'git commit -amend' to amend any edits.\nWhen done, be sure to 'git push --tags' and create a release PR\"",
-    "release:canary": "./node_modules/.bin/lerna publish --no-push --canary --yes",
+    "start:now": "lerna run start --scope apollos-church-api",
```
IsaacHardy: What was the reasoning for switching away from using local .bin to run?

vinnyjth: I find the tradeoffs of using  `.bin` (harder to read package.json) worse than the benefits (forcing users to use local lerna, which is the default as long as they have `npm installed` once)

IsaacHardy: Isn't the trade off to protect devs from having incompatibilities between lerna versions? That seems pretty important especially when TAP becomes more open sourced in a community of devs not familiar with our practices.

vinnyjth: I'm not sure I understand where the incompatible lerna versions would come from. Also, if devs want to use lerna to spin up their own church apps that will be something they have to support :) I'm not using lerna when spinning up other instances of Apollos.

conrad-vanl: I believe that both `yarn` and `npm` now look at the `./node_modules/.bin` folder automatically before defaulting to the globally-installed version. @vinnyjth is going to confirm this is true before we merge 😀

Granted...with that said...working on this monorepo requires `yarn` anyways, and we do know that this is how `yarn` functions.

vinnyjth:
Can confirm that `npm` always looks at local version first. Tested with the following steps: 

1. `npm install -g lerna` 
2. `which lerna` 
  - Confirm my shell is using global lerna
3. `lerna version`
   - Note lerna version is most recent version
4. Added an npm script, `"lerna-version": "lerna --version"`

Issue #694, opened 2019-05-08
First Card on the Home Tab does not fill with content and is not clickable


calebpanza:
### Details

There is a single card that sits above the "For You" section. It's the very first card on the Home Tab. The card continually shows the card as being in loading state and is not clickable.

(I did run it locally in the simulator on my Mac which did result in the card filling with content, although it was still not clickable)

### Expected Behavior

Fill with some content and be able to click into the content to view the Detail View

### Actual Behavior

On my physical phone, no content ever loads in the very first card on the Home Tab.

On my Mac in the simulator, the very first card on the Home Tab fills with content, but is not clickable.

### Possible Fix

It looks like that card never had navigation properties added to it.

<details><summary>Additional Info</summary>

### Your Environment

- iPhone Xs [iOS 12.2]
- MacBook Pro Xcode [10.2.1] iPhone XS Simulator [iOS 12.2]

### Steps to Reproduce

iPhone Build
1.  Go through app on-boarding
2. Wait for content to load
3. Notice that no content loads on the very first card (above "For You")

Xcode Simulator
1. Go through app on-boarding
2. Wait for content to load
3. Click on the very first card and nothing will happen
[Test_Screenshots.zip](https://github.com/ApollosProject/apollos-prototype/files/3159090/Test_Screenshots.zip)

</details>

PR #719, opened 2019-05-15
Makes Campaign Card Clickable

redreceipt:
## DESCRIPTION

Kapture 2019-05-15 at 11 54 44

The campaign card needs to be clickable. Currently, it's not.

Now it's just rendering a `ContentCard` and just passing props instead of a double network call.

### How do I test this PR?

Boot up the sim, click the first card.

None.

## TODO:

- [ ] Closes [#694]
- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
@vinnyjth can you give me a quick explanation of how I would figure out why this coverage is failing?

packages/apolloschurchapp/src/tabs/home/Home.js
```diff
@@ -86,11 +88,33 @@ class Home extends PureComponent {
                           {}
                         );
 
+                        const metrics = [
```
conrad-vanl: Not sure why `metrics` and `coverImage` is added to this component here......it shouldn't be. You're also changing the component from `ContentCardConnected` to `ContentCard`. Why?

redreceipt:
Because using the connected component it was doing a double network call,
one with Isaac's query and one with the standard content query, I just cut
out the middle man. I agree with you and honestly I was hoping it would
spark a discussion. I see two options, combining the queries into one and
just using the content card connected, or making a separate component
called CampaignCard that would be styled a little different. Personally I
think this is the way to go long term because it needs to be different than
a standard content card. But obviously this would require a design. What do
you think?

--
Michael Neeley
Junior Web Developer
NewSpring Church
www.newspring.cc
Cell: (803) 622-6599

conrad-vanl:
@redreceipt it shouldn't result in double network calls if the queries are set up right. Essentially, what you need to do is make sure you're querying for _all_ of the information that's inside of the `ContentCardConnected` query within the `Campaign` query, and then when the `ContentCardConnected` it will be able to read 100% from cache.

redreceipt:
Not sure how we're gonna cover the diff because it's mainly just a `TouchableScale` with an `onPress` handler.

redreceipt:
@conrad-vanl should I extend test coverage here? to make it pass? Not quite sure how if so.

vinnyjth:
@redreceipt I wouldn't worry about it in this case, I think you've got as much covered as you can.

PR #725, opened 2019-05-15
Adds Additional Campus Data

richarddubay:
This PR adds `isActive` to the data returned by the Campus model. This will allow us to filter on our client by only the campuses we need.

None.

### How do I test this PR?
Run the following query:
```
query {
   campuses {
      id
      name
      street1
      street2
      city
      state
      postalCode
      latitude
      longitude
      image {
         uri
      }
      distanceFromLocation
      isActive
   }
}
```

None

## TODO:

- [n/a] Closes [tag-issues-here]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
@richarddubay Do you think this would be better if it was an argument you could pass to the query?

richarddubay:
@vinnyjth I'm not picky on the implementation. We thought if we surfaced the field then we could filter on it on the client. But if it's a better implementation to be able to send an argument that allows us to get "active", "inactive", or "all", that's fine too. The point of this is that on the location finder it currently shows all of our campuses, even those that don't really exist anymore. 

Which is a better way to do it?

vinnyjth:
@richarddubay Given that the use of this field is to filter it on the client (and you won't be using the inactive campuses in any way) I would argue that having a filter in the query so you don't have to do any filtering on the client is the right approach :) 

Another option to consider would be encouraging implementers to extend the dataSource and add the filtering straight to the model here: https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-data-connector-rock/src/campuses/data-source.js#L17

This is probably a common enough use case though, that I do believe it should live in Core.

richarddubay:
@vinnyjth So as I'm sitting here thinking about this issue this morning, I can't help but think that eventually more people than just us would want to be able to filter campuses (or any other data source for that reason) by things like `active` vs. `inactive` states. So while extending the datasource on our end would work, this seems like a lot of people could potentially be writing the same code to do the same thing. 

Now, whether this is done with a filter in the query, or just adding the field as returned data that the client can then filter by on it's own ... I just want to do whatever is the best implementation. But it does feel (to me) like this whole concept should live in core. If only to avoid the need for code redundancy down the line.

vinnyjth:
That sounds great @richarddubay! Do you want to go ahead and add a filter by `IsActive eq True` in the dataSource? I feel like that's a good change that won't impair any existing users of the app.

richarddubay:
@vinnyjth Is there a way we can do this so that we can choose either active, inactive, or all? I'm not sure why we would _necessarily_ need the inactive ones but maybe someone would in the future? I'm just trying to think of options here.

vinnyjth:
@richarddubay I think we totally _could_, but I don't think adding that filtering ability adds any value for the end user. I would rather assume that the church wants to expose all `active` campuses and leave it that. We could use graphql query filters to add `active: True/False` as an input argument if we wanted, but I would argue that it would add needless abstraction/complexity.

Issue #672, opened 2019-05-02
[Production Readiness] When Liking your _first_ ContentItem, the UI never updates to display that new content item.


vinnyjth:
### Details

When Liking your _first_ ContentItem, the UI never updates to display that new content item.

### Expected Behavior

When liking your first content item you should see the item show up in the UI.

### Actual Behavior

You see nothing, until you refresh your app.

### Possible Fix

We are probably not mounting the required queries that are "updated" when liking ContentItems. We should probably make sure the container that loads Liked content items is mounted no matter what. 

<details><summary>Additional Info</summary>

### Your Environment

- Environment name and version (e.g. Chrome 39, node.js 5.4):
- Operating System and version (desktop or mobile):
- Link to your project:

### Steps to Reproduce

1.  Load the app with a logged in user.
2. Unlike all content items.
3. Refresh the app
4. Visit the connect tab
5. Like a content item
6. Visit the connect tab, you shouldn't see any liked content items.
7. Refresh the app.

### Stack Trace

</details>

redreceipt:
Gif me please. Not sure what the issue is here. Here's my experience.

![Kapture 2019-05-15 at 9 56 20](https://user-images.githubusercontent.com/2659478/57781240-c70c9100-76f7-11e9-9735-ed003f4175cd.gif)

redreceipt:
That also captures another bug, the "X" takes me back to Onboarding sometimes. If there's not an issue for that already I can make one.

vinnyjth:
@redreceipt That is fixed in #698

redreceipt:
gotcha, cool. well, I can't seem to reproduce the issue here, have any more details?

vinnyjth:
@redreceipt Did you follow my steps to reproduce above?

redreceipt:
By refresh, you just mean CMD+R?

redreceipt:
ah, I see it now

redreceipt:
ok so this may or may not be related but when liking something, this error comes out:

<img width="613" alt="Screen Shot 2019-05-15 at 1 12 12 PM" src="https://user-images.githubusercontent.com/2659478/57795139-3a6fcc00-7713-11e9-9a03-6a42b6adae64.png">

coming from this file

<img width="697" alt="Screen Shot 2019-05-15 at 1 13 45 PM" src="https://user-images.githubusercontent.com/2659478/57795179-55424080-7713-11e9-94b8-cc7181d5b631.png">

and this function

[...]

PR #731, opened 2019-05-19
Preloads Recent Likes

redreceipt:

The runs the `getLikedContent` query preemptively when we like something. This is simply to load the query into cache since we read it from the cache later in the `updateLikedContent` mutation. Since it can't currently find the query in cache to read from, it just try/catches out the error so the cache was never actually updated the first time.

Don't love this but I don't know of a way around it. If we could avoid using `cache.readQuery` in the update mutation, it wouldn't error out in the first place. Maybe we could pass in what we need instead?

### How do I test this PR?

Clear all your liked content, refresh the app, the like something for the first time and make sure it shows up in the recent liked feed in the connect tab.

None.

## TODO:

- [x] Closes [#672]
- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apolloschurchapp/src/ui/LikeButton/index.js
+++ b/packages/apolloschurchapp/src/ui/LikeButton/index.js
@@ -7,6 +7,7 @@ import { Query, Mutation } from 'react-apollo';
 import Like from 'apolloschurchapp/src/ui/Like';
 import { AnalyticsConsumer } from '@apollosproject/ui-analytics';
 
+import getLikedContent from '../../tabs/connect/getLikedContent';
 import updateLikeEntity from './updateLikeEntity';
 import getLikedContentItem from './getLikedContentItem';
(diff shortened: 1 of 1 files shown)


redreceipt:
@vinnyjth read the top portion of this PR before we merge, I'd love to know your thoughts, I think there's a better way around this. And then we could possibly remove the try/catch from `updateLikedContent` in the process as a bonus.

vinnyjth:
@redreceipt This is an interesting solution, but I think the `try/catch` is okay in this case :) I think this doesn't fix the underlying issue as well. 

When looking into this issue this morning, I managed to find an error in the resolver that, when fixed,  I believe solves this issue

https://github.com/ApollosProject/apollos-prototype/compare/672-fixes-liked-feed-not-populating?expand=1

vinnyjth:
Closing this in favor of #732

PR #735, opened 2019-05-21
Adds custom icon support

### LINKED ISSUE #549: Add custom icon support to `<Icon />`
This can probably be done through theme or by creating an `IconProvider` type component where you pass an icon set into that `<Icon />` reads from.

Also, since there's no good open-source svg-based `react-native` `<Icon />` component, we might consider moving our Icon work into a separate package, perhaps even separate repo with a standard open source license. It'd be good for the larger RN community.

burkeshartsis:
## DESCRIPTION
A technical MVP for custom icon support. This adds the nuts and bolts needed to make it work but these features should not be used until #742 is merged. Consider this part one if you will.

This PR exposes an `iconInput` via the `ThemeProvider` and is modeled after how we approached custom themes with `themeInput`. To add a custom icon just add them to the `/icons` directory in the church app and export them via the directories `index` and you're set. Use them like you do UI-Kit icons.

``` <Icon name="custom-icon-name" />```

>  "#Protip: you can override UI-Kit icons with custom icons that have the same name like `BrandIcon`" – @burkeshartsis 

Note that, like all components, your custom icons should follow `PascalCase` naming but that when you access them via the `Icon` component the name is `kebab-case`

This gives us access to the local (church app) `/icons` directory via theme context in UI-Kit. The down side with how it's implemented now is that there may be a perf  hit and/or bundle size bumb by importing every available icon with every instance of `Icon`. I think there are ways around this though should we run into problems.

### How do I test this PR?
`yarn start` and nothings dies 👌💥
 
There are snapshot tests that cover all the added functionality of this PR. 

## TODO:

- [x] Closes #549 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### CODE REVIEW

**Comment by burkeshartsis** [comment]
Currently blocked by this 😫

**Comment by vinnyjth** [comment]
So hyped for this

File: packages/apollos-ui-kit/src/Icon/index.js [hunk 1]
```diff
@@ -20,19 +20,46 @@ const pascalCase = (string) =>
...
-const Icon = enhance(({ name, size, isLoading = false, ...otherProps }) => {
-  const IconComponent = Icons[pascalCase(name)];
-  return (
-    <Placeholder.Media size={size} hasRadius onReady={!isLoading}>
-      <IconComponent size={size} {...otherProps} />
-    </Placeholder.Media>
+const Icon = enhance(
+  ({ name, size, iconInput, isLoading = false, ...otherProps }) => {
+    const Icons = { ...uikitIcons, ...iconInput };
+    const IconComponent = Icons[pascalCase(name)];
```
vinnyjth: 👍 This is an 0(1) lookup, which is nice.

packages/apolloschurchapp/src/icons/makeIcon.js
```diff
@@ -0,0 +1,10 @@
+import { withTheme } from '@apollosproject/ui-kit';
+
+// Currently used to inject default fill color to icons.
+// Eventually can be used for other shared functionality between icons
+const makeIcon = withTheme(({ theme, fill, ...otherProps } = {}) => ({
```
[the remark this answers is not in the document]
burkeshartsis: Fair enough. This is more fully addressed in #742 but for not I've removed the duplicate and pointed these files (which are deleted in #742) to point to the ui-kit reference.

packages/apolloschurchapp/src/icons/ArrowBack.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Svg, { Path } from 'react-native-svg';
 
-import makeIcon from './makeIcon';
+import makeIcon from '@apollosproject/ui-kit/src/Icon/icons/makeIcon';
```
vinnyjth: Is there any reason you chose to use the path to the function, rather than exporting the function from the module index?

vinnyjth: Burke is going to fix this in a future PR.

richarddubay:
Guys I can't wait for this! I so badly want to get rid of the `NSIcon` component we had to make to use our own icons. Great work!

PR #743, opened 2019-05-24
Adds smarter resetting of Apollo Cache

### LINKED ISSUE #740: Don’t reset Apollos state on each app upgrade
### User Story

As an app user I don’t want to have to re login every time I update my app. 

### Description

I think we can remove the check that resets the Apollo state every time the app is upgraded, and change it to just any time a crash occurs or we record a breaking state scheme change (maybe just a schema version string). 

### Acceptance Criteria

* [ ] It should [SPEC_DETAIL]

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

vinnyjth:
## DESCRIPTION

I added a mechanism by which we can more intelligently purge the Apollo Cache. Previously the cache would only reset under one circumstance - updating the app.

Now the app will reset under two conditions -
1. The API schema changes.
2. The app fails to restore the store to the cache.

This should handle 99% of changes I still have a few open questions around what possible changes could break the cache. The cache busting here still might be a little over eager - non "problem causing" schema changes will still cause a cache bust. 

## TODO:

- [x] Closes #740 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
The net result of this PR means that you will still get logged out anytime the schema changes, correct?

Man...it'd be great if we could just detect breaking changes in the schema and only purge on a breaking change.

Also: thought on this fragmentTypes.json and such that came up while reviewing this -- what happens if we publish a new version of the app with different fragmentTypes.json from the server, but the user hasn't updated their client app yet? Would we run into crashes?

conrad-vanl:
Re: only purging on breaking changes

Wonder if this CLI tool that we're using on startup could use one of the [schema validator](https://github.com/creditkarma/graphql-validator#validate-the-query) scripts out there and instead of storing a schema `hash`, we just store like an integer that gets incremented every time we detect a new breaking schema change.

vinnyjth:
> Also: thought on this fragmentTypes.json and such that came up while reviewing this -- what happens if we publish a new version of the app with different fragmentTypes.json from the server, but the user hasn't updated their client app yet? Would we run into crashes?

I think we are okay here, because the `fragmentTypes` file is to help resolve interfaces we reference by name, and we won't reference new `interfaces` without updating the app :) 

>The net result of this PR means that you will still get logged out anytime the schema changes, correct?

>Man...it'd be great if we could just detect breaking changes in the schema and only purge on a breaking change.

Totally agree with this. I wish I understood better _what_ would be a breaking schema change. My other thought was to manage this manually and increment a variable in `.env.production` or something, but we would need to come up with some process to make sure that gets incremented when it should.

vinnyjth:
>Wonder if this CLI tool that we're using on startup could use one of the schema validator scripts out there and instead of storing a schema hash, we just store like an integer that gets incremented every time we detect a new breaking schema change.

@conrad-vanl I'm halfway with yah. I think one thing that we need to keep in mind is that, 
A. We should never publish backwards incompatible "breaking changes" 
B. I can't identify at this time what type of change, given A is true, should break the schema. 

The CLI tool above would help us if we knew we were going to publish breaking schema changes (like old schema queries !== new schema queries) but given the potential changes are more subtle in this case, I'm worried a tool like the one above would never fail.

vinnyjth:
How do you feel about, as a step forward to merge this PR, changing the strategy to a number we manually increment?

conrad-vanl:
If we change it to a number we manually increment, how will we know when it needs incremented?

vinnyjth:
> If we change it to a number we manually increment, how will we know when it needs incremented?

Hopefully we would discover via testing the PR that we made a change that required a cache flush. We could protect from our own ignorance by having a `ErrorBoundary` that would flush the cache when an error occurred. Crossroads does this

conrad-vanl:
Would it be easy to add a breaking change PR check to our CI pipeline?

It's not just our own ignorance we have to worry about, but our partners and their app teams. I'm happy to push forward with this PR, but I can anticipate that we'd need some other process for handling or catching breaking changes.

vinnyjth:
> Would it be easy to add a breaking change PR check to our CI pipeline?

YES! That's a great idea. 
https://github.com/ApollosProject/apollos-prototype/issues/744

vinnyjth:
@conrad-vanl Switched to static value + env variable that can be optionally configured.

PR #757, opened 2019-05-31
Exports ContentItemInterfaceFields and document usage

### LINKED ISSUE #755: Export the ContentItem interface fields.
### User Story

As a creator of a new ContentType I should be able to import the base `interface` fields from `server-core`.

### Description
Doing this will make it easier to create new implementations of `ContentItem` and also make sure that clients that upgrade to new versions of the schema with a new implementation get the new fields. 

### Acceptance Criteria

* [ ] Export the base fields used in the `ContentItem` `interface`
* [ ] Also use that new variable to define all of our existing ContentItems. 

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

vinnyjth:

As a developer, it's obnoxious to need to include the ContentItem interface fields every time you extend the interface. If we export those fields as a variable, we can give a mechanism by which developers can include all the fields on the interface without needing to copy/paste them. 

This will also make upgrading to new versions of the `ContentItem` interface easier. 

`MediaContentItem` and `SeriesContentItem` support Scripture, but don't actually fetch scripture. I fixed that in this PR. 

### How do I test this PR?

1. Download this PR.
2. Run this PR
 - Your server will crash if the new schema is invalid. 

n/a

## TODO:

- [x] Closes #755
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-data-connector-rock/README.md
+++ b/packages/apollos-data-connector-rock/README.md
@@ -209,6 +209,20 @@ Adding a new ContentItemType is a common operation, and we have some configurati
 
 If determining the `ContentItem` type is more complex than something like` ContentChannelTypeId`, you'll need to override the `__resolveType` function on the `ContentItem` resolver.
 
+To include all the existing fields on the `ContentItem` interface, you can use the following variable. This method is preferred to copy and pasting them.
+
+```
+import { ContentItemInterfaceFields } from '@apollosproject/data-schema';
+
+const newItemSchema = gql`
+type MyNewItem implements ContentItem {
+${ContentItemInterfaceFields}
+
+  someNewField: String
+}
+`
+```
+
 ### F.A.Q
 
 #### Q: How do I figure out what methods are on X data source?
--- a/packages/apollos-data-connector-rock/src/content-items/resolver.js
+++ b/packages/apollos-data-connector-rock/src/content-items/resolver.js
@@ -87,9 +87,23 @@ const resolver = {
   },
   ContentSeriesContentItem: {
     ...defaultContentItemResolvers,
+    scriptures: ({ attributeValues }, args, { dataSources }) => {
+      const reference = get(attributeValues, 'scriptures.value');
+      if (reference && reference != null) {
+        return dataSources.Scripture.getScriptures(reference);
+      }
+      return null;
+    },
   },
   MediaContentItem: {
     ...defaultContentItemResolvers,
+    scriptures: ({ attributeValues }, args, { dataSources }) => {
+      const reference = get(attributeValues, 'scriptures.value');
+      if (reference && reference != null) {
+        return dataSources.Scripture.getScriptures(reference);
+      }
+      return null;
+    },
   },
   ContentItem: {
     ...defaultContentItemResolvers,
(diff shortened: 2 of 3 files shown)


calebpanza:
@conrad-vanl that makes a lot sense to split it up per that modularization branch. At the end of the day, the current problem that is being faced is that the `ContentItem` interface is quite large and when developing on top of the interface, it's very easy to miss attributes in the schema. If the modularization would help solve that issue, my vote would be that we follow through with that sooner rather than later in order to help set up the dev environment for the churches.

vinnyjth:
>How would you feel if I worked on a less but better version of the schema modularization PR? Basically just splitting our current ContentItem interface up into a few smaller interfaces:

I would much prefer this approach, but I don't think it solves @calebpanza's issue. He will still have to list the fields from all the different extensions he implements. 

@calebpanza I think the best answer we can give you in the short term is "learn to read the error messages :P" GraphQL will give you pretty good error messages about what fields are missing, which should help when copy and pasting new fields to your implementation. I like the idea of including fields from a variable, but @conrad-vanl Is totally right, it's a pretty unique and unorthodox way to approach building a schema.

calebpanza:
@vinnyjth the problem isn't knowing what the error is. It's 10000% a convenience problem. For example, on one of our projects, we've got 2 or 3 data types that extend the `ContentItem`. When an update is pushed to Apollos Core, our resolvers & data sources will be all good to go when we update our version number cause they import and inherit from Core, but we will have to go into the Core Project to manually find and copy the new schema into each of our other schemas. For convenience sake, it would be nice if Core schemas could be imported since the Resolvers and DataSources already do. 

So I've read through [this](https://graphql.org/learn/schema/#interfaces) documentation a few times, and I notice that their example keeps the schemas down to a _very_ small number of properties. @conrad would it be "dangerous" or super far outside of Graph's conventions to import schemas so long as those schemas were very, very small in nature?

conrad-vanl:
This is a great conversation.

@vinnyjth would there be a way to do this with a schema directive? I'm not sure I like that _much_ better, but at the least, it's a little more extensible as we could make a directive that would work anywhere on the graph, instead of having to piece-meal strings and export them manually.

vinnyjth:
> would there be a way to do this with a schema directive? I'm not sure I like that much better, but at the least, it's a little more extensible as we could make a directive that would work anywhere on the graph, instead of having to piece-meal strings and export them manually.

I remember looking at this a while back, and I think what I determined was that Schema Directives can change how a schema functions, but not the shape of a schema :/

conrad-vanl:
I just can't help but feel that exporting these strings like this is a code smell. Again, I'm not opposed to go ahead and doing it, especially since I can't well articulate what's so bad about it, other then it goes against GQL convention.

vinnyjth:
> I just can't help but feel that exporting these strings like this is a code smell.

I agree with this. 

@calebpanza I'm going to put this on hold for the time being. I have another idea that I will explore with you to make this easier soon :) This approach is a little too heavy handed and far from GQL convention from my taste.

calebpanza:
@vinnyjth @conrad-vanl makes a lot of sense! And perhaps at the end of the day, maintaining smaller updates to specific fragments will prove to be not as daunting and still live within the proper boundaries of GQL implementation. I like that approach and would love to work out whatever solution you guys are thinking of if you're up for it!

Issue #744, opened 2019-05-24
Test for Breaking Schema Changes in Travis/Test flow


vinnyjth:
### User Story

As a developer, my local schema should be run against the production schema to determine if there are any breaking changes. 
 
### Description
After our 1.0 release, we will not tolerate breaking changes to our core schema. We might allow them in the future against any "2.0" work, but we should have tools to detect breaking changes no matter what.

**A breaking change is defined as a change that would cause errors to occur on old versions of the app pointed at current versions of the API** This is normally changes in the name of fields, changes in input types, and deletions of fields/queries. 

### Acceptance Criteria

* [ ] Developers should be able to run the tool locally to see if there are any breaking changes.
* [ ] Travis.ci should report any breaking changes before merging a PR 

### Other Resources

* Idea for implementing https://www.apollographql.com/docs/platform/schema-validation#service-check-on-ci
* Another idea https://github.com/entria/graphql-find-breaking-changes-cli

vinnyjth:
@redreceipt This might be a fun one for ya!

redreceipt:
Love this. Also, I was reading about the `eslint-plugin-graphql` plugin that lints your client queries against a `.graphql` schema file. Could be really awesome too!

vinnyjth:
Oh that would be pretty sweet! Feel free to add that as a part of this PR.

redreceipt:
*TL;DR* regarding the two implementations above:

1. **Apollo Engine Validation**. Requires trace reporting (query analytics) which is just dropping an Engine API key in our server config. Would register schema with CI check and compare against the last 100 (or whatever) *actual* query calls in the wild. Comments in your PR with a link to a thorough diff report:

### Apollo Service Check

🔄 Validated your local schema against schema tag `staging` on service `engine`.
🔢 Compared **18 schema changes** against **100 operations** seen over the **last 24 hours**.
❌ Found **7 breaking changes** that would affect **3 operations** across **2 clients**

🔗 [View your service check details](https://engine.apollographql.com/service/engine/checks?...).

+ Community supported
+ Thorough with resolution details
+ Determines the real effect on current users
- Potentially wouldn't catch *every* breaking change
- Requires Apollo Engine registration

2. **`graphql-findbreakingchanges`**. Simple package dependency that compares old schema to new schema. For example, useful for comparing any changes to a 1.0 release. Will just add `graphql-findbreakingchanges oldSchema.graphql newSchema.graphql` as a line in Travis.

[...]

PR #759, opened 2019-06-01
Adds GraphQL Schema Diff and Linting

redreceipt:
## DESCRIPTION

Screen Shot 2019-05-31 at 10 21 02 PM

Screen Shot 2019-05-31 at 10 50 07 PM

This PR does two things:
1. Gives us the ability to examine when the schema is updated and test for breaking changes via `yarn` and Travis
2. Enables client-side query lint checking against the schema.

Added a couple scripts and config files. Also adds a new stage in Travis. Added four new packages:
- `graphql-schema-linter`
- `get-graphql-schema`
- `@entria/graphql-findbreakingchanges`
- `eslint-plugin-graphql`

The first three could all be replaced with Prisma's `graphql-cli` except you invoke it with `graphql` and we already have a package called `graphql` so it throws an error 😞 

### How do I test this PR?

1. Make a breaking change to the schema (ex. change a field name)
2. `yarn schema-diff`
3. Write some bad code to a query in the app and make sure `eslint` yells at you.

None.

### Possible Next Steps

1. **Fix schema formatting**. I added a `yarn schema-lint` command as well, very insightful but right now there are 328 errors, mostly just missing descriptions but possibly a few good ones as well. Maybe down the road, we can fix these and include the command in `yarn lint` for CI.
2. **Enable tracing**. This PR makes it impossible to commit breaking changes so we'll have to mark fields deprecated to slowly encourage client changes. When we do get ready to remove fields, we'll need a way to measure impact. This can be done by enabling tracing and acting on the recorded response data. 

## TODO:

- [x] Closes [#744]
- [x] Set a relevant reviewer
- [ ] fix current lint errors
- [x] update `schema.graphql` dynamically

## REVIEW:

**Manual QA**


redreceipt:
@vinnyjth only problem I'm not quite sure what to do about is how and when to update the local schema file. It has to be while `localhost:4000` is running. I could use the `get-introspection-data` script but I think the functionality should be contained to the `apollos-church-api` package. It should be able to update its own file somehow since the `yarn schema-diff` script is going to need it to function properly.

redreceipt:
also I think I totally just solved those weird analytics errors we've been seeing with this linter haha

Screen Shot 2019-05-31 at 11 25 17 PM

vinnyjth:
> also I think I totally just solved those weird analytics errors we've been seeing with this linter haha
> 
> Screen Shot 2019-05-31 at 11 25 17 PM

I wish :/ Those are client queries and so the linter isn't able to find them inside the `.graphql` file we generate. We will need to eslint disable those lines.

redreceipt:
Is a "client query" just a query that pulls from a different GraphQL server? If that's the case, I think we can use multiple schemas so we could lint any query we have.

vinnyjth:
> Is a "client query" just a query that pulls from a different GraphQL server? If that's the case, I think we can use multiple schemas so we could lint any query we have.

A client query is a query that resolves on the device itself, it doesn't hit any server, but is resolved from the on device cache. There isn't really a schema for the client side query, but I'm curious if you can find a way to lint from it.

vinnyjth:
@redreceipt I can review this as soon as you are ready.

redreceipt:
Yea, almost. I need a solution for the self updating `schema.graphql` file. Also would love your thoughts on the possible next steps I've outlined above.

vinnyjth:
@redreceipt I think both of your next steps sound good, but are next steps for future issues. We can create PRs after we merge this.

redreceipt:
🔥 🔥 🔥 love that! I'm gonna make a slight mod so it will automatically happen every `yarn start` using a `--start-server` param to the script that we'll use in Travis but won't have to use locally.

.travis.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         - travis_wait yarn codecov
     - stage: Schema Diff
       script:
-        - cd packages/apollos-chuch-api
+        - cd packages/apollos-church-api
```
redreceipt: 😂

PR #786, opened 2019-06-13
Don't login a user if their phone number has multiple users

### LINKED ISSUE #746: How to handle duplicate phone numbers
### Details

I'm logging this as a bug although it might be considered an enhancement. Essentially NewSpring brought to our attention an issue with duplicate phone numbers. Example:

Frank's son is in Rock, but is associated with Frank's phone number. This is common for churches when you check a kid into childcare, for example.
When Frank signed into the app using sms login, it logged him in as his son.

HMW correct?

One solution might be that whenever there's more than one account associated with a phone number, that we create a new account.

vinnyjth:

New SMS Auth flow: 

1. On login, if multiple people have the same phone number, check to see if any of those people already have a User Account that is the phone number.
2. If Yes, that is the person.
3. If no, create a new person record with the User Account that is the phone number(it would be a duplicate that would have to be merged at some point)

Opted to avoid trying to guess which phone number is the correct phone number.

### How do I test this PR?

- [x] 1. Spin up this PR (make sure you yarn)
- [x] 2. Make sure you have multiple users with your phone number in Rock, login via SMS
  - You should login as a new user
- [x] 3. Logout of that account, and then login via SMS again 
  - You should login as that user. 
- [x] 4. Using the Rock admin interface, give yourself a new phone number (borrow a friend who isn't in Rock?)  Login via SMS
  - You should login as yourself

API Tests covering all new functionality.

## TODO:

- [x] Closes #746 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-church-api/local.graphql
+++ b/packages/apollos-church-api/local.graphql
@@ -85,7 +85,7 @@ scalar Color
 
 type ContentChannel implements Node {
   id: ID!
-  name: String!
+  name: String
   description: String
   childContentChannels: [ContentChannel]
   childContentItemsConnection(first: Int, after: String): ContentItemsConnection
@@ -257,8 +257,8 @@ type PaginationInfo {
 
 type Pass implements Node {
   id: ID!
-  type: PassType!
-  description: String!
+  type: PassType
+  description: String
   logo: ImageMediaSource
   thumbnail: ImageMediaSource
   barcode: ImageMediaSource
@@ -291,7 +291,7 @@ enum PassType {
 
 type Person implements Node {
   id: ID!
-  firstName: String!
+  firstName: String
   lastName: String!
   nickName: String
   email: String
--- a/packages/apollos-data-connector-passes/package.json
+++ b/packages/apollos-data-connector-passes/package.json
@@ -30,7 +30,7 @@
     "apollo-server": "^2.5.0",
     "datauri": "^1.1.0",
     "lodash": "^4.17.11",
-    "passkit-generator": "^1.5.9",
+    "passkit-generator": "1.5.9",
     "qrcode": "^1.3.3"
   },
   "peerDependencies": {
--- a/packages/apollos-data-connector-rock/src/auth-sms/__tests__/__snapshots__/index.js.snap
+++ b/packages/apollos-data-connector-rock/src/auth-sms/__tests__/__snapshots__/index.js.snap
@@ -49,12 +49,12 @@ Array [
 exports[`AuthSms schema logs a new user in via their phone number and pin: try and find phone number / find user login 1`] = `
 Array [
   Array [
-    "/PhoneNumbers?%24filter=Number%20eq%20%275133061126%27&%24top=1",
+    "/UserLogins?loadAttributes=expanded&%24filter=UserName%20eq%20%275133061126%27&%24top=1",
     Object {},
     Object {},
   ],
   Array [
-    "/UserLogins?%24filter=UserName%20eq%20%275133061126%27&%24top=1",
+    "/PhoneNumbers?loadAttributes=expanded&%24filter=Number%20eq%20%275133061126%27",
     Object {},
     Object {},
   ],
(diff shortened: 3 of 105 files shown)


vinnyjth:
| Logging in as a user with a shared number **first time**                                                                    | Logging in as a user with a shared number **second time**                                                                   |   |   |   |
|-----------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------|---|---|---|
|  |  |   |   |   |

vinnyjth:
The flow (if this PR is merged) looks like this

1. On login, if the user has more than one account using his phone number, create a new account with a new phone number. The number has a custom attribute called "ApollosGenerated" 
2. On any subsequent login, if the user has multiple numbers, we check to see if one of those numbers has the ApollosGenerated attribute and picks that account. 
3. If the user has a single phone number, we use that account, and don't create a new account. 

@FrankGrand How does this sound? It's not super sophisticated, but it's safe.

FrankGrand:
@vinnyjth It may just be a vocab thing, but I would love to clarify a couple things:

Each Person in Rock can have multiple User Accounts.  For example, here is the current account that is in Rock for 'my phone number' (it's actually attached to my son's profile)
image

The username is just the phone number.  Two people in Rock cannot have the same username.  So my initial thought was that the flow would be something like this:

1. On login, if multiple people have the same phone number, check to see if any of those people already have a User Account that is the phone number.

2. If Yes, that is the person.

3. If no, create a new person record with the User Account that is the phone number(it would be a duplicate that would have to be merged at some point)

If there is only one person with that phone number, then that is person and we'd just create the User Account and be good to go.

This sounds close to what you mentioned, I don't see a need for an 'attribute' on a phone number though.  We may be saying the same thing, I just wanted to make sure our terminology lined up.

vinnyjth:
Hey @FrankGrand, that's genius and a I love it. It will be a more significant refactor to achieve, but I think we can make it happen. 

We already set the `UserName` of the `UserAccount` to be the phone number, so it's a matter of keying off the `PersonId` of that user account to determine the person, rather than using their PhoneNumber's `PersonId`

conrad-vanl:
@vinnyjth thoughts on having someone on the NewSpring side review this? @redreceipt or @IsaacHardy perhaps? I'm happy to review this, but not sure if I'm equipped to make sure that this handles the logic issues they've been hitting.

vinnyjth:
@conrad-vanl I'm totally fine with that. @redreceipt want to give this a shot?

redreceipt:
Just so I understand the business logic, explain the part in bold from step 1 above: 

> On login, if multiple people have the same phone number, check to see if any of those people already **have a User Account that is the phone number**.

vinnyjth:
>  have a User Account that is the phone number.

In this case, what we mean is that the Username of that UserLogin is literally the phone number. Username: 5133061126, Password: (this cool hash thing)

redreceipt:
Gotcha. Also, I noticed in my poking around, we are not adding a phone number to the account when they log in with a phone number. Shouldn't we be?

redreceipt:
Ok I see how we're doing it now but I'm not sure it's working right...

Screen Shot 2019-06-25 at 3 22 27 PM

redreceipt:
confirmed it's not working right. made a new person.

Screen Shot 2019-06-25 at 3 25 28 PM

Screen Shot 2019-06-25 at 3 25 36 PM

redreceipt:
...which will screw up the person merge

Screen Shot 2019-06-25 at 3 38 34 PM

redreceipt:
The fix may be as simple as adding more data to the POST call or may be as extensive as using different REST endpoints to add phone numbers properly.

redreceipt:
Frank is asking what the preferred method for adding phone numbers is, I think that should be required for this PR since a merge won't work properly until it's done.

vinnyjth:
Michael, look at the top right of the first screenshot you posted. We are adding a phone number, but net setting it as any of the default phone numbers for the Rock person. I can look into doing that.

PR #787, opened 2019-06-13
Remove required flag from certain schema attributes 

vinnyjth:
## DESCRIPTION

I have found that having attributes marked as required can be brittle, for example, parts of our app now throw errors if you don't put in a first name on onboarding. Technically a first name is not great, but GraphQL shouldn't be throwing errors just trying to query for it.

Removes the required marker from `attributes` on types in the schema that have the marker. 

Schema no longer reflects "base case scenario" for data cleanliness as closely. 

### How do I test this PR?

1. Boot the app, it shouldn't fail.
2. Create a new user and opt to avoid setting a first name. The app will no longer throw an error.

n/a

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-church-api/local.graphql
+++ b/packages/apollos-church-api/local.graphql
@@ -85,7 +85,7 @@ scalar Color
 
 type ContentChannel implements Node {
   id: ID!
-  name: String!
+  name: String
   description: String
   childContentChannels: [ContentChannel]
   childContentItemsConnection(first: Int, after: String): ContentItemsConnection
@@ -257,8 +257,8 @@ type PaginationInfo {
 
 type Pass implements Node {
   id: ID!
-  type: PassType!
-  description: String!
+  type: PassType
+  description: String
   logo: ImageMediaSource
   thumbnail: ImageMediaSource
   barcode: ImageMediaSource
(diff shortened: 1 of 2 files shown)


redreceipt:
`apollos-church-api/local.graphql` has been updated, I'll go ahead and push the changes, just FYI

redreceipt:
Not sure if this is the intended behavior or not, the Next button doesn't show up so you have to swipe. If it is, then it works like a charm :)

Kapture 2019-06-23 at 14 14 43

vinnyjth:
Thanks for pointing that out @redreceipt. On the UI side, that field is still required. However, marking it as required in the schema means that GQL won't even try to resolve queries where that field is `null`, which is bad news for us.

redreceipt:
So maybe a separate issue but should we disable the swipe until validation issues are resolved?

vinnyjth:
That's a good question for @burkeshartsis

vinnyjth:
> Would this be considered a breaking schema change? 

It is according to the schema tool, but because clients don't care about if a field is required or not when querying (they don't specify the `!` it's not. The only situation I've found so far when you can't remove a required field, is when using input types. `input { foo: String! }` -> `input { foo: String }` is breaking, because you normally specify inputs like so `query($foo: String!) { thing(foo: $foo) } `

PR #812, opened 2019-06-21
Fixes rock-apollo-datasource null bug

vinnyjth:

"Rock" was returning `{}` for values that were actually `null`. Turns out we were treating `null` as an object to be traversed, causing this issue. 

n/a

### How do I test this PR?

Check the automated tests. Should be no changes in how the API serves data. Endpoints that were doing manual checking of null values won't have their implementation change. 

Wrote a test to cover the `normalizer` in the Rock Apollo Datasource.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-data-connector-rock/src/people/resolver.js
+++ b/packages/apollos-data-connector-rock/src/people/resolver.js
@@ -1,11 +1,6 @@
 import { createGlobalId } from '@apollosproject/server-core';
 import { enforceCurrentUser } from '../utils';
 
-// Rock returns `{}` instead of `null` for null values.
-// This function eliminates the annoyance of checking those values.
-export const ifNotObject = (field) =>
-  typeof field === 'object' ? null : field;
-
 export default {
   Mutation: {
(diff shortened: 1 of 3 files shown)


redreceipt:
You're sure there will no breaking changes in core? We have a couple of things that will break on our end, easy fixes, but just want to make sure we're covering all our bases. Here's an example of something that will break:

```js
  sortPrayers = (prayers) =>
    prayers.sort((a, b) => {
      // if prayerCount is 0, Rock returns an empty object 🙄
      if (typeof b.prayerCount === 'object' || a.prayerCount > b.prayerCount)
        return 1;
      if (a.prayerCount === b.prayerCount)
        return moment(a.createdDateTime) > moment(b.createdDateTime) ? 1 : -1;
      return -1;
    });
```

vinnyjth:
Previously thos fields would return an empty object, now they will return null. Both have a typeof “object” so by the grace of God and the insanity of JS we should be okay

Issue #809, opened 2019-06-19
Disable Notifications


redreceipt:
### User Story

As a user, I want to be able to disable notifications in the app so the API is aware they are disabled.

### Description

Currently, there's no way to disable notifications without manually doing it in the OS settings but then the app and the API have no idea. We can fix this with an in-app switch.

### Why

I may want to turn off notifications and the API needs to know that information.

### Acceptance Criteria

* [ ] It should have a switch somewhere in the settings that disables notifications.
* [ ] (optional) do the same for location services

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

redreceipt:
@vinnyjth care if I take a swing at this and just put a simple toggle in the settings menu for now?

vinnyjth:
@redreceipt Totally! Go for it.

PR #821, opened 2019-06-25
Adds Notifications Toggle

redreceipt:
## DESCRIPTION

Screen Shot 2019-06-25 at 11 24 45 AM

This will give us the ability to toggle push notifications in app

## TODO:

- [ ] Closes [#809 ]
- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
@vinnyjth how do I use the `ui-notifications` queries? None of these are showing up in the schema
Screen Shot 2019-06-25 at 11 44 11 AM

vinnyjth:
@redreceipt Those are all client side. You'll see the resolvers for them inside the `store` for that module.

redreceipt:
gotcha, what exactly do the two functions here do? Do I need to figure out how to disable notifications? Can't tell if your functions already do that with Onesignal.

Screen Shot 2019-06-25 at 12 15 42 PM

vinnyjth:
@redreceipt We don't currently have a way to enable or disable push notifications. You'll need the update the API here https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-data-connector-onesignal/src/data-source.js#L19 to correctly set if a user has enabled push notifications. I don't have an answer for you as to "how" but it should probably be a setting that you set in OneSignal or Rock.

redreceipt:
Ok so here's the game plan:
1. Add `notificationsEnabled` attribute in Rock (include in plugin)
2. Set to `true` in onboarding and the new settings toggle
3. Add `disableNotifications` to our API

Bonus Rock work:
- Have communication service check that attribute first (include in plugin)
- Have communication UI check first (include in plugin)

Thoughts??? @richarddubay missing anything?

redreceipt:
Currently waiting on an answer from Spark about adding an `AllowPushNotifications` attribute to core Rock.

richarddubay:
@redreceipt I think this sounds good. Right now all the transport does (pretty much) is send off the devices to OneSignal and it either sends a notification or not depending on if the device exists there. So the transport will just have to check for the attribute and remove anyone from the list of people it's sending to before it sends it off to OneSignal for processing. 

I like the idea of adding a core attribute so that the communication block can be updated as well. Great stuff.

redreceipt:
@vinnyjth so for now until Rock decides what to do, we could just add an extended attribute called `ApollosNotificationsAllowed`, what do you think?

redreceipt:
@frankgrand is speaking with Spark about getting a `NotificationsAllowed` top level data field on a person similar to SMS so I say we wait and see what they say before adding a custom attribute.

redreceipt:
@vinnyjth can we add attributes with the plugin?

Issue #808, opened 2019-06-19
Android Location Finder Dots Should Match iOS


redreceipt:
### User Story

As a user looking for my campus, I want to see pretty round dots.

### Description

The location dots on iOS are pretty:

<img width="447" alt="Screen Shot 2019-06-19 at 2 50 37 PM" src="https://user-images.githubusercontent.com/2659478/59792195-9e049080-92a1-11e9-8a06-a27ccad8cd46.png">

The Android dots are ugly:

<img width="429" alt="Screen Shot 2019-06-19 at 2 47 34 PM" src="https://user-images.githubusercontent.com/2659478/59792055-5b42b880-92a1-11e9-8ab8-2e25ade9b3b3.png">

They should match.

### Why

Better experience.

### Acceptance Criteria

* [ ] It should match the iOS location dots

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

PR #822, opened 2019-06-25
Redesigns Android Location Markers

redreceipt:
## DESCRIPTION

Screen Shot 2019-06-26 at 12 52 09 PM

Makes the Android location finder pretty like iOS.

The `Locations` code definitely needs to be cleaned up but I restrained myself and only refactored what was necessary for this fix. I took out all code pertaining to the style animation because it wasn't working anyway and I'm not sure what the design requirements are. It doesn't look bad the way it is.

### How do I test this PR?

Use the location finder on iOS and Android.

## TODO:

- [x] Closes [#808]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
Found the culprit, it doesn't seem that you can use absolute positions with Android custom markers

https://github.com/ApollosProject/apollos-prototype/blob/67a6e6a3733d3eb4b415ba86b870260f94073e85/packages/apolloschurchapp/src/user-settings/Locations/Marker.js#L30

vinnyjth:
@redreceipt Code of this PR looks fantastic. You've done a ton of great cleanup work here. I'll test it when I get a chance and hit the approve button, assuming the functionality is the same + the enhanced android markers.

packages/apolloschurchapp/src/user-settings/Locations/index.js
```diff
@@ -46,7 +40,9 @@ class Location extends PureComponent {
       latitude: 39.809734,
       longitude: -98.555618,
       latitudeDelta: 100,
-      longitudeDelta: 100 * ASPECT_RATIO,
+      longitudeDelta:
```
vinnyjth: @redreceipt Can you explain this new math to me?

redreceipt: I don't what `longitudeDelta` does exactly, I simply just removed some unnecessary global variables. `ASPECT_RATIO` was being defined above but I took that out.

redreceipt:
very sneaky...I see those commits

packages/apolloschurchapp/src/user-settings/Locations/index.js
vinnyjth: Gotcha. You actually are doing slightly different math here (look at the parens). I might default this to 100 and move on :)

redreceipt: Oh. Prettier did that. 100*(a/b) is the same as (100*a)/b

Issue #811, opened 2019-06-20
Check-In Image Not Loading


redreceipt:
### Details

The image from check-in pass never loads on Android, a little context, I had a duplicate account on Rock that was recently merged.

### Expected Behavior

Image should load

### Actual Behavior

Image never loads so I can't exit out because I can't see the modal header.

![Kapture 2019-06-20 at 10 57 42](https://user-images.githubusercontent.com/2659478/59859619-0ce9f400-934b-11e9-9a62-a83aa0dc96b5.gif)

### Possible Fix

<details><summary>Additional Info</summary>

### Your Environment

- Environment name and version (e.g. Chrome 39, node.js 5.4):
- Operating System and version (desktop or mobile):
- Link to your project:

### Steps to Reproduce

1.  first...
2.
3.
4.

### Stack Trace

</details>

vinnyjth:
I am currently unable to reproduce this issue naturally, but I can recreate the big gray box by setting the `isLoading` on the `Thumbnail`  in the pass to true. We should fix this by making sure the the placeholder image doesn't take up 100% of the space.

redreceipt:
this is why it's breaking, the thumbnail uri is coming in like this 

<img width="653" alt="Screen Shot 2019-06-26 at 3 46 59 PM" src="https://user-images.githubusercontent.com/2659478/60209960-c48a7480-9829-11e9-9612-38fc31f6e3dc.png">

investigating now...

redreceipt:
so there's two issues here:
1. why is the uri not complete for me?
2. the gray box shouldn't be so big for the thumbnail placeholder, which is a `ui-kit/ConnectedImage`. that file is very scary...

redreceipt:
fixed the gray box thing

PR #827, opened 2019-06-26
Fixes Image Placeholders

redreceipt:
## DESCRIPTION

Screen Shot 2019-07-09 at 4 38 33 PM

This just wraps the thumbnail on the check-in pass in a view that constrains the height, even when loading.

Not sure, don't really know what the intentions were.

### How do I test this PR?

Check the UI pass on Android and iOS, look for the loading state, hardcode it if you have to.

None.

## TODO:

- [x] Closes [#811]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
@burkeshartsis would love your thoughts on this one, `ui-kit/ConnectedImage` is a scary place so I'm sure this fix will have implications that I'm not aware of.

packages/apollos-ui-kit/src/ConnectedImage/SkeletonImage.js
```diff
@@ -4,7 +4,6 @@ import Placeholder from 'rn-placeholder';
 import styled from '../styled';
 
 const SkeletonImage = styled(({ theme, forceRatio }) => ({
-  width: '100%',
```
conrad-vanl: Does this have visible side affects in the other views that use this? There's a reason this width 100% existed originally.

redreceipt: The better question is what's the intended behavior? Seems like it should somehow match the image it's trying to fake instead of filling 100% of its container, which is causing the issue on the check-in screen? I could try to figure out how to make that happen instead if you deem it a better fix?

conrad-vanl: So I assigned this over to @burkeshartsis after we had a discussion about it. I think generally images are supposed to fill their container, which is why this `width: 100%` exists. I believe with taking this line out, these placeholder images will never show up, because now they have no size. But obviously the styling between the placeholder image and the real image is inconsistent, thus the problem. I agree we need a better solution, but I don't think taking this width out without putting anything back is the right solution.

burkeshartsis: Happy to dig into this. This is purely a layout bug that the placeholder renders the way it does in the passes. Should be an easy fix and happy to pair with you on it. @conrad-vanl is right though, removing the `width` will break the placeholder.

redreceipt: So you're saying I should just be able to alter the pass thumbnail placeholder to be small? Or maybe wrap it in a small view?

burkeshartsis: @redreceipt correct! That is likely the correct approach.

redreceipt:
ok @burkeshartsis how about this

vinnyjth:
👀

PR #834, opened 2019-06-28
add <ActionCard /> component

### LINKED ISSUE #824: Create ContentItem ActionCard
### User Story
As an app user I want to see text cards within ContentItems so that I can like and share them.

### Description
This is a simple card component built from existing card components in ui-kit. Depending on how we fetch ContentItem data it can either render purely from props driven by ContentItem data or it can be connected to a query and render all the types of data for a given weekend experience section. See this [link for context](https://sketch.cloud/s/1eRqn/a/eJ1Ekz). That will be dependent on how @vinnyjth feels we should fetch data for these features. If there becomes a need for a connected component I expect we'll need a new `(e)4` issue to track that work.

### Examples
Screen Shot 2019-06-25 at 7 51 06 PM
Screen Shot 2019-06-25 at 7 51 13 PM

### Acceptance Criteria 
- [ ] It should have a customizable label and icon. `ChannelLabel` may be suitable here.
- [ ] It should display some text
- [ ] It should have an clickable action in the bottom right corner.
- [ ] The action should accept customize text and an icon.
- [ ] the component should gracefully render with limited props.
- [ ] There should be sufficient stories and tests to document component functionality.

### Other Resources
- [Design mockup/prototype link](https://sketch.cloud/s/1eRqn/a/eJ1Ekz)

conrad-vanl:
Screen Shot 2019-07-01 at 3 13 16 PM

## DESCRIPTION

@burkeshartsis here's a start to #824 😘 

- Adds a `<ActionCard />` component according to #824 

@michael-riddering would you be open to giving a design signoff / 👍 (or pump the breaks on this) based on the design tradeoffs i talk about below and the image above?

So far:

- The top icon/text label is using the `ChannelLabel` component since we already had this in the ui kit. It uses a different type size though then what's in the design comps. I don't think we are using `ChannelLabel` anywhere else currently, so maybe we either: make this a configurable prop on the component, or just change the type component used

- ~the spacing between ChannelLabel and the text of the card is off.~ After spending time in Sketch and Anima, I'd love to figure out a more consistent spacing solution in UI-Kit then just manually adding margin/padding around things...something that works like Anima's Stack feature would be killer. I feel like the component as is should be spaced correctly without having to add margin/padding to the style components inside of here. We can still do that to get this PR merged, but would be nice to think through. **UPDATE:** I fixed this spacing, but the comment above is still valid - i just added manual bottom padding to the `ChannelLabel` component within the card

- I switched the button text at the bottom right with a Chip. We don't currently have a component that puts text with an icon to the right, and I would've had to write a lot of boilerplate code to do so. I don't think the chip is bad. Other options are to add a prop to ChannelLabel to switch the order of text and icon or create a new component.

## TODO:

- [x] Closes #824 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
@vinnyjth I'm labeling this ready for review, although I'm still looking into why tests are having troubles failing. I can confirm that the `apollos-ui-kit` package tests succeed, which is the only place this PR adds code....

michael-riddering:
@conrad-vanl 

• `ChannelLabel` as it currently stands doesn't look bad. If we really aren't using it anywhere else though I would push for H4.

• Having a hard time telling if the UI-Kit proposed spacing adjustments are for me in Sketch or for you guys in code. Since this is a vertically-scrolling card, I am using 24px padding for the whole component (might be easier said than done on your end) and right now it looks like the padding around the chip is great, and the padding around the `ChannelLabel` is tighter than 24px. I'm not convinced I'm not missing the point of your comment though.

• My vote would be to leave the chip for now and I'll add something about text + icon button into the core proposed changes and we can all talk about it then so that we're not pressed on making unnecessary changes right off the bat.

conrad-vanl:
@michael-riddering yeah the comment about spacing was mainly referring to our code and the way we space things in code, which totally doesn't match the way we are doing it in sketch. So not directed at you at all, sorry

`ChannelLabel` - lets merge as is with the H5 or whatever it is using, then we can discuss changing it in our core conversation that you're gearing up for

:+1: on leaving chip for now.

michael-riddering:
Cool then everything looks good to me 👌

I feel like I got a seat at the lunch room cool kids table now that I’m in the Github organization 😎

vinnyjth:
@burkeshartsis are you planning on reviewing this today?

packages/apollos-ui-kit/src/Card/Actions.js
```diff
@@ -2,13 +2,13 @@ import PaddedView from '../PaddedView';
...
 const Actions = styled(
-  ({ theme }) => ({
+  {
     flexWrap: 'wrap',
     flexDirection: 'row',
     alignItems: 'center',
     justifyContent: 'space-between',
-    paddingTop: theme.sizing.baseUnit / 4,
-  }),
+    paddingTop: 0,
```
burkeshartsis: There are two clear ways to solve the problem you are addressing here and I believe this change to be the correct way. That said, it does introduce the following problem below.


I don't believe we would ever intentionally display a card in this way but we should probably elegantly handle this situation in the event that a title or description text aren't available. Doing so is completely possible, however it will likely introduce more complexity than we would like at the `Card` component level. Possible solutions include but are not limited to:
- making a title or description required.
- not displaying cards without titles or descriptions at the ConnectedComponent "level"
- not displaying cards without titles or descriptions at the feed "level"

One situation I could see above is a church using an image that has the "title" in it (à la sermon series or devotional) in which case they might intentionally leave one or both of these fields blank.

burkeshartsis: Ok, after looking at the code for [`ContentCard` here](https://github.com/ApollosProject/apollos-prototype/blob/action-card/packages/apollos-ui-kit/src/ContentCard/index.js#L74) and the code that renders the content [here](https://github.com/ApollosProject/apollos-prototype/blob/action-card/packages/apollos-ui-kit/src/ContentCard/ContentText.js#L8) it looks like we could handle the situation above by just adding more complexity to an already complex code situation.

packages/apollos-ui-kit/src/ActionCard/index.js
```diff
@@ -0,0 +1,37 @@
...
+  style: { paddingBottom: theme.sizing.baseUnit },
+}))(ChannelLabel);
+
+const ActionCard = ({ label, icon, children, action, ...otherProps }) => (
+  <Card {...otherProps}>
+    <CardContent>
+      {label && icon ? <StyledChannelLabel label={label} icon={icon} /> : null}
+      {children}
+    </CardContent>
+    {action ? <ActionCardActions>{action}</ActionCardActions> : null}
```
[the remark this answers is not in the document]
conrad-vanl: I think this is something that I'd like @vinnyjth to weigh in on from the perspective of "building for our church developers."

To me, I think render props become a no brainer because it not only provides flexibility, but it can drastically simplify our prop structure at the same time because it allows for us to use more components, with each component having very simple and clear prop APIs. Using more components that do one thing.

For example - if we didn't want to put a render prop here, we'd probably at minimum, have to: add a prop for the chip label and the icon to be used. We'd then have to decide - do we want people to be able to use different types/colors of chips? What about if they wanted the chip to be touchable? What if they wanted to add two chips? Or style it slightly differently? Should we open up the api to be able to pass in a `style` object?

I think this example is rather simplistic, but the issue can get compounded quickly in more complex instances. Places where we have to pass things like a `containerStyle` prop because we decided to wrap everything in a container element then later on run into a situation where we need to style that container separately.

packages/apollos-ui-kit/src/Card/Actions.js
conrad-vanl: yeah i'm pretty sure if a content card doesn't have a title or any content it should render the footer on top of the image instead of below it 🤔 

@michael-riddering thoughts on this? How would you like to handle the edge case of a content card having an image, but no title or text, but it still having a "footer" (in this case that little like count thing)

burkeshartsis: Ok took a look at master and visually regardless of this change it is broken so... might just end up leaving this alone all together and addressing it later 💥

packages/apollos-ui-kit/src/ActionCard/index.js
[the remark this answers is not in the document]
vinnyjth: I'm totally on board with passing components as props. I think it provides _just the right_ level of flexibility, and as long as you are making sure developers have access to all of the `default` components I think it's a fairly sane pattern.

PR #843, opened 2019-07-03
Enhances Tile Image

redreceipt:

Allows the aspect ratio of the `ui-kit/TileImage` to be customized (we use it in our prayer feature)

Don't think any. I set the default to 1.

### How do I test this PR?

Look at the new story I made.

One more snap.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-kit/src/TileImage/index.js
+++ b/packages/apollos-ui-kit/src/TileImage/index.js
@@ -10,12 +10,12 @@ import GradientOverlayImage from '../GradientOverlayImage';
 import TouchableScale from '../TouchableScale';
 
 const CardView = styled(
-  ({ theme }) => ({
+  ({ theme, aspectRatio }) => ({
     borderRadius: theme.sizing.baseUnit,
     overflow: 'hidden',
     width: '100%',
     height: '100%',
-    aspectRatio: 1,
+    aspectRatio,
   }),
   'TileImage'
 )(View);
@@ -42,9 +42,9 @@ const enhance = compose(
 );
 
 const TileImage = enhance(
-  ({ image, link, onPressItem, text, theme, isLoading }) => (
+  ({ image, link, onPressItem, text, theme, isLoading, aspectRatio }) => (
     <TouchableScale onPress={() => !isLoading && onPressItem({ ...link })}>
-      <CardView>
+      <CardView aspectRatio={aspectRatio}>
         <SquareGradientOverlayImage
           source={image}
           isLoading={isLoading}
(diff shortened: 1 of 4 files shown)


vinnyjth:
Would love @burkeshartsis to look at this and give his thoughts on this approach.

packages/apollos-ui-kit/src/TileImage/index.js
```diff
@@ -42,9 +42,9 @@ const enhance = compose(
 );
 
 const TileImage = enhance(
-  ({ image, link, onPressItem, text, theme, isLoading }) => (
+  ({ image, link, onPressItem, text, theme, isLoading, aspectRatio }) => (
     <TouchableScale onPress={() => !isLoading && onPressItem({ ...link })}>
-      <CardView>
+      <CardView aspectRatio={aspectRatio}>
```
conrad-vanl: Another option would be to allow `style` to be passed into this component.

But I'm curious as to why we would need to supply a certain aspect ratio to this, since it appears that it just fills its parent (`100%` width and height). Should you just place it inside of a view that has the aspect ratio instead?

I'm not against the change in this PR, I just want to make sure we carefully consider since it is creating new API which is technical debt. I'd like @burkeshartsis to give the sign off / direction on this.

burkeshartsis:
@redreceipt while this change is small and might get you what you need I'm not sure it would be immediately helpful to others not wanting what `TileImage` was built for or expressly what you're trying to achieve. In fact, this component is intentionally built to have a limited use case so as to keep its api signature as single purpose as possible. Couple things that stand out in this component that speak to this,
- name, `TileImage` or `Tile` in general is a naming convention we use (rightly or wrongly) to represent square elements/components. 
- one of this components internal components is named `SquareGradientOverlayImage`
- the `aspectRatio` and the color of the `SquareGradientOverlayImage` are static
- it doesn't accept `children` nor does it allow you to define how they are positioned

All of this suggests to me that what you're trying to do is outside the scope of this component. What I might suggest is simply using a Theme override for the `TileImage` component which would give you access to `AspectRatio`. If you must build your own `CategoryCard/ChickletCard` I would sugges resisting the urge to expose a lot of component api to make this component do more than what you need it to or to be flexible for others currently unknown use cases.

With that said, I'm going to close this PR. You're welcome to open it again if you want or if you update it to with a new component.

redreceipt:
makes sense. our menu has rectangular tile images so currently, we are overriding the theme. just trying to simplify the code a little.

Issue #839, opened 2019-07-02
Add a livestream field to the WeekendContentItem


vinnyjth:
- [ ] Pressing the play button should open the livestream, IF the livestream is live.
- [ ] Otherwise should open the sermon video, if available. 
- [ ] Older sermon items should never open the livestream
- [ ] Might need some way to tie a church online livestream event to a specific content item. @frank Might have ideas?

PR #848, opened 2019-07-08
Adds API support for content items to have livestreams

vinnyjth:

This PR adds some API functionality so that LiveStreams can be associated with ContentItems. Per discussion with @FrankGrand, the following logic is used to determine if a content item has a "live" livestream. 

- If the stream isn't currently live, no content items are live.
- If the stream is live
  - The content item is live if the content item is the most recent sermon content item
  - The content item is **not** live if there are more recent sermon content items than the current content item.

I also attach the `webViewUrl` and a `VideoMedia` to the livestream type.

- I chose to enhance the API design of the ChurchOnline resolver, by moving most functionality to the dataSource. 

- I added a new attribute to the config.yml for the content channel that holds your sermons.

- I added the capability to define ContentItem TypeNames via `ContentChannelIds` as well as `ContentChannelTypeIds`

### How do I test this PR?

Run the following query: 

```
query userFeed {
  userFeed {
    edges {
      node {
        id
        title
        __typename
        ... on WeekendContentItem {
          liveStream {
            isLive
            media {
              sources {uri}
            }
                          webViewUrl
          }
        }
      }
    }
  }
}
```

You should see that one of the ContentItems (the most recent one) should have `isLive: true` IF the livestream is currently live. You can hardcode the livestream to be live by hardcoding `isLive` here:  https://github.com/ApollosProject/apollos-prototype/blob/839-livestream-attribute-on-content-item/packages/apollos-data-connector-church-online/src/data-source.js#L24

Tests over new dataSource methods.

## TODO:

- [x] References #839
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

image

### DIFF
--- a/packages/apollos-church-api/local.graphql
+++ b/packages/apollos-church-api/local.graphql
@@ -219,6 +219,8 @@ input LikeEntityInput {
 type LiveStream {
   isLive: Boolean
   eventStartTime: String
+  media: VideoMedia
+  webViewUrl: String
 }
 
 interface Media {
@@ -458,6 +460,7 @@ type WeekendContentItem implements ContentItem & Node {
   theme: Theme
   isLiked: Boolean
   likedCount: Int
+  liveStream: LiveStream
   sharing: SharableContentItem
   features: [Feature]
 }
--- a/packages/apollos-church-api/config.yml
+++ b/packages/apollos-church-api/config.yml
@@ -23,6 +23,8 @@ ONE_SIGNAL:
   REST_KEY: ${ONE_SIGNAL_REST_KEY}
 CHURCH_ONLINE:
   URL: https://apollos.churchonline.org/api/v1/
+  MEDIA_URLS: []
+  WEB_VIEW_URL: https://apollos.churchonline.org/
 TWILIO:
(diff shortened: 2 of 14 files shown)


redreceipt:
what's the `media` field for?

vinnyjth:
> what's the media field for? 

Some churches will have media URLs for their livestreams, instead of webview urls. That way we can open the native player and play a live video instead of a webview.

vinnyjth:
Waiting on @conrad-vanl to give this a thumbs up.

packages/apollos-church-api/local.graphql
```diff
@@ -219,6 +219,8 @@ input LikeEntityInput {
 type LiveStream {
   isLive: Boolean
   eventStartTime: String
+  media: VideoMedia
+  webViewUrl: String
```
conrad-vanl: Only question I have is wether this `webViewUrl` belongs inside of `VideoMedia` or `LiveStream`. I think here is the right spot.

vinnyjth: I think `LiveStream` is the right place for this, I think it's rare that a Video will have a webView associated with it (unless it's YT....)

PR #857, opened 2019-07-10
Adds "Open Livestream" Capability to the MediaControls

### LINKED ISSUE #839: Add a livestream field to the WeekendContentItem

vinnyjth:

Adds the ability for the MediaControls on the `ContentItem` to
- Open a media stream, if a content item is live and it has a media stream.
- Open a webview, if a content item is live and there is a web view url. 
- Opens a content's media, if there is media attached. 

n/a 

### How do I test this PR?
1. Open the most recent sermon item "Spur One Another" 
2. If you have a livestream running, you should see a "Play" button 
  - Otherwise no button
3. If the play button is visible, tapping it should open the webview
  - You can start a livestream by hardcoding `isLive` to true in `data-connector-church-online/src/dataSource.js` 
4. Open a Content Item that isn't live, but has Media (one of the holy spirit videos, for example) 
  - tapping "play" should open the video. 

## TODO:

- [x] Closes #839
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-church-api/config.yml
+++ b/packages/apollos-church-api/config.yml
@@ -23,6 +23,8 @@ ONE_SIGNAL:
   REST_KEY: ${ONE_SIGNAL_REST_KEY}
 CHURCH_ONLINE:
   URL: https://apollos.churchonline.org/api/v1/
+  MEDIA_URLS: []
+  WEB_VIEW_URL: https://apollos.churchonline.org/
 TWILIO:
   ACCOUNT_SID: ${TWILIO_ACCOUNT_SID}
   AUTH_TOKEN: ${TWILIO_AUTH_TOKEN}
@@ -73,6 +75,7 @@ ROCK_MAPPINGS:
     UniversalContentItem:
       EntityType: ContentChannelItem
     WeekendContentItem:
+      ContentChannelId: [5]
       EntityType: ContentChannelItem
     ContentItem:
       EntityType: ContentChannelItem
@@ -101,6 +104,8 @@ ROCK_MAPPINGS:
 
   CAMPAIGN_CHANNEL_IDS:
     - 11
+
+  SERMON_CHANNEL_ID: 5
 
   HOME_FEATURE_CHANNEL_ID: 13
 
--- a/packages/apollos-church-api/local.graphql
+++ b/packages/apollos-church-api/local.graphql
@@ -219,6 +219,8 @@ input LikeEntityInput {
 type LiveStream {
   isLive: Boolean
   eventStartTime: String
+  media: VideoMedia
+  webViewUrl: String
 }
 
 interface Media {
@@ -458,6 +460,7 @@ type WeekendContentItem implements ContentItem & Node {
   theme: Theme
   isLiked: Boolean
   likedCount: Int
+  liveStream: LiveStream
   sharing: SharableContentItem
   features: [Feature]
 }
--- a/packages/apollos-data-connector-church-online/src/__tests__/__snapshots__/resolver.tests.js.snap
+++ b/packages/apollos-data-connector-church-online/src/__tests__/__snapshots__/resolver.tests.js.snap
@@ -6,6 +6,14 @@ Object {
     "liveStream": Object {
       "eventStartTime": "2018-08-06T17:00:00Z",
       "isLive": true,
+      "media": Object {
+        "sources": Array [
+          Object {
+            "uri": "https://example.org/video.mp4",
+          },
+        ],
+      },
+      "webViewUrl": null,
     },
   },
 }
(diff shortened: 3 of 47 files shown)


calebpanza:
@vinnyjth @conrad-vanl so here's my thoughts on it along with what I spoke about with Rob and Brandi.

**Minimal Valuable Product**
The ability to display a weekend live stream when our main campus is streaming online

**Additional Problems to Solve For**
Currently, we only ever have 1 live stream happening at any given time, but there's no guarantee that the live stream will be related to the weekend services. Live Streams happen for some classes (Prep for Marriage online) and special events (ex. a major student even like VIVE is streamed online).

Some future problems we are going to want to solve for is to how to use contextual live streams when there is more than 1 live stream going on at a time. For example, if each campus is streaming worship, we would want to have a user's campus to be the one available to them. Another problem will be when a live stream is related to a content item like an article or study. For example, if we have a study or group set up for Prep for Marriage, we would want to be able to display the live stream for that study on that Content Item

**Proposal**
Leave the `WeekendContentItem` in there as it currently is as to not mess up any visual changes and /or CMS structuring, but extract the `liveStream` resolver logic to the `defaultContentItemResolver` so that it's accessible by any content item type. This way, we solve the current problem of displaying and managing the weekend items with the necessary preparations for having the code available for any Content Item for future implementations.

Thoughts???

vinnyjth:
> Leave the WeekendContentItem in there as it currently is as to not mess up any visual changes and /or CMS structuring, but extract the liveStream resolver logic to the defaultContentItemResolver so that it's accessible by any content item type. This way, we solve the current problem of displaying and managing the weekend items with the necessary preparations for having the code available for any Content Item for future implementations.

The idea with having the logic specifically on the `WeekendContentItem` is that the `WeekendContentItem` is built around the logic we are envisioning for core. It would be really easy to add a new ContentItem, (that could be a REAL ContentItem in Rock, or a "Fake" ContentItem Apollos creates) that has a totally different strategy for `isLive`. 

We also don't really have any logic on the `WeekendContentItem` specifically right now, beside referencing the `dataSource` in the `resolver`. All business logic is encapsulated in the `DataSource`, so if you want to change the logic behind how `isLive` is generated, you can extend the `dataSource` and modify the functionality as you wish, which would override how `isLive` is implemented for the `WeekendContentItem`

https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-data-connector-rock/src/content-items/data-source.js#L139

vinnyjth:
In other words, I see a clear path forward to handle the case you are describing by small changes to the `ContentItem dataSource`. I think the specific of how that will be implemented will be more obvious when your livestream strategy is more solidified.

conrad-vanl:
@vinnyjth talking to @calebpanza and this is close to what I think he's trying to recommend:

```graphql
type DevotionalContentItem implements Node & ContentItem & LiveStreamContentItem {
   ...
}

interface LiveStreamContentItem {
  liveStream: LiveStream
}
```

```javascript
// LiveStreamContentItem resolver
{
  liveStream: (contentItem) => ({
    isLive: true,
    // ...
  })
}

```

vinnyjth:
@conrad-vanl I like that, except you can't really add an interface to to a type via extend. 

For example, you can do 

```
extend type DevotionalContentItem { 
  liveStream: LiveStream
}
```
but you can't 

```
extend type DevotionalContentItem Interface LiveStreamContentItem
```

The resolvers would look the same either way

conrad-vanl:
but @vinnyjth can you not do a root-level resolver for `LiveStream` type that works for any content item you've added it to?

vinnyjth:
> but @vinnyjth can you not do a root-level resolver for LiveStream type that works for any content item you've added it to?

We could, but I prefer not to do that. The resolver is a single statement so it's easy to add to the other ContentItems if you want, is my argument.

calebpanza:
@vinnyjth that makes a lot of sense! I was trying to avoid having to extend resolvers or create new ones for the `ContentItem`, but I see what you're saying.

vinnyjth:
@calebpanza Awesome :D I think the issue I want to solve here is "I want to avoid having to extend resolvers". A lot of our customization strategy relies on doing this, and so I want to make this as absolutely easy as possible.

conrad-vanl:
@vinnyjth have you tested our media player with a real streaming url yet? I don't know if we ever have....

vinnyjth:
I’ll create a story demonstrating it.

vinnyjth:
@conrad-vanl I added a story showing the media player playing a livestream url.

Issue #851, opened 2019-07-09
[UI] FeaturedCard and HighlightCard


burkeshartsis:
### User Story

As a user I want to experience a "Live" sermon experience that combines sermon information with the livestream.

### Description
This is a `FeaturedCard` and `HighlightCard` to be displayed in the home feed. These cards are visually distinctive from other cards/content in the feed used to highlight their rich media content. 

![image](https://user-images.githubusercontent.com/2053547/60890918-4b781d80-a22a-11e9-8b05-35ebe4c138cf.png)

### Acceptance Criteria
- [ ] It should gracefully handle it's props in a way that if one was missing it might still function.
- [ ] It should accept custom title, description, and label text.
- [ ] The label should accept a custom background color but default to brand or transparent coloring.
- [ ] It should accept a custom gradient color and default to brand coloring.
- [ ] It should display if it has been liked or not
- [ ] It should accept media icon and default to "play"

### Other Resources

- [Basecamp todo](https://3.basecamp.com/3926363/buckets/6947926/todos/1898958850)

michael-riddering:
`ContentCard` never exceeds the height of the tallest Featured Card (in the case of an XR, that is 528px)

<img width="864" alt="Screen Shot 2019-07-10 at 7 38 40 AM" src="https://user-images.githubusercontent.com/41083408/60974245-fc43f280-a2e6-11e9-903d-0ce50187cf89.png">

PR #862, opened 2019-07-11
Adds `minAspectRatio` and `maxAspectRatio` support to ConnectedImage

burkeshartsis:
## DESCRIPTION
How images are handled in React Native is surprisingly hard/complex mostly because the luxuries we are use to on other platforms are not available to us in React Native.

This PR adds `minAspectRatio` and `maxAspectRatio` props to `ConnectedImage` as a way to control how an image is rendered when you don't know it's dimensions until after it's rendered. Often times you do know the aspectRatio you would like that image to conform to. These new props allow you to set one or both to create min/max bounds on the rendered images shape without having to know it's dimensions 💥

Designers and the resulting handoff to engineering is generally in min/max pixel values (or maybe percents). Translating these values to aspect ratios is not immediately intuitive but it's also not hard either. `width/height = aspectRatio`. Given the flexibility the solution this PR creates and the constraints we are operating in I think this is worth working around/getting used to.

*Note*: `minAspectRatio` and `maxAspectRatio` require the `maintainAspectRatio` prop based on the way the code is written currently and does not prompt or warn the engineer if it's not included.

### How do I test this PR?
Check out 'dem stories ✨💥  

All new code is tested 💪 

## TODO:

- [x] Partial work towards #851 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-kit/src/ConnectedImage/ConnectedImage.stories.js
+++ b/packages/apollos-ui-kit/src/ConnectedImage/ConnectedImage.stories.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import { storiesOf } from '@apollosproject/ui-storybook';
+
+import ConnectedImage from '.';
+
+storiesOf('ui-kit/ConnectedImage', module)
+  .add('default', () => (
+    <ConnectedImage
+      style={{ width: 300, height: 200 }} // eslint-disable-line react-native/no-inline-styles
+      source={{ uri: 'https://picsum.photos/600/400/?random' }}
+    />
+  ))
+  .add('maintainAspectRatio', () => (
+    <ConnectedImage
+      source={{ uri: 'https://picsum.photos/600/400/?random' }}
+      maintainAspectRatio
+    />
+  ));
--- a/packages/apollos-ui-kit/src/ConnectedImage/ConnectedImage.stories.js
+++ b/packages/apollos-ui-kit/src/ConnectedImage/ConnectedImage.stories.js
@@ -18,7 +18,10 @@ storiesOf('ui-kit/ConnectedImage', module)
     />
   ))
   .add('min+max aspectRatio', () => (
-    <ScrollView style={{ flex: 1 }}>
+    <ScrollView
+      // eslint-disable-next-line react-native/no-inline-styles
+      style={{ flex: 1 }}
+    >
       <ConnectedImage // should render square
         source={{ uri: 'https://picsum.photos/200/200/?random' }}
         maintainAspectRatio
--- a/packages/apollos-ui-kit/src/ConnectedImage/ConnectedImage.stories.js
+++ b/packages/apollos-ui-kit/src/ConnectedImage/ConnectedImage.stories.js
@@ -36,25 +36,21 @@ storiesOf('ui-kit/ConnectedImage', module)
         source={{ uri: 'https://picsum.photos/600/400/?random' }}
         minHeight={200}
         style={{ width: 50 }}
-        maintainAspectRatio
       />
       <ConnectedImage
         source={{ uri: 'https://picsum.photos/600/400/?random' }}
         maxHeight={100}
-        style={{ width: 50 }}
-        maintainAspectRatio
+        style={{ width: 50, height: '100%' }}
       />
       <ConnectedImage
-        source={{ uri: 'https://picsum.photos/600/400/?random' }}
+        source={{ uri: 'https://picsum.photos/600/100/?random' }}
         minHeight={200}
-        style={{ width: 500 }}
-        maintainAspectRatio
+        style={{ width: 250 }}
       />
       <ConnectedImage
-        source={{ uri: 'https://picsum.photos/600/400/?random' }}
+        source={{ uri: 'https://picsum.photos/600/100/?random' }}
         maxHeight={100}
-        style={{ width: 500 }}
-        maintainAspectRatio
+        style={{ width: 250, height: '100%' }}
       />
     </React.Fragment>
   ));
--- a/packages/apollos-ui-kit/src/ConnectedImage/ConnectedImage.stories.js
+++ b/packages/apollos-ui-kit/src/ConnectedImage/ConnectedImage.stories.js
@@ -15,4 +15,18 @@ storiesOf('ui-kit/ConnectedImage', module)
       source={{ uri: 'https://picsum.photos/600/400/?random' }}
       maintainAspectRatio
     />
+  ))
+  .add('minHeight', () => (
+    <ConnectedImage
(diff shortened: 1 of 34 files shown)


burkeshartsis:
@conrad-vanl this is still a little rough but what do you think of this? Couple things worth noting
- this throws a warning about using min/maxHeight outside of a style so we might need different prop names
- min/maxHeight actually works even without `maintainAspectRatio` being passed with it which suggests that it's being applied as a style and having the intended outcome regardless

packages/apollos-ui-kit/src/ConnectedImage/index.js
```diff
@@ -156,6 +164,17 @@ class ConnectedImage extends PureComponent {
     };
   };
 
+  getImageHeight(imageHeight) {
```
conrad-vanl: I believe we're still getting confused on image height vs rendered height, and the reason why this works in your stories is because of coincidence between image and screen size. I'll try to set up some demos to see if/when this breaks.

conrad-vanl:
Based on your comments i'm wondering if there's some old deprecated support built in for these `min`/`max` `height` props baked into react-native.

I'll check it out.

conrad-vanl:
@burkeshartsis I'm actually not sure what you're solving here

It does appear this works, however I can completely revert your changes to `ConnectedImage` (the `getImageHeight` stuff and all of your stories still work, which makes me think you're not really doing anything.

However, I don't think this solves the need we have for cards. The need we have for cards are that:

1. The aspect ratio of the original image is respected up to certain min/max bounds
2. When the image is outside of the min/max bounds, the image becomes cropped to fill the bounding box

If we enable aspect ratios then I think this breaks down, because with aspect ratios enabled the images will never get cropped. You can see this with the second story I added...these two "smoke test" stories should look identical if we meet the logic above

What happened to trying `min`/`max` `aspectRatio`?

burkeshartsis:
`minAspectRatio` and `maxAspectRatio` require the `maintainAspectRatio` prop based on the way the code is written currently and does not prompt or warn the engineer if it's not included. That said, I’m not sure that logically the prop usage would say it has to be. Granted, I can’t think of ~many~ any situation where one would want a distorted image but I suppose it could be possible. Maybe for a bouncing animation 🙃🤷‍♂ 

Anyways, I think this code could work for such situations but it currently doesn’t. @conrad-vanl do we want to support this usage? If not I think we should add some proptype checking to prompt the user.

conrad-vanl:
@burkeshartsis it worked without providing `maintainAspectRatio` before this commit - f01f5d9

burkeshartsis:
@conrad-vanl so part of the changes you are noting in that commit were because I could not logic a reason for how `min/maxAspectRatio` would work without `maintainAspectRatio` so I refactored it to be more clearly linked. The logic prior to that refactor commit can be seen [here](https://github.com/ApollosProject/apollos-prototype/blob/6345fef3c7641359ef00a90cf8fe5519b82e06bc/packages/apollos-ui-kit/src/ConnectedImage/index.js#L127) and would suggest that `min/maxAspectRatio` required `style.aspectRatio` to be set. `style.aspectRatio` was only set under two conditions with `maintainAspectRatio` or if `isLoading` was true and an image was loading *for the first time*. If I was to understand it correctly cached images would never have set `style.aspectRatio` and as a result not show up. I believe I tested this to be true but am no longer certain of that.

conrad-vanl:
@burkeshartsis got it. Yeah, you're right and I'm tracking now.

So lets just throw a `console.warn` in that `aspectRatio` function that checks for the existence of `minAspectRatio` OR `maxAspectRatio` being set without `maintainAspectRatio` being truthy.

conrad-vanl:
Once the final work on this is done @vinnyjth is going to have to review it since both @burkeshartsis and myself have touched this so much.

burkeshartsis:
@vinnyjth this is ready for you

packages/apollos-ui-kit/src/ConnectedImage/ConnectedImage.stories.js
```diff
@@ -0,0 +1,60 @@
...
+      source={{ uri: 'https://picsum.photos/600/400/?random' }}
+    />
+  ))
+  .add('maintainAspectRatio', () => (
+    <ConnectedImage
+      source={{ uri: 'https://picsum.photos/600/400/?random' }}
+      maintainAspectRatio
+    />
+  ))
+  .add('min+max aspectRatio', () => (
```
vinnyjth: this makes testing SOOOOOOO easy. Thank you.

packages/apollos-ui-kit/src/ConnectedImage/ConnectedImage.tests.js
```diff
@@ -35,21 +35,102 @@ describe('the ConnectedImage component', () => {
     );
     expect(tree).toMatchSnapshot();
   });
-  describe('updateCache', () => {
-    it('updates cache with image uri and sizes', async () => {
-      const source = {
-        url: '//via.placeholder.com/320x240',
-      };
-      await updateCache(source);
+  it('should render with a minAspectRatio', () => {
```
vinnyjth: 💥

packages/apollos-ui-kit/src/ConnectedImage/ConnectedImage.tests.js
```diff
@@ -35,21 +35,102 @@ describe('the ConnectedImage component', () => {
...
+          }}
+          maintainAspectRatio
+          minAspectRatio={0.5}
+          maxAspectRatio={1.5}
+        />
+      </Providers>
+    );
+    expect(tree).toMatchSnapshot();
+  });
+  it('should throw an error if minAspectRatio is used without maintainAspectRatio', () => {
```
vinnyjth: 💥 💥 💥 💥

packages/apollos-ui-kit/src/ConnectedImage/index.js
```diff
@@ -111,15 +133,30 @@ class ConnectedImage extends PureComponent {
...
+      // determine the aspect ratio of an image based on its width and height
       if (firstSource && firstSource.width && firstSource.height) {
         style.aspectRatio = firstSource.width / firstSource.height;
+
+        // account for possible min/max aspectRatio bounds
+        if (this.props.minAspectRatio || this.props.maxAspectRatio) {
+          const maxAspectRatio = this.props.maxAspectRatio || style.aspectRatio;
+          const minAspectRatio = this.props.minAspectRatio || 0;
+
+          style.aspectRatio = Math.max(
```
vinnyjth: Love this logic, nice work.

vinnyjth:
Testing on android, and then approving!

PR #867, opened 2019-07-12
Adds FeaturedCard component

burkeshartsis:
## DESCRIPTION

This PR is blocked by #873 and ~#881~. The `Label` in the screenshots below will display correctly after #881 is merged. 

This PR adds a `FeaturedCard` component that reflects the latest designs for Core. It aims to be as minimal as possible.

This is pretty faithful to the design. There were some changes that were made with the input and consent of the designer (@michael-riddering). 

### How do I test this PR?
There is a story for every prop. Check 'um out!

100% coverager 💥 

## TODO:

- [x] Partial work towards #851 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-kit/src/FeaturedCard/index.js
```diff
@@ -0,0 +1,135 @@
...
+  fill: theme.colors.primary,
+  size: theme.sizing.baseUnit * 1.5,
+  iconPadding: theme.sizing.baseUnit * 0.75,
+  style: {
+    marginLeft: theme.sizing.baseUnit,
+    backgroundColor: theme.colors.text.primary,
+  },
+}))(ButtonIcon);
+
+const Label = withTheme(
```
burkeshartsis: I'm 100% open to better ideas here 😬

packages/apollos-ui-kit/src/FeaturedCard/index.js
```diff
@@ -0,0 +1,135 @@
...
+    mixin={{
+      type: get(theme, 'type', 'dark').toLowerCase(), // not sure why we need toLowerCase
+      colors: get(theme, 'colors', {}),
+    }}
+  >
+    <StyledCard>
+      <Image source={image} />
+
+      <Content>
+        {labelText // only render a label if we have `labelText`
```
burkeshartsis: I'm not a big fan of this nested conditional craziness here. Given that the `Label` component above is already dizzying should I just turn that into a `renderLabel` function and move this logic there where at least it could be written with a more clear `if` statement.

packages/apollos-ui-kit/src/FeaturedCard/index.js
burkeshartsis: In an effort to consolidate component API I used `labelText` here for some conditional logic when I could have otherwise used `isLive`.

packages/apollos-ui-kit/src/FeaturedCard/index.js
```diff
@@ -0,0 +1,156 @@
...
+    PropTypes.arrayOf(ImageSourceType),
+    ImageSourceType,
+  ]).isRequired,
+  title: PropTypes.string.isRequired,
+  actionIcon: PropTypes.string,
+  description: PropTypes.string,
+  isLiked: PropTypes.bool,
+  LabelComponent: PropTypes.element,
+  labelText: PropTypes.string,
+  onPressAction: PropTypes.func,
```
burkeshartsis: It's nice to have these `onPress` props but technically we don't even need them. Should we consider removing them?

packages/apollos-ui-kit/src/FeaturedCard/index.js
```diff
@@ -0,0 +1,156 @@
...
+        <ActionLayout hasDescription={description}>
+          <FlexedView>
+            {// if we have a `description` render it otherwise render a longer but narrower `H2` `title`
+            description ? (
+              <BodyText numberOfLines={2}>{description}</BodyText>
+            ) : (
+              <H2 numberOfLines={4}>{title}</H2>
+            )}
+          </FlexedView>
+          {onPressAction ? (
```
burkeshartsis: Given that technically we don't even need the action button to be clickable using `onPressAction` as the conditional for render is not ideal. That said, if we changed this, what does the API expose that would tell us it "hasMedia"?

packages/apollos-ui-kit/src/FeaturedCard/index.js
```diff
@@ -0,0 +1,156 @@
...
+  iconPadding: theme.sizing.baseUnit * 1.5,
+}))(ButtonIcon);
+
+const Image = withTheme(({ theme }) => ({
+  overlayColor: theme.colors.primary,
+  style: { aspectRatio: 1 },
+}))(CardImage);
+
+const Content = styled(({ theme }) => ({
+  alignItems: 'flex-start',
```
burkeshartsis:
```suggestion
  alignItems: 'flex-start', // needed to make `Label` display as an "inline" element
```

packages/apollos-ui-kit/src/FeaturedCard/index.js
burkeshartsis: So technically, we don't currently support any interactions on cards except tapping them. So exposing these props could make the card more interactive but in practice it's usually less desirable than it sounds since everything lives in a `ScrollView` and we already struggle with handling interactions there. Like sometimes if you scroll slowly a card will "pop" thinking you are tapping on it but then the interaction handler passes the press to the scroll handler for scrolling. Adding two additional buttons to a card might be more detrimental than good.

The other question is with regards to UX. Do we even want people to be able to interact this way without the context of the content the card leads to? Does liking something without reading it make sense? Is like more than like? Is it a bookmark? What is the hierarchy of importance for weekend/card data? Is interacting/consuming media at the top? Etc. Etc. @michael-riddering Thoughts on this? I know you had wanted them to be interactive so they are now 😉the question is do we keep this functionality?

packages/apollos-ui-kit/src/FeaturedCard/index.js
burkeshartsis: so do you like the idea of `hasMedia` triggered by any of those? Or we could do it based on `actionIcon` which defaults to a truthy value but could be explicitly set to `null` That's not the most intuitive option though.

packages/apollos-ui-kit/src/theme/defaultTheme.js
```diff
@@ -139,6 +141,10 @@ export const buttons = ({ colors: themeColors }) => ({
     fill: themeColors.text.primary,
     accent: themeColors.text.primary,
   },
+  overlay: {
```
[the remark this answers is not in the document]
burkeshartsis: @conrad-vanl 's idea!

packages/apollos-ui-kit/src/FeaturedCard/index.js
```diff
@@ -0,0 +1,156 @@
...
+import styled from '../styled';
+import Card, { CardImage, CardLabel, CardContent } from '../Card';
+import FlexedView from '../FlexedView';
+import { H2, BodyText } from '../typography';
+import { ButtonIcon } from '../Button';
+import { ImageSourceType } from '../ConnectedImage';
+
+const StyledCard = withTheme(({ theme }) => ({
+  cardColor: theme.colors.primary,
+}))(Card);
```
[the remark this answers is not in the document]
burkeshartsis: not sure of a better way to do this. I mean we could not use the prop and pass a `backgroundColor` via style. That's all `cardColor` does.

packages/apollos-ui-kit/src/FeaturedCard/index.js
```diff
@@ -0,0 +1,156 @@
...
+    type: PropTypes.string,
+    colors: PropTypes.shape({}),
+  }),
+};
+
+FeaturedCard.defaultProps = {
+  actionIcon: 'play-solid',
+};
+
+FeaturedCard.displayName = 'FeaturedCard';
```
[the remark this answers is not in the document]
burkeshartsis: I think you are correct

packages/apollos-ui-kit/src/FeaturedCard/index.js
```diff
@@ -0,0 +1,156 @@
...
+const LikeButtonWrapper = styled({
+  position: 'absolute',
+  top: 0,
+  right: 0,
+})(View);
+
+const LikeButton = withTheme(({ theme }) => ({
+  size: theme.sizing.baseUnit * 1.5,
+  iconPadding: theme.sizing.baseUnit * 1.5,
+}))(ButtonIcon);
```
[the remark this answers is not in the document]
burkeshartsis: I like this idea if we wanted to keep this as a functional button! I was wondering why we didn't already have a connected `LikeButton` in the UI kit actually. That said, what would we default to here?

packages/apollos-ui-kit/src/FeaturedCard/index.js
```diff
@@ -0,0 +1,156 @@
...
+          ? // if we have a custom `LabelComponent` render it
+            LabelComponent || ( // otherwise default to `Label`
+              <Label
+                customTheme={theme}
+                labelText={labelText}
+                hasDescription={description}
+              />
+            )
+          : null}
+        {// only if we have a `description` render a shorter full width `H2` `title`
```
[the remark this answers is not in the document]
burkeshartsis: I'm 100% *not* opposed to this and I like your naming suggestions here too!

packages/apollos-ui-kit/src/FeaturedCard/index.js
conrad-vanl: Yeah I almost feel like you could just drop `labelText` altogether and just use `LabelComponent`

burkeshartsis: My only concern there is that it would pretty much move almost required visual code to the connected component level. I say almost because while it gracefully renders without a label our design would say it always should.

packages/apollos-ui-kit/src/FeaturedCard/index.js
burkeshartsis: the decision was to remove them for simplicity on all fronts

burkeshartsis:
@vinnyjth I think this is ready for you again. Not really sure its a ton better then before the review started though

packages/apollos-ui-kit/src/FeaturedCard/index.js
vinnyjth: I think this is way better now 👍

packages/apollos-ui-kit/src/FeaturedCard/index.js
```diff
@@ -0,0 +1,170 @@
...
+    mixin={{
+      type: get(theme, 'type', 'dark').toLowerCase(), // not sure why we need toLowerCase
+      colors: get(theme, 'colors', {}),
+    }}
+  >
+    <StyledCard>
+      <Image source={image} />
+
+      <Content>
+        {renderLabel(description, LabelComponent, labelText, isLive, theme)}
```
vinnyjth: I think this is better now 👍 I had an easier time understanding all the permutations.

burkeshartsis:
updated to have the new gradients I think we're good

Issue #806, opened 2019-06-19
Content Shouldn't Load False Children Placeholders


redreceipt:
### User Story

As a user that clicks a piece of content, I shouldn't see false children content in the loading state.

### Description

When you click a `ContentItem` initially, a list of children content shows up in the loading state even when there isn't any.

![20190605_203335](https://user-images.githubusercontent.com/2659478/59790685-7fe96100-929e-11e9-9b1c-6350302962c1.gif)

### Why

It's simply a better user experience.

### Acceptance Criteria

* [ ] It should not show children content tiles in the loading state if there isn't any.

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

redreceipt:
@vinnyjth care if I work on this?

PR #903, opened 2019-07-23
Fixes Poor Loading Content UI

redreceipt:
## DESCRIPTION

Kapture 2019-07-22 at 22 54 34

This removes a very presumptuous loading state. The `HorizontalContentFeed` was passing a loading state to `HorizontalTileFeed` before it knew if there were any children at all. This was causing a poor experience with gray boxes showing up even when there was no child content and then disappearing.

Due to the nature of the data call, we won't know if there's children during the content loading state so I opted to just remove the loading state all together. Doesn't look bad, which I think is an improvement until we come up with a better solution.

Turns off the loading state for children of `UniversalContentItem`

### How do I test this PR?

Click on any content, there's no loading state.

## TODO:

- [x] Closes [#806]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
@vinnyjth in the Gif above, when he taps on "Coming up next" - we shouldn't be getting that big giant placeholder screen. We should be atleast able to show image and title and such, which means we're not handling the cache or the queries correctly. Something we need to look into

PR #904, opened 2019-07-23
New MediaPlayer styles

conrad-vanl:
## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
@burkeshartsis checkout the changes to [this file here](https://github.com/ApollosProject/apollos-prototype/pull/904/commits/b3aa4dbba98ae86914b4ebdd941ad8d003970709#diff-37812f75666669a8da65ab8fce7bd9fc) and lmk what you think. Essentially, it's fixing two issues:

1. The `backgroundColor` propType was mis-named and should be `cardColor`. Not sure how our linter didn't fix this

2. Two issues with `aspectRatio` - first, it was only being applied to the `OverflowFix` component. This meant that there could be instances where the card container was a different size then your card content. This fixes that by applying it to both. Second, aspect ratio was always being passed even if you didn't supply `forceRatio`. This could cause side affects, such as not being able to supply `aspectRatio` via `styled` as it would always clear it out.

Both of these changes should be transparent, but they do cause mild snapshot changes.

burkeshartsis:
@conrad-vanl 

1. I saw this in my last PR that touched this file but didn't want to make the corrective change across that app in my PR. Not sure how we basically got two props that weren't connected.

2. 👇 
    - Have we tested that we still need the overflow fix? It looks like we do 😢
    - Your `aspectRatio` concern is valid but if I'm reading the current code correctly the `OverflowFix` is the child of the `Card` so card should conform to child's `aspectRatio`. I could see an issue here, however, if you passed `aspectRatio` via style or directly and not via `forceRatio` as it would be passed to the wrapping `Card` not the child `OverflowFix` where it would need to be.
        1. What if we changed the name of `forceRatio` to `aspectRatio` (I'm not sure we can do that) and then hijacked it and explicitly passed it to the child.
        2. If the option above doesn't work, what if we keep `forceRatio` as a prop and just explicitly hijack `aspectRatio` by not passing it to anything. Effectively making `aspectRatio` useless and requiring the use of `forceRatio`.

> Both of these changes should be transparent, but they do cause mild snapshot changes.
We should still probably move these changes to a separate PR.

conrad-vanl:
> concern is valid but if I'm reading the current code correctly the OverflowFix is the child of the Card so card should conform to child's aspectRatio

I thought this too, but oddly enough the styling for the `MediaThumbnail` component I added in this PR is not correct without this fix. Essentially does what I described...for whatever reason the card container is rendering a different size than the child.

burkeshartsis:
@conrad-vanl does any part of this PR fix #952 ?

conrad-vanl:
@burkeshartsis I don't know why this never moved towards trying to towards a review/merge.

Do you remember anything about this PR? I'm tempted to close unless we have concerns there are still issues here to fix....

burkeshartsis:
This is so old I wouldn't be able to tell you without spending time to look at the PR's code and test its changes.

PR #918, opened 2019-07-24
Tests App Utils Functions

redreceipt:
## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apolloschurchapp/src/utils/__tests__/utils.tests.js
```diff
@@ -0,0 +1,39 @@
...
+  };
+  test('it returns undefined when no endCursor', () => {
+    expect(
+      fetchMoreResolver({
+        data: { userFeed: { edges: [], pageInfo: {} } },
+        ...args,
+      })()
+    ).toBeUndefined();
+  });
+  // TODO is it possible fetchMore functionality???
```
vinnyjth: @redreceipt Could you call the function with an end cursor and write an expect for the `fetchMore.mock.calls`? 

Additional documentation: https://jestjs.io/docs/en/mock-functions#using-a-mock-function

vinnyjth:
What you have looks great! I left a comment about how we could cover the other 50% of the `fetchMoreResolver`

packages/apolloschurchapp/src/utils/__tests__/utils.tests.js
redreceipt: definitely closer but I need to inspect the return value of the `updateQuery` function, the `fetchMore` function returns nothing, so something like this but I know it's not quite right:

```
    console.log(args.fetchMore.mock.calls[1][0].updateQuery.mock);
```

from here:

Screen Shot 2019-07-29 at 10 47 17 AM

redreceipt: ...wait, I need to mock the `updateQuery` function, hang on...

redreceipt:
ok @vinnyjth super close here, I'm going to try and explain what's happening. here's my test:

Screen Shot 2019-07-29 at 11 32 32 AM

but the log is saying it's expecting this?

Screen Shot 2019-07-29 at 11 34 02 AM

somewhere `0` should be converted to `2` on `baz` but not sure where that's happening in real life.

redreceipt:
got it, I was confused about how JS was handling arrays of objects. there is no spoon.

packages/apolloschurchapp/src/utils/fetchMoreResolver.js
```diff
@@ -6,7 +6,7 @@ export default function fetchMoreResolver({
   variables,
   fetchMore,
   data,
-} = {}) {
```
redreceipt: I figured I'd get a question about this, initializing the arguments here was misleading because we aren't handling it anyway. We want this to throw an exception if they don't pass any args because this function will never be useful without them, and won't be called outside of a developer actually calling it.

vinnyjth: Yah, that makes sense to me. Removing this a good call.

PR #919, opened 2019-07-24
Updates the Test with Apollo Helper

redreceipt:

Updates the `renderWithApolloData` function to be more true to its name. It had nothing to do with Apollo before. And this also makes it a one-stop shop if I don't care about or need any other providers. Now usage is simply:

```
mocks = [{YOUR_MOCKS}]
tree = await renderWithApolloData(<FancyButton>, mocks)
expect(tree).toBeOrDoWhateverYouWant
```

There's just a lot of confusion around how to test connected components and this simplifies it I think. And as far as I can tell, the current `Providers` in the app package doesn't even contain `MockedProviders` so I don't think we're testing *any* connected components in the app.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apolloschurchapp/src/utils/testUtils.js
+++ b/packages/apolloschurchapp/src/utils/testUtils.js
@@ -1,10 +1,34 @@
 /* eslint-disable import/prefer-default-export */
+import React from 'react';
 import renderer from 'react-test-renderer';
+import { MockedProvider } from 'react-apollo';
 import wait from 'waait';
 
-export const renderWithApolloData = async (component) => {
-  const tree = renderer.create(component);
-  await wait(1);
-  tree.update(component);
+export const renderWithApolloData = async (component, mocks = []) => {
+  /*
+   * mocks should be in the form:
+   *
+   *   mocks = [
+   *     {
+   *       request: {
+   *         query: GET_DOG_QUERY,
+   *         variables: {
+   *           name: 'Buck',
+   *         },
+   *       },
+   *       result: {
+   *         data: {
+   *           dog: { id: '1', name: 'Buck', breed: 'bulldog' },
+   *         },
+   *       },
+   *     },
+   *   ];
+   *
+   */
+
+  const tree = renderer.create(
+    <MockedProvider mocks={mocks}>{component}</MockedProvider>
+  );
+  await wait(0);
   return tree;
 };


packages/apolloschurchapp/src/utils/testUtils.js
```diff
@@ -1,10 +1,34 @@
...
+   *       },
+   *     },
+   *   ];
+   *
+   */
+
+  const tree = renderer.create(
+    <MockedProvider mocks={mocks}>{component}</MockedProvider>
+  );
+  await wait(0);
```
conrad-vanl: was the change from `wait(1)` to `wait(0)` purposeful / okay?

conrad-vanl:
any idea why ci tests are failing?

packages/apolloschurchapp/src/utils/testUtils.js
redreceipt: it was on purpose, simply I figured we didn't need to wait for a whole sec and that's what was in the apollo docs. plus I took out the `update`, not sure what it was doing anyway.

vinnyjth: That `update` is needed so that we don't get `loading: true` in our snapshots, but instead get the actual mock.

conrad-vanl: I think it's milliseconds, not seconds, fwiw. It snapshots still pass though it's probably fine?

conrad-vanl: and agree on the `update` part, I think?

vinnyjth:
@michael-riddering I like the idea of this PR. You'll need to refactor the ApolloProvider out of the ` Providers` though (so we don't have two providers when testing). The name is also accurate as is. The code prior to this update did the necessary work so that mocked data would actually show up in the snapshot. Having the provider included is a nice idea, but will require a refactor of existing tests.

packages/apolloschurchapp/src/utils/testUtils.js
redreceipt: but it didn't change any snaps when I took it out?

vinnyjth: @redreceipt something must have changed because tests are failing ;)

redreceipt:
gotcha! I'll make those changes and ping you

redreceipt:
changed my mind, the wrapped component still needs a `client` prop from either `MockedProvider` or `ApolloProvider` but I can't figure out a way to pass down the client prop dynamically :(

Issue #934, opened 2019-07-29
Feature Sharing doesn't work


redreceipt:
### Details

Sharing doesn't work on the text features. the `ShareButton` accepts a piece of content and then goes and pulls the data it needs from the title. And the message is is just `title + url`. the `message` prop isn't even being used.

### Expected Behavior

When you share a feature, it should share the content of the feature. 

### Actual Behavior

It shares the content title.

![Kapture 2019-07-29 at 15 40 20](https://user-images.githubusercontent.com/2659478/62077053-32560180-b217-11e9-93b1-a652b347b5e6.gif)

### Possible Fix

<details><summary>Additional Info</summary>

### Your Environment

- Environment name and version (e.g. Chrome 39, node.js 5.4):
- Operating System and version (desktop or mobile):
- Link to your project:

### Steps to Reproduce

1.  first...
2.
3.
4.

### Stack Trace

</details>

PR #935, opened 2019-07-29
Refactor Share Buttons

redreceipt:
## DESCRIPTION

Kapture 2019-07-29 at 16 19 23

Currently sharing text features doesn't work, this fixes that. The `<Share>` implementation was confusing and not working like we expected it to be. This clarifies what it's doing a little better by changing the naming.

Removes pulling `message` from the API. We weren't using it anyway in the client. We can discuss later on expanding functionality.

### How do I test this PR?

- Share a piece of content
- Share a text feature (Gathering Series, Spur One Another has one)

None.

## TODO:

- [x] Closes [#934 ]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
@michael-riddering I like the direction this PR takes and I feel like it does a good job of cleaning this code up and adding some clarity. 

However, I don't think I understand why you are removing `message` altogether. What's the harm in keeping `message`, and understanding that right now we don't send a `message` over from the API?

vinnyjth:
OH! I see that we don't pass `message` to the `Share.share` call. We should fix that and pass `message`.

vinnyjth:
🤦‍♂ @redreceipt

redreceipt:
the danger in passing it to the share call is that it behaves differently on iOS and Android, but `title` always behaves the same and does what we are expecting `message` to do. which is why we opted to do it like this I think. I think before we just pass `message`, we should define what we want to happen with the share function and then code accordingly.

and I'll always opt for removing complexity and revisiting when there's a need.

richarddubay:
I'm kind of confused here. As best I can tell, according to the RN docs (https://facebook.github.io/react-native/docs/share) the only thing that is iOS only is `url`. `message` and `title` work everywhere. 

In fact, those docs say that one of `message` or `url` is required ... so that makes me believe that `message` actually needs to be there so that sharing would work on Android. Is that not the case?

I'm assuming this is why the code we currently have in `share.js` looks like this:

```
const share = ({ title, url }) => {
  Share.share({
    title,
    message: Platform.OS === 'android' ? `${title}\n${url}` : title,
    url,
  });
};
```

redreceipt:
once #938 gets merged in, this PR will be simplified to just refactor some names

redreceipt:
ok @vinnyjth I think this one still has merit

packages/apolloschurchapp/src/ui/ShareContentButtonConnected/getShareContent.js
```diff
@@ -7,7 +7,6 @@ export default gql`
         id
         sharing {
           url
-          message
```
vinnyjth: We need to roll back this change

packages/apolloschurchapp/src/content-single/ActionContainer/index.js
```diff
@@ -23,7 +23,7 @@ const ActionContainer = ({ itemId }) => (
     <MediaPlayerSpacer>
       <PositioningView>
         <LikeButton itemId={itemId} />
```
vinnyjth: Can you rename LikeButton in a similar way as a part of this PR? Their functionality is very similar and I would like them to match :)

packages/apolloschurchapp/src/ui/ShareContentButtonConnected/getShareContent.js
redreceipt: done

packages/apolloschurchapp/src/content-single/ActionContainer/index.js
redreceipt: done

PR #938, opened 2019-07-30
Fixes sharing text feature

### LINKED ISSUE #934: Feature Sharing doesn't work

vinnyjth:

Updates the `share` function so it accepts a `message` property and fixes a typo that was preventing `message` from being passed down. 

Chose to now engage in a more full featured refactor like @redreceipt did in #935. Trying to find a way to get the intended functionality working without a full scale refactor. 

### How do I test this PR?

1. Open a content item with a text feature
2. Click the share button
3. click "copy to clipboard" 
3. paste somewhere else on your ios device (see screenshot for details) 

## TODO:

- [x] Closes #934
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
in thinking about this, don't you think we should not do anything fancy with the `Share.share` function and handle all the platform stuff at the source? so:

`Platform.OS === android ? share({title, message: message+url}) : share({title, message, url})`

redreceipt:
wait nevermind, I see the value of it, but I have some suggestions...

packages/apolloschurchapp/src/utils/content/share.js
```diff
@@ -1,12 +1,12 @@
...
 
-const share = ({ title, url }) => {
+const share = ({ title, url, message }) => {
   Share.share({
     title,
-    message: Platform.OS === 'android' ? `${title}\n${url}` : title,
+    message:
+      Platform.OS === 'android'
+        ? `${message || title}\n${url}`
+        : message || title,
```
redreceipt:
```suggestion
        : message,
```

packages/apolloschurchapp/src/utils/content/share.js
```diff
@@ -1,12 +1,12 @@
...
-// import getSiteLink from './getSiteLink';
 
-const share = ({ title, url }) => {
+const share = ({ title, url, message }) => {
   Share.share({
     title,
-    message: Platform.OS === 'android' ? `${title}\n${url}` : title,
+    message:
+      Platform.OS === 'android'
+        ? `${message || title}\n${url}`
```
redreceipt:
```suggestion
        ? `${message}\n${url}`
```

packages/apolloschurchapp/src/utils/content/share.js
```diff
@@ -1,12 +1,12 @@
...
-// import { track, events } from 'apolloschurchapp/src/analytics';
-// import getSiteLink from './getSiteLink';
 
-const share = ({ title, url }) => {
+const share = ({ title, url, message }) => {
   Share.share({
     title,
-    message: Platform.OS === 'android' ? `${title}\n${url}` : title,
+    message:
+      Platform.OS === 'android'
```
redreceipt:
```suggestion
      Platform.OS === 'android' && url
```

redreceipt:
I added some comments that will break content sharing as it stands but the fix should be simple, we should be passing `message` to the share function, not `title`. We are intending it to be the message anyway.

vinnyjth:
@michael-riddering I added a `message` to the ContentItem sharing resolver.

redreceipt:
love it, maybe I'm splitting hairs but I still think our implementation is a little confusing. 

the API field should be called `summary`, not `message`
https://github.com/ApollosProject/apollos-prototype/blob/9aa3c6abd4d8634d9916068f5a414c4fba7b4787/packages/apollos-data-schema/src/index.js#L403

the `ContentItem` should be passing `title` as `message` from the API into the `content` prop of `<Share>, it it's odd behavior to share the content and only paste the first sentence of the article. so we are explicit about the fact that we are using the title as the shared message body which I think is a better UX.
https://github.com/ApollosProject/apollos-prototype/blob/9aa3c6abd4d8634d9916068f5a414c4fba7b4787/packages/apolloschurchapp/src/ui/ShareButton/index.js#L18

redreceipt:
if you want, I can make those suggestions apart of a refactor PR, I've already started that other one and I still think we should rename `<Share>` and `<ShareButton>`

vinnyjth:
> the API field should be called summary, not message
The `sharing` interface is meant to be used more broadly than ContentItem, hence why we call that field `message`.

>the ContentItem should be passing title as message from the API into the content prop of ShareButton

~I don't think I agree with this. I think we should pass the `summary` as the message, and the content `title` as the title.~ I misunderstand the title prop. You're right about needing the `title` of the content item in the `message` field.

vinnyjth:
@redreceipt I this this is good now!

Issue #676, opened 2019-05-02
Rock Endpoints for Family, Home Group, and Serving Group Membership


redreceipt:
### User Story

As a user, I want to know if I'm in any groups so that I can do stuff with that info.

### Description

Add queries to `apollos-data-connector-rock` that tell me if I'm in any groups. ex.

`family(personID)`
`homeGroups(personID)`
`servingGroups(personID)`

### Why

For example, in our prayer feature, we don't want to show a list of community prayers if I'm not in any groups.

### Acceptance Criteria

* [ ] It should tell me if I'm in a family
* [ ] It should tell me if I'm in a home group
* [ ] It should tell me if I'm in a serving group
* [ ] (optional) It should give me an array of people in each of those groups

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

redreceipt:
cc @richarddubay

redreceipt:
@vinnyjth you ok if I work on this? the AC would simply be adding a `Group` type and three queries that return an array of groups. Would add tremendous value for us right now and I'm confident the project as a whole in the future.

vinnyjth:
@redreceipt Can you give me the 10 second update on what additions core would see from this? We might need to decide if this is something you guys work on in Newspring and then we move over to core eventually.

redreceipt:
no client-side additions in core for now. I could do it on our API and then move it over but I was trying to avoid double work when it's almost guaranteed that core will have some `Group` implementation in the future. but also understand you don't wanto add scope to your plate. just thought the QA sprint might be a good opportunity

redreceipt:
this first go around is simple: 

```
type Group {
  id
  leader Person
  members [Person]
}

Query {
  families(personID) [Group]
  homeGroups(personID) [Group]
  servingGroups(personID) [Group]
}

vinnyjth:
Gotcha. I agree with you about it being important in core _eventually_. Not sure when that is though. Feel free to give this a stab, but I might not be able to set aside time to review it during this QA sprint.

vinnyjth:
Done in #939

PR #939, opened 2019-07-31
Adds Groups Functionality to API

redreceipt:
## DESCRIPTION

Screen Shot 2019-07-30 at 10 35 46 PM

This adds basic group functionality to the Rock data connector package.

None

### How do I test this PR?

Test the following API calls

- [ ] `leader` field on the `Group` node
- [ ] `members` field on the `Group` node
- [ ] `families` query
- [ ] `homeGroups` query
- [ ] `servingGroups` query
- [ ] `groups` field on a person
- [ ] `groupLead` field on a person

## TODO:

- [x] Closes [#676]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
@michael-riddering Like the direction of this PR. Have a few questions / thoughts on the schema. I'm going to talk to @conrad-vanl about how this should fit into core's plan and then I'll get back to you about the (hopefully amazing) future of this PR.

vinnyjth:
Your code here is great, I think the schema is pretty much 90% there (I might make a few small tweaks to it). 

Before we merge this though, @conrad-vanl and I need to chat for a bit about where this PR belongs. I love to have it in core, but that comes with the cost of maintaining (some really good) code that we don't actually use. This might belong in a separate package, that we would eventually merge into core.

packages/apollos-church-api/local.graphql
```diff
@@ -349,6 +358,10 @@ type Query {
   scriptures(query: String!): [Scripture]
   userPass: Pass
   campuses(location: CampusLocationInput): [Campus]
+  groups: [Group]
```
conrad-vanl: Does this return _all_ groups for the church? 😱

packages/apollos-church-api/local.graphql
```diff
@@ -349,6 +358,10 @@ type Query {
...
+  families(personId: String!): [Group]
```
conrad-vanl: I think this is a potential security flaw - the ability to find families for any arbitrary `personId`. Also, what is this `personId`? Is it an Apollos ID? A Rock ID? etc. Should be type `ID` not `String`.

Eitherway, would prefer a `families`, `homeGroups`, `servingGroups` attribute on the `Person` object, instead of additional root-level queries.

redreceipt:
Could I offer to just address the rightful concerns you've made and simplify the implementation to a minimal footprint?

conrad-vanl:
@redreceipt for sure! I think that would go along ways to giving us comfortability to merge. However, I'm still concerned about this effectively being 'dead code' on the core side, so I would still want to make sure that if we do merge this, this is something that Christ Fellowship would/might also use for whatever group functionality they're working on for MVP.

packages/apollos-church-api/local.graphql
redreceipt: I agree, this should just be the `currentUser` families

packages/apollos-church-api/local.graphql
redreceipt: Yea this one is probably unnecessary

redreceipt:
alright @conrad-vanl how about this, much less invasive

conrad-vanl:
@vinnyjth could you add this to your list to look at some time before the end of the sprint? We don't need to do any work here, just determine what we're comfortable exposing from a schema perspective. If we can align on a schema, we can either merge just the schema in to core, or the whole thing, but at this point we have both NewSpring and CF wanting to build groups functionality on their own and they're afraid that their schemas will conflict, which is valid. 

We will most likely start working on groups in core before the end of the year, so I'm okay getting some initial schema in core as long as it supports both CF's and NS's current use cases, and has a small footprint.

@calebpanza can you give a 👍 or 👎 on whether you support the schema that's defined in this PR, or would want it to look differently for CF's use case?

calebpanza:
@conrad-vanl @redreceipt I think this is a great start! I'd give it a 👍 

I can't say with 100% certainty how we will be using Groups for CF cause we haven't had those conversations yet about design or how exactly out Group Finder is set up. But this seems to hit the main points and will be super easy to extend!

richarddubay:
@vinnyjth Any chance this will make it into the next update?

vinnyjth:
@richarddubay I've got a few other items that are higher priority on my plate, but I'll try and find the time to take a look at this before the end of the sprint.

richarddubay:
@vinnyjth We'd really appreciate it. We just need to know one way or the other whether you guys actually want this or not. We've got two tasks that we need to do before we launch the app that have been waiting on this exact change for a couple months now. So this is a pretty big deal for us. Thanks for taking the time to look at it!

packages/apollos-church-api/local.graphql
```diff
@@ -337,6 +344,11 @@ type Person implements Node {
   birthDate: String
   photo: ImageMediaSource
   campus: Campus
+  groups: [Group]
```
vinnyjth: How likely is it that a person would belong to enough groups that the request should be paginated?

packages/apollos-church-api/local.graphql
```diff
@@ -337,6 +344,11 @@ type Person implements Node {
...
+  groupsLead: [Group]
+  families: [Group]
+  homeGroups: [Group]
+  servingGroups: [Group]
```
vinnyjth: I'm not familiar with the meanings behind "homeGroups" or "servingGroups". I'm assuming normal rock developers would be?

packages/apollos-church-api/local.graphql
redreceipt: average person, 2 to 3

packages/apollos-church-api/local.graphql
redreceipt: serving group, yes. I think the rock term for home groups in "community" groups so we could change it to that? serving group is a group in rock with the purpose set to "Serving". it's a group of people that volunteer together. Community Groups in Rock are groups that get together outside of the context of the church building is the best way to describe it I guess.

redreceipt: what do you think @asleepingpanda

vinnyjth:
This all looks good from my perspective. That said, I have an alternative schema API to suggest that may be more broadly applicable. 

What do you think about replacing the `servingGroups `, `homeGroups`, and maybe the  `groupsLead` and `families` attributes with a field on a Person called `groups` that had a "type" argument, that was an enum. Then you could query like this: 

````
{
  firstName
  lastName
  group(type: FAMILY) {
   id
   members { ... }
  }
}
```

I have a feeling that would be more flexible on the schema side, at the cost of being slightly more complex to implement on the server side (I can walk you through how I've designed resolvers like this in the past if you struggle with it).

redreceipt:
I understand how it would be implemented, it's just a little confusing as the dev writing the client query because "type" here can mean multiple things.

redreceipt:
we could even do a `groupsLead(type:ENUM)` for even more functionality

redreceipt:
orrrr a 2nd arg like this `groups(type: ENUM, lead: true)`

redreceipt:
it's way more explicit as what I am querying

redreceipt:
If you're trying to achieve as few fields on a person as possible, I like my suggestion with two args the best, that way you could pull "servings groups that I lead" or any combo.

vinnyjth:
> If you're trying to achieve as few fields on a person as possible, I like my suggestion with two args the best, that way you could pull "servings groups that I lead" or any combo.

I LOVE your suggestion of two args. I don't understand the group hierarchy as well as you do, so if that makes sense I would aim for that approach.

redreceipt:
@vinnyjth ok give it a look, I think it's ready. I love this implementation, I'm glad you suggested it. I refactored a good bit and sped up most the calls by a lot but the boat anchor is still resolving each group leader in an array of groups. Give the code a once over and see if you spot any enhancements before we merge.

packages/apollos-data-connector-rock/src/groups/data-source.js
```diff
@@ -0,0 +1,59 @@
...
+      .filter(
+        `PersonId eq ${personId} ${
+          asLeader ? ' and GroupRole/IsLeader eq true' : ''
+        }`
+      )
+      .get();
+    const groups = await Promise.all(
+      groupAssociations.map(({ groupId }) => this.getFromId(groupId, true))
+    );
+    return groups.filter(({ groupTypeId }) =>
```
vinnyjth: Would it be better to do this filtering with Odata? I'm sure there's a reason why you're not, maybe you can explain it to me?

packages/apollos-data-connector-rock/src/groups/data-source.js
```diff
@@ -0,0 +1,59 @@
...
+
+  expanded = true;
+
+  groupTypeMap = {
+    Serving: ROCK_MAPPINGS.SERVING_GROUP_TYPE_ID,
+    Community: ROCK_MAPPINGS.COMMUNITY_GROUP_TYPE_ID,
+    Family: ROCK_MAPPINGS.FAMILY_GROUP_TYPE_ID,
+  };
+
+  getFromId = (id, activeOnly = false) =>
```
vinnyjth: Looking good, but let's ditch the positional arguments :) use `{ id, activeOnly }` please! 
 
Reasoning here: https://exploringjs.com/impatient-js/ch_callables.html#named-parameters

I'm trying to be better about making sure we do this across the whole code base. Important for public facing API, not as important in tests/internal API.

packages/apollos-data-connector-rock/src/groups/data-source.js
redreceipt: Unfortunately due to the structure of the database, we have to make a call for each group that's tied to a `GroupMember` individually anyway (see the `Promise.all` above). I hate the number of calls that's necessary but I couldn't figure out a way around it.

redreceipt:
no more positional args. one day I'll catch up to modern programming :)

vinnyjth:
@redreceipt Might need to fix a linter error before I can merge

Issue #852, opened 2019-07-09
[UI] Refactor ContentCard component


burkeshartsis:
### User Story

As a user I want a personalized home feed experience that allows me to engage with content I find relevant to me. 

### Description
Refactor the current `ContentCard` into the simplest combination of props for a given view. This likely would result in two components like a `ContentCard` and a `TileCard`.

### Why
This is meant to simplify the `ContentCard` component api to be easier to reason and more straight forward to use representing a single view with a clear, intentionally limited purpose. 

### Other Resources

* [Basecamp todo](https://3.basecamp.com/3926363/buckets/6947926/todos/1898958850)

PR #967, opened 2019-08-01
Adds HorizontalContentCardConnected and Component prop to ContentCardConnected

burkeshartsis:
## DESCRIPTION

This adds a `Component` prop to `ContentCardConnected` and changes `ContentCardConnected`s default rendered child to `DefaultCard`. Additionally, it adds a `HorizontalContentCardConnected` component. Lastly, we add some missing exports for some of the newer ui-kit cards 🙃 

This makes `ContentCardConnected` way more flexible by letting us pass in other cards to the same query. All of our news cards share the same same basic API so this is really easy 💥

Our new card designs also need to be rendered at the right time. This adds a mapper HOC that selects the right card for the right place.

TONS.

1. This PR breaks things. In particular horizontal card rendering is pretty broken. 
2. We also no longer render 1 card in two ways (vertical vs horizontal scroll). Instead we have separate components for that and there is fair amount of legacy `tile` props still floating around that probably could be refactored.
3. Merging this means committing otherwise broken snapshots to 😢
4. The new cards require images and in a lot of cases titles so there are a bunch of prop warnings.
5. no tests and no stories 😭

### How do I test this PR?
Check that the app now renders the new card styles in the feeds. This PR could use stories for `ContentCardConnected` and `HorizontalContentCardConnected` but they didn't exist before this so...

This PR could use tests for `ContentCardConnected` and `HorizontalContentCardConnected` but they didn't exist before this so...

## TODO:

- [x] Partial refactor work associated with #852 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apolloschurchapp/src/ui/ContentCardConnected/contentCardComponentMapper.js
```diff
@@ -0,0 +1,32 @@
...
+import { DefaultCard, HighlightCard } from '@apollosproject/ui-kit';
+
+const contentCardComponentMapper = (props) => {
+  // check if we have a custom Component prop to use or we'll use the default Component prop.
+  let ComponentToRender;
+
+  // map typename to the the card we want to render.
+  switch (get(props, '__typename')) {
+    case 'MediaContentItem':
+      ComponentToRender = HighlightCard;
```
vinnyjth: You can condense this `switch` statement by removing all but the bottom assignment :) I would also recommend `return`ing inside the case, I find it's easier to parse a switch statement that way vs assignment. Also then you don't need to `break`!

packages/apolloschurchapp/src/ui/HorizontalContentCardConnected/index.js
```diff
@@ -0,0 +1,29 @@
...
+    />
+  )
+);
+
+HorizontalContentCardConnected.propTypes = {
+  Component: PropTypes.func,
+  isLoading: PropTypes.bool,
+};
+
+HorizontalContentCardConnected.defaultProps = {
```
vinnyjth: 👍

burkeshartsis:
@vinnyjth what if instead of defaulting the `Component` prop on `ContentCardConnected` to `contentCardComponentMapper` we exported `contentCardComponentMapper` and passed it in as the `Component` prop when using `ContentCardConnected` that way `ContentCardConnected can always just render a single component and instead we are just extending that functionality via this other function when needed. Granted I’m not sure when we wouldn’t need it but… thoughts?

vinnyjth:
I'm fine the way we are doing it now. Seems the most "sane"

vinnyjth:
Huge step in the right direction. Awesome simplification. Enhanced modularity. I've approved this PR once and I'll do it again!

Issue #974, opened 2019-08-05
Add Sharable to Features


redreceipt:
### User Story

As a user, I want to be able to share every type of feature with the same interface.

### Description

This is would extend `Feature` in the API with `Sharable` the same was that we are doing with content so the sharable message can be resolved and the algorithm to determine what will be shared lives in the API instead of client-side on each `BlahFeature` component.

### Why

Simpler to iterate over features and put the same thing in the `message` prop of the `Share.share()` function.

### Acceptance Criteria

* [ ] It should extend `Feature` to have `sharing.message` as a field.

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

PR #976, opened 2019-08-06
Adds Sharable Features

redreceipt:
## DESCRIPTION

Screen Shot 2019-08-05 at 8 32 42 PM

Adds `sharing` onto the features from the API

None.

### How do I test this PR?

Share a feature from the app

## TODO:

- [x] Closes [#974]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apolloschurchapp/src/content-single/Features/TextFeature.js
```diff
@@ -4,14 +4,15 @@ import PropTypes from 'prop-types';
 import { ActionCard, BodyText } from '@apollosproject/ui-kit';
 import ShareButton from 'apolloschurchapp/src/ui/ShareButton';
 
-const TextFeature = ({ body, contentId }) => (
-  <ActionCard action={<ShareButton message={body} itemId={contentId} />}>
+const TextFeature = ({ body, sharing: { message } = {}, contentId }) => (
+  <ActionCard action={<ShareButton message={message} itemId={contentId} />}>
```
vinnyjth: Fantastic!

packages/apollos-data-schema/src/index.js
```diff
@@ -397,22 +397,42 @@ export const contentChannelSchema = gql`
...
     message: String
     title: String
+    url: String
   }
 
   ${extendForEachContentItemType(`
     sharing: SharableContentItem
 `)}
+
+  type SharableFeature implements Sharable {
```
vinnyjth: 🙌

packages/apollos-data-schema/src/index.js
```diff
@@ -397,22 +397,42 @@ export const contentChannelSchema = gql`
...
+    message: String
+    title: String
+  }
+
+  extend interface Feature {
+    sharing: SharableFeature
```
vinnyjth: We might not want to put this on the interface, because some features (like the action list feature) don't support sharing.

packages/apollos-data-connector-rock/src/features/data-source.js
```diff
@@ -123,4 +123,24 @@ Make sure you structure your algorithm entry as \`{ type: 'CONTENT_CHANNEL', aru
       action: 'READ_CONTENT',
     }));
   }
+
```
vinnyjth: I like the implementation of this method, but I think it might be hard to update when developers create new features. Could you move this functionality into the `createTextFeature` and the createScriptureFeature` methods? Then you won't even need a resolver for it :)

packages/apollos-data-schema/src/index.js
redreceipt: done

packages/apollos-data-connector-rock/src/features/data-source.js
vinnyjth: In that case, write a resolver specifically for `TextFeature` or `ScriptureFeature` sharing. I think you're right about embedded data like that being confusing, but I think that those "catch all" methods that determine what data to return based on a content type are difficult to remember to extend. 

eg 
```
resolvers = {
  TextFeature: {
    sharing: (root) => root.body // or whatever
  }
}
```

redreceipt: love it. done. 

also, why am I adding `__type: "SharableFeature` to the resolver as well? I took it from how you did `SharableContentItem` but I don't know why it's necessary or if it even is.

vinnyjth: It's because of this: https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-data-connector-rock/src/sharable/index.js#L6

You might need to change `__type` to `__typename`

vinnyjth: I think I might not have done things 100% correctly for `__type` with `ShareableContentItem` doesn't look like we run into issues though because of how we use the interface currently. (we never have to figure out if something is a ShareableContentItem or a ShareableXYZ

redreceipt: well, we've been just passing `__typename` down from `...root` so it had what it needed anyway. I'll put in a PR to fix the other in a separate PR.

redreceipt: yea looks like we don't need them, `__typename` is getting resolved correctly from the root interface resolver. Here it is working without them.

Screen Shot 2019-08-06 at 2 16 29 PM

redreceipt: I snuck in the changes to `SharableContentItem` :)

packages/apolloschurchapp/src/content-single/Features/getContentItemFeatures.js
```diff
@@ -18,11 +18,17 @@ export default gql`
...
         features {
+          sharing {
+            message
+          }
           ...FeaturesFragment
         }
       }
       ... on WeekendContentItem {
         features {
+          sharing {
```
vinnyjth: You might need to move these inside the features fragment. I'm going to spin up this PR locally and if it's broken because of this I will fix and push up the changes.

redreceipt:
you marked this "merge" but didn't give an approval

vinnyjth:
silly me

redreceipt:
schema diff worked! it got caught on breaking changes 😄 should i go back and add `@deprecation` to `Sharable.url`?

PR #983, opened 2019-08-08
Adds clarity around current position in devotional list

### LINKED ISSUE #958: Devotional list at bottom of devotionals is confusing
### Details

When you are on the first devotional, and navigate to the next devotional, the number and sort changes at the bottom of the devotional. This is due to us removing the current devotional from the list of all devotionals, and using the index of the devotional in the array to determine it's "number" rather than any real logic. 

Open to various solutions for this issue. 

### Possible Fix

### Your Environment

- Link to your project:

### Steps to Reproduce

1.  first...
2.
3.
4.

### Stack Trace

</details>

vinnyjth:

This PR is a 80/20 approach to bring clarity to a user's position in a list of devotional items. 

I've done two things to accomplish this: 

1. I've made sure that the current devotional you are on displays in the list of devotionals. That way the number ordering should always make sense. 

2. Added support for scrolling to the current item in the horizontal list, if the current item is found in the horizontal list. 

Did a really ugly hack, copied over the width of the `TileCard` into the `HorizontalTileFeed` as a constant. @burkeshartsis or @conrad-vanl let me know if you know of a better solution. 

### How do I test this PR?

1. Open the app.
2. Visit a devotional with multiple sbiliing content items. 
3. Click to the next content item at the bottom. The ordering should make sense and the new item should load with the scroll moved to the current item.

## TODO:

- [x] Closes #958 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apolloschurchapp/src/content-single/HorizontalContentFeed/index.js
```diff
@@ -20,6 +21,10 @@ const loadingStateObject = {
   },
 };
 
+// Hardcoded value taken from CardTile. Expect to change this if you change CardTile.
+// TODO: wish we had a better solution for getting widths of things...
+const CARD_WIDTH = Dimensions.get('window').width * 0.66 + 16;
```
redreceipt: we could create a context at the top and pass that into the different components, including `CardTile`

vinnyjth: What would be in the context? We need the dimensions **from** CardTile, context might be nice if we wanted the dimensions **for** CardTile.

redreceipt:
Works as is. But since this isn't critical, a couple of things I think would make this PR really great:

1. The current devotional should be marked and disabled as a touchable because of INCEPTION
2. If a given devotional exists anywhere in the stack, it should be popped out of the old position when you click it. So at most the nav stack could only ever be `devotions.length` deep. Said another way, clicking 1 -> 2 -> 1 should leave you with a nav stack of [2, 1].

vinnyjth:
> 1. The current devotional should be marked and disabled as a touchable because of INCEPTION
This will probably be followed up in a future PR. @michael-riddering  is working on all new designs for this flow. 

>If a given devotional exists anywhere in the stack, it should be popped out of the old position when you click it. So at most the nav stack could only ever be devotions.length deep. Said another way, clicking 1 -> 2 -> 1 should leave you with a nav stack of [2, 1].

This will require significant work when manipulating the nav stack. I'm not even sure it's possible, because React Navigation doesn't have an `action` that I can see that `shifts` off a route that isn't the current route. I would want to see that this is a pain user's experience before I spend time trying to solve for this use case (although I agree it _could_ be frustrating)

packages/apolloschurchapp/src/content-single/HorizontalContentFeed/index.js
redreceipt: the `CardTile` width is [static](https://github.com/ApollosProject/apollos-prototype/blob/df4669f61af1945f758d78eb04428e2f31a90abe/packages/apollos-ui-kit/src/CardTile/index.js#L34) currently since the `window` width would never change. so we could create a context called `CardWidth.Provider` and pass that into `CardTile` and `HorizontalContentFeed` as props, right?

redreceipt: and then `useContext` so we don't need a `Consumer` :)

redreceipt: I just feel like context is React's answer to global variables which is exactly what you need here

redreceipt:
> > 1. The current devotional should be marked and disabled as a touchable because of INCEPTION
> 
> This will probably be followed up in a future PR. @michael-riddering is working on all new designs for this flow.
> 
> > If a given devotional exists anywhere in the stack, it should be popped out of the old position when you click it. So at most the nav stack could only ever be devotions.length deep. Said another way, clicking 1 -> 2 -> 1 should leave you with a nav stack of [2, 1].
> 
> This will require significant work when manipulating the nav stack. I'm not even sure it's possible, because React Navigation doesn't have an `action` that I can see that `shifts` off a route that isn't the current route. I would want to see that this is a pain user's experience before I spend time trying to solve for this use case (although I agree it _could_ be frustrating)

1. Awesome!

2. Hear me out here, I think it could be done rather easily using `navigation.reset` which wipes the stack and replaces it with a list of actions. We keep track of our own stack in a list we're passing around and then use `navigation.reset` to match our manipulated list of numbers. So have a `mockStack` variable that we pass around and then use `reset` to match that one when you hit the back button.

redreceipt:
... as long as that function only wipes the current nav and not *all* of them

redreceipt:
Screen Shot 2019-08-08 at 4 01 26 PM

packages/apolloschurchapp/src/content-single/HorizontalContentFeed/index.js
vinnyjth: I think that adds two much complexity for the goal of this PR. We'll end up creating a whole new paradigm to satisfy the requirements to render a single flat list :P. Now we have to answer questions like "should HorizontalHighlightCard have a width provider" "What if I want to get the height of a FeaturedCard"? I would rather hardcode the value for now, or come up with some "low key" way to make sure this doesn't break if we change the width of `CardTile`.

vinnyjth:
@redreceipt You are welcome to take a stab at implementing it. It certainly might work well, because RN batches navigations so it should show up as a single navigation. Unless @conrad-vanl or @burkeshartsis say otherwise though, I don't think that adding that feature is blocking for this PR. I don't think it's a user's expectation that their navigations are "de duped"

packages/apolloschurchapp/src/content-single/HorizontalContentFeed/index.js
redreceipt: Word. Could you just export the width out of the `CardTile` from the ui-kit?

`export default CardTile
export CardTileWidth`

vinnyjth: >Word. Could you just export the width out of the CardTile from the ui-kit?

I think that's what I'll end up doing ;) I'm waiting to hear back from @conrad-vanl or @burkeshartsis as well about what they think, because I know they have had to solve this problem before.

redreceipt:
> @redreceipt You are welcome to take a stab at implementing it. It certainly might work well, because RN batches navigations so it should show up as a single navigation. Unless @conrad-vanl or @burkeshartsis say otherwise though, I don't think that adding that feature is blocking for this PR. I don't think it's a user's expectation that their navigations are "de duped"

Cool. actually, the more I think about it, now that the tile list is not confusing, it might be better to not even have a back button. But I agree we can solve that if it becomes a pain point later.

packages/apolloschurchapp/src/content-single/HorizontalContentFeed/index.js
conrad-vanl: we might just consider moving this into theme...if possible

burkeshartsis: all of our "tile" sized cards are just 240x240 now. In fact, I already updated `HorizontalTileFeed` to snap at those points. Why don't we just hard code these and move on. If we want these "tiles" to match our other`HorizontalSomethingSomethingCard`s we'll need to do this anyways.

A theme value is totally an option I'm just not sure how needed of an option it is.

vinnyjth: Thanks gentlemen, I'm going to hard code this in the CardTile file, and we can decide if we want to put this in the theme when we refactor these cards with @michael-riddering's new designs.

Issue #913, opened 2019-07-23
Skip Onboarding


redreceipt:
### User Story

Currently, you have to go through onboarding everytime you get signed out of the app.  Would love to just be able to sign in after verifying the phone number if I've already gone through onboarding

### Description

skips onboarding if you already have all the pre-filled information and permissions checks.

### Why

Faster and better UX.

### Acceptance Criteria

* [ ] It should skip right to the home screen after login

### Possible Implementation & Open Questions

query for all the info and permissions checks up front and skip onboarding on login.

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

vinnyjth:
@redreceipt Can you take a stab at this? We should set a value in local storage that represents if you have done onboarding, and then use that value to direct a user if they login. 

That value should reflect the userId of the user who has done onboarding. So if Jane on Bob's phone signs up, she she do onboarding, even though Bob has already done onboarding.

redreceipt:
then if I get a new phone, I've got to do onboarding again. Shouldn't we just pre-query everything and check permissions and skip if everything is good to go? it'll make the data pop in faster on each screen anyway.

vinnyjth:
New phone means push and location permissions have not been granted. We decided that we wanted new phones to go through onboarding again, for the time being.

redreceipt:
gotcha, cool. probably simpler engineering anyway :)

redreceipt:
I got going on this but the more I think about it, I'm a little unclear. My intention for this issue was to pre-check all the data and permissions and if someone has all that, to skip onboarding. 

I'm not sure what storing a boolean in local storage would do. Currently, if a user is logged in and they've already done onboarding, it already skips ahead. And if someone logs out, it would reset the store anyway so they'd always go back through.

vinnyjth:
> I'm not sure what storing a boolean in local storage would do. Currently, if a user is logged in and they've already done onboarding, it already skips ahead. And if someone logs out, it would reset the store anyway so they'd always go back through.

This might help a bit - I was thinking of storing that flag in `async-storage` local storage, not Apollo's cache :)

redreceipt:
haha yes, that does help. Do you see a problem with just keeping a growing list of `usersCompletedOnboarding`?

vinnyjth:
Nope! I imagine it will rarely be more than 1 user long.

PR #986, opened 2019-08-11
Skips Onboarding When Unnecessary

redreceipt:
## DESCRIPTION

Screen Shot 2019-08-13 at 4 57 48 PM

This will keep track of which users have completed onboarding and skip it the next time they log in on the device.

- Made some changes to some redundant and confusing props in the onboarding connected components.
- Had to add `removeClippedSubviews={false}` to the swiper: https://github.com/leecade/react-native-swiper/issues/416

### How do I test this PR?

Go through onboarding, log out, and log back in.

## TODO:

- [x] Closes [#913]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
@vinnyjth I think I'm close here, how do you recommend I get the `currentUser` here? `Onboarding` is the first component loaded in the app so there's no parent to run a query and pass in as a prop. And it's not in cache yet at this point.

vinnyjth:
@redreceipt One pattern we have done for this in the past, is have a component that renders as a part of onboarding (but doesn't actually render any UI) that uses an apollo query to fetch current user, and then acts on that response. You could create a function called something like `checkIfOnboardingCompleted` that calls `client.query`. You will need to use an `ApolloConsumer`to get the apollo client to pass in, if you do that.

redreceipt:
so....close, currently I'm triggering the `getUser` query on the button pushes but that means that I have to push the first "Next" button before the query is run, is there a way I can run the query on mount? The kicker is I need the `client` from somewhere.

redreceipt:
i'm gonna make a parent `OnboardingConnected` and pass in the client that way

redreceipt:
This one works but I don't think it's 100% there. The first screen hangs until the `currentUser` comes back and then skips once it's got it. And I had to add `removeClippedSubviews={false}` to the `Swiper` because the first screen was not displaying content until you started to swipe it. Very strange. Documented here: https://github.com/leecade/react-native-swiper/issues/416, not really sure what changed and why this is necessary now. 

Hoping someone will have some great ideas about how to push this one over the edge.

redreceipt:
@conrad-vanl @vinnyjth I've thought a little more about this one. If this goal is that they would still have to go through it when they have a new device (which makes sense so we can capture notifications) then the only thing this is getting us is when multiple people are using the same device. Very small edge case for such a complex solution. I say we kill this.

vinnyjth:
>  If this goal is that they would still have to go through it when they have a new device (which makes sense so we can capture notifications) then the only thing this is getting us is when multiple people are using the same device. Very small edge case for such a complex solution. I say we kill this.

Totally agree. I think we can take the core concepts from this PR and refine it to a situation where once a user completes (or skips 🤷‍♂) onboarding, we don't show it to them or any other user again on the same device.

redreceipt:
I'm a little confused, are we killing this or changing the design? I think what you've described is what this already does but I just don't if it's really getting us that much.

vinnyjth:
@redreceipt This PR keeps track of what stage a user completed in onboarding as well as what user ids have completed onboarding on the device. I think we can ignore both of those constraints and end up with some less but better code.

redreceipt:
What's the high-level goal?

vinnyjth:
A user shouldn't have to go through onboarding every time they login on their own device.

redreceipt:
Are we ok with not asking about permissions again if they skipped it?

vinnyjth:
Yep! But @conrad-vanl can override me if I'm wrong

redreceipt:
That's the piece I'm a little worried about because there's currently no other place in the app to enable those permissions if they skipped them the first time.

burkeshartsis:
Jumping in here to provide some UX guidance as this PR have become pretty stale. If a user goes through onboarding and skips a slide without clicking any of the CTA's that prompt for permission we should provide a way for them to come back and enable them. However, if a user **denies** a permission we loose the opportunity to ever prompt again on iOS and as a result should assume the same on Android.

There are lots of ways we could prompt for permissions. Here are some below:

1. check and prompt in place when a permission is needed. This wouldn't take a lot of work but would require some changes around the app. Code is super straight forward. Check for permission > open modal with the correct onboarding slide 💥This is actually a better UX then having an onboarding at all.
2. Add a link to the setting page to "Setup your device" which sends them to onboarding. This is easy to implement but doesn't account for if they have not granted a permission or if they have denied it. Some components to handle permission checking and a component to show would be a welcome PR at some point.

@kevin-birdy you are welcome to weigh in here if you like.

kevin-birdy:
FWIW I've always been 100% behind that line of thinking @burkeshartsis — it's a bit more of a longterm goal, but I would definitely vote to attack the problem via well-timed prompts within the actual experience of the app than go overkill in this onboarding PR.

PR #997, opened 2019-08-14
Allows Username Login

redreceipt:
## DESCRIPTION

Screen Shot 2019-08-14 at 11 01 17 AM

This allows a user to login with a non-email string.

Navigating to `Auth` via `navigate` instead of `push`, seems to be fine. 🤷‍♂️

### How do I test this PR?

Change the value to false in the `NavigationService` and make sure you can login via username.

None.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
@redreceipt I hate to hold this up further, but I've got a change I would like to suggest. 

Rather than passing down an `emailRequired` prop, have you considered passing down the `yup` `schema` instead? 

https://github.com/ApollosProject/apollos-prototype/pull/997/files#diff-41184c938d6428c4a3fd38e40fd65ae4R12

That would give churches the flexibility to write exactly the requirements they desire for email login.

redreceipt:
I don't hate it but I just don't think it's necessary. It's a more complicated API to understand than a simple boolean (could even be an environment variable) and I don't see many other use cases other than "email or not".

redreceipt:
Would love to prove that out as a pain point instead of over-engineering. It's an easy add down the road.

vinnyjth:
Fine by me. Let's pair on a better way to pass down that "emailRequired" prop.

PR #1017, opened 2019-08-21
Adds Search input component

burkeshartsis:
This adds a `Search` input to `UI-KIT`!

It's surprisingly hard to build a search input that is a faithful i̶o̶s̶ ̶r̶i̶p̶e̶ ̶o̶f̶f̶ representation of our designs. I think this is pretty darn close. Only caveat is that it has to use the same background color as the component it is rendered in. That means, you can't render it over an image and it may not look great over all gradients.

### How do I test this PR?
Check the stories!

I have basic snapshot coverage but there is a lot of complexity to this component that is hard to write tests over because it's so interactive.

## TODO:

- [x] Closes 🤷‍♂️
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
@redreceipt i'll be doing a deep-dive code review but if you don't mind running this through your QA testing chops so we have a second set of eyes, that'd be great!

packages/apollos-ui-kit/src/inputs/Search/styles.js
```diff
@@ -0,0 +1,105 @@
...
+)(View);
+
+// We do some memo gymnastics here so that this icon isn't effected by rerenders caused by input state
+const LoopIcon = withTheme(({ theme, isFocused }) => ({
+  fill: isFocused ? theme.colors.action.secondary : theme.colors.text.tertiary,
+  size: theme.helpers.rem(1),
+  style: {
+    marginLeft: theme.sizing.baseUnit,
+  },
+}))(memo((props) => <Icon {...props} />)); // eslint-disable-line react/display-name
```
conrad-vanl: This is so close to being a code smell...

packages/apollos-ui-kit/src/inputs/Search/styles.js
```diff
@@ -0,0 +1,105 @@
...
+const Input = withTheme(({ theme, isFocused, cancelButtonOffset }) => ({
+  placeholderTextColor: theme.colors.text.tertiary,
+  selectionColor: theme.colors.action.secondary,
+  style: {
+    flexGrow: 1, // fixes weird text behind icon (ios) and placeholder clipping (android) bugs
+    height: theme.helpers.rem(2.5), // we have to have a height to make this display correctly. using typographic unit to scale with text size.
+    paddingVertical: 0, // removes weird "default" padding
+    paddingLeft: theme.sizing.baseUnit * 0.5,
+    paddingRight: isFocused // we have to dynamically adjust the padding otherwise it causes the placeholder text to disappear
+      ? cancelButtonOffset + theme.sizing.baseUnit * 5
```
conrad-vanl: where does `theme.sizing.baseUnit * 5` come from? In other words, it is `* 5` instead of `* 4` or something else?

packages/apollos-ui-kit/src/inputs/Search/styles.js
```diff
@@ -0,0 +1,105 @@
...
+      : 0, // `CancelButton` + padding + `ClearSearchButton` + a 🧙‍24px 🤷‍
+    fontSize: theme.helpers.rem(0.875),
```
conrad-vanl: Should we make this a `styled` component so that we can assign it a name and let it be overriden with `overrides`?

packages/apollos-ui-kit/src/inputs/Search/styles.js
```diff
@@ -0,0 +1,105 @@
...
+}))(({ forwardedRef, ...props }) => (
+  <TextInput ref={forwardedRef} {...props} />
+));
+
+// the main reason this component lives here is because we need access to theme colors
+const SmokeAndMirrorsWrapper = styled(
+  ({ theme, screenBackgroundColor }) => ({
+    flexDirection: 'row',
+    alignItems: 'center',
+    // backgroundColor: 'salmon',
```
conrad-vanl: Lets clean up the testing comments

packages/apollos-ui-kit/src/inputs/Search/styles.js
```diff
@@ -0,0 +1,105 @@
...
+  iconPadding: theme.helpers.rem(0.75),
+  style: {
+    opacity: isVisible ? 1 : 0,
+  },
+}))(ButtonIcon);
+
+const CancelButton = styled(
+  ({ theme }) => ({
+    paddingLeft: theme.sizing.baseUnit, // padding away from end of search field
+    color: theme.colors.text.link, // we use UIText here instead of `ButtonLink` becuase onLayout has issues with nested text on Android
```
conrad-vanl: buuuuuut `ButtonLink` is just a styled `UIText` component?

burkeshartsis: [`ButtonLink`](https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-ui-kit/src/Button/ButtonLink.js#L1) doesn't use `UIText` it uses React Native's `Text`

packages/apollos-ui-kit/src/inputs/Search/styles.js
burkeshartsis: We could but I'd still need `withTheme` to pass in to `placeholderTextColor` and `selectionColor`

packages/apollos-ui-kit/src/inputs/Search/styles.js
burkeshartsis: `5` is a semi-magic number. The comment on line 49 attempts to address this. Basically, it breaks down like this.

24px for the "X" button +
16px padding between search bar and "Cancel" +
16px padding between "Cancel" and the edge of the screen +
56px of magic.

I think there is some stuff we could fix here though because the `CancelButtonWidth` should have been the "magic" number.

packages/apollos-ui-kit/src/inputs/Search/styles.js
burkeshartsis: lol. It was that or make an entire component just to memoize the icon.

packages/apollos-ui-kit/src/inputs/Search/styles.js
burkeshartsis: Fixed in cf444e4

packages/apollos-ui-kit/src/inputs/Search/styles.js
burkeshartsis: I agree that being able to override these styles would be nice. I feel that way anytime I pass a `style` prop via `withTheme` but to do so, currently, is cumbersome unless we use `compose()`

packages/apollos-ui-kit/src/inputs/Search/styles.js
burkeshartsis: Ok, I looked at how we could remove this weird magic number and it's possible by changing where we run `onLayout` and how we animate. Totally do able. The only difference is it moves the "magic" number above to a different known value. Specifically, it makes the new value we have to account for:

- 24px for the "X" button +
- 16px padding between search bar and "Cancel"/Edge of screen (depending on `isFocused` state)

The positive here is it's an easy number to calculate and understand as opposed to the seemingly arbitrary number above. The down side is that we would have to get creative with how we pass that value into the places we need it because it's `theme` derived.

conrad-vanl:
@burkeshartsis which of my comments above do you feel are worth addressing? I think this is great, works wonderfully, most of the code is solid. Higher quality then most of what gets approved, so I'm fine moving forward with it as-is.

burkeshartsis:
If you want to continue to support the use of recompose via `compose` I'm happy to address that. We have icons all over that app that gets styles via `withTheme` so it would be come a sort of pattern.

As for the magic number conversation, I'm not sure there is a ton of value in trying to eliminate it at the current juncture. It's well documented and we can address it later time permitting. I left comments above on how we might do that too.

Just let me know about the styles via `withTheme` or merge it!

PR #1037, opened 2019-09-03
Algolia Search

conrad-vanl:
## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
@vinnyjth when you get a chance could you review the schema changes in this [commit](https://github.com/ApollosProject/apollos-prototype/pull/1037/commits/e9510bd8d2755ee5a7923cfc2283635b0b55558d) and provide feedback? You can ignore everything else in this PR for now...

Going to hook up resolvers and dataSource to search the Algolia API then this can be reviewed/merged.

packages/apollos-data-schema/src/index.js
```diff
@@ -397,6 +397,32 @@ export const contentChannelSchema = gql`
...
+    first: Int
+    after: String
+  }
+
+  type SearchHitsConnectionEdge {
+    node: Searchable
+    cursor: String
+  }
+
+  interface Searchable {
```
vinnyjth: Fantastic use of `Node`. However, I'm wondering if a `union` might be a better fit. What if we want `people` to be searchable? With a union we wouldn't have to implement `id, title, summary, coverImage`. Downside is that the client would have to be more flexible.

packages/apollos-data-schema/src/index.js
```diff
@@ -397,6 +397,32 @@ export const contentChannelSchema = gql`
...
 
+export const searchSchema = gql`
+  extend type Query {
+    search(query: String!, first: Int, after: String) {
+      edges: [SearchHitsConnectionEdge]
+      pageInfo: PaginationInfo
+    }
+  }
+
+  input ContentItemsConnectionInput {
```
vinnyjth: Where is this used?

packages/apollos-data-schema/src/index.js
vinnyjth: I might even make `node` have the type of `Node` :)

packages/apollos-data-schema/src/index.js
conrad-vanl: it was a mistake

packages/apollos-data-schema/src/index.js
conrad-vanl: Good questions. So @vinnyjth you might know more than me here with Algolia, but I was trying to keep a consistent "Search" interface that we use to index with Algolia so we can keep these attributes consistent on the Algolia side.

I'm half tempted to create a `SearchResult` type and remove the `Searchable` interface, which would look something like:

```
type SearchResult {
  title: String
  summary: String
  coverImage: Image
  relatedNode: Node
}
```

conrad-vanl: FWIW, if we try to index all attributes on each content item with Algolia, Algolia complains from the objects being too large. I tried to index `contentHtml` and Algolia complained.

vinnyjth: I like this idea better 😬I think it's a little more complex, but it means you can rely 100% on Algolia for data in many cases (a HUGE win for instant-ish search) and the indexing would be consistent.

conrad-vanl:
@vinnyjth fixed the schema in last commit

packages/apollos-data-schema/src/index.js
conrad-vanl: Done. The only hard part might be...our queries will most likely end up looking like this:

```
query {
  search({ query: "blah" }) {
    title
    summary
    coverImage { uri }
    relatedNode {
      id
    }
  }
}
```

Which will cause havoc right now because of the `relatedNode { id }` bit. The way those resolvers would work right now is I believe it would cause each result to query the Rock API to load the full related node....

conrad-vanl: we could just do like a `relatedNodeId` attribute but that's not very graphy

conrad-vanl: which brings us back to the same problem with just making the results be of type `Node` or something. We need a way to make our resolvers to be able to take a `root` input that might be incomplete, and allow for partial loading

vinnyjth: Hmmm - that's true. That's an optimization we could build into the resolvers I think, but it would very annoying. Any reason we have to return the relatedNode id though? for navigation I presume? I'mma think about our options here for a few moments.....

vinnyjth:
```
// proxytest.js
const almostNode = (finder) => ({
  loaded: false,
  loadedObj: {},
  get: async (obj, prop) => {
    if(this.loaded){
      return this.loadedObj[prop];
    }
    if(prop === 'id') {
      console.log('No need to load!')
      return obj['id']
    }
    console.log('Loading up!')
    const loadedObj = await finder(obj);
    this.loadedObj = loadedObj;
    this.loaded = true;
    return loadedObj[prop];
  }
})


async function run() {
  console.log('Running')
  const target = { id: '123' };

  const AlmostPerson = new Proxy(target, almostNode(() => Promise.resolve({ name: 'Nick' })));

  console.log(await AlmostPerson.id);
  console.log('Now attemping load');
  console.log(await AlmostPerson.name)
  console.log('And trying to access again')
  console.log(await AlmostPerson.name)
};

run();
``` 

There might be a better way to do this....

packages/apollos-data-schema/src/index.js
conrad-vanl: yeah for navigation

conrad-vanl:
@vinnyjth yeah I'm thinking through that as well. I'm thinking this would be a really nice feature in our resolvers is if they could dynamically choose whether to hit the server or not for data....

Where this branch is at right now is it's back to using `Node` as the return type for search results and any attribute that's not indexed is just returning as `null` / `undefined`. Next step I was going to start looking into how to hit Rock to load the other attributes only when requested.

conrad-vanl:
@vinnyjth I tried a Proxy experiment at the dataSource level - check out https://github.com/ApollosProject/apollos-prototype/pull/1037/commits/a819e81720f06dacb2f8bf89d3c4b731ffaf2128

While it technically works, it makes all of the attributes promises which sucks.

I'd love to make our resolver objects themselves `Proxy` objects, but the problem is that that means resolvers are no longer simple objects, which suck too. Womp womp.

Still 🤔

conrad-vanl:
Another idea: We go back to having a `type SearchResult` with a `relatedNode` attribute and just use the `@defer` directive on the client?

https://www.apollographql.com/docs/resources/graphql-glossary/#deferred-query

```
query {
  search({ query: "blah" }) {
    title
    summary
    coverImage { uri }
    relatedNode @defer {
      id
    }
  }
}
```

vinnyjth:
> While it technically works, it makes all of the attributes promises which sucks and breaks everything. 

I was really worried about this :/ `Apollos` handles this well, but all our resolver logic bombs on this.

conrad-vanl:
@vinnyjth so here's what I need your help/insight on with this PR, I think:

We need to decide the best schema to move forward with, and then we can optimize it later.

The other part to consider, which is another reason why the `Proxy` idea above doesn't work, is the data that I am currently indexing is not the data that's stored in Rock, rather the data the is the output of one of our feed queries. So the data that is indexed has already gone through our resolvers. If it gets pushed to our resolvers again, they'll just break everything.

We could change how the data is indexed, but we'll have the reverse of this problem, I think.

conrad-vanl:
There's two schemas to consider, I think:

Search results returning `Node`, which I started implementing in this commit - https://github.com/ApollosProject/apollos-prototype/commit/a819e81720f06dacb2f8bf89d3c4b731ffaf2128

Search results returning a `SearchResult` type with a `relatedNode` attribute, which is the current state of this PR (and currently works, albeit with a performance drop if you include the `relatedNode` in your query)

vinnyjth:
@conrad-vanl I like the `SearchResult` type the best. It gives the flexibility of being able to include things like `relevance` or some other search related metadata in the result. I agree that we should design for schema first, and figure out performance enhancement in the future (if needed).

conrad-vanl:
@vinnyjth things like `relevance` I think would traditionally be included on the `edge`? Because it describes the relationship between the query and the node.

```
search(query: "blah") {
   edges {
     node { ... }
     relevance
   }
}
```

conrad-vanl:
FWIW, I think I still _prefer_ the `node` just being the actual `ContentItem`, I'm just concerned about ability to overcome performance implications thereof.

vinnyjth:
I was imagining the `SearchResult` as the `edge`, I neglected to see the actual `edge` in your schema 😂You're totally right, in that case. 

I think that having both a `SearchResult` and an `Edge` might be too many layers. If you do an edge, I think the edge should have a `node: Node` OR the search query should be renamed searchContentItems and have `node: ContentItem`.

I really like the `SearchResult` type and the `relatedNode` type. I've started to do that in a few places across the app where we have similar relationships. I haven't run into a reason not to do it..... yet.

In terms of performance, I think there are some ideas we could run with that would add code complexity but improve performance. doing something like this https://stackoverflow.com/a/50483160 we could check to see if we are only looking for the `id` field, and just return the id in this case rather than fetching.

conrad-vanl:
@vinnyjth so should we move the search result `title`, `summary` and `image` into the `edge`? and keep the `node` just pointing at the `ContentItem`?

conrad-vanl:
@vinnyjth just pushed up schema change that I mentioned above. Thoughts?

vinnyjth:
I wrote a comment regarding your changes last week, but I must not have pushed send 😬 

I like the most out of all your iterations. I think we can solve the performance issues when we get there. Nice work 👍 💯

conrad-vanl:
Nice! I'll add tests and mark it ready for review asap

conrad-vanl:
@vinnyjth added tests, I think this should be good to go.

packages/apollos-data-connector-algolia-search/src/data-source.js
```diff
@@ -0,0 +1,35 @@
+import ApollosConfig from '@apollosproject/config';
```
vinnyjth: 😍 😍 😍 😍

vinnyjth: This file turned out so nice.

packages/apollos-data-connector-algolia-search/src/resolver.js
```diff
@@ -0,0 +1,18 @@
+import { withEdgePagination } from '@apollosproject/server-core';
+
+const resolver = {
+  Query: {
+    search: async (root, input, { dataSources }) =>
```
vinnyjth: I don't think you need `async` here? but not harm in having it.

packages/apollos-rock-apollo-data-source/src/cursor.js
```diff
@@ -1,27 +1 @@
...
-    decrypted += decipher.final('utf8');
-
-    return JSON.parse(decrypted);
-  } catch (e) {
-    throw new Error('Error parsing cursor');
-  }
-}
-
-export { createCursor, parseCursor };
+export { createCursor, parseCursor } from '@apollosproject/server-core';
```
vinnyjth: Smart thinking, totally approve of this change.

PR #1044, opened 2019-09-03
Adds search to Discover tab

burkeshartsis:
This refactors the Discover tab to add a `Search` input and feed.

I considered using [`keyboardDismissMode={'on-drag'}`](https://facebook.github.io/react-native/docs/scrollview.html#keyboarddismissmode) so that when a user searched and then scrolled it would dismiss the keyboard but it breaks the un-focus animation on the input being in sync with the keyboard dismissal. Only way to get this to work would be to use a [keyboard event listener](https://facebook.github.io/react-native/docs/keyboard#addlistener). I deemed this out of scope for this PR.

Also if you tap into the search field but don't type it errors on an empty search string. I think this would be better addressed on the api. Fixing this will allow us to use the designs for the empty search state.

### How do I test this PR?
open the app and try it out.

This has about as much snapshot coverage as you can get in our environment.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apolloschurchapp/src/tabs/discover/Discover.js
```diff
@@ -1,12 +1,48 @@
...
 import TileContentFeed from './TileContentFeed';
 import GET_CONTENT_CHANNELS from './getContentChannels';
 
+const HeaderBorder = styled(({ theme }) => ({
+  paddingBottom: 8,
+  /* It's unclear why this is necessary but without it the layout breaks on both platforms. Limited
+   * research suggest that without a background color the shadows don't know what to blend with so
+   * the view collapses. */
+  backgroundColor: theme.colors.background.paper,
+  // Renders the same shadows that React Navigation does.
```
burkeshartsis: I would love you thoughts on how we might better handle this @conrad-vanl? In some ways we aren't creating a custom header 👍 but now we have to maintain this shadow to always match 👎 which kind of feels like the same thing.

conrad-vanl: Meh. Let me think on this some....

conrad-vanl: Ideas:

1. Could we try getting the search bar to render inside of react-navigation's header? There's a [`headerTitle` prop](https://reactnavigation.org/docs/en/stack-navigator.html#headertitle) which appears that if you rendered a title element and the search bar inside of it, it would allow the header to grow (just by looking at the [source code)](https://github.com/react-navigation/stack/blob/master/src/views/Header/HeaderSegment.tsx#L336)). But that's dicey as bets.

2. Is there a design solution to consider, perhaps removing the search bar from the header and visually rendering it below it, perhaps not fixing it to the top of the screen either?

3. I think this is still better than rendering our own header because react-navigation's header does a lot for animations during scene transitions, the back and left labels, etc. So I think I'd be good with this as is.

burkeshartsis: After lots of conversation and creating an entire working branch for idea 1 above we have decided to go with idea 3 and stick with what we have.

vinnyjth:
Can't reproduce on Android, actually. If you can't reproduce locally let me know and I'll try again, and if I can't reproduce I'll approve this PR.

image

burkeshartsis:
Empty search result is now handled.

burkeshartsis:
@conrad-vanl @vinnyjth this is ready for review again

packages/apolloschurchapp/src/tabs/discover/SearchFeed/NoResults.js
```diff
@@ -0,0 +1,57 @@
...
+  'Discover.SearchFeed.NoResultsText'
+);
+
+const StyledH4 = NoResultsText(H4);
+const StyledBodySmall = NoResultsText(BodySmall);
+
+const NoResults = ({ searchText }) => (
+  <CenteredView>
+    <Title vertical={false}>
+      <StyledIcon name={'umbrella'} />
```
conrad-vanl: Umbrella? Really?

burkeshartsis: I actually kind of like it and I felt like @kevin-birdy liked it too. We can't use the search loop because it's already in use on the page. I felt like the warning icon was a little strong handed for something that isn't really a "DON"T DO THIS" action when the user didn't even do anything wrong. The umbrella is playfully "sad/opps" without being too overly negative. Just my thoughts thought happy to use something else.

conrad-vanl: > We can't use the search loop because it's already in use on the page.

why? 

If @kevin-birdy also says lets use the umbrella, I guess we're going to use the umbrella

burkeshartsis: It just felt a little confusing to have two of the same icons on the page but only one of them is interactive. I think if we could provide tappable suggestions I would have less concern over using an icon. I think with some design/copy changes we might be able use the loop and make it tappable without search suggestions as well. Open to whatever.

kevin-birdy: Ya, honestly I pretty much feel the same as @burkeshartsis outlined above... I guess my suggested alternative if we really don't like the umbrella would be to add a new icon to Core because I think having two search icons looks odd and the alert is overbearing.

I can't help but read the umbrella icon as "Sorry to rain on your parade but there are no results for {!Search}" and I kinda love it 😁

burkeshartsis: I like that copy! should we use it instead? I could move the icon above to and remove the larger "Uh oh" type which I wasn't in love with as it sort of suggested something went wrong.

PR #1051, opened 2019-09-10
Installs Pods Automatically

redreceipt:

Adds a script that updates and installs pods post `yarn install`

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
@redreceipt I apperciate you putting this together. I have some concerns around if this is actually needed, and at present an approach like this will give us trouble in Travis and AppCenter (Travis shouldn't be installing pods, and AppCenter already installs pods via an automatic process).  I won't close this yet though, because I would love feedback from your team about what issues they have ran into with `Pods` thus far (if any) @richarddubay @cdanwards

redreceipt:
what about if I just created a `yarn pods` that could still be abstracted so people wouldn't have to remember about pods but wouldn't get hung up in our other environments

vinnyjth:
> what about if I just created a yarn pods that could still be abstracted so people wouldn't have to remember about pods but wouldn't get hung up in our other environments

This would be the best. In my head, a `yarn pods` would do this: 

1. check to see if you are using the same version of coocapods in the podfile.lock, warn you otherwise. 
2. Make sure your pod repo is up to date.
3. cd ios && pod install 

How does that sound?

redreceipt:
@vinnyjth ok check this out

Issue #1029, opened 2019-08-29
Adds Event Content Single


redreceipt:
### User Story

<img width="491" alt="Screen Shot 2019-08-29 at 11 02 00 AM" src="https://user-images.githubusercontent.com/2659478/63951873-85a4b500-ca4c-11e9-867e-37e588fb89dd.png">

As a `<type of user>` _(who)_ I want to `<perform an action>` _(what)_ so that I can `<achieve a goal>` _(why)_.

### Acceptance Criteria

* [ ] It should display the event time and location
* [ ] It should have a signup button linked to an external form
* [ ] (optional) It should display number of people going (blocked by #939)

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

redreceipt:
"other people going" is blocked by #939

redreceipt:
@conrad-vanl for this guy, I was planning on copy and pasting `WeekendContentItem` and starting there but it looks like there's not much consistency between the three different types we have. Seems like an opportunity to abstract some common pieces. Specifically, the code for `Weekend` seems like the closest, `Devotional` seems like an old design, and `Universal` seems close to `Weekend` but it's missing theme stuff. I can do the work of unifying content items if that's the plan but where should I start?

PR #1059, opened 2019-09-16
Add action list feed feature for upcoming events

vinnyjth:

- Adds the ability to show upcoming events inside the `ActionListFeature`
- Adds a method to fetch "upcoming public" events, or "all events" + a warning if you're not using the Apollos plugin. 
- Adds a method to fetch an image for a given event. 

Will result in a few additional issues being created

- Need to implement "READ_EVENT" action.
- Related #1029 

Chose not to go overboard making the "UPCOMING_EVENT" algorithm hyper customizable. We can take a look at enhanced functionality in future PRs. 

### How do I test this PR?

1. Open the home feed, you should see an event listed in the "This week at Apollos" section.
2. Look at my pretty picture below of a home feed configured just to show a single upcoming event

Screen Shot 2019-09-16 at 3 40 01 PM

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-data-connector-rock/src/events/data-source.js
```diff
@@ -17,13 +18,45 @@ export default class Event extends RockApolloDataSource {
       .filter(`CampusId eq ${id}`)
       .get();
 
-  getName = async (id) => {
+  findRecent = () => {
```
redreceipt:
```suggestion
  getRecent = (numEvents = 3, campusID = null) => {
```

packages/apollos-data-connector-rock/src/events/data-source.js
```diff
@@ -17,13 +18,45 @@ export default class Event extends RockApolloDataSource {
...
+      );
+    } else {
+      request = this.request(
+        `Apollos/GetEventItemOccurencesByCalendarId?id=${1}`
+      );
+    }
+    return request
+      .expand('Schedule')
+      .orderBy('Schedule/EffectiveStartDate')
+      .filter('Schedule/EffectiveStartDate ne null');
```
redreceipt:
```suggestion
      .filter('Schedule/EffectiveStartDate ne null');
      .filter(campusID ? `CampusId eq ${campusID}` : "")
      .top(numEvents)
      .get()
```

packages/apollos-data-connector-rock/src/features/data-source.js
```diff
@@ -63,6 +65,26 @@ export default class Features extends RockApolloDataSource {
     };
   }
 
+  async upcomingEventsAlgorithm() {
+    const { Events } = this.context.dataSources;
+
+    // Get the first three persona items.
+    const events = await Events.findRecent()
```
redreceipt:
```suggestion
    const events = await Events.getRecent()
    
```

packages/apollos-data-connector-rock/src/features/data-source.js
```diff
@@ -63,6 +65,26 @@ export default class Features extends RockApolloDataSource {
...
+      .top(3)
```
redreceipt:
```suggestion
```

packages/apollos-data-connector-rock/src/features/data-source.js
```diff
@@ -63,6 +65,26 @@ export default class Features extends RockApolloDataSource {
...
+      .get();
```
redreceipt:
```suggestion

```

redreceipt:
This is awesome. It was a little confusing to me to have a datasource function that was a request builder so I made it a getter, little more clear in my opinion. Also, the `getByCampus` function is potentially dead code without considering the new endpoint so I extended the other getter. maybe you have some other ideas about future campus implementation?

vinnyjth:
>This is awesome. It was a little confusing to me to have a datasource function that was a request builder so I made it a getter, little more clear in my opinion. 

I understand that returning a cursor rather than a set of data is confusing, but it's important if we ever want to use our pagination helpers or add additional filtering to a request. We do this a lot with ContentItems (pretty much every method on the data source returns a cursor, rather than a feed of content) 

I am open for changing the name of the method to better reflect that, though :) maybe something like `getRecentCursor`?

redreceipt:
gotcha, what about `getRecentRequest`? I'm not making the connection to a cursor, isn't that a field on a `Connection` data type for pagination?

redreceipt:
and this case, we can just extend the `getByCampus` function to first go get the request and then add the filter in the same way. I like it.

vinnyjth:
A `cursor` is our word for a `request` that hasn't been fetched yet. It's pretty much any instance of the `request` class. 

Pagination can be done on any any instance of the `request` class. If you complete the request by calling `get` or `first`, you can't paginate. (Well, you can, you can't use your pagination helpers though)

vinnyjth:
> and this case, we can just extend the getByCampus function to first go get the request and then add the filter in the same way. I like it.

Yep! That's one of the advantages. We do that a lot in `ContentItem`

redreceipt:
just so I'm clear, pagination here and graphql pagination are two different things right? I think that's what's confusing me? same with cursors?

vinnyjth:
What our helper does is allow us to implement graphql pagination easily. You are correct, cursor's are totally different.

redreceipt:
cool. disregard my suggestions then :) rename anything that makes sense to you, if not no sweat. Let me test and give a thumbs up.

vinnyjth:
Here's a good example of pagination: 

https://github.com/ApollosProject/apollos-prototype/blob/browser/packages/apollos-data-connector-rock/src/content-channels/resolver.js#L12

Because `dataSources.ContentItem.byContentChannelId(id)` is a cursor/request, in this example, we can implement the graphql pagination logic very easily using this helper

redreceipt:
for the `subtitle` you're using moment to do something with the schedule but I've already got a function for the start and end date/time

redreceipt:
I say that because it's wrong in the client as is

vinnyjth:
> I say that because it's wrong in the client as is.

Good catch. Little strange that it's wrong, because I'Do you want refactor to use your method and see if it fixes it?

redreceipt:
Yep just give me a bit, reviewing a PR for dan right now

redreceipt:
actually this just made me realize I can improve that function too and give us a slight speed boost. stay tuned.

redreceipt:
ok take a look at this

vinnyjth:
Kicking butt and taking names @redreceipt !

redreceipt:
this one is doing weird things with tests too now...

Issue #1060, opened 2019-09-16
Rename Sharable Data Types


redreceipt:
### User Story

As a developer I want to understand the data types very clearly

### Description

In the schema, we have confused the naming of the `Sharable` implementations.

`SharableContentItem` should be `ContentSharable`
`SharableFeature` should be `FeatureSharable`

This will match the other conventions.

Also, the resolver implementation prevents extending individual fields and doesn't have the `__typename` implemented anywhere. This should fix that.

### Acceptance Criteria

* [ ] It should rename `Sharable` implementations to match the convention
* [ ] it should write `Sharable` resolvers in an extendable way

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

PR #1061, opened 2019-09-17
Reorganizes Sharing in API

redreceipt:

While this adds a bit of complexity, it adds a lot of a clarity too. And makes extending bits and pieces of sharing for churches much more granular.

- Renames sharing data structures to match convention
- Defined sharing resolvers more explicitly to make extending by churches easier

Deprecated the old names

### How do I test this PR?

```
{
  userFeed {
    edges {
      node {
        ... on WeekendContentItem {
          id
          features {
            ... on ScriptureFeature {

              sharing {
                title
                message
              }
            }
          }
          siblingContentItemsConnection {
            edges {
              node {
                id
                sharing {
                  title
                  message
                  url
                }
              }
            }
          }
        }
      }
    }
  }
}
```

None

## TODO:

- [x] Closes [#1060]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
@redreceipt I can tentatively approve this, but need to do some testing first. Normally you can't change the name of GQL types without introducing breaking changes, but because you're not changing any of the field names (and just the names of the types themselves) we might be okay. 

I wholeheartedly approve of the resolver changes based on my glance over them, nice work.

redreceipt:
I left the old ones in there and just deprecated them

vinnyjth:
@redreceipt Ultimately that won't help us, because the types of existing fields changes. Check the backwards compatibility errors coming from Travis.

redreceipt:
Oh shoot, yea I figured that would be a breaking change but wouldn't affect anything since no one calls those field types directly. If there's some way around it just let me know. I guess we could deprecate `sharing` and call the field something else but I think that's the best name for the field. I tested our client with it and still seems to work fine.

richarddubay:
@vinnyjth Any idea when this broke? I'm mostly just curious. We had/have a custom resolver for the URL part of sharing (I _think_ because of the way our links are structured, if I remember right), but it appears that something happened on the core end that made it so that our custom resolver wasn't doing anything? @redreceipt Feel free to correct me if that's wrong. That's concerning. So I'm just trying to track down a why/when so I can understand.

vinnyjth:
@richarddubay I'm not sure I understand the context here. Looks like from here that you should still be able to override the share URL, but how that's accomplished might have changed? 

@redreceipt Do you think you could implement the resolver reorganizing that's in this PR separate from the schema changes? If you do that, I can merge this PR without worry, and we can talk about the best way to change the naming of the types long term.

redreceipt:
Instead of trying to decouple it here and put it back later, I'll just override the `sharing` field in our API temporarily.

I think we still figure out how to rename stuff here. I'm hoping there may be a few interested candidates after the Rock Conference so I want to keep pushing to make things clear and developer-friendly on core.

richarddubay:
@vinnyjth Yeah, I just know sharing and the URL used to work and we have an override. Now it doesn’t seem to work. Which just tells me that something changed, I’m guessing on the core end, that doesn’t allow that override we have to work anymore. I’m just trying to figure out where that happened.

redreceipt:
> @vinnyjth Yeah, I just know sharing and the URL used to work and we have an override. Now it doesn’t seem to work. Which just tells me that something changed, I’m guessing on the core end, that doesn’t allow that override we have to work anymore. I’m just trying to figure out where that happened.

see my comment above. I broke it two months ago.

conrad-vanl:
@vinnyjth @redreceipt what's the latest thinking of this? There's parts of this I like... Looks like latest request from @vinnyjth:

> Do you think you could implement the resolver reorganizing that's in this PR separate from the schema changes? If you do that, I can merge this PR without worry, and we can talk about the best way to change the naming of the types long term.

But it sounds like @redreceipt's response was that you'll just do this separately in NewSpring.

Lets close this if it's no longer interesting / planned to work on.

redreceipt:
@conrad-vanl I still think there's merit in reorganizing without changing names. Let me throw that together and see what you think

redreceipt:
Although I am a little confused about why the naming is a problem if the names are better and the old names are properly deprecated?

redreceipt:
Although I am a little confused about why the naming is a problem if the names are better and the old names are properly deprecated? In order to do this right, I have to break up the SharableFeature schema anyway because the message has to be resolved differently for scripture vs text features.

Issue #1064, opened 2019-09-18
Discover Tabs Images Go In and Out


redreceipt:
### Details

This doesn't happen and every time, and I've only seen it on Android, but when scrolling up and down on the discover tab, the images go and in and out

![20190918_152505](https://user-images.githubusercontent.com/2659478/65179655-3ccb9500-da29-11e9-9f1c-e2ae9768bb52.gif)

### Expected Behavior

Images should stay loaded

### Actual Behavior

Images don't stay loaded

### Possible Fix

<details><summary>Additional Info</summary>

### Your Environment

- Environment name and version (e.g. Chrome 39, node.js 5.4):
- Operating System and version (desktop or mobile):
- Link to your project:

### Steps to Reproduce

1.  first...
2.
3.
4.

### Stack Trace

</details>

redreceipt:
@conrad-vanl @burkeshartsis have you ever noticed this issue on small android phones? it's been around for awhile I'm just now making an issue for it.

burkeshartsis:
So I may have noticed something similar while doing some perf testing on #1044. Try adding a `console.count()` to the discover feed/feeds (it's a feed of feeds) and see if your getting some rerenders. It it seems strange that it would be render related just because you scrolled but it would be the first place I would start.

Another place you could look is to check the docs on the individual feedviews and see if it's doing any kind of "off screen" optimizations by default that we might need to adjust.

redreceipt:
This one is difficult to pin down, I haven't been able to reproduce in a debug environment yet.

redreceipt:
it kinda looks to me like loading states are going in and out

conrad-vanl:
All of our feedviews are flatlists, so yeah, they do render/unrender when scrolling elements off screen. And on top of that, react-native, esp on android, is notoriously slow at displaying "cached" images. It's why things like this exist - https://github.com/DylanVann/react-native-fast-image

If this is why images are doing this, not sure on fix...

conrad-vanl:
FWIW @redreceipt we'd probably strongly consider a PR to core that upgrades our image components in core to using `FastImage` 😉

redreceipt:
I'll certainly give it a shot :)

redreceipt:
what seems odd about this issue is that if `FlatList` renders images on-screen and un-renders them off-screen, it looks like that's happening in reverse.

redreceipt:
[...]

PR #1067, opened 2019-09-22
Refactor ConnectedImage to use FastImage

redreceipt:
## DESCRIPTION

both

This PR swaps out our use of React Native's `Image` component for `FastImage` a third party component that renders images via different native modules. 

This PR works but came with some trade offs. RN's `Image` had a method to get image sizes that it would call before the first render (but presumably after it was renders to the shadow DOM) that we used to use in `ConnectedImage` to calculate `aspectRatio` to make rendering remote images easier. `FastIamge` doesn't expose this method but will give it to you via `onLoad` after the first render cycle. That said, this component has an optimization that will avoid this render if you provide a `width` and `height` in the `source`!

Additionally, `FastImage` doesn't support an array of `source` objects as its `source`. It only accepts one. This is unfortunate but given that our api also doesn't return an array it's a loss of a feature we weren't using. 

### How do I test this PR?
Check the `ConnectedImage` stories

This should have slightly better coverage.

## TODO:

- [x] Closes [#1064]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
@conrad-vanl pretty noticeable difference right off the bat! This is with no other custom features so I'll add back in the ones we need slowly.

packages/apollos-ui-kit/src/ConnectedImage/index.js
```diff
@@ -236,20 +237,23 @@ class ConnectedImage extends PureComponent {
       maintainAspectRatio,
       ...otherProps
     } = this.props;
+    // return (
+    // <SkeletonImage
+    // onReady={!this.isLoading}
+    // forceRatio={forceRatio}
+    // style={style}
+    // >
+    // <ImageComponent
```
conrad-vanl: Instead of commenting all of this out, you should be able to default `ImageComponent` to `FastImage`, which can be done here - https://github.com/ApollosProject/apollos-prototype/blob/5eb848828f1e38e2c0398d5dbbf307369933fee6/packages/apollos-ui-kit/src/ConnectedImage/index.js#L109

The only issue to take into consideration is for some reason we were defaulting to `Animated.Image` instead of `Image`, I'm not sure why 🤔 

Might have to do `Animated.makeAnimatedComponent(FastImage)` to keep things backwards compatible

redreceipt: these are the features the old component had that I see:

1.  intelligent caching

I think `fast-image` has very intelligent caching features. And while I don't fully understand what ours is doing, I don't see a need in keeping any of our caching code. 

2. fade in animation

This was nice when image performance was slow but may not be necessary now??? `fast-image` has a preloading feature that we could strategically use for almost no load times anywhere if we wanted! `Animated.Image` is just a shortcut from `react-native` so yes if we want to keep it, I'll just use their wrapper function. But I'd love to not keep it if we don't need it.

3. skeleton image placeholder

I'm guessing this is to keep space allocated for where the image will be? What does this do that the `rn-placeholder` can't do on its own?

4. something about aspect ratios

@burkeshartsis mentioned we may not want to keep this but also it's important for some new components? What exactly is this code accomplishing, I may be able to get creative with a different implementation if we want to move away from this.

Thoughts??? cc @conrad-vanl

burkeshartsis: > @burkeshartsis mentioned we may not want to keep this but also it's important for some new components? What exactly is this code accomplishing, I may be able to get creative with a different implementation if we want to move away from this.

We have custom `minAspectRatio` and `maxAspectRatio` as a RN friendly way of implementing a similar outcome to using min/max height/width while maintaining image aspect ratio. An example of this would be that in some of the card components we have desired min and max heights for images so that the cards always have enough room to display the min and max amount of text that could be displayed. As far as I know, there is no easier/better way to achieve this effect. This logic may be optimized by moving it to an appropriate `FastImage` callback but you'll need to test if and where that might be possible.

conrad-vanl: My thought is in order to keep the surface area of this PR small, we could start with just swapping out `Animated.Image` with `Animated.createAnimatedComponent(FastImage)`. That still offers a perf improvement while maintaining backwards compatibility. 

Then, we can look at removing functionality in non breaking ways. It's going to be hard to remove a lot of parts of this component because one of the "magic" pieces that `ConnectedImage` does is it calculates each [Image's size and adds it to the cache](https://github.com/ApollosProject/apollos-prototype/blob/5eb848828f1e38e2c0398d5dbbf307369933fee6/packages/apollos-ui-kit/src/ConnectedImage/index.js#L54). This has a big impact in making sure images are displayed correctly by react-native, as well as is the basis for features like `minAspectRatio`/`maxAspectRatio`

burkeshartsis: So based on @conrad-vanl comment above I think you're going to run into the [same issue as this guy ](https://github.com/DylanVann/react-native-fast-image/issues/559). That said, it would be possible to work around it in a similar approach to what we do now with `FastImage`'s `onLoad`. Down side, `onLoad` is called after the image is created in the native code so you will still have the draw back of having to delay render with a placeholder in the same way we do now until after we can calculate aspect ratio. This would eliminate some of our current code and get us closer to our ideal end state but we would still have some, but not all, of the same perf issues we have now. 

I think this is accomplishable in this PR with an ideal outcome for a later PR to have the server return `width` and `height` so we can do aspect ratio calculations without the delay. Then we can keep the `onLoad` code we have now as a fallback if the server doesn't return `width` and `height`

burkeshartsis:
@conrad-vanl this is ready for review except we will need to figure out a bug with why `ConnectedImage` wont render if the `source` provided is a string. We have an internal method to handle this but `FastImage` doesn't appear to be calling `onLoad` to get the `width` and `height`. The `maintainAspectRatio` story can be used to test this. Once this works a bunch of places in the app will work too. 

Last step would then be to update the rest of the snapshots across the app, to which I expect there will be many.

conrad-vanl:
I'll work on this in the afternoon today.

conrad-vanl:
Update: this is currently blocked due to fact that app breaks when actually running it with this PR 😬

redreceipt:
@conrad-vanl care if I pick this up and try to get it working? What's the current issue?

conrad-vanl:
@redreceipt I'd love for you to pick this up. I was just thinking about it and then saw your comment!

Current issue is that when you run the app, none of the images show up, or something like that. Storybook seems to work and tests pass....

So next step is figuring that out :joy:

conrad-vanl:
Also @redreceipt I'd wager that it might be easier to start this PR over (potentially) and work for a smaller footprint.

We could deprecate `ConnectedImage` with a new component. Right now we have several image components:

- `ConnectedImage`: Tries to reliably load network-based images. Displays a loading state, as well as has some niceties for handling different aspect ratios and sizing/layout
- `GradientOverlayImage`: Renders a gradient on top of ConnectedImage
- `OverlayBackgroundImage`:  I can't find anywhere this is actually used
- `ProgressiveImage`: Designed to load a low-res thumbnail quickly then swap it out with the final image. Uses connected image.
- `TileImage`: I think this is what used to be used (maybe still is) for showing children/sibling items on a series. Renders a `GradientOverlayImage`

I do NOT think we should remove all of these components for one massive image component, just wanting to catalogue what we're doing now.

Instead, I think we could deprecate `ConnectedImage` and `ProgressiveImage` and replace it with a `FastImage`-based component.

The hardest part with images is how RN handles sizing and layout. Any `Image` needs to either have a size [specified in the `source` property](https://github.com/facebook/react-native/blob/master/Libraries/Image/ImageSource.js#L72-L73) OR the `<Image />` component must have some sized inferred by the layout before rendering. Otherwise, the `Image` is going to collapse and essentially be invisible.

In other words:

`<Image source={{ uri: '..', width: 100, height: 100 }} />` is fine
`<Image source={{ uri: '..' }} style={{ width: '100%', aspectRatio: 1 }} />` is fine
`<Image source={{ uri: '..'}} />` will usually result in no image appearing

A google or some research online about these "oddities" might produce some more information and recommendations, but overall this is hte main thing that `ConnectedImage` attempts to fix:

1. On mount, it delays rendering the `<Image />` component and instead loads the Image in the background, detecting its size. It adds `width` and `height` into the `source` prop
2. It has props you can pass for forcing preservation of aspect ratio - so say you want an image to be 100% width but the height to be variable on the aspect ratio of the image - it will handle that.

So all this to say... Images are hard. If we had a way to guarantee `width` and `height` coming from the backend - that could be helpful.

Like - the best thing we could do with images:

1. Add `width` and `height` to the `ImageSource` graphql type
2. Use a service/library like [Blurhash](https://blurha.sh/) to compress images and generate thumbnail `data-uri`s that we pass to the [`defaultSource` prop](https://reactnative.dev/docs/image#defaultsource)
3. Create a new `FastImage`-based component that handles some of the default use cases of `ConnectedImage` but in a more efficient way since we now have 1 and 2 above
4. Deprecate usage of `ConnectedImage` and `ProgressiveImage`

redreceipt:
I'd love to take another swing at this.

redreceipt:
@conrad-vanl @vinnyjth could you see us pulling and storing width, height, and blurhash string in Redis or a separate DB? I would think it should be stored somewhere for performance (and it's what blurhash people recommend)

burkeshartsis:
I don't remember why this work stopped but we should probably figure out how to get it finished. This would be a pretty big win.

Issue #909, opened 2019-07-23
Remove Header from User Web View


redreceipt:
### User Story

As a user going to a modal web view, I don't want to see the 'X' button *and* the nav header

<img width="463" alt="Screen Shot 2019-07-23 at 12 48 15 PM" src="https://user-images.githubusercontent.com/2659478/61730907-5e214500-ad48-11e9-8677-6923b3718280.png">

### Why

looks better

### Acceptance Criteria

* [ ] It should [SPEC_DETAIL]

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

redreceipt:
it's broken anyway

![Kapture 2019-09-30 at 13 43 06](https://user-images.githubusercontent.com/2659478/65902582-c44dc800-e388-11e9-9967-4852d7eb2b85.gif)

PR #1074, opened 2019-09-30
Removes Browser Modal View By Default

redreceipt:
## DESCRIPTION

Screen Shot 2019-09-30 at 1 44 51 PM

The Modal view is unnecessary and doesn't work anyway, removing `modal = true` by default, can circle around later if we want to fix.

None

### How do I test this PR?

Click the test link in the test panel

None

## TODO:

- [x] Closes [#909]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
@redreceipt I would prefer to figure out why the ModalView is not working, in this instance :)

richarddubay:
Just my opinion (so take it FWIW), but I don't think the in-app browser should be a modal or have modal behavior. I feel like it should have native in-app browser behavior: comes up with the bar at the top with a button that says "Done" or "Close" (or whatever it is that Android has). The modal "X" and any other modal functionality should be saved for our content.

vinnyjth:
@richarddubay we need to have two modes for the browser. 

1. The one you are describing, I'm calling it a `modal` behavior, I think having the Done or Close button instead of the X would be totally fine here. 

2 The ability to embed an in app browser within another screen, discretely. Imagine a browser that renders a QR code on the profile screen. 

So in essence, we need the ability to have a fullscreen "modal" browser, and an embeddable browser.  Right now we are only doing the latter properly, I think

richarddubay:
@vinnyjth Is the "embeddable browser" something that we had before and is broken with this implementation?

redreceipt:
yea...I'll fix it :)

redreceipt:
ok @vinnyjth here it is

richarddubay:
@vinnyjth Is this something we can discuss getting in pretty soon? Having two ways to close this isn't ideal.

vinnyjth:
Sorry @richarddubay  - didn't realize this was ready for review. These are all changes you can make directly to your app, I would recommend doing that in the meantime while I review and merge this. That way you aren't blocked by me to get these changes.

redreceipt:
@richarddubay I say we just wait. We'll be switching to In-App Browser anyway for all our links, we don't use the web view anymore.

richarddubay:
If we're removing the web view then I'm fine with waiting.

conrad-vanl:
@redreceipt overall I think this looks good. Might have a few questions, but I think we need to resolve conflicts first - which means moving this code into one of our packages and considering what still needs to be done based on anything new we have in core.

redreceipt:
so @conrad-vanl to summarize, to move this forward, I need to move the `user-web-browser` out to a package (probably `ui-connected`) and remove the modal view there?

conrad-vanl:
We have a `RockAuthedWebBrowser` (I think that's what it's called) in one of those packages. There might be another web browser too... I don't recall with 100% certainty. But yeah - you'll haveto move these changes.

Also, I think there was some confusion above on `ModalView` vs `modal` config option on the web browser. Can you explain, perhaps with pictures, the difference between `modal=false` and `modal=true`?

redreceipt:
> Also, I think there was some confusion above on `ModalView` vs `modal` config option on the web browser. Can you explain, perhaps with pictures, the difference between `modal=false` and `modal=true`?

`ModalView` just adds that little "X" but that's the whole point of this PR. It's redundant with also having the back button.

conrad-vanl:
@redreceipt ahhh, I completely agree! I'm good with moving this forward then :)

Issue #1076, opened 2019-09-30
Location Finder Distance Wrong


redreceipt:
### Details

The distance to the campus on the location finder screen always says "0mi". I propose it just shouldn't have distance here, if it's there campus, they don't care at this point how far it is. That information is only useful on the mapview.

<img width="463" alt="Screen Shot 2019-09-30 at 3 21 44 PM" src="https://user-images.githubusercontent.com/2659478/65909303-501a2100-e396-11e9-91c8-345c32336307.png">

### Possible Fix

<details><summary>Additional Info</summary>

### Your Environment

- Environment name and version (e.g. Chrome 39, node.js 5.4):
- Operating System and version (desktop or mobile):
- Link to your project:

### Steps to Reproduce

1.  first...
2.
3.
4.

### Stack Trace

</details>

vinnyjth:
Nice catch.

vinnyjth:
Fixed in #1078

PR #1078, opened 2019-09-30
Removes Campus Distance from Onboarding Screen

redreceipt:
## DESCRIPTION

Screen Shot 2019-09-30 at 3 27 39 PM

Removes distance from location finder screen, not needed at this point (plus currently it's broken). That information is only useful on the `MapView` when the user is trying to decide.

None

### How do I test this PR?

Bring up location finder.

None.

## TODO:

- [x] Closes [#1076]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-kit/src/CampusCard/index.js
```diff
@@ -38,10 +38,12 @@ const CampusCard = memo(
         <FlexedCardContent>
           <Header>
             <H5>{title}</H5>
-            <H6>
-              {Math.round(distance)}
-              mi
-            </H6>
+            {distance ? (
```
vinnyjth: I know this is a bit nitpicky, but let's make sure we are able to render distances of 0 miles correctly. We should check for `distance != null` rather than `falsy`.

redreceipt: good call

redreceipt:
```suggestion
            {distance != null ? (
```

conrad-vanl: Technically you should probably check for `undefined` or other non-number falsy values. If you just don't pass the prop at all, `distance` will be `undefined`, not `null`

conrad-vanl: `!isNaN(distance)` mayne

vinnyjth: `!= null` tests for `undefined` values. `!== null` does not though :)

conrad-vanl: what a sensei move

PR #1086, opened 2019-10-01
Live item polling

vinnyjth:

This PR does the following things
- Removes the LiveButton and the polling that occurs with LiveButton.
- Polls for LiveContent, and updates the badge on live content when it becomes live.

n/a

### How do I test this PR?

Okay - this is a tricky multistep process. 

1. Before you boot your server, change the following line to 
```
isLive: true,
```
https://github.com/ApollosProject/apollos-prototype/blob/live-item-polling/packages/apollos-data-connector-church-online/src/data-source.js#L24
2. Boot your server. Make sure to update build before you boot. `yarn && yarn start`
3. The main content item should have a live badge.
4. Change the line you orignally changed to 
```
isLive: false,
```
5. Wait for your server to rebuild (should happen when you save the file) 
6. Wait for the content-item poll to trigger. You should see a log in your server indicating a network request. 
7. The live badge should go away. 

*If this does not work the first time, try it a second time before you pass it off as broken* Sometimes the poll that happens can happen while the server is rebuilding and rebooting, which causes the content-item not to update. 

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-church-api/local.graphql
```diff
@@ -371,6 +371,7 @@ type Query {
   personaFeed(first: Int, after: String): ContentItemsConnection
   currentUser: AuthenticatedUser
   liveStream: LiveStream
+  activeLiveStreamContent: [ContentItem]
```
conrad-vanl: 😬

Why do you think this is best as a root query vs something inside of `liveStream` node?

vinnyjth: Yah.... not a huge fan of this endpoint either. The `liveStream` node returns a `LiveStream` which seems like the wrong `type` at this point, seeing as a `LiveStream` represents an individual, specific, livestream (not metadata about livestreaming in general)

Open to suggestion about better names / implementations of this.

conrad-vanl: Well i'm just thinking that we're starting to develop this `relatedNode` routine, and if adding `relatedNode` to `LiveStream` may not be a better way to do this.

I know you're trying to handle scenario of having multiple live streams, but we don't right now... and I think a better root query for multiple live streams might be to return multiple live streams, not multiple `ContentItem`. You're asking for live info, but getting back `ContentItem`... just seems weird

vinnyjth: Ah I see what you're saying - maybe add a `relatedItem` to `LiveStream` and then use that. I like that. But we won't be able to support multiple livestreams, unless we add a `liveStreams` root query.

vinnyjth: I'll add a `liveStreams` and a `relatedItem` to livestream to make this PR mergeable. I'm also prioritizing this behind QA issues at this point.

vinnyjth:
@redreceipt Can you review this for me? It replaces the LiveButton (and the polling associated with it) with ContentItems properly polling for their "isLive" state

vinnyjth:
@redreceipt My bad. I made a mistake when fixing the linter blindly

redreceipt:
works! few questions...

1. do we need to the "Start the livestream" button in the test panel anymore?
2. does it only show live if *that* content item is live? so would it show live if a news article was our campaign?
3. shouldn't the content screen also show live in some way on the play button? if the main campaign is not the latest sermon, this will be important.

Screen Shot 2019-10-30 at 11 06 25 AM

vinnyjth:
1. do we need to the "Start the livestream" button in the test panel anymore?
We do need to add it back eventually, but I'm not going to solve that problem for now. It is going to be way harder to get that working. 

2. does it only show live if that content item is live? so would it show live if a news article was our campaign?
Yes. It only shows "Sermon" items as live. Doesn't matter if they are featured or not. 

3. shouldn't the content screen also show live in some way on the play button? if the main campaign is not the latest sermon, this will be important.
I'm not sure, but that's beyond the scope of this PR :)

redreceipt:
if we dropped an old sermon as the featured item, would it show live during a livestream?

redreceipt:
>    I'm not sure, but that's beyond the scope of this PR :)

at the very least, couldn't it just be on the pill in the content item view too?

redreceipt:
this isn't working quite right for me by the way, when I go back to the home screen, it switches back to "Sermon" without me changing API code

Kapture 2019-10-30 at 11 12 07

vinnyjth:
> at the very least, couldn't it just be on the pill in the content item view too?
Yah, in another PR though. 

> this isn't working quite right for me by the way, when I go back to the home screen, it switches back to "Sermon" without me changing API code
I believe that's because of how we are testing this, but I can try and reproduce.

vinnyjth:
@redreceipt I had reverted the poll interval to a value that was way too high. Try now. You can also change the leading `3` to a `1` in the LiveProvider to get a much more testable value.

redreceipt:
I hate being this guy...but when I go back to the home feed, it still changes to "Sermon" and never changes back to "Live"

Kapture 2019-10-30 at 17 21 15

vinnyjth:
No worries at all @redreceipt, if it doesn't work, it doesn't work :) 

I'll take another look and see if I can improve this PR.

packages/apollos-data-schema/src/index.js
```diff
@@ -465,10 +465,12 @@ export const liveSchema = gql`
     eventStartTime: String
     media: VideoMedia
     webViewUrl: String
+    contentItem: ContentItem @cacheControl(maxAge: 10)
   }
 
   extend type Query {
     liveStream: LiveStream
```
redreceipt: shouldn't we deprecate this? seems confusing as to what this would return

PR #1093, opened 2019-10-03
Adds Redis Cache and stores isLiked / likeCount in RedisCache. 

vinnyjth:

- Provides a connector to Redis for storing data to improve performance. 
- Implements the connector for the following module, caches the `likeCount` and `isLiked` for content items. Should improve performance dramatically on the home feed and discover feed.

n/a

### How do I test this PR?

- Open the app to the home feed with redis running.
- Note the number of calls to the `following` endpoint
- Reload the page, you will see that there are no calls to the `following` endpoint in your server console the second time. 
- Like an item, and make sure the like persists. 
- Unlike an item, and make sure the unlike persists. 

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
>  Is this redis cache connector also compatible to use as a Apollos Data Source cache?

Yes, it is :) Because it's easy to access from other `dataSources`. I could have implemented it as a `Model`, but I've been making everything `dataSources` by default because it fits better inside the Apollos Ecoystem.

vinnyjth:
@conrad-vanl waiting on you to see my reply to your comment and respond before I merge this.

conrad-vanl:
@vinnyjth this can be merged, more of a question of why it's not hooked up as the default cache for `Apollos`, whether we should or not vs the in-memory cache we're using.

vinnyjth:
Big difference between this and the `in-memory cache` is this can store arbitrary computed data, and I can edit the cache easily. For example, with this cache, I can invalidate/fix the `likeCount` or `isLiked` of something when it's liked or unliked. 

The `in-memory-cache` is a straight up replay of the requests that were made, so we don't have much control over fixing/nuking/update the cache. So I think the answer is "both and"

PR #1102, opened 2019-10-07
Filter Unpublished Content

erikkgagewhitt:

Adds an additional condition to the LIVE_FILTER which checks that Status is equal to a number specified in the config.yml; the default number is 2 which is the `approved` status according to ROCK.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


erikkgagewhitt:
@vinnyjth I have a couple of questions:

1. How exactly can I test this?

2. This is a little brittle as is; there's no way to specify more than one status (ie. I want published and pending content, but not denied: 1,2,!3). I don't want to take too many liberties, but is this something I could change?

vinnyjth:
1. Let's pair on it :) 

2. You are welcome to come up with an API that would allow you to select exactly the types the user wants! I'm imagining a field in the `yml` called 

```
INCLUDE_CONTENT_TYPES
  - Approved
  - NotYetApproved
```

I'll leave it to you to come up with the implementation.

erikkgagewhitt:
@vinnyjth I'd love to

vinnyjth:
@erikkgagewhitt Once you tackle part 2, I'll show you how to test this.

erikkgagewhitt:
@vinnyjth Okay, we can now have multiple allowed status, but they're not quite enumerated like I'd like them to be.

redreceipt:
can I request that the allowed types are different per environment? I'm not sure how I could conditionally list the types in the `config.yml`

redreceipt:
so for example, in our beta app, I want to see pending and active. gives churches a good way to preview content before it's published.

redreceipt:
I guess a church could just include it all and then go back to take some types out, just want to kick the ball as far as we can the first time

erikkgagewhitt:
@redreceipt I've pushed up a change that should allow you to specify Content Item status based on environment or leave it blank for default.

redreceipt:
> @redreceipt I've pushed up a change that should allow you to specify Content Item status based on environment or leave it blank for default.

how does that work? is it comma separated or `[TYPE,TYPE]`?

erikkgagewhitt:
Sorry, @redreceipt arrays aren't supported in env format. I'll have to come up with another solution.

vinnyjth:
@erikkgagewhitt You might be able to fix that by wrapping this in `JSON.parse()`

https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-config/src/attach-env-variables.js#L15

Give it a shot! We have pretty good tests over the config module.

erikkgagewhitt:
The most recent push _should_ allow the previous adjustment to config.yml to work. The formatting would be `SOME_ENV_VARIABLE=["string", "string"]`

vinnyjth:
@redreceipt Can you glance at the solution to make sure it works for you? You'll need to put something like `CONTENT_TYPES=["APPROVED", "NOT_YET_APPROVED"]` in your alpha server, and `CONTENT_TYPES=["APPROVED"]` in your prod server.

vinnyjth:
Also @redreceipt - this condition is an `AND` with the `publishedAt` dates, is that okay? So if you have content that is unapproved, but has publish dates in the future, you won't be able to see it no matter what.

vinnyjth:
@erikkgagewhitt Testing this locally, I discovered that although the `Status` is exposed as an `Int`, you have to filter it using the `String` values. Thanks Rock! I updated your code and tests to reflect that, and I think we are good to go now!

vinnyjth:
After @redreceipt or @richarddubay gives me a heads up that this will work for them, I will approve and merge this.

redreceipt:
This is all apart of the content preview conversation. I know this is more work but we would love to optionally see future content as well. That's going to be the most common use case for comms, writing content for the future that they want to preview on an alpha build. 😬

vinnyjth:
@redreceipt Would it be fine to ask the content team to not set a "published at" dates for items they want to preview?

redreceipt:
> @redreceipt Would it be fine to ask the content team to not set a "published at" dates for items they want to preview?

We could ask them to hold off but I'd love it if we made their lives as easy as possible in that regard. I'd want them to be able to set it and forget it. Like I'm imagining them doing the content all in one sitting, date and all. And then they are free to preview the content until the start date without having to remember to go back in later and set a date. Especially when a solution doesn't seem overly complex.

If you wanna merge this one, would you be opposed to me going back in a writing a new PR to make that a toggle as well?

richarddubay:
I think it's important that in production we don't show "unpublished" or "in-the-future content". I feel like this would be the case across churches. So I like the concept. For a potential in-app preview, being able to see the items that are "Pending Approval" in our alpha version of the app would be amazing. 

I like the idea of making the date thing a toggle. Then a church could do whatever they wanted to do. As a core team this means that we can provide tools for a church to use and we're not responsible for how they use them. We don't have to make the decision of whether or not they choose to use the tools we provide to break things. Making a decision like "no matter what anything in the future won't show" seems like a individual church based decision and not a core mandated decision ... even if I feel like most churches would choose to do so. 

@redreceipt I guess the toggle would have to be something along the lines of hide future dated things in production, but show them in development?

```
HIDE_FUTURE_DATES
   - production: true
   - development: false
```

... or something?

redreceipt:
actually I think the implementation that would be simplest that would make the most sense is only toggling pending content and future content. So having a single env variable that would accomplish both would cover 95% of use cases:

`PREVIEW_CONTENT=true` would turn on pending and future content by default. No arrays necessary.

vinnyjth:
@erikkgagewhitt Okay I think the feedback from @redreceipt and @richarddubay is really valuable and we need to pivot :) 

Let's orient towards adding a new `config.yml` option called something like `ROCK.SHOW_INACTIVE_CONTENT`. 

When true, we should use no filters, showing all content. When false (which should be the default) we should show only content that is Published and meets the existing `date` filters. 

Sorry for the switch up, the code in this PR is fantastic :D I should have done a bit more research before I scoped out this work.

vinnyjth:
We can keep the JSON parsing in the `config` package :) I think that's a valuable addition that we might need in the future.

redreceipt:
love it, thanks @vinnyjth and @erikkgagewhitt

redreceipt:
is this one ready to review and test?

erikkgagewhitt:
@redreceipt Should be!

redreceipt:
did you add any test content? it's cool if not, I'll throw some in there.

erikkgagewhitt:
I did indeed add some tests.

redreceipt:
> I did indeed add some tests.

like test pending content in Rock?

erikkgagewhitt:
Oh, I see. No. I only added tests around the filtering system. I didn't actually mock out any content.

redreceipt:
something seems to be broken with pulling content now. look where my pointer is, that seems to be the culprit, the request is no longer valid.

Screen Shot 2019-10-08 at 2 34 42 PM

Screen Shot 2019-10-08 at 2 34 00 PM

erikkgagewhitt:
Hmm, yeah; I see the issue. Not sure how that state came up; I'll take a look.

erikkgagewhitt:
I've found the problem - going to get started on trying to fix it.

erikkgagewhitt:
Okay, I believe I've fixed the issue with invalid requests. Updated the test to show that as well.

packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -273,7 +273,8 @@ export default class ContentItem extends RockApolloDataSource {
       .tz(ROCK.TIMEZONE)
       .format()
       .split(/[-+]\d+:\d+/)[0];
-    return `((StartDateTime lt datetime'${date}') or (StartDateTime eq null)) and ((ExpireDateTime gt datetime'${date}') or (ExpireDateTime eq null)) `;
+    const filter = `(((StartDateTime lt datetime'${date}') or (StartDateTime eq null)) and ((ExpireDateTime gt datetime'${date}') or (ExpireDateTime eq null))) and Status eq 2}`;
```
vinnyjth: I think this should be 
```suggestion
    const filter = `(((StartDateTime lt datetime'${date}') or (StartDateTime eq null)) and ((ExpireDateTime gt datetime'${date}') or (ExpireDateTime eq null))) and Status eq 'Approved'}`;
```

I found out yesterday that Rock expects a string for this field :😅

packages/apollos-rock-apollo-data-source/src/request-builder.js
```diff
@@ -84,9 +84,11 @@ export default class RockRequestBuilder {
     return this;
   };
 
-  andFilter = (filter) => this.filter(filter, { operator: 'and' });
+  andFilter = (filter) =>
```
vinnyjth: You might be able to back out your change to `andFilter` and `orFilter`, and make your improvement directly to `filter` :D

packages/apollos-data-connector-rock/src/content-items/data-source.js
erikkgagewhitt: I thought the values we looked at were integers - well, still strings, but strings that represent integers.

vinnyjth: They are. But rock expects strings when you use the `$filter`. I only know this by playing around with the API console, try going to the `ContentChannelItems` endpoint in the Rock API browser and trying the following in the `$filter` field.

`Status eq 2`

`Status eq 'Approved'`

erikkgagewhitt: Oh, okay; so `Status eq '2'` would work? Or is it preferable to use the human-readable string?

vinnyjth: I don't think `Status eq '2'` would work, but you can test in the Rock API console :)

packages/apollos-rock-apollo-data-source/src/request-builder.js
```diff
@@ -77,6 +77,9 @@ export default class RockRequestBuilder {
    * Filter resources by an odata string
    */
   filter = (filter, { operator } = { operator: 'or' }) => {
+    if (!filter) {
```
vinnyjth: Will this be true with an empty string?

erikkgagewhitt: I believe so. I think the empty string is falsey.

packages/apollos-data-connector-rock/src/content-items/data-source.js
erikkgagewhitt: Looks like it works okay, but I'm just going to leave it with the word so its clearer.

packages/apollos-rock-apollo-data-source/src/request-builder.js
vinnyjth: You the man! I always forget.

PR #1110, opened 2019-10-08
Removes Redis warnings and fixes cache keys for like counts

vinnyjth:

- Changes the way Redis is loaded, so that it's only loaded if the REDIS_URL is set, and gives nice warnings if it's not (but only when the redis cache is used) 
- I never fixed some of the cachekeys on the server, so `likedCounts` weren't being cached correctly. Woops! 

n/a

### How do I test this PR?

Run the app with a `REDIS_URL` set in your .env and load the home feed, you shouldn't see any warnings. 
Run the app without a `REDIS_URL` set in your .env and load the home feed, you should see a number of warnings, but only when loading the home feed.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-data-connector-redis-cache/src/__tests__/__snapshots__/data-source.tests.js.snap
+++ b/packages/apollos-data-connector-redis-cache/src/__tests__/__snapshots__/data-source.tests.js.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Redis Cache constructs with Redis 1`] = `
+exports[`Redis Cache constructs with Redis env url 1`] = `
 Cache {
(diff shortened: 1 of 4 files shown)


redreceipt:
I don’t have redis installed and do not have the variable defined and am hung up on build here

Screen Shot 2019-10-09 at 1 37 16 PM

redreceipt:
Works without it. there's still a couple of ugly warning right when you `yarn start` but not during app execution anymore.

What's the `REDIS_URL` that I can test with?

vinnyjth:
`redis://localhost:6379`

redreceipt:
didn't get any warnings when loading the home feed, how can I verify redis is working properly?

PR #1117, opened 2019-10-10
Reduce Warnings

### LINKED ISSUE #1108: Reduce "YellowScreen" warnings.
### Apollos Version: x.x.x

We get an "enjoyable" amount of YellowScreen warnings when booting the app. Some of them we can ignore, and some of them we can (and should) fix. 

image

### Details

From the image above, the warnings I would focus on knocking out are (in order of "bang for buck")

- `OneSignal the configure method has been deprecated`
- `componentWillMount is deprecated` You might have to update Storybook to fix this
- `AsyncStorage has been extracted from react-native core` one of our packages we rely on is using this. 
- `propType title marked as required / propType coverImage marked as required`

### Possible Fix

### Your Environment

- Link to your project:

### Steps to Reproduce

1.  first...
2.
3.
4.

### Stack Trace

</details>

erikkgagewhitt:

Limits the number of warnings that are produced when the app is mounted.

Takes care of prop warnings by handling them dynamically and updates lifecycle methods.

### How do I test this PR?

Load the app and make sure you only see 3 warnings.

## TODO:

- [x] Closes #1108 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-kit/src/ConnectedImage/index.js
```diff
@@ -119,12 +119,12 @@ class ConnectedImage extends PureComponent {
     this.imageOpacity = new Animated.Value(this.isLoading ? 0 : 1);
   }
 
-  componentWillMount() {
+  componentDidMount() {
     this.updateCache(this.props.source);
   }
 
-  componentWillReceiveProps(newProps) {
```
vinnyjth: @burkeshartsis @conrad-vanl based on your recent work on this component, is this an OK change?

packages/apollos-ui-kit/src/FeaturedCard/index.js
```diff
@@ -157,12 +157,37 @@ const FeaturedCard = withIsLoading(
...
+const completedPropTypes = {
   coverImage: PropTypes.oneOfType([
     PropTypes.arrayOf(ImageSourceType),
     ImageSourceType,
   ]).isRequired,
   title: PropTypes.string.isRequired,
+};
+
+FeaturedCard.propTypes = {
+  coverImage: (props, propName, componentName) =>
```
vinnyjth: Nice :D

packages/apollos-ui-kit/src/ConnectedImage/index.js
conrad-vanl: Uhhhh. the only problem is that `this.updateCache` can call `setState`, which moving this do `componentDidUpdate` would result in this component always rendering again. Which...would call `componentDidUpdate` again. Endless loop. No, I'm not comfortable with this change.

erikkgagewhitt: Would it be better to just use the UNSAFE methods?

conrad-vanl: I'm guessing this hsould be refactored to [`getDerivedStateFromProps`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops)

vinnyjth: @conrad-vanl do you have an alternative in mind that might work that's compliant with the latest recommendations from React around the use of `willRecieveProps`?

erikkgagewhitt: I think the problem is that you can't call setState in getDerivedStateFromProps.

conrad-vanl: We could leave this using UNSAFE method until we merge in fast image refactor, I believe tthat PR gets rid of this logic anyways

conrad-vanl: Also it [would appear](https://github.com/facebook/react/issues/13541#issuecomment-418101478) that the master himself says `componentDidUpdate` is the right use case for this.

conrad-vanl: and @erikkgagewhitt i know you can't call `setState` in `getDerivedStateFromProps`, we would have to return then ew state instead, which would be the refactor.

erikkgagewhitt: Oh, okay. I see how that would work.

conrad-vanl: but yeah, anyways @erikkgagewhitt if you can show that this component doesn't re-render endlessly (check out [this comment](https://github.com/airbnb/javascript/issues/1875#issue-345620672)) I'm fine with `componentDidUpdate`. Or we can leave it like it was since #1067 is hopefully coming some time

packages/apollos-ui-kit/src/FeaturedCard/index.js
conrad-vanl: what is this voodoo magic

erikkgagewhitt: XD I thought similarly, even as I typed it.

packages/apollos-ui-kit/src/ConnectedImage/index.js
burkeshartsis: I know you already landed on a solution but I just wanted to clarify one thing for @erikkgagewhitt. You can "set state" or maybe more appropriately said "update state" in `getDerivedStateFromProps` you just don't call `setState`. Instead you return a state object. You can think of `getDerivedStateFromProps` as the `setState` call and logic people commonly used to write in `componentWillUpdate`. If you ever had state logic in there to optimize for render count/control then `getDerivedStateFromProps` is the new way to do it. From the docs when it was released,

> The new static `getDerivedStateFromProps` lifecycle is invoked after a component is instantiated as well as before it is re-rendered. It can **return an object to update state, or null** to indicate that the new props do not require any state updates.

erikkgagewhitt: Alright, so it _looks_ like componentWillMount and componentDidMount both cause the component to render 8 times.

erikkgagewhitt:
This one seems to be a genuine warning; some of the default cards have no background images to render.

vinnyjth:
Cool @erikkgagewhitt we can ignore that. Although I wonder if `isRequired` should be removed if it's technically not actually required. @burkeshartsis might have an opinion.

burkeshartsis:
@erikkgagewhitt @vinnyjth so they can be rendered without images but @kevin-birdy wanted to express that you shouldn't. So this was my way of doing that. Practically speaking the warnings are annoying so I understand if you want to remove them. Just make sure to move the no longer required prop into alphabetical order 😉

vinnyjth:
@erikkgagewhitt let me know when this is ready for review, and I'll review it :D

erikkgagewhitt:
@vinnyjth This is ready to go. The async storage is going to take some more investigation, so I'll do it in another PR.

Issue #1118, opened 2019-10-10
Pending Content Still Showing


redreceipt:
### Details

I'm not specifying the `ROCK_SHOW_INACTIVE_CONTENT` env variable and pending content is still showing up. It should be hiding by default.

<img width="1440" alt="Screen Shot 2019-10-10 at 7 29 38 PM" src="https://user-images.githubusercontent.com/2659478/66613896-fa632700-eb94-11e9-8096-dfec977d1f93.png">

### Possible Fix

<details><summary>Additional Info</summary>

### Your Environment

- Environment name and version (e.g. Chrome 39, node.js 5.4):
- Operating System and version (desktop or mobile):
- Link to your project:

### Steps to Reproduce

1.  first...
2.
3.
4.

### Stack Trace

</details>

redreceipt:
This is due to the variable not being there and it passing through the `config.yml`,  it's a string `"undefined"` so it's truthy. We need the `ApollosConfig` function should convert `"undefined"` to `undefined`

PR #1119, opened 2019-10-11
Sets Missing ENV Variables to null

redreceipt:

Currently, when `env` variables are missing, the `config.yml` reads them as strings: `"undefined"` so they evaluate truthy in code causing misleading results. This sets those values to null.

## TODO:

- [ ] Closes [#1118]
- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
currently, the only thing holding this back is Algolia craps out because those values are `null` so I need to put in some safety checks if those values are missing to not run Algolia functions.

packages/apollos-config/src/attach-env-variables.js
```diff
@@ -16,6 +16,8 @@ function handleValue(value) {
       envVariableRegex,
       (match, p1) => process.env[p1]
     );
+    // set undefined variables to null so they eval falsy
+    if (envVariable === 'undefined') return null;
```
vinnyjth: Good catch!

redreceipt: Algolia is getting hung up somewhere. I tried putting safety checks at the `Search` resolver but that's not the extent of it, where else do I need them?

vinnyjth: Show me the errors?

redreceipt: Screen Shot 2019-10-11 at 11 17 30 AM

redreceipt: that's on `yarn start`

vinnyjth: Try changing `return null;` to `return undefined;`

redreceipt: I did first, no luck. I'm assuming because something is trying to pass an undefined or null value into an Algolia function somewhere. That's the call we need to short circuit.

vinnyjth: We would have to shortcut it here, and also find a way to write `stubs` for all times we call a method on the algolia client. 

@redreceipt Because of the way our schema is written, the algolia module can be removed from the API if you don't want to use it by not adding it to the data object. Let's plan on letting that be the strategy for handling this.

redreceipt: so conditionally include the presence of the `Search` in the data object if the env variables are there?

redreceipt: or just do nothing and force the algolia variables to be required?

vinnyjth: I was thinking more along the lines of `Search` neutering itself it it can't construct algolia.

vinnyjth: >or just do nothing and force the algolia variables to be required?
Yes. Because if someone doesn't use algolia they shouldn't use the algolia data package.

redreceipt: I'm a little confused. I tried making `Search` neuter itself and then shutting off the inclusion of search completely but the problem seems deeper than that. This is the error I'm getting from core when the variables are undefined...

Screen Shot 2019-10-11 at 3 15 28 PM

vinnyjth: I wouldn't worry too much about trying to `neuter` Search :) It's pretty hard, see what I did in the `cache-redis` connector, and not worth it for non essential functionality.

vinnyjth: Based on the error message that you're seeing, it looks to be based on how you neutered Search. It looks like you passed `Search: null` to the data object, where you might need to pass `Search: {}`

redreceipt: That's fine. but with this PR (which I still think is valid) we are just forcing churches to sign up for Algolia or the app won't work. It fails on `yarn start`.

redreceipt: yep that did it

vinnyjth: @redreceipt They can remove the `data-connector-algolia` package and boot the app without issues :)

redreceipt: I don't understand, remove it from where? I don't think we're on the same page, I'm missing something.

Is something missing from this PR that I'm not thinking about?

redreceipt: OOOOO are you just saying we don't even need the conditional I added? If it fails, they can delete the search line and voila. Is that what you mean?

vinnyjth: Totally :D Sorry I didn't make it clear that I was talking about 'but if we made this change, I think this is how we would have to do it'

vinnyjth:
@redreceipt Once you fix the tests, I will merge this.

redreceipt:
Long story short, this is breaking because Travis doesn't have Redis installed *I think*.  And the change to how environment variables are coming seems to be negating your "no Redis" workarounds...

Screen Shot 2019-10-18 at 1 42 00 PM

Do we need to bypass bull creating jobs or something if `REDIS_URL` is undefined?

vinnyjth:
> Do we need to bypass bull creating jobs or something if REDIS_URL is undefined?
Yah I think so :/  I'll take over from here

Issue #1114, opened 2019-10-09
Implement disabled/focused state for content children/sibling


vinnyjth:
### User Story

As a user, when I'm visiting an item with siblings, I should see the current item displayed differently (and it shouldn't be tappable).

### Description

The current item in a list of sibling content items (look at a devotional child item for a good example) should be displayed differently than its peers. Tapping on that item should also not result in a navigation action. 

>All it is = entire component >> 50% opacity and remove the drop shadow/
- @kevin-birdy 

![image](https://user-images.githubusercontent.com/1637694/66510567-446be000-eaa3-11e9-8554-e788e20d33ee.png)

This might be easiest to implement by adding a "Disabled" prop to the Horizontal child that's been rendered here. 

### Acceptance Criteria

* [ ] It should [SPEC_DETAIL]

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

PR #1122, opened 2019-10-11
Adds an active/disabled state to HorizontalHighlightCard

burkeshartsis:
## DESCRIPTION
Partial work towards #1114 along with #1124. One more additional PR will need to be opened to implement the usage of this PR's new `isActive` prop within `HorizontalTileFeed` to complete the work needed for #1114.

This PR adds an `isActive` prop to the `HorizontalHighlightCard` so that when it is used to display a series of the same thing it is possible to differentiate the current or "active" card in the horizontal list. This state is also an effective "disabled" state. 

This PR leverages `theme.overlays.medium` to achieve the desired effect. However, the current `alpha.medium` value used by `overlays.medium` is much less opaque than the updated Apollos style guide intends. For this PR to match the style guide #1124 has to be merged first. 

### How do I test this PR?
Check the stories!

100% test coverage 💥

## TODO:

- [x] Partial work towards #1114 
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
What's the status on this? Anyway I can help out?

burkeshartsis:
This will be ready for review today. I just need to finish documenting the PR.

vinnyjth:
I think `disabled` makes more sense as a prop name. What do you think? Feel free to sell me on `isActive` :)

vinnyjth:
I'll expand on this `isActive` implies additional or enhanced functionality. `disabled` implies reduced or muted functionality, which I think is closer to the truth for how the prop behaves, even if it's farther from our intended use case.

burkeshartsis:
### 💪 The Sell ⤵

So I chose `isActive` because that is the intended use case for the prop to highlight the "active" card in a series. Additionally, as you stated, "disabled" implies a loss of function which the changes in this PR also do not implement mostly because a disabled prop would be passed to the `Touchable` that wraps the card to remove any functionality. It just so happens that in this case the "active" state looks visually like a "disabled" state. 

I could really go either way on this but felt like it was better for the prop to be semantic to how it is used.

vinnyjth:
Let's change this to `disabled`. A quick peek for our app for similar phrases, and I think `disabled` will make more sense to our partner developers. The prop matches how the `button` disabled prop will work (minus the non making this pressable, which you're right, will be handled in the connected or parent component for this).

packages/apollos-ui-kit/src/HorizontalHighlightCard/HorizontalHighlightCard.stories.js
```diff
@@ -31,7 +31,7 @@ storiesOf('ui-kit/HorizontalHighlightCard', module)
           isLiked: true,
           label: 'Custom Label',
           hasAction: true, // may not match actual schema
-          isActive: false, // may not match actual schema
```
vinnyjth: Thank you :D I think this will be easier to understand for our partner churches when reading the proptypes.

PR #1137, opened 2019-10-17
Fix Tappable Areas in Devotional Content Items

erikkgagewhitt:

Fixes an issue that would occur on android where - after swiping - the tabs would cease to function. This was apparently due to the fact that the pager that `react-native-tab-view` uses for android is unreliable, so we use IOS default pager for android since android's is unreliable.

Here is the reference: https://github.com/react-native-community/react-native-tab-view/issues/525#issuecomment-453782020

Also fixes an issue where the full area of the "Scripture" and "Devotional" buttons were not tappable. I fixed this by allowing touches to pass through the overlapping views while still allowing touch events on subviews. This should allow us to tap the tabs while still being able to tap the buttons contained within the overlapping views.

The fix to the tappable areas of the tab view works by allowing taps to pass through the container which contains the navigation action buttons. In doing so, we make swiping less reliable because there is no longer something to grab onto at the top of the screen; we have to rely on the native swiping capabilities.

This is not a problem on screens without images at the top, but the screens which do (most of them) have an animation reliant on touch actions which makes swiping much more difficult.

### How do I test this PR?

**Navigation**

1. Load up an Android device
2. Navigate to a devotional content item
3. Swipe back and forth between the tabs
4. Tap the link on the "Devotional" tab, it should still navigate
5. Tap a tab, it should still navigate

**Tabs**

1. Load up an IOS device
2. Navigate to a devotional content item
3. Tap in the space above the words on the tabs, your tap should pass through to the tabs
4. Tap the back button, it should still navigate back
5. Reenter the devotional
6. Tap the X button, it should still close

## SCREENSHOTS

|Navigation|Tabs|
|---|---|
|Screen Recording 2019-10-17 at 11 37 AM|Screen Recording 2019-10-17 at 03 11 PM|

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


erikkgagewhitt:
@burkeshartsis Yup! Working on writing one up now.

Issue #1184, opened 2019-11-03
HTML Renderer Handles Bold Text


redreceipt:
### User Story

As a content writer, I should be able to use `<b>` tags to bold text.

### Description

Currently, the renderer can't handle `<b>` tags, only `<hX>` tags. The Rock editor only has a "Bold" button so unless you paste in the HTML, you can't bold the text without adding header tags in the code.

### Acceptance Criteria

* [ ] It should [SPEC_DETAIL]

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

vinnyjth:
Fixed in #1185

PR #1185, opened 2019-11-03
Adds <b> and <i> tag support to HTMLView

redreceipt:

Adds the ability to use `<b>` tags in content, that's what Rock uses for bold text in its GUI. We can already handle `<strong>`.

None.

### How do I test this PR?

Add some content with a `<b>` tag.

One snap.

## TODO:

- [x] Closes [#1184]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
> It works and while I understand we don't have a ton of control over what markup Rock decides to 🤮at us but I'll leave this here so at least it's documented.
> 
> https://www.w3.org/International/questions/qa-b-and-i-tags.en
> 
> > You should always bear in mind that the content of a b element may not always be bold, and that of an i element may not always be italic. The actual style is dependent on the CSS style definitions. You should also bear in mind that bold and italic may not be the preferred style for content in certain languages.
> > You should not use b and i tags if there is a more descriptive and relevant tag available. If you do use them, it is usually better to add class attributes that describe the intended meaning of the markup, so that you can distinguish one use from another.

put in a Rock issue citing that article, seems legit

vinnyjth:
Tests are failing here, will need them fixed before I can merge :)

redreceipt:
and while we're here, we might as well cover this too... :(
Screen Shot 2019-11-04 at 3 37 11 PM

Issue #1175, opened 2019-10-30
MapView Zooms In Too Far


redreceipt:
### Details

When you open the `MapView` from the user settings page, the results are very inconsistent. Sometimes it zooms all the way in on just my location, sometimes it's all the way on just one campus, and sometimes it works as intended (showing my location and all campuses zoomed out)

![Screenshot_20191030-132550](https://user-images.githubusercontent.com/2659478/67882719-12d1bc00-fb19-11e9-81bd-3cf4140df9b5.png)

### Possible Fix

<details><summary>Additional Info</summary>

### Your Environment

- Environment name and version (e.g. Chrome 39, node.js 5.4):
- Operating System and version (desktop or mobile):
- Link to your project:

### Steps to Reproduce

1.  first...
2.
3.
4.

### Stack Trace

</details>

vinnyjth:
I believe we fixed this in #1189

PR #1189, opened 2019-11-04
Fixes Map Zoom to Nothing

redreceipt:
## DESCRIPTION

Screen Shot 2019-11-04 at 1 44 12 PM

An empty array of campuses was getting passed to the `MapView` on the first time the app loaded resulting in the view zooming all the way in, only on the user location. This just adds a loading state until the `getCampuses` query comes back.

Map won't load while `campuses` query is pending.

### How do I test this PR?

Uninstall the app and then test on the first time you bring up the locations view, it shows the whole US with all campuses.

None.

## TODO:

- [x] Closes [#1175]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apolloschurchapp/src/user-settings/Locations/index.js
```diff
@@ -64,7 +64,7 @@ class Location extends PureComponent {
           },
         });
       },
-      () => null,
+      (e) => console.warn(e),
```
redreceipt:
```suggestion
      (e) => console.warn('Error getting location!', e),
```

vinnyjth:
Another idea - 

https://github.com/ApollosProject/apollos-prototype/blob/update-load-tests/packages/apolloschurchapp/src/user-settings/Locations/MapView.js#L85-L89

Can you change this line so that it properly `updatesCoordinates` if new campuses are loaded in?

redreceipt:
even better!

Screen Shot 2019-11-04 at 4 07 29 PM

packages/apolloschurchapp/src/user-settings/Locations/MapView.js
```diff
@@ -83,7 +83,10 @@ class MapView extends Component {
   }
 
   componentDidUpdate(oldProps) {
-    if (oldProps.userLocation !== this.props.userLocation) {
+    if (
```
vinnyjth: I think this will be okay for this use case. It becomes inaccurate though if you ever end up in a situation where campuses change, so the length stays the same (updating the elements of an array don't always update the array itself). I've found the most fool-proof way to do this is use `differenceBy ` and make sure the length is 0 :) (ie: check to make sure the arrays have the same content)

vinnyjth: I'm okay merging without this change though, just a tip for future reference :) 

`lodash.differenceBy`

richarddubay:
@vinnyjth In both the core app and the NewSpring app, I'm seeing a distinct bounce on the map when the campuses load. This obviously doesn't look great and isn't a great experience. I'll post a couple of videos here for reference. Anything we can do about that? 

core
newspring

redreceipt:
we can just not run `updateCoordinates` if there's no campuses, @richarddubay if you make an issue, I'll fix it.

PR #1215, opened 2019-11-11
Adds Total Count to Paginated Queries

redreceipt:
## DESCRIPTION

Screen Shot 2019-11-11 at 3 40 36 PM

This adds the total count to all paginated queries and adds it to the schema for `ContentItemsConnection`, useful for knowing how many items are in a series (and many other applications)

Adds one call to every paginated call. Don't think it's that expensive but want to mention it.

### How do I test this PR?

```
{
  userFeed {
    totalCount
    edges {
      node {
        title
      }
    }
  }
}
```

None.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
these tests are failing because we're snapping all calls made from the `paginate` function but when I log `get.mock.calls` only the first one (my ID select call) is showing. The second *real* get call isn't there. 😕

vinnyjth:
I really like the idea of this PR. However, I think we need to be very careful that we only fetch the `totalCount` when the user is asking for the total count. There's a few ways to do that, but you'll most likely have to write a resolver for `totalCount` that fetches the count via a curried function or some other method.

redreceipt:
> I really like the idea of this PR. However, I think we need to be very careful that we only fetch the `totalCount` when the user is asking for the total count. There's a few ways to do that, but you'll most likely have to write a resolver for `totalCount` that fetches the count via a curried function or some other method.

I can do that. It'll make the snaps pass too.

redreceipt:
@vinnyjth ok what do you think of this?

packages/apollos-rock-apollo-data-source/src/index.js
```diff
@@ -73,8 +73,20 @@ export default class RockApolloDataSource extends RESTDataSource {
       }
     }
 
+    // temporarily store the select parameter to
+    // put back after "Id" is selected for the count
+    const tempSelect = cursor.query.select;
```
vinnyjth: Do you still need this line?

vinnyjth: Oh, you might :/ depends on which of those run first. 

@redreceipt How would you feel about updating `cursor` so you can call `.clone` on a cursor or `.count` on a cursor and get the same data back?

redreceipt: don't love it because our count algorithm `select=Id` is specific to rock so it belongs in the `data-rock` package

vinnyjth: Our `request-builder` is specific to `rock/odata`

redreceipt: honestly, i may not need that temp cursor swap anymore, there may a smoother way to do it. but i have a hunch it's still necessary. JS classes are confusing.

redreceipt: if i was gonna update `cursor.count` i'd prefer it be the `$count` param which works in OData but isn't supported in Rock

vinnyjth: I can take a look at doing this in our request builder - I think it would be a nice addition even if we can't use `cursor.count`

redreceipt: I agree, it should be fixed in Rock anyway at which time we can take this hacky stuff out

redreceipt:
@vinnyjth 💪

PR #1221, opened 2019-11-13
Actually logs users with bad cookies out of the app

vinnyjth:

Previously, if a user had an invalid cookie, they wouldn't get logged out of the app properly for two reasons: 

1. Invalid cookies don't throw a Authentication error, they throw a 500 error. 
2. Even when an authentication error was thrown, the `authToken` local storage state wasn't being cleared. 

Both of those issues are fixed in this PR. 

n/a

### How do I test this PR?

🤔
1. Change Auth Datasource to throw an Authentication error, and browse the app and make sure you get logged out? 
2. Hardcode a bad cookie in the client, make sure an Authentication error is thrown in Rock (rather than a 500 error) 

n/a

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-data-connector-rock/src/auth/data-source.js
+++ b/packages/apollos-data-connector-rock/src/auth/data-source.js
@@ -31,13 +31,17 @@ export default class AuthDataSource extends RockApolloDataSource {
     }
 
     if (userCookie) {
-      const request = await this.request('People/GetCurrentPerson').get({
-        options: {
-          headers: { cookie: userCookie, 'Authorization-Token': null },
-        },
-      });
-      this.context.currentPerson = request;
-      return request;
+      try {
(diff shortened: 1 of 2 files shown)


vinnyjth:
@conrad-vanl @richarddubay 

I think this might be one reason we are seeing so many errors. I got into a state locally in NS where I was logged out, but my auth token was still in local storage. I couldn't login, or _really_ logout (I would have had to delete the app) 

After making the changes above to the NS app (I made them via editing the node_modules on the api/client) I was logged out successfully when encountering my first auth error.

vinnyjth:
@conrad-vanl Should clear up some of them. @richarddubay You can get these updates either via a canary build, or I can roll a release for you sometime in the next week.

PR #1222, opened 2019-11-14
Check for Valid Images When Rendering Campus Cards.

erikkgagewhitt:

Adds a check to the CampusCard component that ensures that all images passed to it are indeed valid.

This means it checks that the array exists, that none of the objects in the array are equal to `null`, and that the `uri` key in the objects is not null.

The change is intended to fix an issue that occurs in the New Spring app, but the protection should be added to core, thus this PR!

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-kit/src/CampusCard/index.js
+++ b/packages/apollos-ui-kit/src/CampusCard/index.js
@@ -26,6 +26,9 @@ const CampusImage = styled({
   resizeMode: 'cover',
 })(ConnectedImage);
 
+const hasValidImages = (images) =>
+  images && !images.includes(null) && !images.map((o) => o.uri).includes(null);
+
 const CampusCard = memo(
   ({ title, description, distance, images, ...otherProps }) => (
     <Card
@@ -34,7 +37,7 @@ const CampusCard = memo(
       {...otherProps}
     >
       <HorizontalLayout>
-        {images ? <CampusImage source={images} /> : null}
+        {hasValidImages(images) ? <CampusImage source={images} /> : null}
         <FlexedCardContent>
           <Header>
             <H5>{title}</H5>


erikkgagewhitt:
@conrad-vanl It looks like there's more than just this crash. Some of NewSpring's components (like Marker) require values from a campus.

The campus in question ("Web") has all of its values set to null and so crashes any component that uses it.

conrad-vanl:
Nice @erikkgagewhitt .

@richarddubay what do you think? I know you said the web campus shouldn't show in the list but it looks like that's the issue. Any idea what changed on y'alls end?

We're happy to look at merging this if you want.

richarddubay:
It's weird that "Web" is showing up for you, but I guess that would make sense that it would crash if it does. There is nothing at all set up in Rock for that campus.

Can you show me where you're seeing "Web" show up? I ran the `getAllCampuses` query and it only returns the 14 active and public campuses.

richarddubay:
So I did a little more research. It turns out that I could get the app to freeze on the settings screen after tapping "Location" by changing my campus to "Web". The `getAllCampuses` query then returns the Web campus with an image of "null" for `currentUser`. It seems to make sense that this change would then stop the app from freezing, but I can't think of a really good way to test it.

erikkgagewhitt:
Interesting; when I signed up for the app, I skipped the campus step. Does that cause the user's campus to be set as "Web"?

richarddubay:
@erikkgagewhitt Yep! It's our catch against a person not having any campus set at all.

conrad-vanl:
Woah @vinnyjth I thought `mergify` wouldn't merge if there were any failing checks?

redreceipt:
> Woah @vinnyjth I thought `mergify` wouldn't merge if there were any failing checks?

It adheres to our github rules. you can merge with codecov fails but not travis test fails.

conrad-vanl:
ah ok

Issue #1213, opened 2019-11-11
Scripture copyright information displays multiple times on devotional


richarddubay:
### Details
On the scripture tab of a devotional that has multiple scriptures listed, the copyright for the translation the app is using appears once for each scripture listed, thereby showing the same copyright information multiple times.

### Expected Behavior
The scriptures on a devotional would show the copyright once for all the scriptures listed.

### Actual Behavior
The scriptures on a devotional show the copyright multiple times, once for each scripture listed.

### Possible Fix
Remove the display of the copyright information from the `<ScriptureItem>` and instead display it after mapping over the scriptures in the `<ScriptureTab>` component.

### Steps to Reproduce
1. Change the bible version in your `config.yml` to be something with a copyright. I'm using KJV in the screenshot below.
2. Pull up a devotional that has multiple scriptures listed. You can see that I'm reading "Change starts with a choice" from the Jeremiah study.
3. Go to the Scripture tab and scroll down to the bottom where you can see the end of the first scripture and the start of the second. There you will see the copyright information displayed twice.

### Screenshot
![Screenshot 2019-11-11 08 31 21](https://user-images.githubusercontent.com/3229463/68591453-c9f5fd80-045e-11ea-962b-bb2120aff653.png)

richarddubay:
@vinnyjth @burkeshartsis I wanted to tag you here for reference. I'm going to submit a PR to fix this as I stated above but if you guys have a different/better idea on how to fix this, please let me know!

richarddubay:
Actually, we probably need to leave the ability to pass copyright information to `<ScriptureItem>` because of the scripture feature. But we can change the `<ScriptureTab>` to not pass it down so that it won't display on each scripture in a devotional. Yeah, that sounds like a plan.

PR #1170, opened 2019-10-29
Improved Caching of Content

redreceipt:
### What does this PR do, or why is it needed?

Content was never being updated after the initial pull. This fixes that.

### What design trade-offs/decisions were made?

More data calls.

### How do I test this PR?

- look at some content
- update some content
- look at it again

## TODO:

- [x] PR has a relevant title that will be understandable in a public changelog (ie...non developers)
- [x] Closes [tag-issues-here]
- [x] No new warnings in tests, in storybook, and in-app
- [x] Upload GIF(s) of iOS and Android
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

- [ ] Manual QA on iOS and ensure it looks/behaves as expected
- [ ] Manual QA on Android and ensure it looks/behaves as expected

**Code Review: Questions to consider**

- [ ] Read through the "Files changed" tab _very carefully_
- [ ] Edge cases: what assumptions are made about input?
- [ ] What kind of tests could be written?
- [ ] How might we make this easier for someone else to understand?
- [ ] Could the code be simpler?
- [ ] Will the code be easy to modify in the future?
- [ ] What's one part of these changes that makes you excited to merge it?

> The purpose of PR Review is to _improve the quality of the software._

PR #1230, opened 2019-11-18
Removes Duplicate Copyright Info

redreceipt:
## DESCRIPTION

Screen Shot 2019-11-18 at 11 29 34 AM

This removes duplicate copyright info from showing up on devotionals and exports `LegalText` for the client to use.

Will only show the copyright from the *first* scripture in the array. Not a problem right now but if we ever show scriptures from multiple translations in a devo, we'll have to figure it out.

### How do I test this PR?

Temporailily turn off the "PUBLIC DOMAIN" check (that's what I did for the screenshot)

None.

## TODO:

- [x] Closes [#1213]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apolloschurchapp/src/content-single/DevotionalContentItem/ScriptureTab.js
```diff
@@ -19,10 +19,11 @@ const ScriptureTab = ({ id, scripture, isLoading, navigation }) => (
           reference={ref.reference}
           html={ref.html}
           isLoading={isLoading}
-          copyright={ref.copyright}
           version={ref.version}
         />
       ))}
+      {/* will only show copyright from first reference */}
+      <LegalText>{!isLoading ? scripture[0].copyright : null}</LegalText>
```
richarddubay: This is exactly what I was thinking @redreceipt! 🔥

richarddubay:
This would fix https://github.com/ApollosProject/apollos-prototype/issues/1213

packages/apolloschurchapp/src/content-single/DevotionalContentItem/ScriptureTab.js
burkeshartsis: What if you show two different scripture versions? Wouldn't you need to show the reference for each version used?

burkeshartsis: Also do we need to show "Public Domain" at all?

redreceipt: @burkeshartsis that's the trade-off I mention above, currently, it's not a problem because core only picks one version for the devotionals.

No, we do not need to show "Public Domain" and we don't.

redreceipt:
@burkeshartsis would you prefer your suggested implementation or me fix my version with your comments? It's all good stuff, doesn't matter to me either way.

Also, you're right about the tests, they are messed up. One of the tests is the same with a different title for instance. Should probably be fixed in a separate PR.

redreceipt:
actually i fixed these tests in #1170

burkeshartsis:
@redreceipt if you were to ask me which fix I would prefer I think it would be mine given that it touches less files and preserves the functionality we already have.

redreceipt:
> @redreceipt if you were to ask me which fix I would prefer I think it would be mine given that it touches less files and preserves the functionality we already have.

OK. I agree but I'll have to modify your algorithm slightly as it will currently show the first copyright, not the last.

redreceipt:
> @redreceipt if you were to ask me which fix I would prefer I think it would be mine given that it touches less files and preserves the functionality we already have.

...this is awkward. I forgot I need `LegalText` in order to show the copyright at the bottom of NewSpring's sermon notes instead of in the middle. If it still pleases the court, I'd like to go with my implementation so partners can use core scripture text styling.

For what it's worth, I think it will be useful if we ever do a fancier scripture reading application in core. We may not *always* want copyright info directly after the text (like the Read Scripture app).

redreceipt:
ok @burkeshartsis give this a look

vinnyjth:
After @burkeshartsis dismisses his review, or (requests more changes) we can merge this and run a release.

PR #1025, opened 2019-08-28
Adds Parents to Devotionals and Sermons

redreceipt:
## DESCRIPTION

Screen Shot 2019-08-28 at 12 45 30 PM

### What does this PR do, or why is it needed?

Adds a `series` field to sermon and devotional content items to get their parent series content items.

### What design trade-offs/decisions were made?

None.

### How do I test this PR?

Run this:

```
{
  userFeed {
    edges {
      node {
        ... on WeekendContentItem {
          id
          series {
            title
          }
        }
        ... on DevotionalContentItem {
          id
          series {
            title
          }
        }
      }
    }
  }
}
```

### What automated tests did you write?

None.

## TODO:

- [x] PR has a relevant title that will be understandable in a public changelog (ie...non developers)
- [x] Closes [#1023]
- [x] No new warnings in tests, in storybook, and in-app
- [x] Upload GIF(s) of iOS and Android
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

- [ ] Manual QA on iOS and ensure it looks/behaves as expected
- [ ] Manual QA on Android and ensure it looks/behaves as expected

**Code Review: Questions to consider**

- [ ] Read through the "Files changed" tab _very carefully_
- [ ] Edge cases: what assumptions are made about input?
- [ ] What kind of tests could be written?
- [ ] How might we make this easier for someone else to understand?
- [ ] Could the code be simpler?
- [ ] Will the code be easy to modify in the future?
- [ ] What's one part of these changes that makes you excited to merge it?

> The purpose of PR Review is to _improve the quality of the software._

Issue #1022, opened 2019-08-28
Get Sermon Week Number


redreceipt:
### User Story

As a developer, I want to know what week in a series the current sermon is so I can do things with it

### Description

I'd like the server to return a new field called `seriesIndex` or something so I can know what week number a sermon is in the series. 

### Why

- It will allow us to gray out the current sermon in the horizontal scroll bar
- it will allow us to show "Week Number X of Y" on the content item

### Acceptance Criteria

* [ ] It should expose a new field on the `WeekendContentItem` node called `seriesIndex`

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

redreceipt:
@vinnyjth what do think about the suggestion above? `siblingContentConnection.currentIndex` and `sibilingContentConnection.totalItems`

vinnyjth:
> @vinnyjth don't we already do something like this in core?
Yep! That is working fine in core. 

`sibilingContentConnection.totalItems` is done in Relay, so we're off to a good start there :) In that context, I think I would be okay doing a `currentIndex`.

However, `sibilingContentConnection` is just a `ContentItemsConnection`, which makes this more difficult. I'm still struggling to understand why, if you're going to add a "currentIndex" there, why not do that on the client? I can understand it being necessary to have that data on the server if you ever want to load just the middle of a list and have the correct index, but we don't do that anywhere in the app yet.

redreceipt:
Only for the chance that we may want to know its index outside of the context of a list. We could use the same list of IDs from the `totalCount` algorithm and `index` would just be null if it wasn't in there, as would be the case for `childContentConnection`.

redreceipt:
> `sibilingContentConnection.totalItems` is done in Relay, so we're off to a good start there :)

Not sure what Relay is but isn't this already done with our `totalCount` field?

vinnyjth:
It is :) I would encourage you to try and implement what you need on the front end without a schema index first, and if you run into a situation where it's impossible to get the data you need, and you know that the "sort" index from the API would help you out, we can add that field to the API.

PR #1232, opened 2019-11-18
Adds Index to Paginated Content

redreceipt:
## DESCRIPTION

Screen Shot 2019-11-18 at 3 21 11 PM

Simply adds an `index` field to `ContentItemsConnectionEdge` so you know where in the full list you are. Useful for lots of things.

None.

### How do I test this PR?

```
{
  userFeed(after:"7c613b3d35342b1bb0c53c1089ac11c6") {
    edges {
      cursor
      index
      node {
        title
      }
    }
  }
}
```

None.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-church-api/local.graphql
+++ b/packages/apollos-church-api/local.graphql
@@ -141,6 +141,7 @@ type ContentItemsConnection {
 
 type ContentItemsConnectionEdge {
   node: ContentItem
+  index: Int
   cursor: String
 }
 
--- a/packages/apollos-data-schema/src/index.js
+++ b/packages/apollos-data-schema/src/index.js
@@ -377,6 +377,7 @@ export const contentItemSchema = gql`
 
   type ContentItemsConnectionEdge {
     node: ContentItem
+    index: Int
     cursor: String
   }
 
--- a/packages/apollos-rock-apollo-data-source/src/index.js
+++ b/packages/apollos-rock-apollo-data-source/src/index.js
@@ -80,6 +80,7 @@ export default class RockApolloDataSource extends RESTDataSource {
           .transform((result) =>
             result.map((node, i) => ({
               node,
+              index: i + skip,
               cursor: createCursor({ position: i + skip }),
             }))
           )


vinnyjth:
You're going to have to sell me on this field, I don't see the utility.

conrad-vanl:
@vinnyjth they want to be able to add the number of an item in the horizontal content feeds. Like the current app has a little number on each tile in the horizontal content feeds indicating which number that item is in the series

conrad-vanl:
my worry would be rather this is something that is feasible on large lists

conrad-vanl:
dang, don't accidentally hit "close and comment" instead of "comment", mergify will lay down the law!

redreceipt:
yea it wouldn't be used on big lists. it could I guess and there would be no performance hit if you did.

redreceipt:
> You're going to have to sell me on this field, I don't see the utility.

"Week 1 of 7"

vinnyjth:
I think I would prefer if this was implemented as a `seriesInfo: String` field on the `ContentItem`. Thoughts on that @conrad-vanl?

vinnyjth:
@redreceipt https://facebook.github.io/react-native/docs/flatlist#renderitem `renderItem` passes an Index. I would use that if you're rendering the devotional number. At this point we never fetch the "last 20 devotionals". We always fetch the first 20, and then paginate further as a user scrolls.

redreceipt:
> @redreceipt https://facebook.github.io/react-native/docs/flatlist#renderitem `renderItem` passes an Index. I would use that if you're rendering the devotional number. At this point we never fetch the "last 20 devotionals". We always fetch the first 20, and then paginate further as a user scrolls.

I want to put "Week 4" on the sermon notes too, doesn't use flatlist

redreceipt:
oh but i suppose i could still pass in as a prop

redreceipt:
...wait no, not if it was a campaign item

redreceipt:
so i thought about doing it on `ContentItem` but the paginated field seemed much more multipurpose and *much* simpler

vinnyjth:
If you do it on the pagination the index will be wrong (won’t reflect the week) on the featured item (and everywhere else)

redreceipt:
Yea, you’re right. It should totally be `contentItem.seriesIndex`, similar to my `series` field. If core doesn’t want this one either, you can close, I’ll just put it in NewSpring code.

—
Michael

vinnyjth:
Core would love the ability to tell which week / index an item is, with in a series, I just want to make sure we use the best approach possible to do that :)

redreceipt:
> Core would love the ability to tell which week / index an item is, with in a series, I just want to make sure we use the best approach possible to do that :)

Do you not think it goes hand in hand with merging #1025?

redreceipt:
closing this, it's not the right strategy, we can continue the convo in #1022

PR #1234, opened 2019-11-19
Adds Device Schema

redreceipt:
## DESCRIPTION

Screen Shot 2019-11-19 at 3 33 02 PM
Screen Shot 2020-02-12 at 4 09 04 PM

- Lists devices of a user
- Toggles notifications via Rock

Don't think any.

### How do I test this PR?

Good luck ;)

A few.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
@redreceipt I'm all for micro PR's, but this might be a bit _too_ micro :) Throw a resolver and a datasource with tests in here, and I'll merge this. Your schema looks good, but I would have `Devices`  be it's own data type inside `rock-apollo-data-source`

redreceipt:
I was hoping you'd say that :) I can't quite figure out how we're passing adding devices to Rock so I'm not sure how to resolve it. Wanna point me in the right direction?

redreceipt:
> @redreceipt I'm all for micro PR's, but this might be a bit _too_ micro :) Throw a resolver and a datasource with tests in here, and I'll merge this. Your schema looks good, but I would have `Devices` be it's own data type inside `rock-apollo-data-source`

Also, why not `data-connector-rock`? Help me understand the difference.

vinnyjth:
What do you know! We already have a model for this :D 
https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-data-connector-rock/src/personal-devices/data-source.js#L6

vinnyjth:
> Also, why not data-connector-rock? Help me understand the difference.

Blarg! My bad, I meant `data-connector-rock`

redreceipt:
ok @vinnyjth take a look at this

vinnyjth:
@redreceipt For the two new mutations, do you think you could piggy back off the existing `updateUserPushSettings` mutation instead? Besides that, this is looking good so far! I'll give it a more thorough review tomorrow.

redreceipt:
> @redreceipt For the two new mutations, do you think you could piggy back off the existing `updateUserPushSettings` mutation instead? Besides that, this is looking good so far! I'll give it a more thorough review tomorrow.

I saw that and was tempted but that field is in reference to OneSignal settings. This would be dealing with Rock bit flips. I suppose we could but now that we know Rock has data we can play with, maybe it would be better to strategize some better naming for push notification stuff and go from there?

vinnyjth:
@redreceipt 

That `PushSettingsInput` field has an `enabled` field inside it that currently isn't used. I would prefer you wire up the `enabled` field instead of adding two new mutations. 

While the implementation of that mutation is based on OneSignal settings, that isn't exposed in the schema, so I think we're okay.

redreceipt:
> @redreceipt
> 
> That `PushSettingsInput` field has an `enabled` field inside it that currently isn't used. I would prefer you wire up the `enabled` field instead of adding two new mutations.
> 
> While the implementation of that mutation is based on OneSignal settings, that isn't exposed in the schema, so I think we're okay.

So you think for that field, even though it lives in `data-connector-onesignal`, just call a ds function from `PersonalDevice` class? Doesn't that create a cross dependency for churches to have OneSignal *and* Rock npm packages? I very well could be overthinking...

vinnyjth:
> So you think for that field, even though it lives in data-connector-onesignal, just call a ds function from PersonalDevice class? Doesn't that create a cross dependency for churches to have OneSignal and Rock npm packages? I very well could be overthinking...

We already have encountered that issue with our existing rock implementation, and solved it by overriding the resolver explicitly in `apollos-church-api`

https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-church-api/src/data/oneSignalWithRock.js#L5

redreceipt:
Nice!! Do you agree that we should be toggling notifications at the database and not the push service? I think so but wanna make sure it’s the best move because this will be relatively set in stone.

—
Michael

vinnyjth:
> Nice!! Do you agree that we should be toggling notifications at the database and not the push service?

Yep! We need to be toggling the notifications enabled/disabled using rock. Toggling in onesignal doesn't let Rock know that user has or hasn't enabled push notifications.

redreceipt:
@vinnyjth I finally finished this one to get it out my inbox 😆 it actually turned out pretty cool

Screen Shot 2020-02-12 at 4 09 04 PM

conrad-vanl:
@redreceipt question - obviously this is written from our end to require a user. I know the Devices entity in Rock doesn't actually require them to be connected to a person. I know all of the current Apps require sign in. We're starting to get requests from community to not requires sign-in for all apps, I know atleast Watermark Porch is going to launch without sign in at all. What would you think about allowing the creation of devices without having to be tied to a user?

redreceipt:
> @redreceipt question - obviously this is written from our end to require a user. I know the Devices entity in Rock doesn't actually require them to be connected to a person. I know all of the current Apps require sign in. We're starting to get requests from community to not requires sign-in for all apps, I know atleast Watermark Porch is going to launch without sign in at all. What would you think about allowing the creation of devices without having to be tied to a user?

I think that's a valid use case but I don't think it would be wise to allowing for *all* devices anyway. If you're suggesting the ability to add devices without a userID attached, we *may* already be able to do that. But the schema here is valid I think so I can enable/disable notifications. Or else we're gonna need an `UnauthenticatedUser` on the schema with the ID being the OneSignal generated user ID or something.

conrad-vanl:
I want to review this PR. But I feel I'm only understanding a subsection of it. I'll have to push back over to @vinnyjth for a review.

redreceipt:
@vinnyjth added some more coverage, this one is ready now

FrankGrand:
@vinnyjth @conrad-vanl Can we prioritize this getting looked at?

FrankGrand:
@vinnyjth @conrad-vanl When would be a reasonable expectation for this to be looked at again?

packages/apollos-data-schema/src/index.js
```diff
@@ -112,6 +113,14 @@ export const peopleSchema = gql`
...
 `;
 
+export const deviceSchema = gql`
+  type Device implements Node {
+    id: ID!
+    pushId: String!
+    notificationsEnabled: Boolean!
+  }
+`;
+
```
vinnyjth, line 123: rather than having 

`device: [Device]` inside the People schema, do this inside the `deviceSchema`

```
extend type Person {
  devices: [Device]
}
```

Then you avoid needing to include `deviceSchema` inside all your tests :)

redreceipt: brilliant

redreceipt:
@vinnyjth I made that change, seemed simple enough but now a test is failing and I don't fully understand why, any ideas?

Screen Shot 2020-07-22 at 1 28 39 PM

redreceipt:
oh weird, it's failing locally...

redreceipt:
nuked and it's passing locally now

Issue #901, opened 2019-07-22
Sermon Thumbnails


redreceipt:
<img width="447" alt="Screen Shot 2019-05-30 at 10 57 22 AM" src="https://user-images.githubusercontent.com/2659478/61663385-82c0e280-ac9e-11e9-8057-3315c3b5f989.png">

### User Story

As a user I want to see an engaging image to click in sermon series content for each sermon.

### Description

Currently sermons are just gray boxes. They should be pretty.

### Why

Because it's ugly.

### Acceptance Criteria

* [ ] It should give sermons images on the tiles inside sermon series content

### Possible Implementation & Open Questions

Maybe they could just default to the parent unless a thumbnail is in Rock.

### Other Resources

* Design mockup/prototype link
* Basecamp todo or link to "epic" on GitHub

redreceipt:
@conrad-vanl you care if I take a swing at this? My strategy is:

- Add `ContentItem.thumbnailImage` field
- Pull from a custom attribute in Rock
- Add a prop to `ContentCardConnected` to use thumbnail instead of `coverImage`

conrad-vanl:
sounds good to me! Should this prop also be used for the video thumbnail player?

conrad-vanl:
and i'm not sure that `ContentCardConnected` shuold always use `thumbnailImage`, I think just horizontal content cards should?

redreceipt:
> and i'm not sure that `ContentCardConnected` shuold always use `thumbnailImage`, I think just horizontal content cards should?

I'm sure you're right. I meant the ones at the bottom of a content-single, took a guess :)

redreceipt:
> sounds good to me! Should this prop also be used for the video thumbnail player?

not sure what that is...

conrad-vanl:
The "Play" thumbnail on content singles that you tap on to start the video

conrad-vanl:
> Currently sermons are just gray boxes. They should be pretty.

They are only grey boxes on the newspring app 😉

redreceipt:
> The "Play" thumbnail on content singles that you tap on to start the video

Word. I made an [issue](packages/apollos-ui-scripture/src/index.js) for that too. This solves half of it.

redreceipt:
> > Currently sermons are just gray boxes. They should be pretty.
> 
> They are only grey boxes on the newspring app 😉

Dang. Cold-blooded. I'll look into why we don't match.

redreceipt:
[...]

PR #1237, opened 2019-11-19
Adds Thumbnails

redreceipt:
## DESCRIPTION

Screen Shot 2019-11-19 at 2 19 31 PM

Adds `thumbnailImage` to content items so we can show them instead of series artwork in the app.

None.

### How do I test this PR?

Bring up the "Gathering" series, you should see the "Spur One Another" sermon have a thumbnail.

None.

## TODO:

- [x] Closes [#901]
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-church-api/local.graphql
+++ b/packages/apollos-church-api/local.graphql
@@ -272,6 +272,7 @@ type MediaContentItem implements ContentItem & Node {
   parentChannel: ContentChannel
   theme: Theme
   scriptures: [Scripture]
+  thumbnailImage: ImageMedia
   isLiked: Boolean
   likedCount: Int
   sharing: SharableContentItem
@@ -528,6 +529,7 @@ type WeekendContentItem implements ContentItem & Node {
   siblingContentItemsConnection(first: Int, after: String): ContentItemsConnection
   parentChannel: ContentChannel
   theme: Theme
+  thumbnailImage: ImageMedia
   isLiked: Boolean
   likedCount: Int
   liveStream: LiveStream
--- a/packages/apollos-data-connector-rock/src/content-items/data-source.js
+++ b/packages/apollos-data-connector-rock/src/content-items/data-source.js
@@ -87,6 +87,19 @@ export default class ContentItem extends RockApolloDataSource {
     }));
   };
 
+  getThumbnailImage = ({ attributes, attributeValues }) => ({
+    __typename: 'ImageMedia',
+    key: 'videoThumbnail',
+    name: attributes.videoThumbnail.name,
+    sources: attributeValues.videoThumbnail.value
+      ? [
+          {
+            uri: createImageUrlFromGuid(attributeValues.videoThumbnail.value),
+          },
+        ]
+      : [],
+  });
+
   getAudios = ({ attributeValues, attributes }) => {
     const audioKeys = Object.keys(attributes).filter((key) =>
       this.attributeIsAudio({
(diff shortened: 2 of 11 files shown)


vinnyjth:
@redreceipt Help me understand this: why not use the cover images? Do you want the cover image to always show the series image, and then be able to show a different image inside the horizontal card view?

vinnyjth:
Here's what I see in core right now
image

redreceipt:
I would assume we'd want the artwork on the content single header and save the thumbnails for the "Play" button and the horizontal tile list.

@kevin-birdy have thoughts on needing thumbnails and series artwork?

vinnyjth:
@redreceipt Yah, I would be curious to know what @kevin-birdy thinks about needing two images for cards. For now @redreceipt you might be better off doing this in your own app.

redreceipt:
Actually I don't think we do now that I think about it. Client side, you could either show the `coverImage` or the parents `coverImage`

vinnyjth:
Yep :D So I think we are good here. Feel free to reopen if I'm wrong.

conrad-vanl:
FWIW, while i'm not sure this PR is the right way to go about it, I do think a secondary image in our schema might be beneficial. and `thumbnailImage` might be the right way to think about it - essentially we're learning that our ui has two use cases for images - one where the image is presented big and boldly, and the other where it's presented small with text laid on top of it. It is hard to produce images that work well for both use cases.

kevin-birdy:
seconding @conrad-vanl 's comments. It's basically impossible to have a catch-all solution, so two images provides some flexibility that will prove useful.

One other thing I want to add on to this. It might be a separate issue but it feels the same. For video content items, having a `thumbnailImage` would be great because then we wouldn't have two of the same images stacked on top of each other.

redreceipt:
yes and yes, but the question is couldn't we just use the parent's cover image as the artwork and the sermon cover image as the thumbnail?

redreceipt:
...unless you would ever want a sermon series to have a trailer or something

redreceipt:
> One other thing I want to add on to this. It might be a separate issue but it feels the same. For video content items, having a `thumbnailImage` would be great because then we wouldn't have two of the same images stacked on top of each other.

@conrad-vanl this is why I thought it might be good to have a `videos.thumbnail` field instead of a `ContentItem.thumbnailImage` field

kevin-birdy:
Does every sermon have a parent? As far as sermons go that makes a lot of sense to me. Also probably means that it is in fact a slightly different discussion. How it gets implemented is up to you guys. I am all for the value of a `thumbnailImage` though.

conrad-vanl:
Yes, you could use the parent and in that case I would want you to point the items coverImage to the parent item. You shouldn’t do that mapping on the client.  It sounds like we all agree content items need two images, at the least. Our schema does not support that clearly - it only has the one coverImage. @vinnyjth we should add this as a Validated feature request somewhere. Not something we need to implement immediately but something we should handle on core.

redreceipt:
I think a better solution is a key/value pair for videos in Rock:

```
videoUrl: video.mp4
thumbnailUrl: image.jpg
```

That's what wistia gets us and I got it working on our app last night with a very simple schema.

redreceipt:
that way you could have a separate thumbnail for every video on the content

redreceipt:
or else we probably shouldn't have `videos` be an array

PR #1246, opened 2019-11-21
Notifications hasPrompted State

erikkgagewhitt:

Handles the state where the user has already been prompted for notifications and has declined. I previously tried [this](https://github.com/ApollosProject/apollos-prototype/pull/1243), but found that I could not enable notifications, if a user had ever declined them (makes sense from a user protection standpoint, but is very obnoxious). I also tried using listeners, but to no avail.

Now instead, I disable the notifications button to show the choice has already been made; I disable the"skip" button and enable the "finish" button 

Not sure this solution is great design-wise, but it seemed to look alright and make sense to me.

### How do I test this PR?

1. Delete the app from your sim
1. In onboarding tap Enable Notifications and reject the hard ask
2. Log out
3. In onboarding the state of this screen should change

## SCREENSHOTS

Screen Shot 2019-11-21 at 3 47 54 PM

## TODO:

- [ ] Closes
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


erikkgagewhitt:
@vinnyjth Not quite as good, but fixes the same issue as https://github.com/ApollosProject/apollos-prototype/pull/1243 did

vinnyjth:
@burkeshartsis can you comment on the flow of this PR vs how you imagined the final flow?

conrad-vanl:
@erikkgagewhitt @vinnyjth I think I could be okay with this but we need to chnage the button text to communicate to user why it's disabled. +1 would be some help text going into more detail with a link that opened up settings.

@erikkgagewhitt do we know for sure in this case that notifications are disabled? IF so, I think MVP approach would be:

> Notifications disabled in settings

With a +1 Helptext added somewhere that says like:

> Open **settings** and navigate to APP_NAME to enable notifications

vinnyjth:
@erikkgagewhitt I would go ahead and make the changes @conrad-vanl Suggested to this PR before we merge :D

erikkgagewhitt:
@vinnyjth Button text added!

vinnyjth:
@erikkgagewhitt I'm testing this by building for Testflight (best way I've found to genuinely test notification setup) 

https://appcenter.ms/orgs/Apollos-Church/apps/Apollos-Church/build/branches/notifications-prompted-state/builds/712

erikkgagewhitt:
I just want to quickly layout my plan. I'm going to try my hand at a Provider/Consumer approach; hopefully this will fix the issues with cached data.

There are still some other issues you mention and I'm not sure this approach will fix them on its own. The issue is OneSignal subscriptions. I can find no way to reenable them after a user has disabled them. Even the user turning notifications back on does not enable subscriptions.

To fix these issues, I'd like to use the Permissions package to handle notifications permission. If a user rejects notifications, we will manually disable the subscription (this is the only method that can be reversed of disabling a subscription). We should then be able to use a Onesignal permission listener to manually enable/disable subscriptions as the user changes their settings.

erikkgagewhitt:
@vinnyjth I've not messed with a context/provider format before; when I do this, will I keep the same queries that we have now? It seems like I would, but I just want to make sure.

vinnyjth:
@erikkgagewhitt 
> I've not messed with a consumer/provider format before; when I do this, will I keep the same queries that we have now? It seems like I would, but I just want to make sure.

I think you can go either way :) I think one of the issues we are seeing is that the queries don't ever rerun against the resolvers, so you might run into issues if you try and use the queries as the backbone for the provider. I don't care either way, whatever you think produces the cleanest and most functional API I'm good with!

erikkgagewhitt:
Closed in favor of https://github.com/ApollosProject/apollos-prototype/pull/1266

PR #1248, opened 2019-11-22
Ensures that n+3 redis connections aren't created

vinnyjth:

`Bull` has some best practices for reducing the number of connections being created. Now, a maximum of 4 connections per server will be made to Redis. 

n/a

### How do I test this PR?

1. Run the api locally.
2. have redis setup locally. 
3. run `redis-cli client list` 
4. You should see 5 clients connected (4 for the server + 1 for the `client-list` command)
Screen Shot 2019-11-22 at 8 49 44 AM

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-data-connector-algolia-search/src/__tests__/__snapshots__/jobs.tests.js.snap
+++ b/packages/apollos-data-connector-algolia-search/src/__tests__/__snapshots__/jobs.tests.js.snap
@@ -36,9 +36,15 @@ exports[`Alolia search jobs must create a job and add it to the queue 3`] = `
 Array [
   Array [
     "algolia-full-index-queue",
+    Object {
+      "createClient": [Function],
+    },
   ],
   Array [
     "algolia-delta-index-queue",
+    Object {
+      "createClient": [Function],
+    },
   ],
 ]
 `;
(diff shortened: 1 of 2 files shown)


conrad-vanl:
@richarddubay you might notice this PR and this explanation for Redis connections:

Currently, bull, our redis queue manager (that is currently only used for search indexing) is using 3 redis connections per queue. We have two queues (the delta indexer and the full indexer). So that alone represents 6 redis connections per dyno. Then there's also 1 connection for the main redis connection, for a total of 7 redis connections per dyno. We were seeing 1 extra connection on top ofthat, perhaps for the heroku dashboard monitoring, but not sure yet. So anyways, if you have 5 dynos running, expect 35 connections.

This PR reduces that to 4 connections per dyno, dropping that down to 20 connections with 5 dynos running.

richarddubay:
@conrad-vanl @vinnyjth This is great. Still probably going to go with the next step up from free ... just in case. If I go look at our `staging` environment, I see that we have a max connections of 8 ... which makes sense based on what you just explained. Cutting that in half is 🔥.

PR #1249, opened 2019-11-25
Fix Featured Card Loading State

richarddubay:
This PR give a default loading state color to the `FeaturedCard`.

Screenshot 2019-11-26 07 56 07

It removes `theme.colors.primary` as the default loading state color. I used `theme.colors.lightSecondary` because the line color is `theme.colors.background.inactive` and I wanted the line to stand out from the background color.

### How do I test this PR?
Load the app and see the default loading state.

None.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
@richarddubay so after checking out core design resources, it appears @kevin-birdy purposefully designed this card to have a primary color loading state:

image
https://www.figma.com/file/YeoeIuFfIiQmUAUFKEGwS6/Apollos-Core-UI-Kit?node-id=74%3A1671

Therefore, I need to get @kevin-birdy's sign off on this change. @kevin-birdy  - the issue we're running into is on both Willow Creek and NewSpring, the featured item has a dynamic color based on the content that is being featured. We don't know what that color is until after the item is loaded. So for example on the NewSpring app, during the loading state the featured card is the newspring green, then it might jump to a different color (depending on what content is currently featured). That can create a jarring experience. @richarddubay's suggestion was to use a different color for the loading state. What are your thoughts?

@richarddubay could you provide a screenshot of what the card looks like during loading in this PR?

richarddubay:
Yep ... totally meant to put a screenshot. 🤦‍♂️

So this is what we currently see on the NewSpring app when the home feed is loading:
Screenshot 2019-11-26 07 53 44

As you can see, it uses our `theme.colors.primary` and that's great, if everything we had in the featured card was that color.

But now that we're using a dynamic `primary` color, once it finishes loading, it switches out to something like this:
Screenshot 2019-11-26 07 53 52

Which is quite different than the original `primary` color.

So what I'm proposing is something that more closely resembles the loading states everywhere else in the app:
Screenshot 2019-11-26 07 56 07

The color switch may still feel "jarring", but I don't feel like it's as big an issue since all the loading states do this.

Whatcha think?

kevin-birdy:
Makes total sense! Let's just use `lightSecondary` so that these loading states match `HighlightCard` then.

Screen Shot 2019-11-26 at 9 37 55 AM

richarddubay:
@kevin-birdy Sweet! `lightSecondary` was what I chose! 😍

conrad-vanl:
@richarddubay are you able to make the text placeholders white? And would you know how we can drop the gradient during loading, to more closely match teh screenshot?

richarddubay:
@redreceipt Do you think this is something that you could take over for me?

PR #1255, opened 2019-12-02
Allow During Onboarding the Skipping of Birthdate, Gender, or Both

erikkgagewhitt:

Allows users to skip the birthdate, gender, or both fields when onboarding. I did this by slightly changing the form and mutation that manage these. The form now checks the values and dynamically builds the input to the mutation.

- The structure of the mutation on the client had to change since it was previously static
- Selecting "Prefer not to reply" will not actually save any data for gender, so it will be as if the user never input anything. This also means that changing from "male" or "female" to "prefer not to reply" will not actually change gender.

### How do I test this PR?

Best done on a new user

1. In onboarding with nothing selected, ensure that the user can skip
2. In onboarding, select "prefer not to reply"; no data should be saved
3. In onboarding, select a birthdate, but not a gender; you should advance and the birthdate should be saved.
4. In onboarding, keep your birthdate and select a gender; you should advance and your data should be saved

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
@erikkgagewhitt I did some peeking in the codebase, and it looks like we do have a gender value for 'Unknown' 😬 

I would keep all the fantastic code you wrote around sending form values only as they are populated, and send over the gender as "unknown" if a user selects preferr not to say". That way if someonee unsets their gender it will actually be unset.

vinnyjth:
https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-data-connector-rock/src/people/data-source.js#L82

vinnyjth:
One more bug @erikkgagewhitt. "Unknown" doesn't seem to map back into "Prefer not to reply". if you select it in onboarding and then reopen onboarding.

erikkgagewhitt:
@vinnyjth Do you mean that "Prefer not to reply' is unselected?

vinnyjth:
Correct, when it should be selected.

erikkgagewhitt:
Okay, yeah; I knew that happened - wasn't sure if it would be an issue. Should be easy to fix!

erikkgagewhitt:
Initial value of 'Unknown' should work as intended now!

PR #1260, opened 2019-12-05
Adds baseBorderRadius value to the defaultTheme

burkeshartsis:
## DESCRIPTION
This adds and implements a value to the `defaultTheme` called `baseBorderRadius`.

This PR also fixes a few random things I found while making these changes. A small `state` reference change and updating some components to use theme values.

Before this there was no easy way to change the `borderRadius` value across the app. In some cases it was impossible to change.

N/A

### How do I test this PR?
Click through the app. Run tests. Everything look the same? You're g2g 💥 

N/A

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-kit/src/OverlayBackgroundImage/index.js
```diff
@@ -65,7 +65,8 @@ class OverlayBackgroundImage extends Component {
     width: DeviceWindow.width,
   };
 
-  handleLayout = ({ nativeEvent: { layout } = {} }) => this.setState(layout);
+  handleLayout = ({ nativeEvent: { layout: { width } = {} } = {} }) =>
+    this.setState(width);
```
vinnyjth: This doesn't seem right, this looks like it would set the whole `state` object to an integer, instead of setting it to `{ width: int } `. What's the goal of this change?

packages/apollos-ui-kit/src/inputs/Radio/RadioButtonIndicator.js
```diff
@@ -6,9 +6,9 @@ import styled from '../../styled';
 
 const OutlinedCircle = styled(
   ({ theme, isSelected }) => ({
-    height: 20,
```
vinnyjth: Nice!

packages/apollos-ui-onboarding/src/OnboardingSwiper/index.js
```diff
@@ -11,7 +11,7 @@ import { SafeAreaView } from 'react-navigation';
 const dotStyles = ({ theme }) => ({
   width: theme.sizing.baseUnit / 2,
   height: theme.sizing.baseUnit / 2,
-  borderRadius: theme.sizing.baseUnit / 4,
+  borderRadius: theme.sizing.baseBorderRadius / 4,
```
vinnyjth: I'm struggling with this and the change to `RadioButtonIndicator.js `. This is supposed to be a circle, so the borderRadius should be a value that causes this to be round, correct? Thus, I think the `borderRadius` should scale with `baseUnit`. If someone changes `baseBorderRadius` without changing `baseUnit` they would see ovals here I think

packages/apolloschurchapp/src/user-settings/Locations/Marker.js
```diff
@@ -7,17 +7,17 @@ import Color from 'color';
 import { styled } from '@apollosproject/ui-kit';
 
 const MarkerView = styled(({ theme }) => ({
-  width: 8,
-  height: 8,
-  borderRadius: 4,
+  width: theme.sizing.baseUnit * 0.5,
```
vinnyjth: Same circle issue here. I think should stay as baseUnit, (but I like that you tied the value to `baseUnit`, instead of leaving the static values)

packages/apollos-ui-kit/src/OverlayBackgroundImage/index.js
burkeshartsis: My bad. Missing some brackets for shorthand object assignment.
```suggestion
    this.setState({ width });
```

vinnyjth: Sweet! That's what I was expecting to see

burkeshartsis: The goal is to only set the one layout value we need rather than the entire layout object in state.

packages/apollos-ui-onboarding/src/OnboardingSwiper/index.js
burkeshartsis: Good catch. I 100% agree!

packages/apollos-ui-kit/src/OverlayBackgroundImage/index.js
vinnyjth: Yah I figured, makes perfect sense with this change. Good catch.

packages/apolloschurchapp/src/user-settings/Locations/Marker.js
burkeshartsis: fixed in 5d72f8d

packages/apollos-ui-onboarding/src/OnboardingSwiper/index.js
burkeshartsis: fixed in b920046

burkeshartsis:
@vinnyjth I addressed the issues you mentioned. Should be g2g now.

vinnyjth:
I propose we upgrade to `baseBorderRadius: 2`

Simulator Screen Shot - iPhone 11 - 2019-12-06 at 14 35 35

PR #1265, opened 2019-12-10
Updates description style for HorizontalDefaultCard

burkeshartsis:
This updates the `description` to use `BodySmall` for the `HorizontalDefaultCard`.

This change is an approximation of the desired design. To for this to be visually correct we would have to change the `BodySmall` which is currently 14px to 12px across the app.

### How do I test this PR?
stories

🍪

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


kevin-birdy:
@burkeshartsis I'm sorry I think I might've confused myself talking with jordan and made a mistake here.

Was this previously using H6? `BodySmall` is definitely 14px... but I just combed the UI Kit and it looks like this is actually H6 and the real discrepancy that Jordan found was that H6 in Github is bold and H6 in Figma is regular.

This is the only place I can find `BodySmall` in its current state (`CampusCard`). So, if that's the case...I am probably not utilizing `BodySmall` as well as I should be.

Screen Shot 2019-12-11 at 9 10 30 AM

So...now we have a couple of options that I see:
1. Keep `H6` bold and make `BodySmall` 12px regular
2. Update `BodySmall` to 14px and change `H6` to regular

Given the above, I actually think you are heading down the right track and we go with option 1. The only thing that I can see that I need to update is `CampusCard` (which, isn't called that anymore? Help me find it in Storybook so I can see what we are currently using for the description?)

kevin-birdy:
Basically every instance of `H6` is currently being used as a description text (hence why I had it as `regular`... so one thing to note is that if we did make `BodySmall` 12px regular it would essentially nullify `H6` other than the possibility of a little caption/disclaimer or something like that.

burkeshartsis:
@kevin-birdy so for better or worse combing the app for `H6` usage reveals more references than I think you have in the "design kit." We can totally audit this later but I agree the safest, easiest, and most semantic route is to update `BodySmall`. This would leave `H6` to continue to be a "header" and let `BodySmall` be the caption/subtext/smaller bodied `BodyText` brother it was always meant to be.

kevin-birdy:
Sounds like a plan to me

burkeshartsis:
So for clarity, we are going to keep this PR as it stands. It updates the `HorizontalDefaultCard` to use `BodySmall` even though the styling on `BodySmall` isn't correct. This #1267 updates `BodySmall` to the correct size of 12px.

kevin-birdy:
Yes! 100% on board.

PR #1266, opened 2019-12-11
Push Provider/Consumer/Context

erikkgagewhitt:

Handles the state where the user has already been prompted for notifications and has declined. I previously tried [this](https://github.com/ApollosProject/apollos-prototype/pull/1243), but found that I could not enable notifications, if a user had ever declined them (makes sense from a user protection standpoint, but is very obnoxious). I also tried using listeners, but to no avail.

Now instead, I disable the notifications button to show the choice has already been made; I disable the"skip" button and enable the "finish" button 

This version of Push handling uses Provider/Consumer/Context in order to better handle push settings. It should address the problems [here](https://github.com/ApollosProject/apollos-prototype/pull/1246)

- Not sure this solution is great design-wise, but it seemed to look alright and make sense to me.
- There is a loading state for the button, but the change is so quick it's hard to see
- Changing your settings through the settings menu will not update the UI;  this is because RN does not have access to permission listeners. If there is a way to track background/foreground, we can fix this.

### How do I test this PR?

1. Delete the app from your sim
1. In onboarding tap Enable Notifications and reject the hard ask
3. In onboarding the state of this screen should change
4. Repeat the previous steps, but accept the hard ask

## SCREENSHOTS

Screen Shot 2019-11-21 at 3 47 54 PM

## TODO:

- [ ] Closes
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


erikkgagewhitt:
@vinnyjth Improved version of https://github.com/ApollosProject/apollos-prototype/pull/1246

Sorry it took me so long. This should be much better!

vinnyjth:
The UX on this LIT 🔥💣 🎉 🙌 !!! I couldn't get myself into a state that I thought would be a bad user experience. I'm going to go through and do a code level review of this PR, and then we should be good to go :D :D :D

vinnyjth:
The UX on this LIT 🔥💣 🎉 🙌 !!! I couldn't get myself into a state that I thought would be a bad user experience. I'm going to go through and do a code level review of this PR, and then we should be good to go :D :D :D

vinnyjth:
The UX on this LIT 🔥💣 🎉 🙌 !!! I couldn't get myself into a state that I thought would be a bad user experience. I'm going to go through and do a code level review of this PR, and then we should be good to go :D :D :D

vinnyjth:
The UX on this LIT 🔥💣 🎉 🙌 !!! I couldn't get myself into a state that I thought would be a bad user experience. I'm going to go through and do a code level review of this PR, and then we should be good to go :D :D :D

vinnyjth:
The UX on this LIT 🔥💣 🎉 🙌 !!! I couldn't get myself into a state that I thought would be a bad user experience. I'm going to go through and do a code level review of this PR, and then we should be good to go :D :D :D

packages/apolloschurchapp/src/ui/Onboarding/index.js
```diff
@@ -63,22 +66,30 @@ function Onboarding({ navigation }) {
...
-                    <StyledGradient
-                      source={'https://picsum.photos/640/640/?random'}
-                    />
-                  }
-                />
-              )}
-            </ApolloConsumer>
+            <PushProvider>
+              <AskNotificationsConnected
+                onRequestPushPermissions={(update) => {
```
vinnyjth: This might flow a tad easier if you rewrite it using `async/await` :D

packages/apollos-ui-notifications/src/index.js
```diff
@@ -6,3 +6,4 @@ export {
   GET_PUSH_ID,
 } from './permissionUtils';
 export updatePushId from './updatePushId';
+export PushProvider, { PushConsumer } from './pushProvider';
```
vinnyjth: I can see it being confusing for our partner churches when we export two different providers from UI-Notifications. Do you think you could merge the two providers and export only a single Provider/Consumer?

vinnyjth:
@erikkgagewhitt 

Let's get rid of these `resolvers` 
https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-ui-notifications/src/store.js#L42
https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apollos-ui-notifications/src/store.js#L17
We don't need it, and none of our partner churches use it :D Let's also remove all the functions that are defined in the `permissionUtils` and used only in those resolvers.

erikkgagewhitt:
@vinnyjth I think I understand what you want me to do with the providers; could you give me some input on how you might combine them? They're a bit dissimilar in the way they function and seem to be used in different places.

PR #1267, opened 2019-12-11
Updates typographic styles

burkeshartsis:
## DESCRIPTION
This PR changes the apps vertical rhythm.

Previously we used a `baseLineHeight` of `23.04` which was derived by a `baseFontSize` of `16` with a vertical Rhythm (read: ratio) of `1.44`. This PR changes the `baseLineHeight` to `24` creating a perfect fifth ratio of `1.5` or `2:3`. Aside from being what @kevin-birdy wanted it also is more likely to produce whole numbers and that makes him feel warm and fuzzy inside.

**Note:** the `lineHeight` on `H1`, `H2`, and `H3` are unchanged and continue to use a ratio of `1.15` or `20:23`. The `margin/padding` on these components **has changed** to reflect the new `1.5` everything else uses.

No trade offs but this does "loosen" the app a bit. All typographic elements will have a touch more `lineHeight` and vertical `padding/margins`. This also increases the paragraph spacing.

### How do I test this PR?
There are a TON of snapshot changes here. My request is to just have @kevin-birdy give this branch a look over and make sure he is happy with the changes. Then an engineer can just review that everything works as expected. Given that this is entirely a frontend UI change I'm not even sure it's necessary for an engineer to review this.

Existing snapshot coverage.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


burkeshartsis:
@kevin-birdy something worth considering as an addition to this PR would be if you want to update header padding. These will automatically adopt your new desired `1.5` vertical rhythm ratio but depending on how you like to position things we might want to consider changing these values.

I will also add that, we only use `verticalRhythm` for typographic spacing, that is between one typography component and another. An example of this can be seen on any article in the app. Importantly, it could be used everywhere in the app but it would require the ratio to match or be some derivative of the "grid unit" or what we currently call `baseUnit`. This would mean that type would always fall on the grid and elements around it would be positioned correctly but it would require strict adherence to a fixed grid in design. The benefits would be that we would use a single unit for positioning and sizing elements/components. Additionally, aligning objects would be somewhat trivial as positioning on a grid is, generally speaking, obvious to the eye. Lastly, the entire layout would be scalable to whatever ratio we use (tighter <> looser). The hard part would be accounting for images/media that is "full width." Obviously, we need to preserve media aspect ratio but we can not guarantee it will always be at a ratio that "fits" on the grid.

PR #1269, opened 2019-12-13
Fixes auth thrashing

vinnyjth:

Current code:
1. When you logout, we call `resetStore` 
2. `resetStore` reruns all queries in the app
3. Some of those queries provide an `unauthenticated` code without a current user
4. Receiving an unauthenticated code triggers an `onAuthError`, which calls resetStore and so on and so on. The app becomes unusable and loops forever failing. 

FIx: 

**If we already are resetting auth, don't try and reset auth again.** 

I also made the `authReset` function reusable. This mess of `react-navigation` code will not only navigate back to Auth, but will reset the navigation state (so you don't land on the page with the pin filled in)

n/a

### How do I test this PR?

1. Login to the app, click around.
2. Find your UserLogin in Rock and delete it. 
You can do this by querying for UserLogins with a `$filter` like `UserName eq '5133061126'`, then use the ID to call a `delete`
3. Take an action that you need the current user to complete (like liking an item)
4. You should end up in the Auth flow.

n/a

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apolloschurchapp/src/NavigationService.js
```diff
@@ -15,6 +15,23 @@ const navigate = (routeName, params) => {
...
+const resetToAuth = () => {
+  _navigator.dispatch(
+    StackActions.reset({
+      index: 0,
+      key: null,
+      actions: [
+        NavigationActions.navigate({
+          routeName: 'Auth',
+          action: NavigationActions.navigate({
+            routeName: 'AuthSMSPhoneEntryConnected',
```
conrad-vanl: Should we take the back to the landing screen instead?

vinnyjth: That does make more sense :)

PR #1273, opened 2019-12-17
Auth api frontend

conrad-vanl:
## TODO:

  - iOS: https://v.usetapes.com/SWq1K5pQnc
  -Android: https://v.usetapes.com/isFZuPnn3q
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


jordanwade:
Just wanted to give a quick update on the frontend side of this PR. Everything is in place and is working as expected. I'm just trying to fix a couple Android visual issues then I'll add the review tag.

vinnyjth:
Only one visual bug that I found so far, it looks like the border radius on the two tabs is incorrect on Android (is rounded on the inside when it should be straight). There might not be anything we can do about that :/ 

Screen Shot 2020-01-13 at 8 39 15 AM

One flow bug I encountered was hitting "back" from the Pin entry page. You end up on the first/last name page with nothing to do. This is something I can fix in the provider I think :) 

Screen Shot 2020-01-13 at 8 42 09 AM

There's a bit of cleanup that I need to do from when I worked on the frontend provider (need to move queries files into a better spot) but besides that, this looks great @jordanwade :D Very smooth and slick, and I think this will have a huge impact on the amount of deduping that Willow will need to do, and will save Christ Fellowship quite a bit of work and bring them back closer to core.

packages/apollos-ui-kit/src/Button/Button.js
```diff
@@ -107,7 +107,8 @@ const Button = enhance(
         {loading ? (
           <InlineActivityIndicator color={accent} />
         ) : (
-          children || <H4>{title}</H4>
+          // Passing down styles to the H4, but not the background color to fix Android issue.
```
vinnyjth: @jordanwade Do you happen to have a link describing the behavior you are avoiding here?

jordanwade: @vinnyjth basically I was just trying to have the color of the button text to be white. From what I could tell the h4 doesn't inherit the `color` if you use the color prop on the button. Is there something I missed in the Button component api that handles text color if text is passed as a in the `title` prop?

vinnyjth: I don't think so :D I was wondering what the "Android issue" mentioned in the comment was, and wanted to update the comment to cite more details.

jordanwade: @vinnyjth Yeah, the Android issue was that I was initially passing all the style props, which passed the background styles of the button to the text. This caused a weird semi-transparent background on the text.

jordanwade:
@vinnyjth I think this is ready for another look. I took Burke's suggestion of refactoring the tab buttons to be a `Touchable` instead of a `Button` and it seems to fix the issue on Android.
image

PR #1278, opened 2020-01-10
Fixes StretchyView bug on overscroll effect

burkeshartsis:
This PR fixes a bug in StretchyView that was causing a crash on iOS. It does this by:
- correcting an api syntax error
- using `Animated.ScrollView`

Coincidentally, this might actually improve `StretchyView` performance on iOS by actually passing its animation to the native thread, where it might not have been doing so previously.

Additionally, this PR adds a commit (92c0151), **that you are welcome to keep or revert**, that removes all `StretchyView` functionality from Android. It basically just renders a `View` with no additionally renders or computational logic. This is because `StretchyView` doesn't actually work on Android to begin with because Android doesn't overscroll. No overscroll, no way to animate the image with a stretchy overscroll effect. So might as well ditch all the code right 🤷‍♂ Like I said, keep it or revert it. I leave that decision to the reviewer.

### How do I test this PR?
- open app
- tap on an Article/Sermon/Series (anything but a Devo single)
- everything look correct?
- pull down to see the overscroll stretchy effect on ios and scroll on android.
- No error/crash? We're good!

All the changes in this PR are unable to be covered with jest snapshots.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-kit/src/StretchyView/index.android.js
```diff
@@ -0,0 +1,27 @@
...
+    <View style={[background ? StyleSheet.absoluteFill : null, style]}>
+      {children}
+    </View>
+  );
+
+  render() {
+    return (
+      <View style={StyleSheet.absoluteFill}>
+        {this.props.children({
+          Stretchy: this.StretchyPortal,
```
vinnyjth: We might want to pass the same arguments down as the ios version, but with empty or `noop`, just in case a client decides to reference each argument individually instead of spreading the arguments.

burkeshartsis: fixed in db6cba7

PR #1284, opened 2020-01-15
Adds caching to requests made to api.bible

vinnyjth:

To avoid the 5,000 daily request limit set by API.bible, we want to start caching the requests we make to their API. To do this, we are going to use the apollo datasource inmemory cache, because it will capture uses of the scripture API no matter the source, and should be robust enough for our needs. 

The TTL is set to 1 day, because it seemed like a sane place to start. We can lengthen it if we run into no issues, and shorten it if we discover a disadvantage to having the cache length lower.

### How do I test this PR?

There's a rather decent way to test this PR. 

1. Open up the datasource for the `data-connector-bible`, then create a (postbin)[https://postb.in/]
2. Drop the link to the postbin into your local datasource
3. Run the following two queries multiple times in GQL
```
query getScripture {
  scriptures(query:"SNG.1.1") {
    id
    html
    version
  }
}

query getId {
  node(id:"Scripture:27ee1e80fe669dc64351715914e68dab417253a3558db00cd7edbaa6b058176e71ce3edec3df90cb5154ba4f0c0c06d70fcf6b35b9f96d7fe5eb6ebfd8902095") {
    id
    ... on Scripture {
      html
      version
    }
  }
}
```
4. You should see only two requests total in your postbin dashboard (you will have to refresh after making the requests)

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-data-connector-bible/src/data-source.js
+++ b/packages/apollos-data-connector-bible/src/data-source.js
@@ -3,6 +3,7 @@ import { RESTDataSource } from 'apollo-datasource-rest';
 import ApollosConfig from '@apollosproject/config';
 
 const { BIBLE_API } = ApollosConfig;
+const ONE_DAY = 60 * 60 * 24;
 
 export default class Scripture extends RESTDataSource {
   resource = 'Scripture';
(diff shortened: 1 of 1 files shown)


redreceipt:
Is there a difference in what you did here vs. `this.ttl(ONE_DAY).get()`?

PR #1292, opened 2020-01-17
Use default import in root index

conrad-vanl:

@vinnyjth explain to me why I had to do this to get this to work but it seems to work as is for everyone else? Shouldn't it `.default`?

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apolloschurchapp/index.js
+++ b/packages/apolloschurchapp/index.js
@@ -7,7 +7,7 @@ import Storybook from './storybook';
 
 const useStorybook = ApollosConfig.STORYBOOK === 'true';
 
-const MainApp = require('./src');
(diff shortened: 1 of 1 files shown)


vinnyjth:
This is very strange @conrad-vanl, I believe that the behavior is undefined / changes across node versions? I believe your version is safer / more correct and assuming it works fine for me I'll merge it. 

The use of `require` is important because it loads the file synchronously to ensure all of the config has been loaded before we load the files that use the config. f

conrad-vanl:
>  I believe that the behavior is undefined / changes across node versions?

I thought this too, but I tried multiple node versions and even made sure I was using same version as @burkeshartsis

PR #1304, opened 2020-01-28
Clears apollo store instead of reseting store

vinnyjth:

There's a problem in the current app where any action that triggers a store reset (logout or auth error) causes all of the queries to be rerun. This can cause a vast number of queries to be rerun in some cases, as evidenced by watching your API console after logging out of the app (even after not clicking on a single content item or piece of UI besides the logout button!) 

This causes problems for two reasons: 

1. These queries take time and energy to be ran and re-ran
2. Some of those queries aren't designed to be run without a user active, and return auth errors. 

We have a workaround to prevent infinite redirects because of the second issue, but Bugsnag is still hard to sift through at least in part because of the number of unnecessary errors. 

By switching to `client.clearStore` and `client.onClearStore` we can wipe the cache without reloading the data. You can see that in action on the attached GIF of logging out. 

I needed to update apollo-client, because our version of `onClearStore` would fire the callback _before_ the store had been wiped, meaning the data we would write with `onClearStore` would immediately be erased. 

### How do I test this PR?

1. Login to the app.
2. Logout of the app.
  - Clear the API console before you do so, ensure that only 1 or 2 queries are fired.
3. Log into the app.
4. Play a media player video 
  - In the past, messing with this functionality has caused the media player to stop working. 

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
Is the only way to get core working to `sudo` install bundler v2?

vinnyjth:
@redreceipt I don't think any of us have had to `sudo`.  Do you use `rvm`?

redreceipt:
> @redreceipt I don't think any of us have had to `sudo`. Do you use `rvm`?

no, quick and dirty explanation?

vinnyjth:
@redreceipt I don't really know why it helps. `rvm` and and `rbenv` both change where gems are installed by default and I think everyone on our team uses one or the other.

redreceipt:
Screen Shot 2020-01-28 at 11 51 51 AM

PR #1315, opened 2020-01-30
Fixes location finder preferring Cincinnati without a location

vinnyjth:

The Location Finder should no longer pretend that Cincinnati is the magic location for everyone without a location. There's an additional fix we can perform on Willow to make sure that Android Willow actually asks for Location Permissions, but this work needed to happen in Core. 

Part of this PR involved marking the latitude and longitude as non-required arguments. 

### How do I test this PR?

1. To begin, I would recommend commenting out the following lines: 
https://github.com/ApollosProject/apollos-prototype/blob/master/packages/apolloschurchapp/src/user-settings/Locations/index.js#L60-L64

This will simulate the situation that we are seeing on Android. 

2. Open up the location finder. Switch between the different campuses.

n/a

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
@burkeshartsis Tests are fixed, snapshots are back to where they were before

jordanwade:
Seems to be working as expected. I'll let @burkeshartsis have the finally approval since he reviewed this first. ✅Nice work!

packages/apolloschurchapp/src/user-settings/Locations/__snapshots__/Locations.tests.js.snap
```diff
@@ -109,11 +109,6 @@ exports[`Display Native Map with Locations Render a native map view 1`] = `
     }
   }
   onLocationSelect={[Function]}
-  userLocation={
-    Object {
-      "latitude": 39.104797,
-      "longitude": -84.511959,
-    }
-  }
+  userLocation={null}
```
burkeshartsis: So this test works better now but for posterity I duplicated it and passed in `userLocation` values to lat/long and I got the same result. Am I doing something wrong or is something amiss?

vinnyjth: `userLocation` is not a valid prop, it's is passed down from `state`. It gets populated into state via the `Geolocation` library.

burkeshartsis: It's a prop on MapView [here](https://github.com/ApollosProject/apollos-apps/blob/location-finder-without-current-location/packages/apollos-ui-mapview/src/MapView/index.js#L64,L66) but you are correct that it isn't on the `Locations` component. I wish there was a way to test this but it's all on the native side so there isn't a difference when you write a test for it in `ui-mapview`.

PR #1318, opened 2020-01-31
Improved Content Card API

redreceipt:

This PR came out of trying to understand and customize our app to fit specific needs while also trying to stick as close to the core implementation as possible. The things that were changed are mostly a simple restructuring of components with a few deprecations to simplify the API. It provides many benefits and improvements:

- Removes many unnecessary data calls and cache hits
- Removes redundant mappers and components
- Abstracts UI elements from data components
- Consolidates location of content cards
- Exposes much simpler API to developers
- Improves many snapshots
- Reduces the amount of code necessary in current partner apps

Specifics:

- Made `ContentCard` the only API necessary to style in a variety of ways. Left others as deprecated.
- Removed need for `ContentCardConnected` in the app template. Possibly could deprecate in the future.
- All content card types now live in the same place in `ui-kit`.
- Horizontal variations can now be achieved by simply adding a `horizontal` prop.

Screen Shot 2020-01-31 at 2 04 42 AM
Screen Shot 2020-01-30 at 11 01 49 PM

No functionality is lost. Some UI and data components have been deprecated or removed in favor of additional props to similar components.

### How do I test this PR?

Spin up the app and go through all the different content card styles.

None. Improved snaps by default.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
Alright @redreceipt! I decided I'd go ahead and respond to this PR. @burkeshartsis and I have been having some back and forth conversations on our thoughts about this approach, as well as making sure we understand the problem you're trying to solve for. 

I do not believe we have a 100% certain answer yet, but I wanted to update you with our (really **my**) current thoughts and see what you think, as well as solicit thoughts from @vinnyjth and @jordanwade.

So taking a step back, our current "Content Card" solution in core is comprised of three conceptual parts:

- A `Connected` Content Card
- A mapper function that renders the correct UI Kit Card component based on both props and data form the server
- The underlying UI-Kit Card Component
- All of the above for both Horizontal and Vertical variants

The problems I am hearing you articulate with this current approach I can sum up as:

- Confusing mapping of data and props. There's prop mapping going on in both the connected component as well as the mapper function
- Hard to understand or customize why and when a specific content card is rendered
- Data querying seems to be happen more then it should

Your approach appears to tackle this problem by looking at reduction of code as the primary solution. Namely, it:

- Keeps a single `Connected` Content Card component (and uses it for both Horizontal and Vertical). It also removes prop mapping and passes data down pretty much un-altered into a new ui-kit component
- Creates a new `ContentCard` component in the UI-Kit. This component is pretty much doing the same thing as the existing mapper functions, but it's now a single function.

Here's **my** initial feedback on this approach, and the code you have in this PR:

- I think you did do a good job finding a solution that works while reducing a lot of code. There's a lot of value in that!
- This new `ContentCard` component really achieves the same thing that the mapper function did before, but it's now called a Component, which is a win!
- This new `ContentCard` component has too much of a direct relationship to data from the server... in other words its prop design is designed around the server-side data model and not ui/ux concerns. UI-Kit components are largely meant to be decoupled from the shape of our data model, and looked at from a ui-first design mentality. I would say this currently is utilizing the opposite mindset, and results in a more complex prop design for the component. Specifically, using something like `_typename`s as a prop within a `ui-kit` component would be looked at as an anti-pattern.
- Shuffling all of the card components into the Content namespace I think can create some issues: these card components aren't necessarily content-specific (they could represent events, groups, etc) and also there are other content-related components sprinkled throughout the ui-kit - the decision on whether or not to put them within the content namespace becomes a decision fatigue potential

With that said...I wanted to take a step back and make sure we are tackling the right problem.

My hypothesis is that the real problem here is not data concerns or lack of customizability, but those are rather symptoms of the real problem... which is that we are trying to create a one-size-fits-all component that renders other components based on parameters that are inconsistent.

For example - with this content card component - sometimes we want to use a certain card component based on where the component is rendered in the app. Such as if we are rendering a content card in a featured section, we want to use the `FeaturedCard` component. Other times, it is based on parameters from the server, such as deciding between a `DefaultCard` or a `HighlightCard`. Lastly, we have the decision between `Horizontal` and `Vertical` flavors.

Abstracting all of this logic behind a single component, whether it's contained within mappers or not, I believe is the problem, and I'd wager the true anti-pattern here.

My proposal: we should instead seek a solution with a simpler prop footprint, but an increased component footprint. This probably means having multiple connected card components, but it should focus on allowing more specificity (and thus clarity) on the apolloschurchapp side of when and how which content cards are rendered.

I think I have a few ideas for how this could look. Neither of these are completely flushed out and are simply thought starters.

One:

- Create a `FeedContentCardConnected` component for rendering items of varying `_typename`s in a content feed (and that's all it does!). This component only switches between `DefaultCard` and `HighlightCard` based on the `_typename` of the content. It can handle vertical/horizontal flavors, or those are separate components (`HorizontalFeedContentCardConnected` and `VerticalFeedContentCardConnected`) Otherwise, this component is a simple, straight-forward connected component.
- Create a `ConnectedFeaturedContentCard` component (that is a simple, straight-forward connected component)
- Leave UI-Kit unchanged


Two: 

- Create a `ConnectedContent` version of _every_ `ui-kit` card: `ConnectedFeaturedCard`, `ConnectedHighlightCard`, `ConnectedDefaultCard`.
- The logic for which component to render is moved into the `Feed` components, but most of the time you'll just render the specific connected card component that you want.

Third:

- There could be value in having some combination of the above two. Basically, still have `Connected` versions of all of the content cards, but use `FeedContentCardConnected` at the high level to switch between `DefaultContentCardConnected` and `HighlightContentCardConnected`. That would mean it doesn't have to do any prop-mapping...

burkeshartsis:
I agree with @conrad-vanl's summary and will jump back in to discuss the finer details of a particular direction when you're ready @redreceipt. 

Additionally, I will add that we have not really documented our architectural decisions which I think makes it difficult for contributors to make higher level component api changes like you are purposing here.  As we grow we should aim to address this to increase transparency, knowledge sharing, and to help others contribute more effectively.

redreceipt:
Thank you for spending time on this. I'm simply trying to give the dev fewer API hooks to look up. I agree with you that the UI component probably shouldn't be data-dependent. I did it for reducing unnecessary data calls. So maybe I pull out data fields and map them to UI props in `ContentCardConnected` and we give that component a prop to decide if it should pull fresh data or not. so `<ContentCardConnected update />` would pull fresh data using the `id`, otherwise, it simply maps data fields to UI props and calls my `ContentCard`.

conrad-vanl:
>  decide if it should pull fresh data or not

It shouldn't pull fresh data if the query is already cached, which it will be if you use the fragments we provide.

redreceipt:
> > decide if it should pull fresh data or not
> 
> It shouldn't pull fresh data if the query is already cached, which it will be if you use the fragments we provide.

But I need to modify that data so it looks different to the UI "NewSpring - Sermons" becomes "Sermons"

redreceipt:
this whole PR allows me to do `<ContentCard labelText={channel.name.split()} {...item} />` sooooo much better from a dev perspective

redreceipt:
If I modify my original PR to put the data fields back and add an `update` prop, the API signature becomes `<ContentCardConnected labelText={channel.name.split()} {...item} />` sooooo much better from a dev perspective

conrad-vanl:
@burkeshartsis if someone needs to modify the query in a `connected` component, is the plan for them to have to make their own connected component? The `update` prop seems like an anti-pattern, at that point it's no longer a `connected` component.

conrad-vanl:
So here's an idea @burkeshartsis @redreceipt for connected component design going forward...

What if our api allowed for passing in a custom `query` prop as well as a custom `children` render-function prop? Something like...

```
<ConnectedContentCard query={myCustomQuery}>
  {({ labelText, ...props }) => (
    <ContentCard labelText={labelText.split()} {...props} />
  )})
</ConnectedContentCard>
```

conrad-vanl:
```
const ConnectedContentCard = ({ children, ...props }) => (
  <Query {...props}>
    {({ data: { node = {} }, loading, error }) => children({
      ...node,
      hasAction: !!get(node, 'videos.[0].sources[0]', null),
      labelText: get(node, 'parentChannel.name', ''),
      // ...etc
    })}
  </Query>
);

ConnectedContentCard.defaultProps = {
   children: ContentCard,
   query: GET_CONTENT_CARD
}

**Comment by redreceipt** [comment]
I think that would solve our issue! We'd still making unnecessary cache hits but I don't suppose that matters. And I'm biased but I think my suggested changes to `ContentCard` would still be very helpful in simplifying the `ui-kit`.

**Comment by conrad-vanl** [comment]
> my suggested changes 

which changes specifically

**Comment by redreceipt** [comment]
> > my suggested changes
> 
> which changes specifically

having `ContentCard` be the one-stop shop instead of mapper and using the the `type` and `horizontal` props I suggested.

**Comment by conrad-vanl** [comment]
So @redreceipt you can do exactly what I suggested today using the `Component` prop, it works the same way:

```jsx
<ConnectedContentCard Component={({ labelText, ...props }) => <ConnectedContentCard.defaultProps.Component labelText={labelText.split() />} />
```

conrad-vanl:
> having ContentCard be the one-stop shop instead of mapper and using the the type and horizontal props I suggested.

I still think this is a design problem - this means we are trying to create a one-size-fits-all component that renders other components based on parameters that are inconsistent. Sometimes those parameters are data bound (ie...`typename`) and sometimes those parameters are based on UX (a `featured` item)

conrad-vanl:
After seeing both the `Component` example and the `children` prop example, I think I favor the `Component` example as it's more explicit.

redreceipt:
I think I'm following. So you're saying no changes necessary or you're saying we'll still to change to the `ContentCardConnected` slightly?

conrad-vanl:
Hah... I think the amount of changes we should make should be scaled back. I agree there is probably work to be done with the way prop and component mapping is happening. 

1. `ConnectedContentCard` is still very much a "one-size-fits-all" component which I think we need a better solution for.
2. `ConnectedContentCard` and `contentCardMapper` are still both mapping props in both places, and that can be confusing

burkeshartsis:
> @burkeshartsis if someone needs to modify the query in a connected component, is the plan for them to have to make their own connected component?

@conrad-vanl the general for how we have things is if you need to make small changes to a query we suggest you make them to the fragment so your changes are made cached across all the places that query might be used. If you need to make wide sweeping changes to a query you probably need to change how that data maps to the ui component to meaning you will need a custom `ConnectedComponent`. All that said, this is already maybe the third or fourth iteration to this part of the ui and I fully expect new behavior/use cases to guide us towards possible changes. 

FWIW, I have discussed a `query` prop with @vinnyjth before and I think he is generally in favor of it but I am not sure if adding it will still create the scenario I mentioned above.

conrad-vanl:
@burkeshartsis how can one change the fragment though??

vinnyjth:
@conrad-vanl Like this: https://github.com/Differential/willow-prototype/blob/master/packages/WillowCreekApp/loadConfig.js

redreceipt:
Ok :) I was simply scratching my own itch. I'd love to help make the API better but it doesn't sound like there's consensus and if I can solve my problem by using the `<ContentCardConnected.defaultProps.Component labelText...` solution then we'll just do that (_I can do that right now with no core changes, correct?_).

My last two cents. I don't think it's confusing to have `ContentCardConnected` with props to adjust UI. In fact, I think it's better than making me know about all the different variations.

redreceipt:
so @conrad-vanl for this PR to continue, I need to move data mapping back to the `Connected` component. What else? I know it's been mentioned to make fewer API and prop changes. You do or don't like the `type` and `horizontal` props on the `ContentCard`?

PR #1325, opened 2020-02-04
Adds Algolia Custom Configuration

redreceipt:

There are two things that Algolia advises in order to properly configure search and this adds the ability for churches to define their own custom configuration with some reasonable defaults.

None.

### How do I test this PR?

Log into the Algolia dashboard and make sure the settings have taken effect.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-data-connector-algolia-search/src/data-source.js
```diff
@@ -47,6 +48,7 @@ query getItem {
       objectID: id
       __typename
       coverImage { sources { uri } }
+      likedCount
```
vinnyjth: I like the rest of this PR, but would prefer not to add likedCount to the indexed data. That's something you can override as needed :)

packages/apollos-data-connector-algolia-search/src/data-source.js
```diff
@@ -18,6 +18,7 @@ export default class Search {
 
   initialize({ context }) {
     this.context = context;
+    this.index.setSettings(ApollosConfig.ALGOLIA.CONFIGURATION);
```
vinnyjth: 👍

packages/apollos-church-api/config.yml
```diff
@@ -35,6 +35,12 @@ ALGOLIA:
   API_KEY: ${ALGOLIA_API_KEY}
   # TODO: should we support multiple indices? how?
   SEARCH_INDEX: test_ContentItems
+  CONFIGURATION:
```
vinnyjth: Maybe throw a link to what the possible options are for this block? Love this idea.

packages/apollos-data-connector-algolia-search/src/data-source.js
redreceipt: I mainly just included it to show what's possible without commenting out a bunch of stuff. Is there another we could demonstrate capability? Maybe a less intensive field than likedCount?

redreceipt: I could do by date???

redreceipt: since date is cached?

vinnyjth: Could you use one of the fields that is already being indexed?

packages/apollos-church-api/config.yml
redreceipt: done

packages/apollos-data-connector-algolia-search/src/data-source.js
redreceipt: custom rankings are either booleans or numbers

redreceipt: i could use isliked?

redreceipt: I have an idea...just a sec

vinnyjth: That's specific to a user (so won't work in this context)

redreceipt: what would be a better way to do this?

Component 1

vinnyjth: I'm personally opposed to doing this, that `mapItemToAlgolia` method is designed to be overridden, and as of right now I would prefer that overriding that method is the default for adding indexable fields. 

Happy to be convinced otherwise though, why do you want to have this data in the config.yml instead of overridden in the datasource?

redreceipt: just a lot of complicated code debt to simply add another word to a string. but that's fine. every package and data-source that has to be overridden gives churches more potential bugs to write customer support tickets for. I'm probably the farthest along out of any outside partner and the thought of overriding that function just to add a field gives me a headache.

redreceipt: ideally, people don't have to copy and paste anything and of course, that's not possible but as much as we can do to minimize it I think serves both parties very well.

vinnyjth: My view is kinda the opposite, adding new config.yml fields can quickly lead to a situation where we end up with tickets because you can't do what you expect in the `config`, and those types of features we've found are harder to maintain that the code itself. Would love to hear from @conrad-vanl in this situation if he has an opinion as well. 

> I'm probably the farthest along out of any outside partner and the thought of overriding that function just to add a field gives me a headache.

Curious to hear more about this. Why does the thought of overriding that field give you a headache? Our long term goal is that overriding core functionality via the datasources is easy and painfree, and I agree that we aren't there yet. Would love to troubleshoot what we can do to make that easier,.

redreceipt: just spitballing here, what if we had a CLI function that dropped in all the boilerplate needed and then you just drop in your custom code? If as a dev I could just write the actual function without having to think about the `index.js` exports, folder structure, and where to put the `data-source.js` (and in other cases schema and resolver).

something like `apollos-cli --override algolia` would give you all the files you need to start writing the custom functionality.

vinnyjth: Ohhh I like that idea a lot. Some way to use the CLI to generate the overrides for you, and then you fill in the blanks. I think that's absolutely something we could build. 

Also one more bit of context - what fields you include in that list can have huge consequences on the performance of this indexer, and potentially cause problems (`likeCount` on large sets of content items _could_ massively increase the time it takes to index). Hence, I'm fine with forcing a bit more than a config tweak to change what data is indexed.

redreceipt: gotcha. I pushed new commented out code.

vinnyjth:
Looks good @redreceipt! 

We do need to comment out the unused line in the config.yml I think 

image

PR #1340, opened 2020-02-12
Adds Multidex to Android

redreceipt:
## DESCRIPTION

Screen Shot 2020-02-12 at 3 12 33 PM

Adds multidex support to our Android app because the package size too big not too. I don't know the implications.

None.

### How do I test this PR?

Boot the Android app.

None.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


burkeshartsis:
Unfortunately, our [`minSdkVersion=16`](https://github.com/ApollosProject/apollos-apps/blob/80e78b633368e98b025da5da90525cefe058eccc/packages/apolloschurchapp/android/build.gradle#L6) so I think you'll have to jump through a few more hoops to do this.

https://developer.android.com/studio/build/multidex.html#mdex-gradle

vinnyjth:
Yah, this will be hard to test because our app won't actually build as a multi-dex app because everything fits in a single `.dex`. I would recommend trying to copy the setup on Willow or your app and then double checking with the docs @burkeshartsis linked above.

redreceipt:
Our min SDK version is 16 too, this seems to suffice for us.

https://github.com/NewSpring/corinth/blob/develop/android/build.gradle <https://github.com/NewSpring/corinth/blob/develop/android/build.gradle>

—
Michael

redreceipt:
I’m guessing Caleb is too and it worked for him.

—
Michael

redreceipt:
@burkeshartsis still have an issue with this? What I've done here works for us and Caleb and we're both on min 16.

conrad-vanl:
I'm confused by this one. @vinnyjth and @burkeshartsis seems to be saying this won't work. @redreceipt is saying that this does work, and claims it does for Caleb as well.

Has anyone tried to build this locally?

vinnyjth:
@conrad-vanl It will work locally because our app doesn't need multidex, however @burkeshartsis is suggesting that if your app is big enough to need multidex, you need to complete the remainder of the changes. 

Willow needs multidex and I did the complete changes (more than this PR). I haven't tried backing the changes out to just be the contents of this PR.

conrad-vanl:
So is the suggestion for @redreceipt to complete the rest of the changes that Android recommends for multidex, and including that on core?

vinnyjth:
Yah, per my comment above: 

> I would recommend trying to copy the setup on Willow ~or your app~ and then double checking with the docs @burkeshartsis linked above.

redreceipt:
Since we're not sure what the bare minimum is, I'd rather just close this instead of adding unnecessary changes and cross this bridge when we come to it. I don't want to add a bunch of redundant native code in our app because the docs say we need it when it doesn't appear that we do.

Ultimately we're solving problem that isn't there for this codebase anyway.

vinnyjth:
> Since we're not sure what the bare minimum is, I'd rather just close this instead of adding unnecessary changes and cross this bridge when we come to it. I don't want to add a bunch of redundant native code in our app because the docs say we need it when it doesn't appear that we do.

I think this is the right choice for now :) We'll need to re-create this PR either way, since it would belong in the template repo.

PR #1354, opened 2020-02-20
Allows Configurable Browser

redreceipt:
## DESCRIPTION

Screen Shot 2020-02-20 at 12 42 01 PM

This allows churches to be able to configure their web browser with UI options using the `theme.js` file on their app. For example changing "Cancel" to "Close".

None.

### How do I test this PR?

Uncomment the example object and export it from the `theme.js` file.

None.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
Technically I'm not opposed, but I want to see what Burke thinks about using the theme like this :). 

This also leads me to another wish, which is to add override support to `withTheme`. Then you wouldn't need a whole section in the theme.js

packages/apollos-ui-connected/src/RockAuthedWebBrowser/index.js
```diff
@@ -28,17 +34,23 @@ const RockAuthedWebBrowserWithClient = ({ children, client, paper, primary }) =>
...
-  }))(RockAuthedWebBrowserWithClient)
+  withTheme(
+    ({
+      theme: { colors: { paper, primary } = {}, browserOptions = {} } = {},
+    }) => ({
+      paper,
+      primary,
+      browserOptions,
+    })
+  )(RockAuthedWebBrowserWithClient)
```
burkeshartsis:
```suggestion
    withTheme(
    ({ theme: { colors: themeColors = {} } = {} }) => ({
      themeColors,
    }),
    'ui-connected.RockAuthedWebBrowser'
  )(RockAuthedWebBrowserWithClient)
```

Line 51 above enables prop overrides in theme!

packages/apollos-ui-connected/src/RockAuthedWebBrowser/index.js
```diff
@@ -6,7 +6,13 @@ import { withTheme } from '@apollosproject/ui-kit';
...
 import RockAuthedInAppBrowser from './RockAuthedWebBrowser';
 
-const RockAuthedWebBrowserWithClient = ({ children, client, paper, primary }) =>
+const RockAuthedWebBrowserWithClient = ({
+  children,
+  client,
+  paper,
+  primary,
+  browserOptions,
+}) =>
```
burkeshartsis:
```suggestion
const RockAuthedWebBrowserWithClient = ({ children, client, themeColors }) =>
```

packages/apollos-ui-connected/src/RockAuthedWebBrowser/index.js
```diff
@@ -28,17 +34,23 @@ const RockAuthedWebBrowserWithClient = ({ children, client, paper, primary }) =>
             forceCloseOnRedirection: false,
           },
         }),
+        ...Platform.select(browserOptions),
```
burkeshartsis:
```suggestion
```

redreceipt:
@burkeshartsis sounds good to me. So given the updated API, how would I change the button text from "Cancel" to "Close"? I'm not drawing the connection from `themeColors` to the other options. I'm not familiar with theme overrides.

Also, guessing this negates the comments I made in the `app/src/theme/theme.js`?

burkeshartsis:
@redreceipt yeah you can remove your changes to the theme file. I missed those. 

To use them on your app all you would do is move the `browserOptions` object to be a prop on the browsers overrides. You can see an example and some new documentation [here](https://github.com/ApollosProject/apollos-apps/blob/fd03415e49189823a384bcc258d87baa952c1520/packages/apollos-ui-kit/src/theme/defaultTheme.js#L303). In that example, note that `fill` is a prop not a style and that while the documentation is waiting to be merged the functionality is already in core.

redreceipt:
@burkeshartsis I'm still a little unclear. Are you suggesting closing this PR entirely? The goal of this PR is to allow churches to configure global options for their browser. Even with attaching `browserOptions` as a prop to the `RockAuthedWebBrowser`, some additional work is required I believe (but could be wrong).

redreceipt:
From what I can tell, your suggestions removed all mention of `browserOptions` and just throw colors into a wrapper. but we still need to deal with options in the core component too, right?

conrad-vanl:
Hey @redreceipt! I think @burkeshartsis's suggested changes should enable everything that you're asking for. Totes not asking to close this PR...asking to approve his suggested changes and make sure you understand what's going on.

Basically what you're trying to do with this PR is override the props/options of a single component. You accomplished this by adding new API to the global theme (via `browserOptions`). Instead, you can now accomplish this using our theme overrides system, which was originally designed to override the `styles` of individual components. We recently expanded them to work with theme and props as well.

If you're unfamiliar with the overrides system, we use them pretty extensively on Willow Creek: https://github.com/Differential/willow-prototype/blob/master/packages/WillowCreekApp/src/theme/theme.js#L53-L133

They work by allow you to override the styles anytime the `styled` HOC is used, so if I made a component like:

```es6
const MyStyledText = styled({ color: 'red' }, 'MyStyledText')(Text);
```

You could later override the styles of that single component using our themes override system:

```es6
const overrides = {
  MyStyledText: { color: 'blue' }
}
```

We made two improvements to overrides recently:

1. You can now pass in functions instead of just objects (ie .... `MyStyledText: ({ props }) => ({ color: props.textColor })`)
2. You can use overrides anywhere `withTheme` is used, instead of just with `styled`

So with @burkeshartsis 's suggested changes, you can override `browserOptions` like this:

```es6
// in your custom theme...
const overrides = {
  'ui-connected.RockAuthedWebBrowser': Platform.select({
    ios: { dismissButtonStyle: 'close' },
    android: { showTitle: true },
  }),
};
```

redreceipt:
Thank you! Sorry, I wasn’t drawing the conclusion in my mind you could use `Platform.select` there.

—
Michael

packages/apolloschurchapp/src/theme/theme.js
```diff
@@ -42,5 +42,17 @@
...
+ * https://github.com/proyecto26/react-native-inappbrowser
+ * Must be separated by OS, ios or android
+ *
+ * Example:
+ *
+ * const browserOptions = {
+ *   ios: { dismissButtonStyle: 'close' },
+ *   android: { showTitle: true },
+ * }
+ */
```
redreceipt:
```suggestion
```

packages/apollos-ui-connected/src/RockAuthedWebBrowser/index.js
```diff
@@ -6,7 +6,7 @@ import { withTheme } from '@apollosproject/ui-kit';
 
 import RockAuthedInAppBrowser from './RockAuthedWebBrowser';
 
-const RockAuthedWebBrowserWithClient = ({ children, client, paper, primary }) =>
+const RockAuthedWebBrowserWithClient = ({ children, client, themeColors }) =>
```
conrad-vanl: Why changing `paper, primary` to `themeColors`?

conrad-vanl: Why changing `paper, primary` to `themeColors`?

packages/apollos-ui-connected/src/RockAuthedWebBrowser/index.js
```diff
@@ -35,10 +35,12 @@ const RockAuthedWebBrowserWithClient = ({ children, client, paper, primary }) =>
   );
 
 const RockAuthedWebBrowser = withApollo(
-  withTheme(({ theme: { colors: { paper, primary } = {} } = {} }) => ({
-    paper,
-    primary,
-  }))(RockAuthedWebBrowserWithClient)
+  withTheme(
+    ({ theme: { colors: themeColors = {} } = {} }) => ({
```
conrad-vanl: I'm pretty sure we can walk back this change without breaking anything.

packages/apollos-ui-connected/src/RockAuthedWebBrowser/index.js
```diff
@@ -35,10 +35,12 @@ const RockAuthedWebBrowserWithClient = ({ children, client, paper, primary }) =>
...
+      themeColors,
+    }),
+    'ui-connected.RockAuthedWebBrowser'
```
conrad-vanl: @burkeshartsis @vinnyjth I can see a potential confusion with this new theme overrides system...

We're naming both `withTheme` and `styled` overrides the same - so for example in this case `ui-connected.RockAuthedWebBrowser` could easily be a `withTheme` override _or_ a `styled` override, and from a church-developer perspective if I was looking at a list of components that could be overridden, it wouldn't be clear / can be easily confusing. I don't think there's any designation on the `overrides` system between the two?

Another solution might be to make overrides always override props, instead of just `withTheme` overrides (so make `withTheme` and `styled` overrides work the same). That'd mean if you want to override styles, you have to pass it in a `styles` object:

```
const overrides = {
  'ui-connected.RockAuthedWebBrowser': {
    style: { background: 'red' },
  }
}
```

Or did you already think about this and I'm missing something?

redreceipt: From my perspective, I *think* I agree with this. It's confusing to me to need `styleOverrides` and `propOverrides` and keep track of which goes to `withTheme` and `styled` when as far as I can understand, it's a lot more flexible than the docs on the `theme.js` file (and the different functions and variable names) would suggest.

I propose deprecating `withTheme` if `styled` can get it done and just using `overrides` on the client file.

packages/apollos-ui-connected/src/RockAuthedWebBrowser/index.js
redreceipt: I'll wait for confirmation one way or the other to change it back. Just doing what I was told here.

packages/apollos-ui-connected/src/RockAuthedWebBrowser/index.js
burkeshartsis: You can totally change this back @redreceipt . That was my suggestion. My original thinking was the specific colors don't make sense as prop names and it would make them harder to override. However, my thinking was wrong as they are just the values passed to the `preferredBarTintColor` (etc.) props and you can still access the entire theme within the overrides.  I support changing these back.

packages/apollos-ui-connected/src/RockAuthedWebBrowser/index.js
vinnyjth: This is something that @burkeshartsis and I talked about. 

In the future we are aiming to get complete coverage over all ui components with both `withTheme` and `styled` (or really, we're going to make a wrapping hoc that adds the overrides froom both) so you can know that if you want to pass props, you can use `propOverrides`, and if you want to override styles, you can use `styleOverrides` and not have to check with component supports what.

burkeshartsis: So we could hardcode a solution for this or just set a pattern we think people should use. Either way they should work even if you have both a `withTheme` and `styled` override with the same names. One thing we did to create some distinction was we moved style and prop overrides into separate files. That PR was merged already. Another thing @vinnyjth recommended was adding `.props` to our current override naming convention for `withTheme` overrides (e.g. `ui-connected.RockAuthedWebBrowser.props`). If you all think it would be helpful I'd be fine with adopting this across the codebase. I'm just not entirely sure its necessary given the [separate](https://github.com/ApollosProject/apollos-templates/blob/master/apolloschurchapp/src/theme/propOverrides.js) [files](https://github.com/ApollosProject/apollos-templates/blob/master/apolloschurchapp/src/theme/styleOverrides.js) we now how.

packages/apollos-ui-connected/src/RockAuthedWebBrowser/index.js
redreceipt: done

packages/apollos-ui-connected/src/RockAuthedWebBrowser/index.js
redreceipt: I'll wait for this thread to be wrapped up and then merge this puppy

burkeshartsis: @redreceipt you're probably good to merge. I think we are all saying the same thing.

PR #1359, opened 2020-02-24
Adds capability to styled overrides; they can now be a function of component props

vinnyjth:

I wanted style overrides to be able to look at the component's props. It will make styling some crucial components (like radio buttons and other interactive styled elements) way easier. 

We now have functions stored in the theme, but we already have them anyway! So I think we're okay. 

### How do I test this PR?

Write yourself some overrides :) or look at the test I wrote.

Test demonstrating new functionality.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-kit/src/styled/index.js
+++ b/packages/apollos-ui-kit/src/styled/index.js
@@ -59,7 +59,15 @@ const styled = (styleInput, fqn) =>
           theme,
         });
 
-        const themeOverrides = fqn ? get(theme, `overrides['${fqn}']`, {}) : {};
+        const themeOverridesValue = fqn
+          ? get(theme, `overrides['${fqn}']`, {})
+          : {};
+
+        const themeOverrides =
+          typeof themeOverridesValue === 'function'
+            ? themeOverridesValue(ownProps)
+            : themeOverridesValue;
+
         const { style: ownPropsStyle = {} } = ownProps;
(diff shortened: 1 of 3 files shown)


packages/apollos-ui-kit/src/styled/index.js
```diff
@@ -59,7 +59,15 @@ const styled = (styleInput, fqn) =>
...
         });
 
-        const themeOverrides = fqn ? get(theme, `overrides['${fqn}']`, {}) : {};
+        const themeOverridesValue = fqn
+          ? get(theme, `overrides['${fqn}']`, {})
+          : {};
+
+        const themeOverrides =
+          typeof themeOverridesValue === 'function'
+            ? themeOverridesValue(ownProps)
```
conrad-vanl: does `theme` also get passed in? Should it? `{ theme, ...ownProps }` type thing?

vinnyjth: I realized crafting this PR the theme is already available! It makes the syntax of the overrides a bit odd (but I left it that way because it's consistent). Here's my example from the tests in this PR. 

```
    const overrides = {
      StyledView: (theme) => (props) => ({
        backgroundColor: props.active ? theme.green : theme.blue,
      }),
    };
```

PR #1360, opened 2020-02-25
With theme works with overrides

vinnyjth:
## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-connected/package.json
+++ b/packages/apollos-ui-connected/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@apollosproject/ui-connected",
-  "version": "1.3.0-beta.6",
+  "version": "1.3.0-alpha.8",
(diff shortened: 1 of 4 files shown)


vinnyjth:
>because technically we can use withTheme to pass any props to a component do we think we should isolate "style" overrides from withTheme overrides? Just so we can make sure people see the difference between valid style attributes (style overrides) and prop overrides like fill

I would :) I would probably name the `withTheme` override something like 'ui-Auth.LoginButton.Props' and always use `ui-Auth.LoginButton` for the `styled` name. I think it's fine that with theme supports styles, but I would separate them if we control it. 

> ideas on ways to avoid having to curry this function and avoid this syntax?
No :/ I thought through a few options, and all of them are either messy or break from the existing paradigms of how you configure themes.

PR #1363, opened 2020-02-26
Adds props handle to onboarding button

redreceipt:

This adds the ability to pass arbritrary button props to the primary buttons in onboarding. The specific use case I have in mind is to allow integration testing by passing the `testID` prop.

Exposes the button to be customized further.

### How do I test this PR?

Pass a button prop into `<Slide>` from the app package and make sure it takes effect.

Snap coverage.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-onboarding/src/Slide/Slide.js
+++ b/packages/apollos-ui-onboarding/src/Slide/Slide.js
@@ -76,6 +76,7 @@ const Slide = memo(
     onPressSecondary,
     primaryNavText,
     primaryNavIcon,
+    primaryButtonProps,
     secondaryNavText,
     isLoading,
     pressPrimaryEventName,
@@ -98,6 +99,7 @@ const Slide = memo(
                 trackEventName={pressPrimaryEventName}
                 onPress={onPressPrimary}
                 loading={isLoading}
+                {...primaryButtonProps}
               >
                 <>
                   <H5>{primaryNavText}</H5>
@@ -135,6 +137,7 @@ Slide.propTypes = {
   primaryNavIcon: PropTypes.string, // optional custom icon name or empty string for no icon at all
   secondaryNavText: PropTypes.string, // text link
   isLoading: PropTypes.bool,
+  primaryButtonProps: PropTypes.shape({}), // can be anything
 };
 
 Slide.defaultProps = {
--- a/packages/apollos-ui-onboarding/src/Slide/Slide.tests.js
+++ b/packages/apollos-ui-onboarding/src/Slide/Slide.tests.js
@@ -108,4 +108,17 @@ describe('The Onboarding Slide component', () => {
     );
     expect(tree).toMatchSnapshot();
   });
+  it('should add a test ID', () => {
+    const tree = renderer.create(
+      <Providers>
+        <Slide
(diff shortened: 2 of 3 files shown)


packages/apollos-ui-onboarding/src/Slide/Slide.js
```diff
@@ -135,6 +137,7 @@ Slide.propTypes = {
   primaryNavIcon: PropTypes.string, // optional custom icon name or empty string for no icon at all
   secondaryNavText: PropTypes.string, // text link
   isLoading: PropTypes.bool,
+  primaryButtonProps: PropTypes.shape({}), // can be anything
```
vinnyjth: @burkeshartsis what are your thoughts on this? I know we've pushed back against it in the past, but maybe it's time?

burkeshartsis: Still not in love with this. Fortunately we just [merged a PR](https://github.com/ApollosProject/apollos-apps/pull/1360) that will let you do this without having to extend a different components API in order to pass props to one of its children. That PR lets you pass overrides via `withTheme` which means all we have to do is add an override "name" to that button

redreceipt: Not sure I understand, so could I pass a testID prop down through the Slide to the button with that method?

burkeshartsis: yes!  We just need to update the button in core to support the prop overrides then you will be able to achieve exactly what you want to do from the `theme.overrides` section 💥

redreceipt: Oh ok, is that on the table for y'all or can I take a swing at it?

redreceipt: wait a second, that will allow me to pass a prop to _all_ buttons but I need the prop to be unique per button, will that still be possible?

burkeshartsis: So the default `Button` will totally get this functionality so if you wanted to add some accessibility props or something to every `Button` you totally could. We'll also add it to a bunch of other places, like the exact button you referenced in this PR, in the same way that we expose style overrides.

redreceipt: so (pseudocode) I could do `<Onboarding.LandingScreenButton testID="landing-button" />`

burkeshartsis: it would be via the `theme` in the `overrides` object. Something like...

```
overrides: {
  ui-onboarding.Slide.LandingScreenButton: {
    testID: 'landing-button',
  },
}
```

You can see examples in [this updated documentation](https://github.com/ApollosProject/apollos-apps/pull/1367/files#diff-6707fd674436c5f18e9d0f40b5ea8a38) or in [this test](https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-ui-kit/src/theme/withTheme.tests.js#L68).

redreceipt: Love it!

PR #1368, opened 2020-03-03
Template split

vinnyjth:

Sooooo..... I feel like I'm that guy who's always like "THIS PR IS URGENT" and guess what.... this PR is kinda urgent. Only because it's highly susceptible to merge conflicts and will get very hard to merge very quickly. That said...

This PR requires probably the biggest change to how you develop on the apollos project yet. This removes the a`apolloschurchapp` and `apollos-church-api` from the `apollos-apps` repo. 

While this will require a change in process for development (a minor change) it will also greatly increase the flexibility of the template project, and force us to narrow the template down to code that our clients should be copying.

### How do I test this PR?

Most of the details on how to test this PR are included in the template repo, but I'll copy and paste those instructions here (and amend them for this PR)

#### Setting up "linked" development

1. Clone the [apollos-templates](https://github.com/ApollosProject/apollos-templates) folder
2. From apollos-templates repo, cd into apollos-church-api and apolloschurchapp and run `yarn install`.
3. Edit your `.env. in apolloschurchapp, and add a relative or direct path to your apollos-apps root under the `APOLLOS_APPS_LOCATION` variable. ex: `APOLLOS_APPS_LOCATION='../../apollos-prototype'.
4. From the root of `apollos-apps`, run `yarn link-packages`. **You only need to run this once**
5. From the root of apollos-templates repo, run `yarn link-packages`. **You only need to run this once**

#### Running the linked environments

1. If you are working on API packages, cd into the `apollos-apps` project and run `yarn build:api-watch`.
   If you don't do this, you can still run the app, but you won't get live updates on changes to the API packages.
2. In another terminal, run `yarn start` in apollos-templates. Just like the good old days.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
Will this break the upgrade tool?

conrad-vanl:
Also @vinnyjth it appears there are a lot of "ready for review" PRs. We should probably merge any of those PRs that touch template files first? Then make sure those changes get replicated over to the template project?

vinnyjth:
> Also @vinnyjth it appears there are a lot of "ready for review" PRs. We should probably merge any of those PRs that touch template files first? Then make sure those changes get replicated over to the template project?

Either way is the same work (either merging them now, and copying the changes over, or merging them later, and copying the changes over). They will only be a problem if they touch the template files.

vinnyjth:
> Will this break the upgrade tool?

It won't break using the tool, but it will break my ability to deploy new updates to the update database. The fix is easy, and involves pointing the update tool at the new repo and updating paths (now that those folders aren't nested under `packages/`

vinnyjth:
@conrad-vanl I'm going to wait to resolve conflicts until this PR has a `merge` tag, because the conflicts will act as audit log for changes I need to port over to the apollos-templates. 

Once this is merged, I will also post a Basecamp post describing the "why" again, and repeating the steps for local apollos-apps development.

jordanwade:
Just thought I'd add my 2 cents. The template worked pretty well initially spinning up the the Watermark church app 'The Porch'. So, ✅ from me. I'll let @conrad-vanl give the final approval since he raised some questions.

conrad-vanl:
i just want to make sure we completely think through impact on current church partners. If their lives will be unaffected by this PR, I'm good to give it my 👍

vinnyjth:
@conrad-vanl I couldn't think of a way their lives would be changed by this :D I will have to fix the update tool, but once that happens it will be biz as usual.

conrad-vanl:
Something I just thought about. With this, the upgrade tool would be tied to versions of the apollos-apps template repo, not versions of the apollos packages themselves?

conrad-vanl:
Or...how do we make sure we create some clarity between how `apollos-templates` and `apollos-apps` are versioned

vinnyjth:
My thoughts on this! 

Apollos Templates is the source of truth for the most recent Apollos Version. You know what version of Apollos you are on by checking your apollos.json. You may be using different versions of apollos packages (for whatever reason), but in general, they will coordinate. 

We can now release packages more frequently than we release apollos versions. I think it makes it easier to move forward with package version numbers, but publish specific dot releases of the core template that the upgrade tool can upgrade people too (for example, if there's a bad "default" for how we added some core component). 

Regarding clarity: the `apollos.json` is the current source of truth for your current apollos version. When that goes up, you've upgraded. 99% of your time, your specific packages will match your `apollos.json` version. 

I'm still eeking out the impacts of that. I think the biggest positive is more flexibility of exactly what the template is, and can contain (like if we have to downgrade one of our own packages for whatever reason, we now can) 

The biggest downside is the additional burden of making sure we are keeping the template projects up to date. That will fall on me right now, as I'm cutting the release. I'll make sure my process is documented so others can follow along If I'm out.

burkeshartsis:
This might seem silly but why do we need to track an apollos version in `apollos.json`? If the main developer interface is the template repo when setting up a project why not just use a git construct (labels, releases, branches, etc.) or the root `package.json` version?

vinnyjth:
Because the apollos.json transfers to client repos when the `tag`, `branch`, `label`, `package.json` version doesn't.

burkeshartsis:
is that not something organizations can maintain themselves? The way I understand it:

1. we will update `apollos.json` in the template
2. `apollos.json` dictates the package versions used for the org.
3. the update tool is used by the org to get the updated `apollos.json`

vinnyjth:
> apollos.json dictates the package versions used for the org.

This is not true, the versions are pulled from the package.json

PR #1372, opened 2020-03-04
Async Live Content Filter

redreceipt:

This gives churches the ability to have dynamic live content filters.

@vinnyjth I know this is daunting but I do think it will be beneficial in the long run for folks. I just made this so you could a take a look and make a decision one way or the other. I tried to deprecate all the non-async functions and then on 2.0, we could put the names back. I didn't change the tests (there's a ton) hoping that 2.0 might be sooner than later and because the functionality of the filter didn't change at all.

Don't think any.

### How do I test this PR?

Boot up the app and make sure everything looks ok.

None.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
@vinnyjth I fixed the tests here if you wanted to give this one a look.

vinnyjth:
I went over this PR this morning @redreceipt. 

I think your approach is a valid one, and under certain conditions I think it would be the right one for Apollos. But at this time, I think it doubles the landscape of the content-item datasource without much immediate gain. I think. if we could omit backwards compatibility, it would be easier to tolerate a change like this and be able to handle that change across the growing number of repos. 

However, I think there are a few universally beneficial changes that would make what you're trying to do in the Newspring app easier to accomplish. For example: 
- https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-rock-apollo-data-source/src/index.js#L78 The `.paginate` method doesn't `await` it's cursor, but it could without breaking anything! 
- https://github.com/ApollosProject/apollos-apps/pull/1372/files#diff-b177d4dac4e73f6ec162e470c50adaa9R244 Methods like this (that you've already fixed) can remain fixed. If something might be a async cursor, and you're in an async method, no reason not to await! 

I think if you start from that point, you might be able to get close enough to the point where you can override core methods on your app to accomplish what you need to accomplish. For now, I would rather try and fix the building blocks rather reconfigure the components that those blocks built.

redreceipt:
@vinnyjth I think I understand. So you're saying on the core side, I could make `paginate` await the cursor, leave the `getLiveContentFilter` async function in there, and then on the NewSpring side run all our content queries' cursors through the new filter function? so `userFeed`, `userFeedFeatures`, child and sibling connections, search queries, liked content connections, and any more I can think of. Is that what you're saying?

redreceipt:
actually I wouldn't even need to put the async filter function in core. Only core change needs to be `paginate` to await its cursor. Right?

redreceipt:
so really i'm just overwriting every resolver instead of every data source function right?

PR #1379, opened 2020-03-10
Deprecates Redundant Live Stream Field

redreceipt:

Simplifies the schema a bit. Hopefully this will be the first step in a multi-part improvement of the live streams API.

None. We're not using that field anywhere as far as I know.

### How do I test this PR?

Make sure the yellow squigly shows up in the playground I guess.

None.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
Throwing this @vinnyjth's way to confirm this is a 👍 schema change

vinnyjth:
@redreceipt Maybe we need to talk more 🙃I don't think this is depreciated. We don't query for this field in the app, currently, but it's a field we've used in the past and I can see us use again to check if a specific content item is live.

vinnyjth:
I'm open to hearing about why you think we should depreciate it :D

redreceipt:
Just trying to clean up the confusing live stream logic flow. Thought this was a good first step. You can always get the exact same thing with no extra cost by querying `liveStreams` and comparing the ID.

vinnyjth:
Let's see if we can clean up the `dataSources` first, and then clean up the schema to match :) I think this unidirectionality is fine and useful. I think Willow might use this field, also.

PR #1383, opened 2020-03-12
Moves User Web View to Packages

redreceipt:

Step one of two PRs to move the user web view out of the app package. Next step will be to reference the new one in the template package. This also cleans up a bit and does everything in #1074

- Instead of using `NavigationService`, you just pass in `navigation`.
- Exports context instead of individual provider and consumer
- Removed deprecated query field and replaced with new one

### How do I test this PR?

Look at the story

Added some coverage.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
@redreceipt i tried to mention this in the other PR, but we do have a [`RockAuthedWebBrowser`](https://github.com/ApollosProject/apollos-apps/tree/master/packages/apollos-ui-connected/src/RockAuthedWebBrowser) already in this package that I think we should use here. It's the same thing right?

redreceipt:
> @redreceipt i tried to mention this in the other PR, but we do have a [`RockAuthedWebBrowser`](https://github.com/ApollosProject/apollos-apps/tree/master/packages/apollos-ui-connected/src/RockAuthedWebBrowser) already in this package that I think we should use here. It's the same thing right?

That's an inappbrowser

conrad-vanl:
@redreceipt ahhhhhh now that's confusing 😉 Sorry about that.

So this is a "WebView" - what's the difference? It's technically a browser...that's in the app. So what's the difference in use case? What does this do that RockAuthedWebBrowser doesn't?

I think this is opportunity to provide clarity and create something that's easier to understand. If we had a `RockAuthedWebBrowser` and `UserWebView` - as a developer by those names I have no clue what the difference is. If we had `RockAuthedWebBrowser` and `RockAuthedWebView` ... that might give a little clarity (as then the only distinction is `Browser` vs `View`).

Thoughts on how we can better provide that clarity?

redreceipt:
I like that idea. I'll go with that unless @vinnyjth has other thoughts.

vinnyjth:
> If we had RockAuthedWebBrowser and RockAuthedWebView ... that might give a little clarity (as then the only distinction is Browser vs View).

💯

redreceipt:
Ok how about now

PR #1386, opened 2020-03-12
Adds 'up next' support to GraphQL API

vinnyjth:

This PR adds the ability for ContentSeriesContentItems to query for the "up next" item. That item is the first item that comes after the last item a user completed it. Example: (bold represents items that have been completed

A. B. **C.** D. E. <- `C` would be picked

A. B. C. D. E. <- `A` would be picked

**A. B. C. D. E.** <- `null` would be picked

There's some code design choices I had to make that will have an impact on performance. Those have been documented through comments in the PR. 

### How do I test this PR?

You can simulate this PR in action by using the query/mutation below.

1. Login to GraphQL via mutations and storing the auth token.
2. Run the "series" query. Note that title and `id` of the upNext item.
3. Plug that `id` into the `readNode` mutation and run it. 
4. Now run the series query again. Note that the result has changed.

```
query series {
  node(id:"ContentSeriesContentItem:b74c2d295a0cf1340232ffa8be856c00"){
    id
    ... on ContentSeriesContentItem {
      upNext {
        id
        title
      }
    }
  }
}

mutation readNode {
  interactWithNode(action:COMPLETE, nodeId:"WeekendContentItem:836a837a1aec93fc163d45315340ff0c"){
    success
  }
}
```

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
I need to write tests for this, and then it will be ready for review.

packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -459,6 +460,48 @@ export default class ContentItem extends RockApolloDataSource {
...
+      .orderBy()
+      .sort([
+        { field: 'Order', direction: 'desc' },
+        { field: 'StartDateTime', direction: 'desc' },
+      ])
+      .get();
+
+    // Returns the item _after_ the most recent item you have interacted with.
+    let lastItem = null;
+    for (let i = 0; i < childItems.length; i += 1) {
```
vinnyjth: This method in particular is something I am keen to optimize long term. I thought about a number of potential implementations, and I picked this `for` loop because it will do the least amount of querying possible, and it's simple. 

Other ideas I had: 

1. Query for all interactions at once, using `Promise.all()` and use roughly the same logic (will use `.find` instead of a for loop). Simple, but potentially crippling for a 365 day journal
2. Use the logic below but combined with a paginated approach. Query for pages of 20 interactions, check to see if the most recent item in that chunk, then query for the next chunk. Complex, but best balance of performance in best and worst case scenarios.  
3. Use a Rock Plugin that we write.

vinnyjth: If we are okay with this initial implementation, I think we can move forward with it for. I wanted to bring to light that there _are_ other solutions and start the conversation for potential rock plugin updates.

vinnyjth:
@conrad-vanl @burkeshartsis This PR is ready for review.

packages/apollos-data-connector-rock/src/content-items/data-source.js
conrad-vanl: What about a binary search?

conrad-vanl: Actually, I don't think a binary search would work

conrad-vanl: Could we store extra data on the interactions object so we could query for interactions directly?

conrad-vanl: For example, storing the `parentContentItemId`

vinnyjth: I don't think it applies here :/. Binary search would help us find an item that has been "read" X times, in a list sorted by the number of times read. Please prove me wrong.

vinnyjth: We could kinda work with it if we assumed that the user had read every single item up to the item they stopped at (if the list was was sorted) but it's not :/

vinnyjth: > Could we store extra data on the interactions object so we could query for interactions directly?

Yes we can, but then we have to know "oh if we have an interaction on a content item, we have to store it's parentChannelId" and I'm not sure if it's dangerous territory or not. What if we really want to do something similar, but by the ContentType? 

We are limited to 3 pieces of data, each in a field named CustomField1, CustomField2 and CustomField3 that we have to keep track of. We could use those! I'm not opposed to it, but I would honestly rather build a Rock Plugin to query better before relying on those.  

We could also store data as serialized JSON. Not a bad option, but then we can't query for data by properties on that serialized JSON.

vinnyjth: Also item's can have multiple `parentContentItemIds` 😭 I don't know of a way to store more than one of those on an interaction and still be able to query for them

vinnyjth: A rock plugin could do one of two things

1. Take a list of `ApollosIds` and return a list of interactions (optimal, imo) 
2. Take a `parentContentItemId` and return a list of interactions. Better in some ways, (query string will be smaller) but ties us to rock.

conrad-vanl: This solution really makes me nervous due to our previous issues with likes... it seems very similar. 

Another user case I could see us using a lot is detecting % progression within a series (Series A is 50% completed). We might even want that info to display in feeds.

What if we cached this kind of stuff in redis, by user, without an expiration date?

vinnyjth: > This solution really makes me nervous due to our previous issues with likes... it seems very similar.

I agree, I think this at best a hack and shouldn't end up in use in production. 

>  Another user case I could see us using a lot is detecting % progression within a series (Series A is 50% completed).

Also agree. I'm using roughly the same mechanism [in this PR](https://github.com/ApollosProject/apollos-apps/pull/1389)

> What if we cached this kind of stuff in redis, by user, without an expiration date?

I think this is a valid situation, _except_ for situations with high user onboardings. If we are showing any of this data on the home feed, or in a place where we drive new users, we can expect Rock to take a massive one time performance dive as we populate our Redis cache. We could take the strategy of saying "if there's no data in Redis, that means they haven't interacted", but then we lose our ability to flush redis. 

I would like to propose this path forward. 

1. Review this PR as is. I'm writing PR's on top of it, and I would hate for there to be a PR pile up that someone has to work through in the next two weeks. I'm also totally fine not. 

2. Today and tomorrow I'll release an update to our Rock plugin, meaning that we'll be able to make a single query for a comma separated list of ForeignIDs. We've done exactly the same thing for ContentItem's, and it hasn't let us down yet. 

3. I'll write a PR that swaps out the mechanism used to fetch all content items in this PR, for the Rock Plugin one, using the approach in this PR if you don't have the plugin endpoint (with plenty of warnings)

PR #1387, opened 2020-03-13
Fixes issue with live video when duration is detected as 0 on the live player

conrad-vanl:

@asleepingpanda found a bug with our player on certain live streams that don't have a duration set. So our video player detects a video as "loaded" with a `duration=0`. this causes a division by 0 RSOD.

- This is going to show a video length of `0:01` in our playhead. 

### How do I test this PR?

This is the stream url that was broken on master: https://christfellow2-lh.akamaihd.net/i/christfellow_1@144364/master.m3u8

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-media-player/src/MediaPlayer/PlayheadState.js
```diff
@@ -42,7 +42,8 @@ class ProviderWithoutApollo extends Component {
   }
 
   handleOnLoad = ({ duration }) => {
-    this.setState({ duration, isLoading: false });
+    if (duration > 0) this.setState({ duration, isLoading: false });
```
burkeshartsis: If I'm reading this correctly this is just protecting against the same thing as line 10 which sets default value of `this.state.duration` to `1`. The problem is that `handleOnLoad` was passing its `duration` of `0` and thus `setState` set `duration` to `0`.  Let's document this so we can keep track of it better moving forward.

```suggestion
    // Only call setState if `duration > 0` to prevent division by zero in other places (`Seeker`, etc.)
    if (duration > 0) this.setState({ duration, isLoading: false });
```

burkeshartsis:
You know in an ideal world a test would have caught this 😉

conrad-vanl:
Adding dont merge tag, @asleepingpanda discovered a minor issue with the play/pause/forward/rewind controls, that while is not a breaking experience (we should still merge this PR), I think I have a fix for it that he's testing, and if it works, we'll add that to htis PR before we merge.

@vinnyjth do we have the ability to do patch releases yet?

conrad-vanl:
> You know in an ideal world a test would have caught this 😉

@burkeshartsis yeah, but we can only test what we know. It was an assumed given that `duration > 0`. We tested that with other live streams and `duration` always had a value. I blame type-less javascript, not tests 😉

vinnyjth:
@conrad-vanl I could easily cut 1.3.1 or we could provide the canary version to @asleepingpanda

asleepingpanda:
@vinnyjth that would be awesome if we could get a patch build out! We're currently running 1.3.0-beta.6 if that effects anything

PR #1388, opened 2020-03-16
Decrease the cache time for the livestreams query

vinnyjth:

Didn't realize this was GraphQL's behavior, but it makes sense 🤷‍♂ If a field isn't included in a response, then it's `maxAge` value is not respected for the returned result. 

## Before

Screen Shot 2020-03-16 at 8 20 11 AM
Screen Shot 2020-03-16 at 8 21 12 AM

## After

Screen Shot 2020-03-16 at 8 26 18 AM

Long term there's a better fix for this PR, but we'll need to be clever about it because it will cause breaking changes. 

1. Setup the `LiveStreams` query so that it _always_ returns content items, even if the `liveStream.isLive` is `false. 
2. Change the client LiveStream provider so that it uses the `true/falseness` of `LiveStream.isLive` rather than the presence of the item in the array. **I'll include that fix in this PR**. 

If we do step 1, without first deploying step 2, old versions of the app will display content as live perpetually. 

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-data-schema/src/index.js
+++ b/packages/apollos-data-schema/src/index.js
@@ -526,7 +526,7 @@ export const liveSchema = gql`
   extend type Query {
     liveStream: LiveStream
       @deprecated(reason: "Use liveStreams, there may be multiple.")
-    liveStreams: [LiveStream]
+    liveStreams: [LiveStream] @cacheControl(maxAge: 10)
   }
 
   extend type WeekendContentItem {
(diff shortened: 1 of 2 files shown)


redreceipt:
If I may disagree, I think the max age would suffice here. returning content items even when there's no active live stream is confusing. plus this requires an app update. I thought the API functioned better as is.

redreceipt:
So essentially it makes the `liveStreams` query function as a `mostRecentSermon` query

vinnyjth:
Okay - I buy that @redreceipt. I don't think we need to back out the client changes in this PR though, even if it's not necessary it is more correct.

redreceipt:
> Okay - I buy that @redreceipt. I don't think we need to back out the client changes in this PR though, even if it's not necessary it is more correct.

Agreed.

PR #1390, opened 2020-03-16
Safer handling of getting horizontal content item fetch more results

conrad-vanl:

Have been getting off-and-on errors navigating content and content children on The Porch App. Such as:

Screen Shot 2020-03-16 at 2 40 24 PM

and 

Screen Shot 2020-03-16 at 2 41 06 PM

Looking at network stacks, it appears that all queries were successful. I am unsure why `null` is being returned in this scenario yet, but it seems like something the UI should handle without throwing errors. Thoughts?

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-connected/src/HorizontalContentSeriesFeedConnected/HorizontalContentSeriesFeedConnected.js
+++ b/packages/apollos-ui-connected/src/HorizontalContentSeriesFeedConnected/HorizontalContentSeriesFeedConnected.js
@@ -106,7 +106,8 @@ class HorizontalContentSeriesFeedConnected extends Component {
               const connection = isParent
                 ? 'childContentItemsConnection'
                 : 'siblingContentItemsConnection';
-              const newEdges = get(fetchMoreResult, `node.${connection}.edges`) || []);
+              const newEdges =
+                get(fetchMoreResult, `node.${connection}.edges`) || [];
 
               return {
                 node: {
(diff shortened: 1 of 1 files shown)


conrad-vanl:
Needs someone to run `--fixlint`. Next time I have this repo pulled up I'll run it...otherwise whoever reviews this if you could be so kind 😅😇

packages/apollos-ui-connected/src/HorizontalContentSeriesFeedConnected/HorizontalContentSeriesFeedConnected.js
```diff
@@ -106,7 +106,7 @@ class HorizontalContentSeriesFeedConnected extends Component {
               const connection = isParent
                 ? 'childContentItemsConnection'
                 : 'siblingContentItemsConnection';
-              const newEdges = get(fetchMoreResult.node, connection, []).edges;
+              const newEdges = get(fetchMoreResult, `node.${connection}.edges`) || []);
```
vinnyjth: @conrad-vanl any reason you didn't continue to rely on the `default` arg of `get? 
aka
```suggestion
              const newEdges = get(fetchMoreResult, `node.${connection}.edges`, []));
```

conrad-vanl: because it doesn't work the same way. The `default` arg of `get` is only returned if the value is `undefined`, wheres ` || []` returns `[]` if the value is `falsey`.

If we depend on an array later on in the code, you should _not_ use the default arg of `get`.

vinnyjth: 👍 Makes perfect sense. That nuance of the default value has bit me before.

PR #1391, opened 2020-03-16
Allows Custom Liked Feed

redreceipt:

Simply allows for a custom tile component to be passed to `HorizontalLikedContentFeed` similar to usual pattern.

None.

### How do I test this PR?

Look at the tests and stories.

None.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-connected/src/HorizontalLikedContentFeedConnected/HorizontalLikedContentFeed/index.js
```diff
@@ -66,18 +71,15 @@ class HorizontalLikedContentFeed extends Component {
     coverImage: [],
   };
 
-  titleImageItem = ({ item }) => (
+  titleImageItem = (TileComponent) => ({ item }) => (
```
conrad-vanl: Nice usage of a HOF! Buuuuut... I don't think it's the right answer here. Since this function is defined on the class, _I believe it's_ better to access `this.props.Component` from within this function, rather then using an anonymous function like this. I'll explain why below, but first, here's an example of what I think you should do instead:

```es6
titleImageItem = ({ item }) => {
  const { Component } = this.props;
  return (
    {/* ... */}
      <Component isLoading={item.isLoading} contentId={item.id} />
    {/* ... */}
  );
}
```

Reason is: when you call `this.titleImageItem(Component)` below, everytime it renders, a new function is generated. If `StyledHorizontalTileFeed` is a pure component ([it is!](https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-ui-kit/src/HorizontalTileFeed/index.js#L9)) then it would re-render _every time_ this component re-rendered, even if it didn't need to, simply because a newly created function is generated each render.

Sometimes this is unavoidable, and it's probably not going to cause any noticeable performance issues. However, for class-based components, it's always a good idea to stick to using `this` in methods when you can, keep your argument API as small as possible, and try to keep props you're passing into child components as pure as possible.

redreceipt:
> PS...A test would be _really_ nice to add, showing a custom Component being passed in 😉
> And a story with a custom Component would be even 💯

Ok ok. I haven't yet tried the linking stuff to get the app to boot yet. I was secretly hoping it would just magically work. I guess this is as good a time as any...

packages/apollos-ui-connected/src/HorizontalLikedContentFeedConnected/HorizontalLikedContentFeed/index.js
```diff
@@ -58,6 +58,11 @@ class HorizontalLikedContentFeed extends Component {
     content: PropTypes.arrayOf(
       PropTypes.any // this component doesn't care about the shape of `node`, just that it exists
     ),
+    Component: PropTypes.node,
```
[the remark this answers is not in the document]
redreceipt: done

packages/apollos-ui-connected/src/HorizontalLikedContentFeedConnected/HorizontalLikedContentFeed/index.js
redreceipt: done.

PR #1392, opened 2020-03-17
Allows Custom Chip Icon Color

redreceipt:
## DESCRIPTION

Screen Shot 2020-03-17 at 10 49 20 AM

You couldn't change the color of the chip icon. Now you can.

None.

### How do I test this PR?

Tests and storybook.

Snaps.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-kit/src/Chip/__snapshots__/Chip.tests.js.snap
+++ b/packages/apollos-ui-kit/src/Chip/__snapshots__/Chip.tests.js.snap
@@ -41,6 +41,163 @@ exports[`The Chip component should be selected 1`] = `
 </View>
 `;
 
+exports[`The Chip component should render a red icon 1`] = `
+<View
+  bordered={false}
+  disabled={false}
+  pill={false}
+  style={
+    Object {
+      "alignItems": "center",
+      "backgroundColor": "#A5A5A5",
+      "borderColor": "#A5A5A5",
+      "borderRadius": 16,
+      "borderWidth": 2,
+      "elevation": 2,
+      "flexDirection": "row",
+      "height": 32,
+      "justifyContent": "center",
+      "opacity": 1,
+      "overflow": "hidden",
+      "paddingHorizontal": 12,
+    }
+  }
+>
+  <RNSVGSvgView
+    align="xMidYMid"
+    bbHeight={16}
+    bbWidth={16}
+    height={16}
+    meetOrSlice={0}
+    minX={0}
+    minY={0}
+    style={
+      Array [
+        Object {
+          "backgroundColor": "transparent",
+          "borderWidth": 0,
+        },
+        Object {},
+        null,
+        Object {
+          "flex": 0,
+          "height": 16,
+          "width": 16,
+        },
+      ]
+    }
+    vbHeight={24}
+    vbWidth={24}
+    width={16}
+  >
+    <RNSVGGroup
+      fill={
+        Array [
+          0,
+          4278190080,
+        ]
+      }
+      fillOpacity={1}
+      fillRule={1}
+      font={Object {}}
+      matrix={
+        Array [
+          1,
+          0,
+          0,
+          1,
+          0,
+          0,
+        ]
+      }
+      opacity={1}
+      originX={0}
+      originY={0}
+      propList={Array []}
+      rotation={0}
+      scaleX={1}
+      scaleY={1}
+      skewX={0}
+      skewY={0}
+      stroke={null}
+      strokeDasharray={null}
+      strokeDashoffset={null}
+      strokeLinecap={0}
+      strokeLinejoin={0}
+      strokeMiterlimit={4}
+      strokeOpacity={1}
+      strokeWidth={1}
+      vectorEffect={0}
+      x={0}
+      y={0}
+    >
+      <RNSVGPath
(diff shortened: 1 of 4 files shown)


packages/apollos-ui-kit/src/Chip/index.js
```diff
@@ -74,6 +82,7 @@ Chip.propTypes = {
   icon: PropTypes.string,
   iconStyles: PropTypes.object, // eslint-disable-line react/forbid-prop-types
   iconSize: PropTypes.number,
+  iconColor: PropTypes.string,
```
conrad-vanl: `iconFill` might be a better name, since it more closely matches to what gets passed to the underlying `<Icon/>` component

redreceipt:
> However, I'd expect to see `red` somewhere in your snapshot, and there's not. Which is hinting that the test isn't actually working for some reason...

I linked the packages but Storybook doesn't seem to be working. Has anyone gotten Storybook to boot on the new templates package?

packages/apollos-ui-kit/src/Chip/__snapshots__/Chip.tests.js.snap
```diff
@@ -41,6 +41,163 @@ exports[`The Chip component should be selected 1`] = `
...
+    }
+    vbHeight={24}
+    vbWidth={24}
+    width={16}
+  >
+    <RNSVGGroup
+      fill={
+        Array [
+          0,
+          4278190080,
```
redreceipt: @conrad-vanl this may be where the icon is red but I don't speak SVG

packages/apollos-ui-kit/src/Chip/index.js
redreceipt: I changed the names

packages/apollos-ui-kit/src/Chip/__snapshots__/Chip.tests.js.snap
conrad-vanl: Screen Shot 2020-03-17 at 10 03 07 AM

Doesn't look like it.

redreceipt: sorry, wrong line

packages/apollos-ui-kit/src/Chip/__snapshots__/Chip.tests.js.snap
```diff
@@ -41,6 +41,163 @@ exports[`The Chip component should be selected 1`] = `
...
+      vectorEffect={0}
+      x={0}
+      y={0}
+    >
+      <RNSVGPath
+        d="M13.13 20.2h-2.26l1.4-1.4.03-.02.03-.03c6.37-5.6 8.06-7.67 8.06-10.27 0-2.2-1.7-3.9-3.9-3.9-1.23 0-2.48.6-3.28 1.54L12 7.57l-1.23-1.45C9.97 5.18 8.72 4.6 7.5 4.6c-2.2 0-3.9 1.67-3.9 3.88 0 2.6 1.7 4.66 8.07 10.27l.03.03.04.03 1.4 1.4zM10.8 4.03l-.03.03h.1c.4.3.8.64 1.13 1.03.33-.4.72-.75 1.14-1.05h.1c-.02 0-.03-.02-.04-.03.97-.66 2.13-1.03 3.3-1.03C19.6 3 22 5.4 22 8.47c0 3.8-3.4 6.88-8.6 11.46l-1.4 1.4-1.4-1.4C5.4 15.36 2 12.27 2 8.48 2 5.38 4.4 3 7.5 3c1.17 0 2.33.37 3.3 1.03z"
+        fill={
+          Array [
+            0,
+            4294901760,
```
redreceipt: @conrad-vanl I meant this one

Screen Shot 2020-03-17 at 1 10 59 PM

conrad-vanl:
@burkeshartsis I _think_ I agree with you, and I'm also willing to accept this PR at the same time 😆 

I'd like you to say what we should do here - if you want to do something different, lay that out for @redreceipt otherwise I think this is a felt pain point and a reasonable solution. FWIW, if that solution is just simply - instead of passing more props to `<Icon>`, just add an `IconComponent` prop type deal - I'd be good with that!

burkeshartsis:
I 100% agree it's a felt need and a hindsight on our part for not exposing this functionality sooner. I would vote for the following changes:

- [ ] Allow users to pass in `children` in the same way we do in `Button` to handle this situation.
- [ ] Keep the current conditional `title` logic.
- [ ] Remove the `withIcon` prop/styling from `TitleText` and move it to the current icon's style. In the future this will be handled by the `Icon` component passed in through `children`.
- [ ] Keep the current conditional icon configuration/props for backwards compatibility.
- [ ] Add depreciation warnings to the icon related props suggesting users use `children` to pass an `Icon`
- [ ] Add a `TODO:` to the deprecated props to remind us to remove these in the future.
- [ ] Copy the related `Button` stories/tests that for passing an icon to the `Chip` stories/tests
- [ ] Update the `ActionCard` test and story that uses `Chip` to reflect the api change (this is the only place in Apollos Core that we use `Chip`).

_Note: this may not be a complete list of necessary changes_

If these changes are made correctly, we should have 100% backwards compatibility + depreciation warnings and a single new `children` prop to flexibly handle icons and many other situations.

redreceipt:
I can start on that. So I'm clear, in order to change the color of an icon, I'll need to pass a custom `<Icon>` and a custom `<Text>` component through the `children` prop?

burkeshartsis:
@redreceipt correct, except I would keep the current conditional `Title` logic so that if you pass a `title` it will render the styled `TitleText` component. But in theory, if you wanted a different `TitleText` component you could not pass a `title` and pass a custom component as part of the `children` prop. Less of a this or that more of a this and/or that. Only down side to this is that you can't easily put the `Icon` after the `TitleText` but maybe we'll cross that bridge when we get to it considering you couldn't do that before and at least now it would be possible albeit by way of copying `TitleText`.

```render() {
  return (
    {children}
    {title ? <TitleText>{title} : null}
  );
}
```

redreceipt:
closing in favor of #1394

PR #1393, opened 2020-03-17
Exports RockAuthedWebView

redreceipt:

Forgot to export these last time 😬

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-connected/src/RockAuthedWebView/index.js
+++ b/packages/apollos-ui-connected/src/RockAuthedWebView/index.js
@@ -1,5 +1,5 @@
 // Provider API for WebBrowser that injects theme values and defaults to the web browser:
-import React, { createContext } from 'react';
+import React from 'react';
 import gql from 'graphql-tag';
 import { Query } from 'react-apollo';
 import { ModalView } from '@apollosproject/ui-kit';
--- a/packages/apollos-ui-connected/src/RockAuthedWebView/index.js
+++ b/packages/apollos-ui-connected/src/RockAuthedWebView/index.js
@@ -48,7 +48,6 @@ const RockAuthedWebView = ({ url, navigation }) => {
         const cookie = get(data, 'currentUser.rock.authCookie', '');
         return (
           <Browser
-            testID="browser"
             cookie={cookie}
             url={uri}
             modal={modal}
--- a/packages/apollos-ui-connected/src/RockAuthedWebView/index.js
+++ b/packages/apollos-ui-connected/src/RockAuthedWebView/index.js
@@ -48,6 +48,7 @@ const RockAuthedWebView = ({ url, navigation }) => {
         const cookie = get(data, 'currentUser.rock.authCookie', '');
         return (
           <Browser
+            testID="browser"
             cookie={cookie}
             url={uri}
             modal={modal}
@@ -67,10 +68,10 @@ RockAuthedWebView.navigationOptions = ({ navigation, navigationOptions }) => ({
   header: navigation.getParam('modal', false) ? null : navigationOptions.header,
 });
 
-const OpenWebView = ({ url, navigation }) =>
+const openWebView = ({ url, navigation }) =>
   navigation.navigate('RockAuthedWebView', { url });
 
-const WebViewContext = createContext(OpenWebView);
+const WebViewContext = createContext(openWebView);
 
 export default RockAuthedWebView;
-export { GET_USER_COOKIE, WebViewContext, OpenWebView };
+export { GET_USER_COOKIE, WebViewContext, openWebView };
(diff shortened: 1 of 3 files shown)


packages/apollos-ui-connected/src/RockAuthedWebView/index.js
```diff
@@ -48,6 +48,7 @@ const RockAuthedWebView = ({ url, navigation }) => {
         const cookie = get(data, 'currentUser.rock.authCookie', '');
         return (
           <Browser
+            testID="browser"
```
redreceipt:
```suggestion
```

conrad-vanl: ?

packages/apollos-ui-connected/src/index.js
```diff
@@ -35,6 +35,11 @@ export MediaControlsConnected, {
   MediaControls,
 } from './MediaControlsConnected';
 export RockAuthedWebBrowser from './RockAuthedWebBrowser';
+export RockAuthedWebView, {
+  GET_USER_COOKIE,
+  WebViewContext,
+  openWebView,
```
conrad-vanl: do we need to export all of these other things? Should we keep the exports namespaced under `RockAuthed`?

packages/apollos-ui-connected/src/RockAuthedWebView/index.js
```diff
@@ -67,10 +68,10 @@ RockAuthedWebView.navigationOptions = ({ navigation, navigationOptions }) => ({
   header: navigation.getParam('modal', false) ? null : navigationOptions.header,
 });
 
-const OpenWebView = ({ url, navigation }) =>
+const openWebView = ({ url, navigation }) =>
   navigation.navigate('RockAuthedWebView', { url });
```
conrad-vanl: This assumes that the app has a `RockAuthedWebView` route. I don't know if this code belongs here, on second thought. Furthermore, and I should've caught this beofre, but I _think_ we're trying to remove dependency on `react-navigation` from `ui-connected`. @burkeshartsis is that true?

packages/apollos-ui-connected/src/RockAuthedWebView/index.js
```diff
@@ -67,10 +68,10 @@ RockAuthedWebView.navigationOptions = ({ navigation, navigationOptions }) => ({
...
 
-const WebViewContext = createContext(OpenWebView);
+const WebViewContext = createContext(openWebView);
```
conrad-vanl: Can you explain to me what this does?

packages/apollos-ui-connected/src/index.js
redreceipt: Ok so client-side, we'll need the component (for navigation) and the `open` function at the least. Don't think we'd actually need the query or the context to use the functionality.

packages/apollos-ui-connected/src/RockAuthedWebView/index.js
redreceipt: it was there before but I can't think of a reason it's needed. But I don't know all the places this code is currently being used. I believe you can get all the same stuff by just using `open` function directly.

conrad-vanl:
@redreceipt what would you feel about only exporting `RockAuthedWebView`, and not the other things? I don't think we _need_ them and I'm hesitant exposing more API / debt then we need to here.

redreceipt:
> @redreceipt what would you feel about only exporting `RockAuthedWebView`, and not the other things? I don't think we _need_ them and I'm hesitant exposing more API / debt then we need to here.

So you would just navigate client-side. I like that the best. I could probably clean up more code if we did it that way.

redreceipt:
Ok how about this? Simple to use I think. I can't test on the app right now because I'm having difficulties linking again but it seems pretty straightforward.

packages/apollos-ui-connected/src/index.js
burkeshartsis: So the general thinking behind what our connected components export is that they export themselves, their query, and the pure component that the connected component renders. This is so that if you want to use it as is you use the connected component. We export the query so that if you want to write your own connected component with our query and pure component you can do that too. This just affords flexibility.

That said, we should probably be thinking namespace more.

packages/apollos-ui-connected/src/RockAuthedWebView/index.js
burkeshartsis: Generally speaking, you are correct @conrad-vanl. I would export the Component you navigate to and leave the navigation to the screen/location/component that is rendering the `RockAuthedWebView`.

packages/apollos-ui-connected/src/index.js
conrad-vanl: I don't think we should export `openWebView` at all. All it does is call a `navigate` function for a route that's not actually defined anywhere in this code - for that function to work, you have to specify that exact route in your own navigator.

PR #1394, opened 2020-03-17
Allows Passing Custom Chip Children

redreceipt:
## DESCRIPTION

Screen Shot 2020-03-17 at 3 10 47 PM

See #1392 

Deprecated `icon` prop

### How do I test this PR?

Stories

Snaps

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-kit/src/Chip/Chip.stories.js
```diff
@@ -37,4 +39,17 @@ storiesOf('ui-kit/Chip', module)
...
+  ))
+  .add('Custom children', () => (
+    <Chip onPress={() => {}}>
+      <Text>Custom children</Text>
+    </Chip>
+  ))
+  .add('Custom icon', () => (
+    <Chip onPress={() => {}} title="Custom Icon">
+      {/* eslint-disable-next-line */}
+      <View style={{ paddingRight: 8 }}>
```
conrad-vanl: @burkeshartsis only problem I see with using this `children` props approach is that you now have to know how to pad your icon to show up right

burkeshartsis: I agree but if you are a developer using our `ui-kit` at this level this feels like a reasonable ask. The developer is already importing `Chip` and positioning it on the screen. I don't think it's crazy to assume that if they pass an `Icon` via `children` that they would see the need for some padding and know how to accomplish that. Plus we don't even know how much padding they need. This gives the user more control and at this level of the platforms api that feels right.

redreceipt: We could just give the title a bit of `paddingHorizontal`? similar to before but it would keep it centered

redreceipt: or even better, we could give the whole view `paddingHorizontal` and then `flex: space-between` the elements

burkeshartsis: adding `paddingHorizontal` without a condition would make the chip "wider" if there were no other children. This would change its default "shape." 

You could conditionally add padding to `TitleText` if there were `children` but, as I mentioned above, you would be making a blind assumption as to how much padding you needed.

burkeshartsis: The other option here, which @conrad-vanl suggested early on is to use a `Icon` prop which takes a component. By being more specific you can make stronger assumptions like padding the text. The down side to this, outside what I have mentioned already, is that we already have an established api pattern for this with `Button`. I'd just assume keep it consistent.

redreceipt:
Up to you @burkeshartsis to merge as is or tell me to change something.
I'll do whatever.


--
Michael Neeley
Junior Web Developer
NewSpring Church
www.newspring.cc
Cell: (803) 622-6599

redreceipt:
@burkeshartsis what's the decision here?

PR #1397, opened 2020-03-24
Adds Google Cast

### LINKED ISSUE #448: Implement Chromecast
## TODO

- [ ] Needs design
- [ ] Needs library verification

## User Story

* split out from #250

* [ ] It should [SPEC_DETAIL]

redreceipt:
## DESCRIPTION

Kapture 2020-03-24 at 1 33 54

Adds Google Cast Support. The media player interface stays synced with the cast remote device.

NOTE: Airplay doesn't work on the sim and Chromecast doesn't work on the Android sim. Will probably want to test on real devices. With this PR: https://github.com/ApollosProject/apollos-templates/pull/19 

Moved icons above the seeker bar. On Android, it will just show one icon in the middle, on iOS it will show two icons even spaced out.

The only thing I don't love about my solution is in order to keep the cast and the on-device seeker in sync, I have to keep the video playing on the device with the screen turned off. *Maybe* a slight performance hit?

### How do I test this PR?

Boot it up on both device types with Airplay/Chromecast available/not available.

Mocked the button.

## TODO:

- [x] Closes #448
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
Any thoughts on the test failure here? Confusing because *we are* mocking `Animated`. The only thing that changed is I'm passing an animated value as a prop into `FullScreenControls` and then using `stopAnimation` to get the value later.

Screen Shot 2020-03-24 at 9 46 11 AM

packages/apollos-ui-media-player/src/MediaPlayer/FullscreenControls.js
```diff
@@ -104,6 +112,19 @@ class FullscreenControls extends PureComponent {
     }),
     showAudioToggleControl: PropTypes.bool,
     showVideoToggleControl: PropTypes.bool,
+    cast: PropTypes.shape({
```
vinnyjth: Nice propTypes shape 👍

conrad-vanl:
@vinnyjth are you reviewing this, or should I jump in?

vinnyjth:
@conrad-vanl Feel free to jump in. I started but got pulled off into other stuff 👍

redreceipt:
> 1. Move the casting logic out to its own functional container...a separate component, a state controller, or something else.
> 2. API added so that you can atleast disable the cast button, and maybe so that you can modify the data that's sent to be cast (title, url, etc)

#2 is done. #1 is going to be pretty tricky because of how closely I've synced the controls UI with the casting status and functionality. Not saying it isn't possible but I may have to add a lot of complicated code to pass functions and props up and down. I'll take a look...

redreceipt:
@conrad-vanl is there a reason why some things in `FullScreenControls` are kept up with using state (`this.state.isScrubbing`) and some are just static attributes (`this.isMuted` and `this.isVideo` among others). That's one thing that's making it difficult to bounce the cast stuff out to a separate component.

conrad-vanl:
So `isMuted,`, `isPlaying`, etc come from the `GET_CONTROL_STATE` Query from local graphql state. It's not a great design, and setting those values [here](https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-ui-media-player/src/MediaPlayer/FullscreenControls.js#L252-L254) really is a hack and a massive code smell. We absolutely do not want to call `setState` here though because that would introduce a nasty render loop.

`isScrubbing` I believe is the only value that uses local component state. I think we might've originally tried to use Apollo GraphQL state for this but it caused performance issues. 

I believe you should be able to write a Cast component that just "consumes" state via the `GET_CONTROL_STATE` mutation, that fires graphql mutations (available [here](https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-ui-media-player/src/MediaPlayer/mutations.js)) to respond to Cast events.

conrad-vanl:
One option I think you could do is:

- Create a `<CastButton />` that starts the casting flow

- Create a `<CastingController />` component that listens to `GoogleCast.EventEmitter` and keeps the player state in-sync via our available [graphql mutations](https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-ui-media-player/src/MediaPlayer/mutations.js). You might have to add some of your own data to the player state, such as setting [`isCasting: true`](https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-ui-media-player/src/Provider.js#L130), which would be fine.

redreceipt:
> One option I think you could do is:
> 
> * Create a `<CastButton />` that starts the casting flow
> * Create a `<CastingController />` component that listens to `GoogleCast.EventEmitter` and keeps the player state in-sync via our available [graphql mutations](https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-ui-media-player/src/MediaPlayer/mutations.js). You might have to add some of your own data to the player state, such as setting [`isCasting: true`](https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-ui-media-player/src/Provider.js#L130), which would be fine.

Is there a reason all the logic couldn't happen in one component?

conrad-vanl:
Well you could render a CastButton in multiple places, but there should only be _one_ listener for maintainer of the casting state. A better model would probably be a `<CastProvider />` component that's rendered at a higher level and then the `<CastButton>` just becomes a consumer of that provider.

In other words... if you rendered two `<CastButton/>`s and they both were listening to cast events and maintaining state I think it would get into some funky edge cases.

conrad-vanl:
But these comments come from a bit of a heightened view - I'm not as in the weeds on it as you are. If you think you see a better direction go for it. Ultimately we need to make sure we're maintaining separation of concern.

redreceipt:
@conrad-vanl ok everything is moved over except four lines in `FullScreenControls`, 188, 193, 271, and 285. Also, I'd like to be able to move around the seeker head to keep in sync with cast but either I'm real dumb...or the media player code is a mess. I suppose both could be true but I'm having a very hard time following the flow of everything.

Wanna take a look at where I'm at so far and point me in the right direction? I think it's almost to the place you requested.

conrad-vanl:
> the media player code is a mess

This is a safe assumption.

redreceipt:
so @conrad-vanl for this to move forward, you'd like to see this `VideoWindow` get swapped out when media is being cast, perhaps with a simple poster of some sort. Is that correct?

Screen Shot 2020-04-13 at 8 35 57 PM

conrad-vanl:
@redreceipt `VideoWindow` takes a `VideoComponent` prop. I believe the `VideoComponent` prop should be switched to a `Casting`-type component that controls the different states while casting, as well as displaying some poster/cover artwork. If you look inside of the `VideoComponent` component ([such as the one we built for WillowCreek](https://github.com/Differential/willow-creek/blob/master/packages/WillowCreekApp/src/ui/MediaPlayerYoutube/YoutubeVideoWindow.js)) you'll see that there are hooks for managing all of the different video state - playing, pausing, scrubbing, playhead position, etc. I believe this should be your entry point for keeping the mediaplayer in-sync with the cast.

conrad-vanl:
And I would simplify the `CastingButton` component to just trigger the start of the cast / swapping out the `VideoComponent` prop

conrad-vanl:
Said in a different way, the `VideoWindow` component is ultimately what "runs" the entire `MediaPlayer`. Currently, it uses `react-native-video` to display the video on the screen. When the user presses the `pause` button, that makes a state update that the `VideoWindow` component detects, and tells `react-native-video` to pause. While `react-native-video` is playing, `react-native-video` sends status updates to `VideoWindow` which `VideoWindow` uses to inform the rest of the `MediaPlayer` of the video status. These are all functions you need to complete for `Casting`. It makes sense to me to just replace the `VideoComponent` with a `Casting`-based component since we are no longer using `react-native-video` to play the media, we are using the google cast to play the video.

redreceipt:
Sounds good. I’ll get started and let you know if I have questions.

—
Michael

redreceipt:
ok @conrad-vanl this is by no means done, but look over the architecture changes I made and let me know if I'm on the right track please

conrad-vanl:
@redreceipt this is on my list to look at

redreceipt:
I can certainly move the query and poster inside the controller, I like that a lot. But the `FullScreenPlayer` has to switch between local video and cast when cast connects/disconnects. So I think something like this would do the trick `isCasting ? GoogleCastController : VideoWindow`

the `googleCastEnabled` prop would only enable/disable showing the button.

conrad-vanl:
Ah ok I think that makes sense!

redreceipt:
Sike, I moved the query.

redreceipt:
Ok @conrad-vanl give this one a look

conrad-vanl:
@redreceipt this is looking good! I think I'm okay with the code where it's at. Need to look at it a little closer, but I want to do some device testing too. May not be able to get to that until Monday but I will work towards pushing this through.

conrad-vanl:
@redreceipt still working on testing this. All of my chromecast devices were still packed up from the move. I do notice that travis is reporting lint and test failures - might be good to go ahead and fix that

redreceipt:
Tests are failing because I'm using a hook but it works fine in real life. Have you seen this error before?

Screen Shot 2020-04-28 at 4 20 31 PM

redreceipt:
So this is close I think. Currently, I can't sync up the controls. the google cast status event listener doesn't work on app reboot (https://github.com/react-native-google-cast/react-native-google-cast/issues/151) so I forked the repo (https://github.com/ApollosProject/react-native-google-cast/commit/e074e8c2a287131cdf81c98a07fc748f8ae705f6) and added in this bug fix (https://github.com/react-native-google-cast/react-native-google-cast/pull/188).

Could I get someone to publish that repo to npm so we can use it until that PR gets merged into the real thing?

Another option is to not have a working play/pause button, seeker, or skip buttons. And we just wait for the fix.

redreceipt:
so using that "fix" did nothing. I can't get the `MEDIA_STATUS_UPDATED` event to work right after app restart. I disabled the play/pause and seeker until it works right. For now, you can play/pause by clicking the cast button. I left in the code that syncs everything up, it works right the first time you connect and play around with everything.

conrad-vanl:
So @redreceipt am I to understand correctly that casting doesn't work once you leave the app and come back?

burkeshartsis:
What if you just paused the casting when the app is backgrounded or closed. That could be a temporary solution till they release a fix. It also means we wouldn't have to maintain a fork of the package while we wait for the fix to be released.

redreceipt:
Still won't help because when you open it back up, it won't be able to start playing again. It loses the device. I like that you can start casting and close the app. that's how I will use it weekly. And then when you open the app back up, you can click the cast button to play/pause/disconnect.

what we lose until it's fixed is the ability to integrate our own buttons.

and as far as maintaining, it still doesn't work with the forked repo after my testing so I might as well just put it back to the external lib. I'm talking with the guy to get it resolved here. https://github.com/react-native-google-cast/react-native-google-cast/pull/188

redreceipt:
What do we want to do here? two options:

1. Wait for library fix, forked repo "fix" did nothing
2. Ship as is with all media player UI disabled when casting

If we go with 2, a little more device testing is warranted before I'd feel 100% confident. I've seen the sim crash in between casting sessions occasionally. But it may be just native iOS libraries switching back and forth, causing weirdness.

conrad-vanl:
I'm kind of at the opinion of waiting for library fix or trying to fix library ourselves, even if that means making (ehem...forking) or own `apollos-google-cast` library, vs shipping an experience that is fundamentally broken.

However, I am also a fan of "just ship it" and would like to rely on @FrankGrand, @asleepingpanda, @redreceipt, etc thoughts on where we stand on that.

conrad-vanl:
@redreceipt for others looking at this, would you mind recording a quick screencast/video of where this PR is at now? So they can see what the experience is like

asleepingpanda:
So a couple of thoughts on this. I'm not sure what CF's level of urgency is on getting Chrome Cast supported. If the urgency is incredibly high, then I think shipping immediately with known issues is doable so long as we have backlog items in place to watch the repo and put out a patch release immediately when the fix is available (not waiting for a major release of Apollos to fix it)

If our urgency isn't super high, then we would likely just disable Chrome Cast from our app to avoid the potential feedback of Chrome Cast users who experience the bug. We've gotten some very, very specific bug reports recently, so if Chrome Cast is not a high urgency item, we would probably be a bit cautious on rolling out a feature like this with such known issues.

That being said, I see that Chrome Case can be toggled on/off with properties, so I wouldn't have an issue with pushing this repo out if NS needs it asap. (just want to confirm that the enable/disable with props works correctly so we can adjust as need be)

vinnyjth:
My unsolicited 0.02. It’s been a year or two since I’ve used chrome cast, but last time I used it the experience was very poor on iOS, even with apps like Netflix. Desync was frequent and the controls on the controller apps would either work with a large delay or not work, and I would end up jumping over to a the chrome cast app to play/pause/ volume up down.

redreceipt:
@conrad-vanl I've tried to get a GIF, currently stuck wrestling with yarn in the templates repo. 

@FrankGrand @asleepingpanda current status is no bugs, it's the same behavior as this GIF but with no seeker bar and UI buttons except the cast button. When you disconnect, they all reappear.

Kapture 2020-03-24 at 1 33 54

PR #1403, opened 2020-04-01
Dark theme fixes

jordanwade:
## DESCRIPTION
This fixes places in the app that are not using the theme types to control colors. It also adds a few stories to bordered buttons to make sure they are working correctly. And finally, it removes shadows when dark theme.

image

To setup a solid foundation for dark theme to be used in The Porch app.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-auth/src/Entry.js
+++ b/packages/apollos-ui-auth/src/Entry.js
@@ -66,7 +66,11 @@ const Entry = ({
         Platform.OS === 'android' ? StatusBar.currentHeight : 0
       }
     >
-      {BackgroundComponent}
+      {React.isValidElement(BackgroundComponent) ? (
+        BackgroundComponent
+      ) : (
+        <BackgroundComponent />
+      )}
       <FlexedSafeAreaView>
         <ScrollView keyboardShouldPersistTaps="always">
           <PaddedView>
--- a/packages/apollos-ui-auth/src/Entry.js
+++ b/packages/apollos-ui-auth/src/Entry.js
@@ -32,12 +32,8 @@ import {
 } from './styles';
 
 const FullScreenGradientBackground = styled({
-  bottom: 0,
-  left: 0,
-  position: 'absolute',
+  ...StyleSheet.absoluteFillObject,
   resizeMode: 'cover',
-  right: 0,
-  top: 0,
 })(BackgroundView);
 
 const Entry = ({
--- a/packages/apollos-ui-auth/src/Entry.js
+++ b/packages/apollos-ui-auth/src/Entry.js
@@ -31,9 +31,13 @@ import {
   TitleText,
 } from './styles';
 
-const FullScreenImage = styled({
-  resizeMode: 'cover',
+const FullScreenGradientBackground = styled({
+  bottom: 0,
+  left: 0,
   position: 'absolute',
+  resizeMode: 'cover',
+  right: 0,
+  top: 0,
 })(BackgroundView);
 
 const Entry = ({
@@ -66,7 +70,7 @@ const Entry = ({
         Platform.OS === 'android' ? StatusBar.currentHeight : 0
       }
     >
-      <BackgroundComponent />
+      {BackgroundComponent}
       <FlexedSafeAreaView>
         <ScrollView keyboardShouldPersistTaps="always">
           <PaddedView>
@@ -162,7 +166,7 @@ Entry.defaultProps = {
   authTitleText: 'Have we met?',
   promptText:
     'Sign in for a personalized experience that helps you grow and connect with God.',
-  BackgroundComponent: FullScreenImage,
+  BackgroundComponent: <FullScreenGradientBackground />,
   alternateLogin: false,
 };
 
--- a/packages/apollos-ui-auth/src/Password/EmailEntryConnected.js
+++ b/packages/apollos-ui-auth/src/Password/EmailEntryConnected.js
@@ -10,7 +10,10 @@ import Entry from '../Entry';
 class EmailEntryConnected extends Component {
   static propTypes = {
     alternateLoginText: PropTypes.string,
-    BackgroundComponent: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
+    BackgroundComponent: PropTypes.oneOfType([
+      PropTypes.element,
+      PropTypes.elementType,
+    ]),
     // Custom component to be rendered. Defaults to Entry
     Component: PropTypes.oneOfType([
       PropTypes.node,
--- a/packages/apollos-ui-auth/src/Password/EmailEntryConnected.js
+++ b/packages/apollos-ui-auth/src/Password/EmailEntryConnected.js
@@ -10,6 +10,7 @@ import Entry from '../Entry';
 class EmailEntryConnected extends Component {
   static propTypes = {
     alternateLoginText: PropTypes.string,
+    BackgroundComponent: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
     // Custom component to be rendered. Defaults to Entry
     Component: PropTypes.oneOfType([
       PropTypes.node,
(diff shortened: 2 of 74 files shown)


conrad-vanl:
> And finally, it removes shadows when dark theme.

@asleepingpanda is this intended?

conrad-vanl:
@asleepingpanda i think you will like this

kevin-birdy:
@conrad-vanl I was playing with shadows on Porch at first but I don't think it's worth the effort. You'd basically have to create a whole new set with _way_ higher opacities and even then you'll rarely be able to see them. I dig it ✌️

asleepingpanda:
@conrad-vanl this makes me very happy cause we’re probably only a few weeks out from launch. We also removed all shadows when in dark mode. They were either awkwardly light or not visible. Can’t wait for this to be merged into core!

conrad-vanl:
@asleepingpanda do you have any list of issues that you ran into with dark mode?

asleepingpanda:
@conrad-vanl the two major ones are

1. Open the app, change the mode of your phone (light to dark or dark to light) and you’ll see that while most colors change, TEXT COLORS do not change until you close & reopen the app

2. Media player had no theming support at all

I’ll let you know if I can find any other specific ones

conrad-vanl:
Got it. @asleepingpanda this isn't "dark mode" support, it is just "dark theme" support. We'll still need to resource dark mode support separately. I think the issues you're seeing there is more around the switching of theme dynamically.

conrad-vanl:
@jordanwade did you test the media player yet?

asleepingpanda:
Ah! Makes sense. Ok so then just checking out the Media player would be to on the list. Maybe doing a once over on inputs? I’m sorry I don’t have an exact list, but Media Player, Auth and Inputs would be the top ones that immediately come to mind

jordanwade:
@conrad-vanl working on Media Player now 👍 and I'll take a look at the other ones @asleepingpanda mentioned.

packages/apollos-ui-auth/src/Entry.js
```diff
@@ -31,9 +31,13 @@ import {
...
 
-const FullScreenImage = styled({
-  resizeMode: 'cover',
+const FullScreenGradientBackground = styled({
+  bottom: 0,
+  left: 0,
   position: 'absolute',
+  resizeMode: 'cover',
+  right: 0,
+  top: 0,
```
conrad-vanl, line 40:
```suggestion
  ...StyleSheet.absoluteFillObject,
  resizeMode: 'cover',
```

packages/apollos-ui-auth/src/Entry.js
```diff
@@ -66,7 +70,7 @@ const Entry = ({
         Platform.OS === 'android' ? StatusBar.currentHeight : 0
       }
     >
-      <BackgroundComponent />
+      {BackgroundComponent}
```
conrad-vanl, line 73: This is technically a breaking API change. Might have to do something like what [`react-native` core](https://github.com/facebook/react-native/blob/master/Libraries/Lists/VirtualizedList.js#L970-L975), would look something like:

```suggestion
      {React.isValidElement(BackgroundComponent) ? (
        BackgroundComponent
      ) : (
        < BackgroundComponent />
      )}
```

packages/apollos-ui-auth/src/Password/EmailEntryConnected.js
```diff
@@ -10,6 +10,7 @@ import Entry from '../Entry';
 class EmailEntryConnected extends Component {
   static propTypes = {
     alternateLoginText: PropTypes.string,
+    BackgroundComponent: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
```
conrad-vanl, line 13:
```suggestion
    BackgroundComponent: PropTypes.oneOfType([PropTypes.elementType, PropTypes.element]),
```

packages/apollos-ui-auth/src/SMS/PhoneEntryConnected.js
```diff
@@ -10,6 +10,7 @@ import Entry from '../Entry';
 class PhoneEntryConnected extends Component {
   static propTypes = {
     alternateLoginText: PropTypes.string,
+    BackgroundComponent: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
```
conrad-vanl, line 13:
```suggestion
    BackgroundComponent: PropTypes.oneOfType([PropTypes.elementType, PropTypes.element]),
```

packages/apollos-ui-kit/src/ActionBar/ActionBarItem.js
```diff
@@ -39,7 +39,10 @@ ActionBarItem.propTypes = {
 };
 
 export default withTheme(({ theme, ...props }) => ({
-  tint: theme.colors.action.primary,
+  tint:
+    theme.type === 'light'
```
conrad-vanl, line 43: By convention we shouldn't be checking for a theme type here. A church can define any number of theme types.... we need to map to the same theme value regardless of type.

packages/apollos-ui-kit/src/Avatar/index.js
```diff
@@ -121,5 +121,8 @@ Avatar.propTypes = {
 
 export default withTheme(({ theme, size }) => ({
   themeSize: get(theme.sizing.avatar, size, theme.sizing.avatar.small),
-  iconFill: theme.colors.action.primary,
+  iconFill:
+    theme.type === 'light'
```
conrad-vanl, line 125: By convention we shouldn't be checking for a theme type here. A church can define any number of theme types.... we need to map to the same theme value regardless of type.

packages/apollos-ui-media-player/src/MediaPlayer/MiniControls.js
```diff
@@ -58,7 +58,10 @@ const Controls = styled(({ theme }) => ({
   justifyContent: 'space-between',
   alignItems: 'center',
   paddingRight: theme.sizing.baseUnit * 0.75,
-  backgroundColor: theme.colors.white,
+  backgroundColor:
+    theme.type === 'light'
```
conrad-vanl, line 62: By convention we shouldn't be checking for a theme type here. A church can define any number of theme types.... we need to map to the same theme value regardless of type.

packages/apollos-ui-auth/src/Password/EmailEntryConnected.js
conrad-vanl: FYI: I _think_ these are the right proptypes - https://github.com/facebook/prop-types/pull/211

packages/apollos-ui-auth/src/Entry.js
jordanwade: 👍

packages/apollos-ui-auth/src/Entry.js
jordanwade: Nice! 👍

packages/apollos-ui-auth/src/Password/EmailEntryConnected.js
jordanwade: I'll double check, last time I attempted to use `element` it wouldn't work because of the version of react we were on... 👍

packages/apollos-ui-kit/src/ActionBar/ActionBarItem.js
jordanwade: @conrad-vanl or @burkeshartsis Do you or anyone else have a suggestion on how to handle a situation like this where the primary action looks good on light theme, but not so great on the dark theme? 

Ideas:
1. Should we try remapping the dark theme colors? 
2. Create a switch statement to support light and dark and a default? 
3. Leave it as is using `theme.colors.action.primary` for any type?
4. Other ideas?

conrad-vanl: > Leave it as is using theme.colors.action.primary for any type?

Is a valid option, and we change the dark theme values.

Or, we create a new theme value that all types must support that is still used on all themes.

Basically, none of our components should ever access `theme.type`

PR #1408, opened 2020-04-07
Updated Shadow Value

ridderingand:
## DESCRIPTION
This PR updates the default shadow value for Core

The initial shadows were darker than intended because they were using `.fade(alpha.low)` which is effectively 40% opacity. This matches what is in Figma by using `colors.black` instead of `colors.darkTertiary` and having transparency be 6% instead of 40%.

The current opacity that is being faded is hard-coded instead of being tied to the theme. This is because our lightest alpha value is 40%.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-kit/src/theme/types/light.js
+++ b/packages/apollos-ui-kit/src/theme/types/light.js
@@ -18,8 +18,8 @@ const light = ({ barStyle, colors, alpha }) => ({
       inactive: colors.lightTertiary,
     },
     shadows: {
-      default: Color(colors.darkTertiary)
-        .fade(alpha.low)
+      default: Color(colors.black)
+        .fade(0.06)
         .string(),
     },
     action: {
--- a/packages/apollos-ui-kit/src/theme/types/light.js
+++ b/packages/apollos-ui-kit/src/theme/types/light.js
@@ -19,7 +19,7 @@ const light = ({ barStyle, colors, alpha }) => ({
     },
     shadows: {
       default: Color(colors.black)
-        .fade(0.06)
+        .fade(alpha.high)
         .string(),
     },
     action: {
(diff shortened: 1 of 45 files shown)


burkeshartsis:
Where you intending these shadows to be as heavy has you have them here?

ridderingand:
lol nope. So much for my 0.1 transparency, 🤦‍♂️ I'm going to have to spin up a local environment to be able to get this to work. Kinda in the dark right now.

burkeshartsis:
I got you dawg. More like this?

ridderingand:
Unfortunately, a .gif is probably the only file format that makes it impossible to see what a drop-shadow looks like :)

It's 2px height, 8px blur, black, 6% opacity. Not sure what part of that I messed up, but if it looks like that then yes...more like that 😇

Appreciate the help 🙏

burkeshartsis:
You're not wrong. We can sync up on this tomorrow

packages/apollos-ui-kit/src/theme/types/light.js
```diff
@@ -18,8 +18,8 @@ const light = ({ barStyle, colors, alpha }) => ({
       inactive: colors.lightTertiary,
     },
     shadows: {
-      default: Color(colors.darkTertiary)
-        .fade(alpha.low)
+      default: Color(colors.black)
+        .fade(0.06)
```
burkeshartsis, line 22: If your looking to have and end transparency of 10% than this would work and use our current theme values.
```suggestion
        .fade(alpha.high)
```

ridderingand: Ok. Definitely didn't understand .fade like I though then :) That's ok! Appreciate it dude.

burkeshartsis: You'll want to accept that commit suggestion and then run `yarn test -- -- -u` from the root directory in core to update snapshots. Then look over the changes to make sure you are seeing the correct values. Should just be a change to something like `rgba(0,0,0,0.1)`.

burkeshartsis:
Feel free to merge this if you can ever get tests to pass 😭

PR #1422, opened 2020-04-15
Adds support for Features to be fetched via their ID

vinnyjth:
Adds support for `getFromId` to the Features dataSource, and adds "lazy fetching" for actions/cards on the server side, so getting the definition of a feature doesn't also fetch all the data for a feature. 

I had to rename `Features` -> `Feature`. Not fun, but there's no way around it. We can alias the naming when updating clients, if they are in a rush and use the old naming in a bunch of places. 

### How do I test this PR?

Best way to test this is going to be running the two queries below in GQL playground. The first query demonstrates fetching a list of features, and the second query demonstrates fetching a single feature by ID. You will need to sign in via GQL playground. 

```
query getFeed {
  userFeedFeatures {
    id
    __typename
    ... on VerticalCardListFeature {
      cards {
        id
      }
    }
    ... on HorizontalCardListFeature {
      cards {
        id
      }
    }
    ... on ActionListFeature {
      id
      actions {
        id
      }
    }    
  }
}

query getFeature {
  node(id:"VerticalCardListFeature:9724110384a5109532f35a4980ea9b77f95c2a2b8a0943252964d15273d8a51775529404f74998d4ef79928a0fe0be06dd6972a92140611b25edecac03c8c03e"){
    ... on VerticalCardListFeature {
      cards {
        id
      }
    }
    ... on HorizontalCardListFeature {
      cards {
        id
      }
    }
    ... on ActionListFeature {
      id
      actions {
        id
      }
    }  
  }
}
```
New tests over new functionality. 

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-data-connector-rock/src/index.js
```diff
@@ -32,6 +32,6 @@ export {
   Group,
   Utils,
   BinaryFiles,
-  Features,
+  Feature,
```
conrad-vanl, line 35: Do we need to technically still export `Features` for this to be nonbreaking?

packages/apollos-data-connector-rock/src/features/data-source.js
```diff
@@ -59,16 +73,19 @@ export default class Features extends RockApolloDataSource {
     isFeatured = false,
   }) {
     // Generate a list of cards.
-    const cards = await this.runAlgorithms({ algorithms });
+    const cards = () => this.runAlgorithms({ algorithms });
```
conrad-vanl, line 76: `runAlgorithms` stil appears to be promise based. Why are we scoping this in a function now?

packages/apollos-data-connector-rock/src/features/data-source.js
```diff
@@ -17,6 +17,18 @@ export default class Features extends RockApolloDataSource {
     USER_FEED: this.userFeedAlgorithm.bind(this),
   };
 
+  getFromId(args, id) {
+    const type = id.split(':')[0];
```
conrad-vanl, line 21: I'm confused....the `id` that's passed into this function should generally already be parsed and split?

packages/apollos-data-connector-rock/src/index.js
vinnyjth: If we export it, their app will still boot but they will get errors on certain querues. because we've renamed the dataSource methods internally. Removing the export will cause an error to be thrown if they don't fixing the naming, which seemed safer to me.

packages/apollos-data-connector-rock/src/features/data-source.js
vinnyjth: We want to avoid making calls to rock if we don't want the `cards`. Keeping it as a promise outside a function scope would still make the rock calls (even if the promise was never `awaited`) while wrapping it in a function will ensure we don't make the calls unless we ask for `cards`.

packages/apollos-data-connector-rock/src/features/data-source.js
vinnyjth: Yah that's the `args` argument. I also need the typename, which is only available by splitting the `apollosID`. We could update `Node.get()` to pass it, but 🤷

vinnyjth: FIrst arg is "parsed id" and the second arg is "apollosId"

PR #1423, opened 2020-04-15
Refactors FeaturedFeedConnected to be a feed of feeds using FlatList

burkeshartsis:
## DESCRIPTION
This is the companion PR to https://github.com/ApollosProject/apollos-templates/pull/40. Merge this PR first.

Based on how the features load I'm not entirely sure if this is of huge benefit but we needed to do the essence of this work (migrating the entire Home tab to use features) no matter what and this accomplishes that. At least this way if there is an advantage to using `FlatList` here we are getting it 🙃

I'm still not getting loading states for the various features on the home screen. I think this that is work to be done somewhere else. There is also probably some UX optimization that can be done with regards to how things load or are perceived to load but at this time I'm not sure I fully understand the order in which everything loads given that these features are dynamically driven.

### How do I test this PR?
Open app. All the same features in the same spot? 💥 

There was no coverage over this before and it will continue to be that way for now. I may try to address this during the cool down.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-connected/src/FeaturesFeedConnected/FeaturesFeedConnected.js
```diff
@@ -1,35 +1,50 @@
...
-};
+  loadingStateObject = {
+    isLoading: true,
+  };
 
-FeaturesFeedConnected.defaultProps = {
-  Component: FeaturesFeed,
-};
+  renderFeatures = ({ item }) =>
+    featuresFeedComponentMapper({
```
vinnyjth, line 26: I really like this PR. However, I wonder if we can improve the API of this `mapper` to handle the 99% use case. 

Most churches will be adding additional features, and will want to map a `__typename` straight to a `Component`. Do you think there's a way we could extend the mapper so that you could pass in a simple object describing that mapping: ```
{
  CustomFeedFeature: CustomFeedFeatureComponent
}
```

That way churches could contribute their own features in an additive manner, rather than needing to rewrite existing rules + more.

vinnyjth: If you want to tackle this question in a future PR, that's fine. This seems like a good place for it but there's no reason we can't change this API moving forward

burkeshartsis: So I think the problem is that the data is always a 1<>1 map to the component particularly for cards were there is data on the `relatedNode`. As a result even if I exposed an object like you describe they would likely run into that problem. My solution very basic solution was if you want to add a component pass a new/custom `featuresFeedComponentMapper`.

vinnyjth: fair enough. this won't be a problem soon because each `feature` will be fetching it's own data, hence there won't be any mapping in this method function.

vinnyjth: we can rewrite this in that PR.

burkeshartsis: I'm just not sure we are at the point where simply adding to the config will update the client ui. I think there is more problem statements to identify and work through

vinnyjth: The place where that data mapping will happen will chance once we have each feature load it's own data. There won't be any data to map here, even if we wanted to. All we will have is `__typename` and `id`, so an object lookup table should be fine.

vinnyjth: Well, we might have a few more bits of data than that :) but I think the mapping will still be straightforward.

PR #1425, opened 2020-04-17
Adds Features feed incremental loading

vinnyjth:

This enables the home feed to load in two distinct stages: 

1. Fetch features that will be loaded into the page. 
2. Each of those features fetches all the data it needs to render. 

This enables our loading state to look 100x better, and in general will improve performance and caching of this page. 

In terms of understanding and debugging any loading behavior you are seeing, here's what you need to know. 

1. The ` FeaturesFeed` fires off a query, passing down a loading state that represents a `VerticalCardFeature` with isFeatured set to true, an `ActionListFeature`, a `HorizontalCardFeature`, and another `VerticalCardFeature`

2. When the data for the `FeaturesFeed` query is returned, the features being rendered now correspond to the features coming from the API. They won't have all the data they need yet, so they'll still show a loading state. That loading state may look different, if the api driven features aren't in the same order as the `FeaturesFeed` loading. 

3. As each `Feature` gets it's data, it will stop showing a loading state and start displaying it's real data.

And one more thing, I reduced the ComponentMapper so you can easily add your own features, by passing them in an `additionalFeatures` prop. 

A few interesting things going on. 

I had to override the loading logic of the feed view in a few places, because I didn't want the feed view to generate my loading data ten different times. I wanted a CampaignFeature to always render a single bit of data, and I wanted the `FeaturesFeed` to load 4 distinct bits of loading data. 

the normal UI tests. 

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
I improved things a bit since I recorded this gif, so definitely try for yourself if you think things look funky.

packages/apollos-ui-connected/src/HorizontalCardListFeatureConnected/HorizontalCardListFeatureConnected.js
```diff
@@ -0,0 +1,45 @@
...
+          {...props}
+          isLoading={loading || isLoading}
+        />
+      )}
+    </Query>
+  );
+}
+
+HorizontalCardListFeatureConnected.propTypes = {
+  Component: PropTypes.node,
```
burkeshartsis, line 36: This throws a proptype warning 😢

packages/apollos-ui-connected/src/HorizontalCardListFeatureConnected/HorizontalCardListFeatureConnected.js
```diff
@@ -0,0 +1,45 @@
...
+  Component,
+  isLoading,
+  ...props
+}) {
+  return (
+    <Query query={GET_HORIZONTAL_CARD_LIST_FEATURE} variables={{ featureId }}>
+      {({ data, loading }) => (
+        <Component
+          {...get(data, 'node')}
+          cards={get(data, 'node.cards', []).map(({ actionIcon, ...card }) => ({
```
burkeshartsis, line 20: we should consider moving this function as it will cause `HorizontalCardListFeature` to render because it's create a new object every time. Memoizing this function would solve that.

packages/apollos-ui-connected/src/VerticalCardListFeatureConnected/VerticalCardListFeatureConnected.js
```diff
@@ -0,0 +1,61 @@
...
+}) {
+  return (
+    <Query query={GET_VERTICAL_CARD_LIST_FEATURE} variables={{ featureId }}>
+      {({ data, loading }) => {
+        const featured = get(data, 'node.isFeatured') || isFeatured;
+        const ComponentToRender = featured ? FeaturedComponent : Component;
+        return (
+          <ComponentToRender
+            {...get(data, 'node')}
+            cards={get(data, 'node.cards', []).map(
```
burkeshartsis, line 25: we should consider moving this function as it will cause `VerticalCardListFeature` to render because it's creating a new object every time. Memoizing this function would solve that.

packages/apollos-ui-connected/src/index.js
```diff
@@ -1,4 +1,6 @@
-export ActionListFeature from './ActionListFeature';
+export ActionListFeatureConnected, {
+  ActionListFeature,
```
burkeshartsis, line 2: Is there a specific reason we are choosing not to expose the queries here (for all the components)? If someone wanted to write their own ConnectedComponent wouldn't they need to import the query?

packages/apollos-ui-connected/src/VerticalCardListFeatureConnected/VerticalCardListFeatureConnected.js
```diff
@@ -0,0 +1,57 @@
...
+  FeaturedComponent,
+  isFeatured,
+  isLoading,
+  ...props
+}) {
+  return (
+    <Query query={GET_VERTICAL_CARD_LIST_FEATURE} variables={{ featureId }}>
+      {({ data, loading }) => {
+        const featured = get(data, 'node.isFeatured') || isFeatured;
+        const ComponentToRender = featured ? FeaturedComponent : Component;
```
burkeshartsis, line 21: Want to explain what's going on here 🧙

burkeshartsis: I'm assuming this means we can delete `CampaignItemListFeature`. Is this passing a Component name via an argument from config?

packages/apollos-ui-connected/src/CampaignItemListFeature/index.js
```diff
@@ -111,8 +75,7 @@ const CampaignItemListFeature = memo(
       <FeedView
         onPressItem={onPressItem}
         ListItemComponent={ListItemComponent}
-        loadingStateObject={loadingStateObject}
-        content={cards} // {getContent({ cards, isLoading })}
+        content={isLoading ? [loadingStateData] : cards}
```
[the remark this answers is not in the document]
vinnyjth: yep

packages/apollos-ui-connected/src/FeaturesFeedConnected/featuresFeedComponentMapper.js
```diff
@@ -1,81 +1,48 @@
...
+  VerticalCardListFeature: VerticalCardListFeatureConnected,
+};
+
+const featuresFeedComponentMapper = ({
+  feature,
+  onPressActionItem,
+  additionalFeatures = {},
+}) => {
+  const { id, __typename, ...featureData } = feature;
+  const map = { ...MAPPINGS, ...additionalFeatures };
```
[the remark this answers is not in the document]
vinnyjth: I don't understand what you're suggesting here. Can you share bit of pseudo code?

packages/apollos-ui-connected/src/index.js
vinnyjth: The queries are all pretty simple - it's a `query node()....` wrapping the fragment, so I don't think about it the first time around. Nothing wrong with exporting them

packages/apollos-ui-connected/src/VerticalCardListFeatureConnected/VerticalCardListFeatureConnected.js
vinnyjth: We still need the `CampaignItemListFeature` The `FeaturedComponent` in this case is the `CampaignItemListFeature`. We could potentially eliminate it by passing a different `componentMapper` component to the `FeaturedFeed` here instead, I presume.

packages/apollos-ui-connected/src/VerticalCardListFeatureConnected/VerticalCardListFeatureConnected.tests.js
```diff
@@ -0,0 +1,97 @@
...
+            },
+          },
+        ],
+      },
+    },
+  },
+};
+
+describe('The VerticalCardListFeatureConnected component', () => {
+  it('should render', async () => {
```
burkeshartsis, line 54: This test also doesn't render the image component in the test that your data above would suggest it should.

packages/apollos-ui-connected/src/VerticalCardListFeatureConnected/VerticalCardListFeatureConnected.js
burkeshartsis: I saw later why we still need the component. Sorry about that.

Are you suggesting rendering a second `FeatureFeedConnected Component` with different mapper?

packages/apollos-ui-connected/src/HorizontalCardListFeatureConnected/__snapshots__/HorizontalCardListFeatureConnected.tests.js.snap
```diff
@@ -0,0 +1,1264 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`The HorizontalCardListFeatureConnected component should render 1`] = `
```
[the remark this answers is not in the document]
vinnyjth: I think you need to do a triple render in order to get images to show up. We've seen this same problem in a lot of our content card snapshots.

packages/apollos-ui-connected/src/index.js
burkeshartsis: Well I know @conrad-vanl has advocated for being conscious of the API we expose. Logically the more we expose the more we have to support "publicly." That said, I think the idea has always been to export the "parts" of our connected components so that custom connected components could be built with dependencies on our code so that they still get some of the benefits.

vinnyjth: I agree

packages/apollos-ui-connected/src/VerticalCardListFeatureConnected/VerticalCardListFeatureConnected.js
burkeshartsis: So in practice this is only an issue on the first load. It totally render thrashes which could create scrolling performance during the loading state on a slow connection. Once this has loaded once the data doesn't change and there are no side effects. So while true I'm pretty sure the effect on the UX is only felt during the loading state.

burkeshartsis: TL; DR: You can probably disregard this for now.

packages/apollos-ui-connected/src/HorizontalCardListFeatureConnected/HorizontalCardListFeatureConnected.js
burkeshartsis: TL; DR: You can probably disregard this for now.

So in practice this is only an issue on the first load. It totally render thrashes which could create scrolling performance during the loading state on a slow connection. Once this has loaded once the data doesn't change and there are no side effects. So while true I'm pretty sure the effect on the UX is only felt during the loading state.

vinnyjth: Yah, I don't see anywhere to move it too. Anywhere we move it we still create a new array every time we render.

packages/apollos-ui-connected/src/FeaturesFeedConnected/featuresFeedComponentMapper.js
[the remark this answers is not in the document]
burkeshartsis: Thinking through this further. I see the following scenario, let's say someone wanted to add a custom feature component and remove one of the defaults.

1. We expose a `featuresMap` prop.
2. We inline require the current "default components" based on if they passed a `featuresMap`. Our package code looks something like this:

```
const MAPPINGS = !!featuresMap ? 
  {
    ActionListFeature: require('../ActionListFeatureConnected`).default,
    HorizontalCardListFeature: require('../HorizontalCardListFeatureConnected`).default,
    VerticalCardListFeature: require('../VerticalCardListFeatureConnected`).default,
  }
:
  featuresMap;
```
3. If a user wanted to to add a component and remove one of the defaults they would do something like this and pass p

```
import {
  ActionListFeatureConnected as ActionListFeature,
  VerticalCardListFeatureConnected as VerticalCardListFeature,
} from '@apollosproject/ui-connected`;

const featuresMap = {
  ActionListFeatureConnected,
  VerticalCardListFeature,
  TheirOwnAwesomeCoolComponentThatMatchesTheirAwesomeCoolTypename
};
```
All I'm suggesting is a way for us to structure our code in a way that if they were to do something custom (which we are building for already) that they are not penalized with imports/memory usage they don't need.

packages/apollos-ui-connected/src/HorizontalCardListFeatureConnected/__snapshots__/HorizontalCardListFeatureConnected.tests.js.snap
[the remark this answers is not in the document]
burkeshartsis: 😢 never mind then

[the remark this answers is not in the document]
vinnyjth: I was wrong! Looked back and realized we fixed a lot of the card tests by examining the logs and fixing the tests, and that was what was needed here.

PR #1427, opened 2020-04-17
Refactors ActionListCard to be ActionList component

burkeshartsis:
## DESCRIPTION
This PR is the scaffolding for https://github.com/ApollosProject/apollos-apps/pull/1428 and is a blocker.

This PR forks the `ActionListCard` component and refactors it to be a more flexible `ActionList` component. It cleans up a lot of stuff and I'm thinking we'll deprecate `ActionListCard`. **I'm not 100% confident in the direction this PR went so please consider this an open call for feedback!**

We are starting to introduce a version of this UI element that doesn't render in a card. This PR mainly refactors `ActionListCard` to render as either a card or a straight list. It defaults to `isCard: true` so it is backwards compatible with `ActionListCard`.

This also updates to some new design spacing and other minor style changes.

I went back and forth on the following possible implementation paths. 

1. Separate but otherwise identical components for rendering in a card and not. 
  - This felt too wet. Even if I shared components between the two it just seemed unnecessary.
2. Single component that doesn't render in a card. When you want in in a card you just wrap it in one.
  - I think I like this idea the most. It's the most flexible but it is also the least backwards compatible option, requiring the most changes for client apps.
3. Single component that supports rendering in a card or not.
  - This is the usual route most people go and the one I generally avoid. Adding additional API to a component is easy but also complicates maintenance and support. It also opens the door to adding more "features/support" down the road, exacerbating the problem further. That said, the actual implementation felt easy and, while I don't have a strong opinion on this yet, if we are going to continue to add otherwise identical ui elements that don't render in a card this might be the best way to support that.

So I ended up going with option 3. **I'm not sold on it or the `isCard` prop and would happily entertain other ideas**. We could also delete `ActionListCard` now and export `ActionList` as both names since it's backwards compatible.

### How do I test this PR?
Check out the stories for the `ActionList` component.

Everything should be covered 💥 

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-kit/src/ActionList/index.js
```diff
@@ -0,0 +1,103 @@
...
+  static defaultProps = {
+    isCard: true,
+    onPressActionListButton: () => {},
+  };
+
+  RenderAsCard = ({ children }) =>
+    this.props.isCard ? (
+      <Card isLoading={this.props.isLoading}>{children}</Card>
+    ) : (
+      <>{children}</>
```
vinnyjth, line 56: Why is this fragment needed?

packages/apollos-ui-kit/src/ActionList/index.js
```diff
@@ -0,0 +1,103 @@
...
+
+  render() {
+    const {
+      onPressActionItem,
+      onPressActionListButton,
+      actions,
+      header,
+    } = this.props;
+
+    const RenderAsCard = this.RenderAsCard; // eslint-disable-line prefer-destructuring
```
vinnyjth, line 67: Any reason you're not destructuring here?

packages/apollos-ui-kit/src/ActionList/index.js
```diff
@@ -0,0 +1,103 @@
...
+        <Content cardPadding={this.props.isCard}>
+          {actions.map((item) => (
+            <ActionListItem
+              action={item.action || ''}
+              key={item.id}
+              label={item.subtitle || ''}
+              onPress={() =>
+                onPressActionItem({
+                  action: item.action,
+                  relatedNode: item.relatedNode,
```
vinnyjth, line 81: Looking good! I might consider spreading `...props` in here. One thing I've discovered is it can be hard to grant this function access to something like the `UserWebView` without.

packages/apollos-ui-kit/src/ActionList/ActionListItem/index.js
```diff
@@ -0,0 +1,69 @@
...
+
+const CellImage = styled(({ theme }) => ({
+  width: theme.sizing.baseUnit * 4,
+  height: theme.sizing.baseUnit * 4,
+  borderRadius: theme.sizing.baseBorderRadius,
+  marginRight: theme.sizing.baseUnit,
+}))(ConnectedImage);
+
+// eslint-disable-next-line react/prop-types
+const RenderAsTouchable = ({ children, onPress }) =>
```
vinnyjth, line 39: I like the style of this Component, but is it necessary? How does `Touchable` handle not having an `onPress` prop? I think it acts just like a normal view, but I could be mistaken.

packages/apollos-ui-kit/src/ActionList/index.js
burkeshartsis: This is a fragment so that if a `header` is passed it will render inside something. I didn't actually test if it was necessary but in head the render order would suggested it is.

packages/apollos-ui-kit/src/ActionList/ActionListItem/index.js
burkeshartsis: The problem was that `TouchableScale` still animates even if there is not `onPress` function. I agree though that the fix should live in `TouchableScale`! I'm going to open a PR to fix this there! I'll also add a `TODO` here to remove this fix in that PR.

packages/apollos-ui-kit/src/ActionList/index.js
burkeshartsis: You know... I was sort of blanking on the syntax for that for whatever reason since it was a class member. Of course the syntax is the same no matter what 🙃 Fixed in 2a3154b

burkeshartsis: Actually I reverted that last commit because it doesn't work if you use destructuring 🤷‍♂️

packages/apollos-ui-kit/src/ActionList/index.js
burkeshartsis: Tested this and you were correct. Removed the fragment in ed96ff8

packages/apollos-ui-kit/src/ActionList/index.js
burkeshartsis: when you say `props` do you mean the component props or the `item` from the `.map()`?

packages/apollos-ui-kit/src/ActionList/index.js
burkeshartsis: Ok actually working here 0816caa

burkeshartsis:
@vinnyjth this is ready for review again. I think I addressed all of your comments.

Do you have any input my the "What design trade-offs/decisions were made" section before you approve this?

packages/apollos-ui-kit/src/ActionList/ActionListItem/index.js
burkeshartsis: I got a working solution in this component in 2f1745b. I'll look to open a long term fix in another PR as noted above.

PR #1432, opened 2020-04-23
Adds ability for HTMLView to handle non-http links

vinnyjth:

Previously we were locked strictly into HTTP links when parsing with the HTMLView. This PR does two things 

- Still renders a link for non HTTP Links
- Updates the default HTMLViewConnected to open non-http links with Linking.openURL, if the protocol is supported. 

You might see links in content that don't work, where previously there would be plain text.They shouldn't crash the app or throw an error when tapped. 

### How do I test this PR?

I added stories to both HTMLView and the new safeUrl function so you can test the functionality of the handlers on the sim/device. 

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-connected/src/safeOpenUrl/index.js
```diff
@@ -0,0 +1,20 @@
+import { Linking } from 'react-native';
+import InAppBrowser from 'react-native-inappbrowser-reborn';
+
+async function safeHandleUrl(url, { external = false } = {}) {
+  try {
+    if (url.startsWith('http') && !external) {
```
conrad-vanl, line 6: What about the case of the URL being a universal link - in other words, a link that is a valid web link `http` but that link is supposed to open the app? Thinking about Liz's use case. I also don't see `external` referenced anywhere else in this PR?

vinnyjth: @conrad-vanl Yah - I think the next step is to take advantage of that `external` arg. I noticed Rock has a setting when creating `a`'s in the WYSIWYG editor for "open externally". I'm assuming that adds an "target="_blank"" attribute, which we can test for to drive that `external` arg.

I'm fine shipping as is, because for now clients that don't also use rock to drive their website can use deep links in their content, and those links will now work correctly.

conrad-vanl: We just aren't documenting this prop anywhere, nor is it included in test coverage. Feels weird to include if we aren't using it

vinnyjth: But if you want to support the `external` functionality in this PR, I think we can make it happen. I need to dig into exactly what that Rock "open externmally" checkbox does and make sure we can read it..

conrad-vanl:
I'd like to see test coverage for `external` added or it removed from the code base....but I'll approve for now and let you work that out.

PR #1437, opened 2020-04-27
Adds Pull to refetch on Features feed

vinnyjth:

This PR has made a few notable changes to add pull to refetch. 

1. Enables the `FeedView` to accept loading state data directly, rather than having to ignore the `loading` prop. 

2. Adds a `refetchRef` method the FeaturesFeed that collects all the `refetch` functions from each of the children, making sure not to collect any single ref more than once. 

I'm not in love with the `refetchRef` concept. However, I don't see any better options besides building a `Provider` to handle this, which seems overkill. At least `refetchRef` is based on an established react pattern. Open to suggestions. 

### How do I test this PR?

1. Pull down this PR, link repos, boot the app.
2. Pull down from the home feed, the feed should refetch and you can can confirm in the API console that new data is indeed loading. 

New tests where appropriate. 

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+
+v1.4.0-beta.2: April 27, 2020
+
+- Adds CalendarActionImages from Willow and for Willow (#1434) (e5538c45)
+- Adds ability for HTMLView to handle non-http links (#1432) (8e5478fa)
+- Adds Features feed incremental loading (#1425) (8db9fbbf)
+
+- Refactors ActionListCard to be ActionList component (#1427) (255d7778)
+- Pulls new content items when opening content single (#1431) (d4bd596b)
+- Updates token.js for longer auth expiration (#1429) (dc5501d9)
+
+- Fixes barStyle to be correct for light and dark theme types (#1433) (e1dc6df1)
+- Fixes Html content doubling issue (#1430) (67d59a91)
+
 v1.4.0-beta.1: April 17, 2020
 
 - Refactors FeaturedFeedConnected to be a feed of feeds using FlatList (#1423) (5143cd80)
--- a/lerna.json
+++ b/lerna.json
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "1.4.0-beta.1",
+  "version": "1.4.0-beta.2",
   "npmClient": "yarn",
   "useWorkspaces": true
 }
--- a/packages/apollos-apollo-server-env-mock/package.json
+++ b/packages/apollos-apollo-server-env-mock/package.json
@@ -1,13 +1,13 @@
 {
   "name": "@apollosproject/apollo-server-env-mock",
-  "version": "1.4.0-beta.1",
+  "version": "1.4.0-beta.2",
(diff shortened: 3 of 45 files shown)


conrad-vanl:
Q: the passing of the `refetchRef` is so that you can refetch all of the child queries, right? Would it not be enough to just refetch the parent-most query?

vinnyjth:
Q: the passing of the refetchRef is so that you can refetch all of the child queries, right? Would it not be enough to just refetch the parent-most query?

It won't because none of the props of the child features will change. I tried that route first, you can test  by changing this line https://github.com/ApollosProject/apollos-apps/pull/1437/files#diff-e47c1b17d119b8039f445d030628658aR69 to `refetch={refetch}`

burkeshartsis:
@vinnyjth if you are looking to force a change on the children try passing something (anything?) into [`extraData`](https://reactnative.dev/docs/flatlist#extradata)

vinnyjth:
@burkeshartsis I don't think that will work, because Apollo will only refetch if the a piece of data used in the query variables changes. I can try though.

conrad-vanl:
@vinnyjth so it's not enough to set the `fetchPolicy` to `cache-and-network` and then to `refetch` the parent component?

Would a `refetchQueries` type of call that is outside of the component props landscape be cleaner? Probably not just a thought.

vinnyjth:
> @vinnyjth so it's not enough to set the fetchPolicy to cache-and-network and then to refetch the parent component?

Correct. That was my first stop, but because the `featureId`s of the children never change, the parent query would be refetched but the children would remain untouched. 

> Would a refetchQueries type of call that is outside of the component props landscape be cleaner? Probably not just a thought.

I thought about that! It's probably possible, but it would be harder to extend, if you added a custom component you would have to somehow get your query into that `refetchQueries` array. At least with this route you can opt to inject your `refetch` by calling that `refetchRef` function, or not if you don't care....

conrad-vanl:
Oh hey... @vinnyjth I thought of an error on this. What happens if a feature is removed? Like it refreshes and the feature is no longer in the feed? It will still exist in the `this.refetchFunctions` object and throw an error when you pull to refresh again, since that component no longer exists?

vinnyjth:
>  What happens if a feature is removed? Like it refreshes and the feature is no longer in the feed? 

This shouldn't shouldn't happen under normal circumstances, but I can totally see it as a possibility in the feature. Right now an API deployment with a new config.yml is the only way this condition could arrise. 

> It will still exist in the this.refetchFunctions object and throw an error when you pull to refresh again, since that component no longer exists?

I'm honestly not sure what will happen, but I think it's worth trying to find a way to reproduce and at least know the consequences for.  I have some guesses, either ....

An error will be thrown and silently ignored because those refetch promises aren't awaited
An error will be thrown and it will break the feed, only fixed by closing and opening the app (scary!)
The refetch will point to the new component, with a new data to be pulled. Refetch will be called twice with what imagine will be minimal consequences. 

Good callout, I'll test now before we merge.

PR #1438, opened 2020-04-27
Fixes TouchableScale animating when there is no "action/handler"

burkeshartsis:
## DESCRIPTION
`TouchableScale` is smart enough not to do anything if there is no `onPress` function however, it still animates suggesting it should do something when it can't. This fixes that.

it checks if a `onPress` prop exists and conditionally passes `onPressIn` and `onPressOut` only if `onPress` exists.

This requires some gymnastics still. Now when using `TouchableScale` if you want to ensure it doesn't animate if it can't you have to conditionally pass `onPress` like this.

```
<TouchableScale
  {...(handleOnPress ? { onPress: handleOnPress } : {})}
/>
```

I think there might be a better way by possibly checking if `onPress` is a function instead.

### How do I test this PR?
Unfortunately there is no place to show you this working.

No tests before... no tests after 🤠 

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-kit/src/TouchableScale/index.js
```diff
@@ -43,12 +44,23 @@ class TouchableScale extends Component {
...
   render() {
-    const { minScale, style, children, ...touchableProps } = this.props;
+    const {
+      minScale,
+      style,
+      children,
+      onPress,
+      ...touchableProps
+    } = this.props;
     return (
```
vinnyjth, line 54: Nitpick, would this look better outside the component props? 

```
const animationHandlers = onPress ? {. onPress, onPressIn: this.handlePressIn, onPressOut: this.handlePressOut } : {}
... 

< 
   {...animationHandlers}
>
```

burkeshartsis: I like it! Fixed in 879791b

burkeshartsis: I like it! Fixed in 879791b

PR #1439, opened 2020-04-28
Fix padding when no header on ActionListCard

conrad-vanl:

When there's no `header` element provided to `ActionListCard`, there is currently an empty `<CardContent/>` that is rendered, which produces undesirable padding at the top of the card. This only renders `<CardContent />` if there's a header provided.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-kit/src/ActionListCard/index.js
+++ b/packages/apollos-ui-kit/src/ActionListCard/index.js
@@ -51,7 +51,7 @@ class ActionListCard extends PureComponent {
 
     return (
       <Card isLoading={isLoading}>
-       {headerContent ? (<CardContent>{headerContent}</CardContent>) : null}
+       {headerContent ? <CardContent>{headerContent}</CardContent> : null}
         <Content>
           {actions.map((item) => (
             <ActionListItem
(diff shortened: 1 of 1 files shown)


packages/apollos-ui-kit/src/ActionListCard/index.js
```diff
@@ -51,7 +51,7 @@ class ActionListCard extends PureComponent {
 
     return (
       <Card isLoading={isLoading}>
-        <CardContent>{headerContent}</CardContent>
+       {headerContent ? (<CardContent>{headerContent}</CardContent>) : null}
```
conrad-vanl, line 54:
```suggestion
       {headerContent ? <CardContent>{headerContent}</CardContent> : null}
```

burkeshartsis:
This is actually fixed in the new/refactored `ActionList` component.

conrad-vanl:
@burkeshartsis I'm confused, this PR is against master. Is there another PR?

burkeshartsis:
There was another PR that created `ActionList` and was already merged. It also purposed possible deprecation strategies for `ActionListCard` but then other PR's like yours were opened against `ActionListCard` to fix or add functionality. I'm only mentioned this because, I'd like to see us pick a depreciation strategy and move towards that rather than continuing spend time on a component we are actively trying to move away from. 

All that said, the `header` prop isn't widely used so in my refactor I actually removed `CardContent` from the header all together so that headers could pad themselves. There were other reasons for that change too but I might advocate for removing the offending padding all together, rather than making it conditional, and patching the 1 or 2 places that used that padding.

PR #1442, opened 2020-04-30
Webview feature

jordanwade:
## DESCRIPTION
Set up a webview feature that will allow the embedding of Spotify playlists or other url embeds.

Simulator Screen Shot - iPhone 11 Pro Max - 2020-05-05 at 15 33 02

### How do I test this PR?

Check out the story `ui-connected/WebviewFeature example`

Added webview feature tests to `ContentSingleFeaturesConnected`

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
As much as I'd love to review this, this is totally within @vinnyjth's wheelhouse and would like to kick this over to him. I'm more the give you balloons and cake person with an `A+` sticker as from my knowledge of features, this looks overall great!

packages/apollos-ui-connected/src/ContentSingleFeaturesConnected/WebviewFeature.js
```diff
@@ -0,0 +1,71 @@
...
+import PropTypes from 'prop-types';
+import { SideBySideView, H3, UIText, styled } from '@apollosproject/ui-kit';
+import { WebView } from 'react-native-webview';
+import { Platform, View, TouchableOpacity, Linking } from 'react-native';
+// import { SafeAreaView } from 'react-navigation';
+
+const StyledCard = styled(
+  ({ theme }) => ({
+    backgroundColor: theme.colors.black,
+    height: 400,
```
vinnyjth, line 11: Curious about this height. I imagine it's really hard to decide on what this value should be, and 400 is totally right for the Spotify View. I think this is fine for now, because we have an override for this component (good catch on that, btw) but in the future maybe we should expose height as a prop?

conrad-vanl: We could easily expose `height` as a prop now and just set the `defaultProp` for it to 400.

jordanwade: Cool, I didn't like putting a fixed height. I'll update to just expose the height and default it to 400. Good idea!

vinnyjth: Agh! I merged this too soon. Feel free to do this in a follow up PR

jordanwade: Actually, since the height is set in the styles and the styles for the webview feature card are exposed with `ui-connected.WebviewFeature.StyledCard` should we still expose the height in the props too?

image

conrad-vanl: overriding in styles would allow a church to override it globally, instead of on a case-per-case basis. Technically a church could still use a style override in a case-per-case basis since style overrides can be functions, and when they are a function, they receive the same props of the component. Generally we'd opt for both in case like this. 

I think for now we can just leave it as is and add the prop when someone requests it.

jordanwade: Again good point, thank ya! Adding it 👍

PR #1443, opened 2020-04-30
Exposes Component prop on ContentCardComponentMapper

burkeshartsis:
## DESCRIPTION
👇 Rendering normal as it should / 👇Rendering with a Component override via theme 

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-connected/src/ListItemCard/ListItemCard.js
```diff
@@ -0,0 +1,33 @@
+import React, { memo } from 'react';
+import { get } from 'lodash';
+import PropTypes from 'prop-types';
+
+import { DefaultCard, HighlightCard } from '@apollosproject/ui-kit';
+
+import listItemCardMapper from './listItemCardMapper';
+
+const ListItemCard = memo(({ Component, ...props }) => {
+  if (Component) return <Component {...props} />;
```
conrad-vanl, line 10: Since `Component` is defaulted to `listItemCardMapper` already, can we just simplify this component to just always return `<Component {...props} />` ?

Also still need to throw a wrapping `withTheme` in so we can override it 😅

burkeshartsis: I was actually planning on removing the `defaultProp` since the mapper is one line below this and then deleting that mapper as its own file/component. But now my thinking is to go back to the way we have been doing this (as you point out) because then if you pass a `Component` the mapper will never even mount/import it's components. Fixed in ece9429

conrad-vanl: Cool 😃

Since this is just a component that renders the component prop... I would say we could remove `memo` 😬 I think there's potential it could produce unintended consequences or actually make performance worse since it's going to do a prop check every re-render? With the way our components are going most likely whatever component is passed to the `Component` prop is going to be `memo`'d or `PureComponent`'d ??

PR #1456, opened 2020-05-07
Adds Rookout Debug Information

redreceipt:

Rookout needs sourcemaps to convert src to lib. And also needs a config file to tell rookout that this repo will be used as an npm dependency.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
I'm fine with adding the `sourcemaps`. @vinnyjth should we just do that on all packages?

As far as adding the `rookout` specific package. I'm not against doing that for this case, but fundamentally I don't want to add a bunch of service-specific configuration to each pacakge. Is there not a way you could do this on your end? Maybe a `postinstall` script that copied that file into the `node_modules` file directory?

redreceipt:
> Maybe a `postinstall` script that copied that file into the `node_modules` file directory?

I understand. I could do the `postinstall`. May not even be necessary. They said, "it would help" so I'll try without first.

conrad-vanl:
I'll approve this, but assigning it to @vincent as it'll need an NPM release for it to be helpful for @redreceipt

redreceipt:
I can use canary. 

the `postinstall` won't work but there's another way, working on that now

vinnyjth:
If you're good with a `canary` then I'm going to merge this :) `postinstall` should be what you need? If you run into trouble with your solution send me the deets and we can figure something out.

PR #1457, opened 2020-05-07
Adds PrayerCard and PrayerInput component

burkeshartsis:
## DESCRIPTION
Currently tests don't pass and I'm not sure how to fix it without removing hooks. Which I could totally do if that was my only option.

### What does this PR do, or why is it needed?
This PR adds `PrayerCard` and `PrayerInput` components. Together these make up the card UI for prayer input and viewing.

This PR adds some extra code for viewing these cards to the `PrayerSwiper` stories as well just to get things on the screen. This story code will become the bases for the `PrayerScreen` component in #1466.

### What design trade-offs/decisions were made?
I intentionally made these components more flexible than I really needed to. It didn't create any more work and it will allow these components to power a wider "prayer" experience including anonymous prayer and personalized prayer prompts.

### How do I test this PR?
Check out the stories! You'll need to run this PR on [this branch](https://github.com/ApollosProject/apollos-templates/pull/51).

### What automated tests did you write?
Should have pretty good coverage.

## TODO:

- [x] PR has a relevant title that will be understandable in a public changelog (ie...non developers)
- [ ] Closes [tag-issues-here]
- [x] No new warnings in tests, in storybook, and in-app
- [x] Upload GIF(s) of iOS and Android
- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**

- [ ] Manual QA on iOS and ensure it looks/behaves as expected
- [ ] Manual QA on Android and ensure it looks/behaves as expected

**Code Review: Questions to consider**

- [ ] Read through the "Files changed" tab _very carefully_
- [ ] Edge cases: what assumptions are made about input?
- [ ] What kind of tests could be written?
- [ ] How might we make this easier for someone else to understand?
- [ ] Could the code be simpler?
- [ ] Will the code be easy to modify in the future?
- [ ] What's one part of these changes that makes you excited to merge it?

> The purpose of PR Review is to _improve the quality of the software._

PR #1464, opened 2020-05-21
Adds PrayerFeature ui

burkeshartsis:
This PR primarily adds a "PrayerFeature" ui that might be used on a home feed or within a piece of content.

There are a few things that come to mind around the avatar size. This is faithful to the designs provided but breaks from our established theme patterns. While this probably meets the needs of the MVP I think components like `AvatarList` could be further abstracted and moved to live within the `ui-kit`.

There are also some potential edge cases like how we might handle a user without a profile image vs. an anonymous prayer. That said, neither of thus are MVP features so it's kind of outside the scope of work.

### How do I test this PR?
Check out the stories.

Should be basically 100% for the `PrayerFeature` component. `AvatarList` is missing test coverage as I am waiting to see if we are going to move it to `ui-kit`

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


burkeshartsis:
@conrad-vanl I want to get your feedback one this where it is right now. I think it's done form an MVP perspective but there is lots of room to improve. Maybe just give it a look and see if you like this direction. I expect the way data flows to change a bit when we start to build out the connected components.

packages/apollos-ui-prayer/src/PrayerFeature/AvatarList/index.js
```diff
@@ -0,0 +1,97 @@
...
+  />
+);
+
+AvatarList.propTypes = {
+  avatars: PropTypes.arrayOf(ImageSourceType).isRequired,
+  onPressAdd: PropTypes.func,
+  onPressAvatar: PropTypes.func,
+};
+
+export default AvatarList;
```
conrad-vanl, line 97: Is there a reason this is defined here instead of in UIKit? I could see this being used for more than Prayer and it only depends on UIKit components. Plus it's named generally (AvatarList vs PrayerAvatarList or something like that). Curious on intentions

packages/apollos-ui-prayer/src/PrayerFeature/AvatarList/index.js
```diff
@@ -0,0 +1,97 @@
...
+    showsHorizontalScrollIndicator: false,
+  }),
+  'ui-prayer.PrayerFeature.AvatarList.AvatarFeed'
+)(FlatList);
+
+const keyExtractor = (item) => item && item.id;
+
+// eslint-disable-next-line react/display-name, react/prop-types
+const renderItem = (onPressAvatar) => ({ item }) => (
+  <Touchable onPress={() => onPressAvatar()}>
```
conrad-vanl, line 63: If you don't pass any arguments to `onPressAvatar`, how will the feature know which avatar was pressed?

conrad-vanl: I think we might consider atleast `onPressAvatar({ item })` ... but I'd suggest looking at our `FeedView` to see what arguments we pass to `onPressItem`, since it's basically the same thing (and keeping API paradigms consistent between components will be a good thing)

packages/apollos-ui-prayer/src/PrayerFeature/AvatarList/index.js
burkeshartsis: I actually ended up opening a PR for exactly. See #1471

packages/apollos-ui-prayer/src/PrayerFeature/AvatarList/index.js
burkeshartsis: 100% agree! What your suggesting is our normal practice

burkeshartsis: fixed in 05f2d71

burkeshartsis:
@conrad-vanl I think this PR is ready for approval now.

conrad-vanl:
Not sure if this is blocked for #1457 or not, but not going to add merge tag unless @burkeshartsis says we can.

PR #1472, opened 2020-06-02
Adds an AvatarCloud component to ui-kit

burkeshartsis:
## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-kit/src/Avatar/Avatar.js
```diff
@@ -0,0 +1,123 @@
...
+}))(ButtonIcon);
+
+const ButtonIconPositioner = styled({
+  position: 'absolute',
+  bottom: 0,
+  right: 0,
+})(View);
+
+const LoadingSpinnerContainer = styled(({ theme }) => ({
+  backgroundColor: 'white',
```
conrad-vanl, line 58: `'white' `=> code smell, yeah?

packages/apollos-ui-kit/src/Avatar/Avatar.js
```diff
@@ -0,0 +1,123 @@
...
+  // The following three measurements are used to match those of the ButtonIcon container
+  width: 43,
+  height: 43,
```
conrad-vanl, line 61: woah, why all these magic numbers?

packages/apollos-ui-kit/src/Avatar/Avatar.js
```diff
@@ -0,0 +1,123 @@
...
+  padding: 9.6,
```
conrad-vanl, line 62: 😳

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
```diff
@@ -0,0 +1,121 @@
...
+class AvatarCloud extends PureComponent {
+  static propTypes = {
+    avatars: PropTypes.arrayOf(ImageSourceType).isRequired,
+    maxAvatarSize: PropTypes.number, // a percentage represented as a whole number
+    minAvatarSize: PropTypes.number, // a percentage represented as a whole number
+    primaryAvatar: ImageSourceType, // The source to render a larger avatar at the center of the cloud
+  };
+
+  static defaultProps = {
+    maxAvatarSize: 50,
```
conrad-vanl, line 44: So when we express percentages as whole numbers, usually the scale is `0` to `1`, yeah? `0.5` = `50%` and so on. Also, I wonder if these prop names can express percentage in the prop name, so as to be more self documenting? `maxAvatarScale` might be nice

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
```diff
@@ -0,0 +1,121 @@
...
+    minAvatarSize: 10,
+  };
+
+  getOrientation = () => {
+    if (Dimensions.get('window').width > Dimensions.get('window').height) {
```
conrad-vanl, line 49: this is assuming that the component fills the screen, yeah?

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
```diff
@@ -0,0 +1,121 @@
...
+import ConnectedImage, { ImageSourceType } from '../../ConnectedImage';
+import styled from '../../styled';
+
+const CenteredAvatar = styled(
+  ({ orientation, size }) => ({
+    aspectRatio: 1,
+    borderRadius: 1000, // For simplicity we are just going to use a very large magic number 🙃🧙
+    zIndex: 100,
+    ...(orientation === 'landscape' // Avatar size is based on the smallest device dimension
+      ? { height: `${size}%` }
```
conrad-vanl, line 15: hm.....I think you can simplify this!

Since `aspectRatio: 1`, it doesn't matter if you use `height: $size` or `width: $size`, as `width` will always equal `height`! So...I think you can ditch all of the `orientation` code?

packages/apollos-ui-kit/src/Avatar/Avatar.js
burkeshartsis: so this is actually not something I wrote. This is our `Avatar` component just moving directories.

packages/apollos-ui-kit/src/Avatar/Avatar.js
burkeshartsis: so this is actually not something I wrote. This is our `Avatar` component just moving directories.

packages/apollos-ui-kit/src/Avatar/Avatar.js
burkeshartsis: I would purpose we fix this in a different unrelated PR

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
burkeshartsis: so really my whole number choice here was out of laziness given that all the styles need it represented this way. Meaning I would have to do the math every time or write a function for it. I'm happy to do that if you think it's better.

burkeshartsis: As for naming, I like scale! I think it potentially makes the component api more clear but I'm not sure it does anything for the rest of the code. I use the word "size" a lot 😬

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
burkeshartsis: so we only use the window width to determine the orientation of the device which is fine. We don't use any of those values to determine any elements size as they are all percentages.

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
burkeshartsis: I was hoping the comment in the code would help with understanding here. Basically since we are using percentages and say your `maxAvatarSize` is 50%.  That's fine when the device is in portrait mode. There is lots of white space because it's 50% of the width! But if you rotate the device and do nothing to account for that then all of a sudden that 50% width is basically 100% of your viewport height. This code says, use the smallest phone dimension to determine which what the avatars percentage is based on so that it doesn't completely blow up the UI.

As another example, let's say you render `AvatarCloud` in a card in portrait mode. This code will render fine. Then you rotate the phone to landscape. Normally your card would get get very wide and while not likely fitting on the screen it's reasonably scrollable. If we didn't account for rotation the avatar would be huge and would make the card huge with a lot of scrolling. This code the way it is felt like it resulted in the preferred behavior of adapting and making the avatars smaller so that they were reasonably visible on the screen.

conrad-vanl: Got it. But still, that's assuming the orientation of the component matches the same orientation of the screen. In other words, this wouldn't work well for these cards I imagine -

Screen Shot 2020-06-04 at 8 20 29 AM

The container for these cards is essentially landscape - wide but not tall. But since the phone is in portrait, it would be using the wrong size selector.

Not a big deal for now just calling it out.

conrad-vanl:
Not sure if blocked by #1457 so not automerging

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
burkeshartsis: dang it... you're right. The only way for this logic to work correctly then is to use `onLayout`.  I suppose we could just remove this logic like you suggested given that the "container" is sizable via `style` and you can adjust the `max/minAvatarSize` to account as needed.

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
burkeshartsis: @conrad-vanl Did you feel strongly about this one way or the other?

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
conrad-vanl: Oh I like the idea of removing the logic then. Less gotchas and you'd just have to adjust `max/minAvatarSize` to account as needed. You could even change the prop to be like `minAvatarWidth` or something

conrad-vanl: to be more clear that it's `50%` of the width of the container instead of the height or something

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
conrad-vanl: I would prefer `0` to `1` on the percentage. and I would prefer renaming this prop to be more self documenting.

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
```diff
@@ -0,0 +1,121 @@
...
+    const positionBoundry = 100 - avatarSize;
+
+    return {
+      left: `${Math.round(Math.random() * positionBoundry)}%`,
+      top: `${Math.round(Math.random() * positionBoundry)}%`,
+    };
+  };
+
+  renderRandomAvatars() {
+    return this.getRandomAvatarSizes().map((size, i) => (
```
vinnyjth, line 86: Given this method calls a method that uses `Math.random()`, would it be better to avoid using `Math.random()`. If I understand correctly, multiple re-renders would cause the bubbles to move around each time the component re-rendered. Would it be better to to init a few random seeds using `Math.random()` in the `constructor` and then reference those seeds whenever you need a random value?

conrad-vanl: one of the things I thought about was doing like a memoized getter that just generated a random value the first time a given url was passed

conrad-vanl: A seed would be nice but can't think of an easy way to do a seed on construction?

vinnyjth: `Math.sin` might be too deterministic 🤔 but there's a few other patterns out there of varying complexity we could try out. 

Memoizing using a getter is also a great idea

burkeshartsis: Guys, I love that we are engaging on this! So I basically came to we can either create our own memoized solution using something like the url as the seed. The other options could be to add this [seedRandom](https://github.com/davidbau/seedrandom) package with the url as the seed. I like your idea @vinnyjth of generating our own seed with `Math.random()` in the constructor. That would fix the rerender problem since it's only fired once.

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
burkeshartsis: this is fixed

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
burkeshartsis: fixed

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
burkeshartsis: @vinnyjth @conrad-vanl what do you guys think of this approach... 6daf6de

burkeshartsis:
@conrad-vanl originally I was thinking this probably wouldn't need additional review after your first pass but I feel like a lot has changed since then. If you don't mind giving this another once over that would be great.

burkeshartsis:
Updated screenshot with the blur effect.
<ing src="https://user-images.githubusercontent.com/2053547/83934850-94f62300-a781-11ea-8a86-88a005cfc8ea.png" width="30%" />

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
```diff
@@ -0,0 +1,164 @@
...
+  static defaultProps = {
+    blur: true,
+    maxAvatarWidth: 0.5,
+    minAvatarWidth: 0.1,
+  };
+
+  constructor(props) {
+    super();
+
+    this.randomSeeds = props.avatars.map(() => Math.random());
```
conrad-vanl, line 69: So this is a good idea...but if the number of avatars change we won't get more avatar seeds and it'll throw an error lower. For example you mount with one avatar then re-render with 2 or more.

burkeshartsis: Are you suggesting I need to write a `componentDidUpdate` call? I'm not really sure why the number of avatars would change but... I could see it being expected purely from a that's how components work" sort of way. I could also write it so that when it updated it only appended new values instead of regenerate them all. This would mean when the avatars wouldn't move when props updated.

conrad-vanl: `componentDidUpdate` is too late - that happens after a render. `componentWillUpdate` is what you want but it's deprecated.

I think what you're saying is a fallacy - we can write validation to ensure the length of `avatars` don't change, but right now if it did the component would throw an error. I would rather support the length of the avatars array changing.

Newest react documentation shows we should be using some sort of memoization function instead of `componentWillUpdate`.

Loading state is a valid example - say you tried to mount an Avatar cloud with 3 avatars as a loading state, and then once the data loaded there was 4 users in actuality to show. Then you'd trigger an unhandled error that would likely result in a RSOD.,

conrad-vanl:
```
const seeds = {};

const getMemoizedSeedForUrl = (url) => {
   if (!seeds[url]) seeds[url] = Math.random();
   return seeds[url];
}
```

might be simple enough...Could still put this within the component and the seeds would change every time the component is mounted, or could put at root and seeds would persist for the length of the app being opened.

conrad-vanl: Example of doing it in the component, also you could do it off of the position in the array instead of the url, if you wanted:

```
class MyComponent {
   seeds = {}

   getSeedForAvatarPosition(i) {
      if (!seeds[i]) seeds[i] = Math.random();
      return seeds[i];
   }
}
```

conrad-vanl: the second example might be better for the loading state example - your loading orbs would stay in the same place once they're replaced with actual data since the momoization key is the array index instead of the url

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
```diff
@@ -1,4 +1,4 @@
-import React, { PureComponent } from 'react';
+this.getRandomPositionValue(i)import React, { PureComponent } from 'react';
```
conrad-vanl, line 1:
```suggestion
import React, { PureComponent } from 'react';
```

conrad-vanl:
image

One little comment on this - there's not a way to actually blur the layer is there? The edges of the circle are still sharp - it's just blurring the photo. The effect isn't the same.

PR #1485, opened 2020-06-15
Adds Closed Captioning

redreceipt:
## DESCRIPTION

Screen Shot 2020-06-16 at 8 34 05 AM

Adds ability to turn on subtitle tracks if they are included in the stream

Replaced the mute button

### How do I test this PR?

Go to campaign content and turn on subtitles

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
> Couple quick things:
> 
> * Not sure we should completely remove the mute logic - just that if CC is available, the mute icon disappears and is replaced by CC
> * Does this work if no CC is provided?
> 
> I like how you integrated this...feels like it'll be much easier to approve 😃 Main concern is making sure this continues to be backwards compatible for apps that don't use CC

Yea couple of thoughts there. I considered having mute available when there's no CC but it feels to me like a secret feature that you get by accident. I personally liked the mute button so it would be frustrating to me knowing that the feature is still there, I just can't get to it most of the time.

As far as intelligently determining CC, that one is tricky and I figured to opt out for MVP. It's tricky because my implementation simply chooses the first subtitle track available embedded in the HLS. which gets processed by `react-native-video` but not by us

conrad-vanl:
@redreceipt this PR is adding a requirement for churches to add CC. I can't make that a requirement. This PR is non-mergeable if as soon as upgrading to this version of Apollos your player is broken if you don't have CC available.

redreceipt:
I definitely agree. Another UI option is to put it on the row above the seeker with the cast buttons. In a similar fashion, it could be disabled with a prop?

conrad-vanl:
image

I don't mind the way iOS lays it out... we have more buttons though.

I also wonder if we need the "toggle video" control either. 

What would be trick would be instead of separate "cast" and "airplay" buttons, to have a single button that opens a dropdown/overlay control that allows you to select between airplay and casting, defaulting the option if we only detect you have one or the other. That would free up real-estate. I agree that the mute/volume controls offer more value then the video show/hide control. 

So maybe we kill the video show/hide control and put closed captioning there, but we still need to have it only show up if closed captioning is available. Do any of the [event props](https://github.com/react-native-community/react-native-video#event-props) from react-native-video have information about the presence of text track? Maybe we can initially hide the CC button by default always, then onLoad see if there's a text track, if so, show the CC toggle button.

So, V1: we can replace the "show/hide" video button instead of the mute button and just get rid of that feature all together, like you did with mute.

vinnyjth:
> So, V1: we can replace the "show/hide" video button instead of the mute button and just get rid of that feature all together, like you did with mute. 

I agree with this. Bonus points for doing this conditionally based on if CC is available.

packages/apollos-ui-media-player/src/MediaPlayer/VideoWindow.js
```diff
@@ -128,6 +128,7 @@ class VideoWindow extends PureComponent {
         onProgress={this.handleOnProgress}
         style={StyleSheet.absoluteFill}
         volume={mediaPlayer.muted ? 0 : 1}
+        selectedTextTrack={mediaPlayer.ccOn ? { type: 'index', value: 0 } : {}}
```
vinnyjth, line 131: Defaulting to 0 is a solid start, but it won't fly long term. Churches like Willow will probably want to broadcast subtitles in Spanish, if they exist, alongside the option for english subtitles. I think we can merge this as is, but I would love if you could figure out what level of visibility we have into what subtitles are available.

redreceipt:
So if there's no text track, you're suggesting an empty hole on the right side of the fullscreen controls UI?

redreceipt:
Oh I see, for V1, don't worry about detecting the track. Yea I can do that.

redreceipt:
like this...

Screen Shot 2020-06-22 at 11 23 09 AM

conrad-vanl:
I think I'd rather replace the mute button with the CC button, instead of the video show/hide. And for V1, detect if there's CC available (or have a prop for it that a church has to add) - if CC is available the CC button displays. If it isn't, either an empty hole on the bottom left or we can keep the mute/un-mute button.

redreceipt:
This seems like a lazy design hack. Like some churches don't want CC so we're just appeasing the ones that do by replacing a feature. Some may want both. Video toggle seems completely irrelevant.

conrad-vanl:
@redreceipt a future version of the player controls will likely do away with both mute and video toggle, and instead have a single "Cast/Airplay" button and a "CC" button. This brings us closer to native player controls, except native player controls still has a volume control. That is why I'm suggesting what I am....

conrad-vanl:
If you want to replace the video toggle button instead in this PR, that's fine. But I believe the other points are still valid

conrad-vanl:
@redreceipt just as a reminder this PR will be closed EoD if it's not ready. That's fine, you can re-open when it is ready for review again.

Sometime between now and the release of iOS 14, I think we will be refactoring the player all together, likely migrating to using [ApollosProject/react-native-apollos-player](https://github.com/apollosproject/react-native-apollos-player). Not sure if it makes sense to wait on this if it's going to be difficult to get this over the finish line. Up to you.

redreceipt:
What would a conservative estimate be on the timeline for a re-factored media player? It may be that it’s worth waiting for that.

conrad-vanl:
Conservative effort would be we would sync it up with the iOS 14 general release.

Otherwise, I could see this being something we release in the cycle after next (so we'd be waiting at least 6-8 weeks).

Next Wednesday during the betting meeting I'll get more insight.

conrad-vanl:
This pull request is being closed to clear out "backlog" post cycle ending. You can push to have it reprioritized in a future cycle or re-open once changes have been made and it is ready for review.

PR #1487, opened 2020-06-16
Adds core support for URL type and opening URLs Actions in w/ Features

vinnyjth:

This PR sets up the context for being able to link feature cards / action list items to URLs or Navigation actions. I also setup a fragment for the `relatedNodes` on all the feed features, so that they can be more easily extended. Right now it's very difficult to make a straightforward change to the `relatedNodes` 

This will help eliminate custom code in Grace, Willow, and the The Porch (and probably NS and CF as well)

n/a

### How do I test this PR?

1. Open the app
2. Make sure the feature feed loads. (I will follow up with a template PR demonstrating a link)

n/a

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-data-schema/src/index.js
+++ b/packages/apollos-data-schema/src/index.js
@@ -676,7 +676,6 @@ export const featuresSchema = gql`
     READ_CONTENT
     READ_EVENT
     OPEN_URL
-    OPEN_ROUTE
   }
 
   type Url implements Node {
--- a/packages/apollos-data-schema/src/index.js
+++ b/packages/apollos-data-schema/src/index.js
@@ -675,6 +675,13 @@ export const featuresSchema = gql`
   enum ACTION_FEATURE_ACTION {
     READ_CONTENT
     READ_EVENT
+    OPEN_URL
+    OPEN_ROUTE
+  }
+
+  type Url implements Node {
+    url: String
+    id: ID!
   }
 
   type ActionListAction {
--- a/packages/apollos-ui-connected/src/ActionListFeatureConnected/getActionListFeature.js
+++ b/packages/apollos-ui-connected/src/ActionListFeatureConnected/getActionListFeature.js
@@ -8,4 +8,6 @@ export default gql`
     }
   }
   ${ApollosConfig.FRAGMENTS.ACTION_LIST_FEATURE_FRAGMENT}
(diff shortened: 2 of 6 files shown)


redreceipt:
Conceptually I don't think I'm the right person to review this. I have no idea what the new feature/node stuff is or how to judge the merits of this approach or whether this works.

packages/apollos-data-schema/src/index.js
```diff
@@ -675,6 +675,13 @@ export const featuresSchema = gql`
   enum ACTION_FEATURE_ACTION {
     READ_CONTENT
     READ_EVENT
+    OPEN_URL
+    OPEN_ROUTE
```
conrad-vanl, line 679: Should this be _less_ abstract / tied to the way the frontend is composed? 

I'm just wondering if this should be more specific, like have it be `OPEN_TAB`, or more backend data structure specific, like `OPEN_NODE`? `OPEN_ROUTE` feels like it could be dangerous...what if a new app changes the way navigation is structured? Not sure that should be a "contract" we want to sign between api and frontend for them to be dependent on each other

vinnyjth: @conrad-vanl I don't have an answer to this question, and I haven't written any code to actually handle Navigation actions from features, so I'm going to remove this specific `ACTION_FEATURE_ACTION` for now.

packages/apollos-data-schema/src/index.js
```diff
@@ -675,6 +675,12 @@ export const featuresSchema = gql`
   enum ACTION_FEATURE_ACTION {
     READ_CONTENT
     READ_EVENT
+    OPEN_URL
+  }
+
+  type Url implements Node {
+    url: String
```
conrad-vanl, line 682: `url` vs `uri` naming? we use `uri` elsewhere just curious if it matters

vinnyjth: All of the partner apps use `url` with their implementations 😬  I think URL is fine here, but I also agree that I should have used URI to begin with. 

Arguably, per my understanding, we should have started with URL in the app. I believe URI is normally too broad. URI encompasses URL's and URN's, and our app only deals in URLs.

PR #1489, opened 2020-06-18
Create Apollos App

Lepozepo:
Adds a single command to create a "blank" apollos project

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
This is amazing. One thought: 

Lots of times clients will setup their repo by clicking the "clone template" button in Github on the templates project. That's been working out well for them so far, do you think you could configure this tool to function given an already setup repo? 

Something like ...

1. Click "Use Template"
2. Fetch code from your new Github repo
3. Run this tool, which renames the app and whatever else it needs to do.

Lepozepo:
@vinnyjth Yeah! I think that would make sense BUT I do think we should try to enforce a single path for everyone long term

Lepozepo:
@vinnyjth @conrad-vanl this should be in a decent spot now to at least get the end user running with a single command `yarn create apollos-app <church-name>` if the project was cloned we can also handle that with `yarn create apollos-app <church-name> --cloned` but that should be discouraged.

Not too sure how you guys are handling package publishing so I'll leave that up to you.

Lepozepo:
@conrad-vanl @vinnyjth have you guys had a chance to test this out? To test it out locally, you can do `./bin/run --help` to view available arguments and `./bin/run <nameOfChurch>` to see it build for you.

conrad-vanl:
I think we'd prefer users to use the "Use this template" approach, actually. IT's good advertisement since it adds this to the new github repo:

Screen Shot 2020-07-20 at 9 39 33 AM

So maybe the template project we can just have a `bootstrap` or `setup` command added to it that runs what's in this package?

Lepozepo:
@conrad-vanl we can do that ^_^ hopefully we won't run into a situation where a church does not use github

conrad-vanl:
In that case, I think a church could still `clone` the repo from github and then run the commands.

conrad-vanl:
@vinnyjth how do you want to handle this based on our preferred template repo approach? I don't mind keeping a `create` command in place just so we have it. Would this still be "ready" to go and adding to the template repo would be as simple as like... adding the following yarn command to the package.json:

```
"bootstrap": "yarn create apollos-app <church-name> --cloned"
```

maybe we need to add interactivity to prompt for <church-name> ?

Honestly, I think I might prefer interactivity prompts to get bundle name, package name, app name all separately, but defaulted. That way we can have some good help text and create a slightly better setup experience.

Lepozepo:
@conrad-vanl we'll want to patch this line https://github.com/ApollosProject/apollos-apps/pull/1489/files#diff-d9ce5fef4ca4e4866c7ea36befc66b96R65 also before merging, passing the `-b` flag makes it only set up for android and not iOS. We'll need to fork and patch `react-native-rename` though to be able to run on both android and iOS properly I think.

Lepozepo:
@conrad-vanl also, I think we can use https://developer.github.com/v3/repos/forks/ to have the CLI create the fork unless otherwise specified

conrad-vanl:
Okay @Lepozepo, thanks. In that case - I'm going to close this PR for now. We're doing a PR backlog push today and closing everything that's not going to be merged by end of cycle. We're reprioritizing work for next cycle and this will almost certainly be a part of it, so we'll re-open then. Thanks!

Lepozepo:
@conrad-vanl awesome

PR #1497, opened 2020-06-30
Ui-kit override names

christinamegura:
The purpose of this PR is to make sure all instances of using `withTheme` and `styled` in core's ui-kit package provides a named override. With "names", church apps are able to override that component's style. For example:

```
const MyButtonOnCore = styled({
   backgroundColor: 'green',
}, 'MyButtonOnCore')(Button);

```
And then in an app's overrides, you could override that button's styles:

```
const overrides = {
  'MyButtonOnCore': {
    backgroundColor: 'red',
  }
}
```


packages/apollos-ui-kit/src/typography/UIText/index.js
```diff
@@ -16,7 +16,7 @@ const UIText = compose(
       fontWeight: bold ? '700' : null,
       color: theme.colors.text.primary,
     }),
-    'UIText'
+    'ui-kit.UIText'
```
conrad-vanl, line 19: Example of name changing. Previously if you were overriding `UIText` your override would no longer work, as you'll have to rename your override to `ui-kit.UIText`. That's fine, but we should just document that.

christinamegura: Aah, yep. I'll document in the updating markdown. Thanks!

PR #1500, opened 2020-07-02
Refactor connected image

conrad-vanl:

Ground-up refactor on `ConnectedImage`. The main problem with `ConnectedImage` that this PR tackles is that it's slow. We have other issues with other Image components that we don't need ~(`ProgressiveImage`, for example)~, which I'd be happy to remove / consolidate in this PR, but wanted to start with this. Edit: I went ahead and removed `ProgressiveImage`

My main priority was to simplify the code structure and load order of `ConnectedImage` with hopes of improving load times as much as possible while still using the native `Image` component (the why on not using `react-native-fast-image` below).

The primary benefit from this refactor is that images are no longer delayed in mounting until the cache for that image is loaded. Instead, we immediately mount and render the image, and hydrate the cache for that image after the image is mounted and visible. This will trigger a second render when the cache round-trips, but that is okay as one of two things will happen:

1. The image will resize slightly to the size it should be, based on props like `maintainAspectRatio`
2. The tree won't change and the re-render is a no-op
3. (bonus!) The original implementation still re-rendered at this point, just for different reasons (In fact, I believe this component will have less total renders, and a quicker time to paint) 

1. Removing the "fade in" animation on iOS. We can't actually reliably detect when to play this. Often, this fade in animation is played when it doesn't need to be (such as opening a content-item when the image is already visible). Browsing most Apollos apps you'll see images flicker - that's bad! Instead, we'll let react-native image use it's native fade controls to fade the image in, which is currently only supported on Android. FWIW, I looked at a bunch of iOS apps and what I see is images don't usually fade in, but entire loading states will crossfade. Like facebook will have a loading state for a card, when that card's content loads, the entire loading state crossfades (text too). Images don't actually fade in ever though if they are delayed at all in loading. They just appear.

2. Doesn't use `react-native-fast-image`. See https://github.com/DylanVann/react-native-fast-image/issues/685

3. Changes every single snapshot. This is because I removed the SkeletonImage component. Believe it or not, rendering this component causes a noticeable slow-down in time-to-paint. And, we don't even need it! `ConnectedImage` is already rendering a default background-color of `inactive`, the same color skeleton uses. From a snapshot perspective, I consider this a win-win as images actually show up in snapshots now, instead of the skeleton.

These trade-offs actually are the key reason in a huge net code reduction. **However,** if I might brag a bit: the new `ConnectedImage` component is about half of the lines of code as the original. The original component definition didn't even __start__ until line `93`(!), wheres the current component __ends__ at line `116` 💪 . And, I _think_ it maintains 100% api compatibility, only losing the fadeIn animation.

### How do I test this PR?

Honestly, best thing you can do is open the Porch App, scroll to the podcast on the home screen, open it up, and see how slow the items in that action list are to load. Compare that to the 100 item story I added in this PR. 

Relying on existing test coverage. 

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-kit/src/ConnectedImage/index.js
```diff
@@ -45,30 +41,79 @@ const getCachedSources = (_sources = []) => {
...
+      );
+    }
+  }
+
+  if (forceRatio) {
+    aspectRatioStyle.aspectRatio = forceRatio;
+  }
+
+  const handleOnLoad = (e) => {
+    if (onLoad) onLoad(e);
```
burkeshartsis, line 97: I really like that you are keeping `onLoad`. This is probably the most useful api that this component exposes and would allow someone to "extend" the functionality of this component. As an example, if you wanted to add a fade-in effect you would use this prop (I'm not advocating for that, it's just an example). That said, would you be opposed to passing props and the images size along with the errors?

packages/apollos-ui-kit/src/ConnectedImage/index.js
```diff
@@ -18,18 +15,17 @@ export const ImageSourceType = PropTypes.oneOfType([
...
-  if (source.size && source.fileLabel)
-    return `${source.size}-${source.fileLabel}`;
   if (source.url) return source.url;
   if (source.uri) return source.uri;
+  if (typeof source === 'string') return source;
   return undefined;
 };
 
-const getCachedSources = (_sources = []) => {
+export const getCachedSources = (_sources = []) => {
```
burkeshartsis, line 28: I wonder if it's worth refactoring this to use [`queryCache`](https://reactnative.dev/docs/image#querycache)? Probably could reduce this components footprint even further.

packages/apollos-ui-kit/src/ConnectedImage/index.js
```diff
@@ -45,30 +41,79 @@ const getCachedSources = (_sources = []) => {
...
+  }
+
+  const handleOnLoad = (e) => {
+    if (onLoad) onLoad(e);
+    if (!imageInCache) {
+      const {
+        nativeEvent: { source: loadedSource },
+      } = e;
+      updateCache(loadedSource);
+      setImageSize([loadedSource.newWidth, loadedSource.newHeight]);
```
burkeshartsis, line 103: You might be able to refactor this native event code away by using [`resolveAssetSource`](https://reactnative.dev/docs/image#resolveassetsource). You'll still need state but I think there are a few functions you have written that RN Image now has methods for.

packages/apollos-ui-kit/src/ConnectedImage/index.js
conrad-vanl: All of those props/attributes are contained within the event object (‘e’) that’s being passed

packages/apollos-ui-kit/src/ConnectedImage/index.js
conrad-vanl: `queryCache` doesn't return size properties, I already looked. Also, `queryCache` is async/promise-based, wheres this is syncronous.

packages/apollos-ui-kit/src/ConnectedImage/index.js
conrad-vanl: I actually originally wrote the component this way, it was more complex and less efficient since it effectively "loaded" the image twice. This route loads the image once, and then re-renders. `resolveAssetSource` wasn't as friendly here, I promise.

PR #1511, opened 2020-07-13
Ui packages override names

christinamegura:
The purpose of this PR is to make sure all instances of using `withTheme` and `styled` in all core packages provide a named override. With "names", church apps are able to override that component's style. For example:

```
const MyButtonOnCore = styled({
   backgroundColor: 'green',
}, 'MyButtonOnCore')(Button);

```
And then in an app's overrides, you could override that button's styles:

```
const overrides = {
  'MyButtonOnCore': {
    backgroundColor: 'red',
  }
}
```

### DIFF
--- a/packages/apollos-ui-scripture/src/Scripture/typography/index.js
+++ b/packages/apollos-ui-scripture/src/Scripture/typography/index.js
@@ -4,24 +4,36 @@ import { styled } from '@apollosproject/ui-kit';
 
 import ScriptureText from './ScriptureText';
 
-const VerseNumber = styled(({ theme }) => ({
-  fontSize: theme.helpers.rem(0.6),
-  color: theme.colors.text.secondary,
-}))(ScriptureText);
+const VerseNumber = styled(
+  ({ theme }) => ({
+    fontSize: theme.helpers.rem(0.6),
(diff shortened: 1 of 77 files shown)


conrad-vanl:
This has been marked as approved but tests are failing. Please update tests.

packages/apollos-ui-scripture/src/Scripture/typography/index.js
```diff
@@ -4,24 +4,36 @@ import { styled } from '@apollosproject/ui-kit';
...
-const VerseNumber = styled(({ theme }) => ({
-  fontSize: theme.helpers.rem(0.6),
-  color: theme.colors.text.secondary,
-}))(ScriptureText);
+const VerseNumber = styled(
+  ({ theme }) => ({
+    fontSize: theme.helpers.rem(0.6),
+    color: theme.colors.text.secondary,
+  }),
+  'ui-scripture.Scipture.typography.VerseNumber'
```
vinnyjth, line 12: A typo in this file I think :) `Scipture` instead of `Scripture`

christinamegura: Aah, yep, thank you!

PR #1517, opened 2020-07-23
MVP of Auth Refresh (instead of error) 

vinnyjth:
## DESCRIPTION

Screen Shot 2020-07-23 at 2 52 21 PM

Makes the following changes: 

- If an auth attempt fails because Rock denies the cooke
  1. Check to see if we have a reference to a sha1 hashed version of the cookie in Redis. 
  2. If we do, lookup the cookie and grab the User Name (stored on future logins) 
  3. With the username, fetch the UserLogin, then fetch the associated Person (wish we could do this in one step)
  4. If any of those fail, raise a Useful error to make debugging in the future easier. 

- On login
 1. Once we have a cookie, make a SHA1 version of the cookie, and use it as the key, with the value being a user's username. 

Q: Why store the username in redis instead of the PersonId? 
A: I wanted there to be some form of rock based bail out. if a user login gets changed or deleted in rock, we can properly log the user out. 

Q: So the user will be failing login, and then getting their data from redis forever? eww. 
A: Baby steps. A client change will be required to actually send a user a new cookie to use. We should get this code out as early as possible so we can start collecting cookies/usernames in Redis. 

Q: How soon will this have an impact? 
A: A user will get logged out one more time, theoretically, but any future time they login we'll track their cookie. We could add preemptive code that creates that cookie/username association on ever `getCurrentPerson` call, but that feels like overkill. 

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+
 language: node_js
 node_js: '10.19.0'
 cache:
@@ -29,8 +30,8 @@ jobs:
       script:
         - yarn lint
     - script:
-        - travis_wait yarn ci-test
-        - travis_wait yarn codecov
+        - yarn ci-test
+        - yarn codecov
     - stage: Deploy Api
       script: skip
       deploy:
--- a/packages/apollos-data-connector-rock/src/auth/__tests__/__snapshots__/data-source.tests.js.snap
+++ b/packages/apollos-data-connector-rock/src/auth/__tests__/__snapshots__/data-source.tests.js.snap
@@ -69,6 +69,47 @@ Array [
 ]
 `;
 
+exports[`Auth should return a person while retrying getCurrentPerson 1`] = `
+Object {
+  "id": 123,
+}
+`;
+
+exports[`Auth should return a person while retrying getCurrentPerson 2`] = `
+Array [
+  Array [
+    Object {
+      "key": ":userLogins:234d7a42b877ed9b07eec6880c3576deac311057",
+    },
+  ],
+]
+`;
+
+exports[`Auth should return a person while retrying getCurrentPerson 3`] = `
+Array [
+  Array [
+    "People/GetCurrentPerson",
+    Object {},
+    Object {
+      "headers": Object {
+        "Authorization-Token": null,
+        "cookie": "some cookie",
+      },
+    },
+  ],
+  Array [
+    "/UserLogins?%24filter=UserName%20eq%20%27username%40example.com%27&%24expand=Person&%24top=1",
+    Object {},
+    Object {},
+  ],
+  Array [
+    "/People?%24filter=Id%20eq%20123&%24top=1",
+    Object {},
+    Object {},
+  ],
+]
+`;
+
 exports[`Auth should throw an error when creating an invalid user 1`] = `
 Array [
   Array [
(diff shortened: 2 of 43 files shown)


vinnyjth:
@conrad-vanl I'm going to write tests for this, and then try to hack together a demo of the login failed side. Can you give it a preliminary glance over in the meantime?

conrad-vanl:
Can @richarddubay @redreceipt or @asleepingpanda 👀  on this a little and give input from the Rock side / security aspect?

conrad-vanl:
> Q: Why store the username in redis instead of the PersonId?
> A: I wanted there to be some form of rock based bail out. if a user login gets changed or deleted in rock, we can properly log the user out.

@vinnyjth if we used an impersonation token or a person token instead of a username would that be better from this perspective?

conrad-vanl:
Also, will the Willow update atleast verify that we are in fact getting tokens denied from Rock? I'd hate to put this out there and still not know whether or not this will fix anything.

vinnyjth:
@conrad-vanl It would be different. We would have the option of setting an expiration on that token directly, but I don't think a user changing their login would invalidate the impersonation token. I'm open to either option. I like the UserLogin option better, because i understand the all the implications of using UserLogin. I'm still a little weary of the ImpersonationToken, because I don't want to run into the same "they were invalidated and we don't know why" problems.


What's a person token?

vinnyjth:
> Also, will the Willow update atleast verify that we are in fact getting tokens denied from Rock? I'd hate to put this out there and still not know whether or not this will fix anything.

Yep! The build is submitted, we are now waiting for approval. I think we can safely hold off merging this until we get data from Willow

conrad-vanl:
Screen Shot 2020-07-23 at 3 26 39 PM

Screen Shot 2020-07-23 at 3 26 59 PM

vinnyjth:
Oh okay - so same thing.

redreceipt:
waiting makes sense but wanted to throw out there that even if this doesn't fix anything for our current bug, it will solve the issue where jwts will expire in 400 days and everyone is going to get logged out again.

conrad-vanl:
@vinnyjth with this approach, why use cookies at all? Aren't they effectively not used now? Like, if the cookie is invalid, we just circumvent the cookie. Why still use it at all?

vinnyjth:
> @vinnyjth with this approach, why use cookies at all? Aren't they effectively not used now? Like, if the cookie is invalid, we just circumvent the cookie. Why still use it at all?

Because it is a drop in update that doesn't require any architectural change. Also, we can patch this version to send updated cookies to the client in the future (so that cookies, in some sense, do matter).

That said, I understand the sentiment here. I would like to talk to Jon about moving to People tokens long term (replacing cookies altogether with people tokens) and removing this "retry" logic.

vinnyjth:
@conrad-vanl I think the next step, no matter what, is to equip the client to update the auth token that it stores. That's going to be a requirement if we want to change what is stored in auth token, no matter what direction we go, long term. I'm hoping that this PR will, for now, stop the bleeding.

vinnyjth:
Added tests, and so I'm now considering this "ready for review"

conrad-vanl:
FYI - the above comment was 100% meant to be suggestive and open for discussion and not "this is what we're doing" ;-) 

I'm not sure of the best solution, I'm just trying to figure out what's going to help us the most, introduce the least amount of debt, and be lowest risk. Arguably, what I'm proposing could be looked at as more debt, or less debt. I'm not really sure.

vinnyjth:
> FYI - the above comment was 100% meant to be suggestive and open for discussion and not "this is what we're doing" ;-)

I'm all for your suggestion :) I think it's slightly more vulnerable from a security standpoint (someone could, if they know the `SECRET`, login as anyone if they knew their ID, or could cycle through IDs sequentially. 

I like the simplicity of your approach, and I think beyond being slightly more performant it will function identically. 

I need to jump back over to integration work, but when I wrap up that I can jump back over here and rewrite this PR.

vinnyjth:
In terms of debt, I think both approaches have equal debt. They both touch the codebase in the same way, at the same places. I think both have logic that is easy to follow. The branches paths of my PR are arguably a little more complex, but I hope I've mitigated through clear annotation and specific error messages.

vinnyjth:
@conrad-vanl I just realized one big issue with the `personId` approach. It will become inaccurate if a user is deduped post login. That _shouldn't_ happen, now that we try and de-dup up front, but if a de-dup is performed after login, a user will never end up associated with their new profile until they logout and login again.

conrad-vanl:
@vinnyjth based on your explanation I'm a fan of your `username` approach then. If you get this PR in that state and tag me I'll do the final review so we can merge.

vinnyjth:
@conrad-vanl We're back at the "username" state.

PR #1520, opened 2020-07-30
Adds Remote campus support

vinnyjth:

On at least two of our other apps (LCBC, The Porch), we have code that supports remote campuses. The code used is identical in both cases, and doesn't have any negative trade offs, so I figured it would be handy to bring it over to core.

n/a

### How do I test this PR?

Download LCBC or The Porch ;) My tests should also show the code functioning. 

Updated existing tests and added new tests.

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-apollo-server-env-mock/src/index.js
```diff
@@ -256,9 +256,9 @@ const createApolloServerEnvMock = (apolloServerEnv) => {
         return resolveWith([]);
       }
 
-      if (url.match('api/Campuses/\\d')) {
-        return resolveWith(rockMocks.campuses()[0]);
-      }
+      // if (url.match('api/Campuses/\\d')) {
```
conrad-vanl, line 259: Can we remove this code altogether?

packages/apollos-data-connector-rock/src/campuses/data-source.js
```diff
@@ -9,21 +9,40 @@ export default class Campus extends RockApolloDataSource {
 
   getFromId = (id) =>
     this.request()
-      .find(id)
+      .filter(`Id eq ${id}`)
```
conrad-vanl, line 12: Is there a reason we move from `find` to `.filter(...).first()`? Does `expand('CampusTypeValue)` not work on `find()` I'm guessing?

packages/apollos-data-connector-rock/src/campuses/data-source.js
```diff
@@ -9,21 +9,40 @@ export default class Campus extends RockApolloDataSource {
...
   getByLocation = async ({ latitude, longitude } = {}) => {
     let campuses = await this.getAll();
+
+    const onlineCampuses = campuses
+      .filter(({ campusTypeValue }) => campusTypeValue?.value === 'Online')
+      .map((campus) => ({
+        ...campus,
+        location: {
+          ...campus.location,
+          street1: 'No locations near you. ',
```
conrad-vanl, line 36: These values should really come from the config.yaml....

packages/apollos-data-connector-rock/src/campuses/data-source.js
```diff
@@ -9,21 +9,40 @@ export default class Campus extends RockApolloDataSource {
...
+          city: "When there's one",
+          state: "we'll let you know!",
```
conrad-vanl, line 38: Also, I kind of wonder if we should generalize this field or something.... deprecate `location` and make like `locationText` that allows for line breaks and such?

packages/apollos-apollo-server-env-mock/src/index.js
vinnyjth: Yep

packages/apollos-data-connector-rock/src/campuses/data-source.js
vinnyjth: Yah, `expand` doesn't work at all with `.find`. None of the extra data was never being fetched.

packages/apollos-data-connector-rock/src/campuses/data-source.js
vinnyjth: I'm on board with that. I don't think we actually use any of the fields for geolocation besides lat/long

packages/apollos-data-connector-rock/src/campuses/data-source.js
vinnyjth: What do you feel about something like this? https://github.com/ApollosProject/apollos-apps/pull/1520/commits/7119dc0439635d49200ed5cff8257ae1a91fa14b

Adds ability to modify those fields easily, as needed, and without needing to change all of them. Also is backwards compat because we aren't required to add new mandatory fields to the config.yml

PR #1524, opened 2020-08-05
Update updating.md

christinamegura:
Update the updating.md file to include more succinct instructions regarding the styled-components override names.

### DIFF
--- a/UPDATING.md
+++ b/UPDATING.md
@@ -1,4 +1,4 @@
-# Updating from 1.4.3 to 1.5.0-beta.0
+# Updating from 1.4.3 to 1.5.0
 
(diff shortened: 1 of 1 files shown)


UPDATING.md
```diff
@@ -1,87 +1,17 @@
-# Updating from 1.4.1 to 1.4.2
+# Updating from 1.4.3 to 1.5.0-beta.0
```
vinnyjth, line 1: Maybe we can reference 1.5.0 (without the beta version) instead? That way it's more clear for users who aren't interested in the beta builds :)

christinamegura: Doh! Lol, dropped it

PR #1531, opened 2020-08-18
perf: Adds keepAlive support to the DataConnectorRock class.

vinnyjth:
## DESCRIPTION

A non-trivial chunk of time in any HTTP requests is spent setting up the request. Normally, that cost is nominal and a simple fact of "doing HTTP business" but, you are expecting to make several HTTP requests over time to a single HTTP server it makes sense to keep the connection open. 

This is something you have to opt into in the node fetch/http module by providing a custom agent. By default, Node uses a `globalAgent` which avoids keeping any connection open. 

This PR will hopefully provide meaningful improvements to the performance of all Church Apps that use Rock extensively. Small tests locally using load testing show a speedup of around 30%. This should hopefully improve the throughput of Rock as well, as Rock will also no longer need to spend time tearing down connections we have made.

The downside is agents that are used, and then abandoned, have the potentially to hold onto system resources. This would be the case if a Rock request was made only once every five minutes or so. 

### How do I test this PR?

I tested this PR by opting into the functionality with the config.yml (add `USE_AGENT: true` to the `ROCK` section) and then running load tests a few times against my local instance and comparing performance. 

None :/ I can't think of a good way to test this with either unit or integration tests. 

Before: 
Screen Shot 2020-08-18 at 10 48 42 AM

After:
Screen Shot 2020-08-18 at 10 47 47 AM

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,0 +1,28 @@
+name: Linting
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [10.19.x]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm install -g yarn
+
+    - name: Install Dependencies
+      run: yarn
+
+    - name: Lint
+      run: yarn lint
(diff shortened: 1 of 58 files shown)


redreceipt:
I feel like we should just turn this on by default instead of the added complication of another config variable

vinnyjth:
@redreceipt I'm a little worried that this might have unintended consequences on performance that only become evident at scale - like we might hog all of a churches rock connections and their website stops working, or something similar. I'm fine turning it on by default, but I also want to make sure we have an easy to deploy kill switch :)

vinnyjth:
Turned it on by default 👍

PR #1535, opened 2020-08-18
chore: Adds canary release to Github Actions

redreceipt:

Completely automates CHANGELOG generation, Github releases, and NPM publishes. I really tried to cover all our bases here but I think `lerna` makes this suuuuper easy.

I removed a lot of the old scripts and infrastructure that we were using to tiptoe through the process. Don't think we'll need them any more which is awesome but this may be trial and error, unfortunately. 

Worst-case scenario, git tags, and releases are published wrong (we can delete them) or NPM canaries are published wrong (we can delete them). If the canaries work via Actions, I'm confident the regular release versions will work as well.

This is merely the first step in getting canaries to publish on every push to master. The next steps will be to think through templates and the upgrade tool.

Will still prompt for "are you sure" on regular releases but it will in fact push to Github for us.

### How do I test this PR?

🤷 

None.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
@redreceipt I'm going to review and merge this earlier rather than later so hopefully we can get this refined by the end of the cycle. I know how hard this logic is to get right when it's not actually live :)

package.json
```diff
@@ -18,10 +18,8 @@
     "lint": "lerna run lint",
     "test": "TZ=utc lerna run test",
     "ci-test": "lerna run test  --stream -- --maxWorkers=2",
-    "release": "lerna publish --no-push --force-publish && echo \"Release has been tagged!\nPlease look over the CHANGELOG.md file, and use 'git commit -amend' to amend any edits.\nWhen done, be sure to 'git push --tags' and create a release PR\"",
-    "release:canary": "lerna publish --canary --yes",
-    "preversion": "./scripts/update-change-log.sh",
-    "prepublishOnly": "lerna run build && lerna run generate-stories",
```
vinnyjth, line 24: We should probs keep this line, we will still need to build the lib/ folder and the stories to publish.

redreceipt: yep!

PR #1536, opened 2020-08-21
feat: Initial draft of included interfaces

vinnyjth:

This PR is small, but rather foundational, so I'm open to all feedback in order to get this right. 

As you're reviewing this PR, the questions you should be asking yourself are ...

- Do the fields fragments make sense?
- Do the names of the fragments make sense? 
- Do any of these fragments feel useless? 
- As a developer, would I be confused by any of these fragments? 

This PR is intended to provide support for fragments in a way that you can opt out of and extend easily. That's why the fragments are contained in their own data object. You can omit the object, and not have fragments generated, or include the object, and have the fragments added to the schema. 

Any new ContentTypes added to the `config.yml` don't get the fragments by default. You have to opt into them. That's an intentional change because the interplay between the config.yml is confusing, and I want to move away from magical schema generation. 

### How do I test this PR?

There's a companion PR you can pull down: 

https://github.com/ApollosProject/apollos-templates/pull/92

The best way to test this PR would be to run a query that mimics the queries that client will run, like this

```
{
  node(id:"ContentSeriesContentItem:9712de06eab8adc3050f8f96d72b79b3"){
    id
    ... on ContentNode {
      title
    }
    ... on LikableNode {
      isLiked
    }
  }
}
```

Screen Shot 2020-08-27 at 1 56 01 PM

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-data-schema/src/index.js
```diff
@@ -1,6 +1,61 @@
...
+  interface ProgressNode {
+    percentComplete: Float @cacheControl(maxAge: 0)
+    upNext: ContentItem @cacheControl(maxAge: 0)
+  }
+
+  interface ScriptureNode {
+    scriptures: [Scripture]
+  }
+
+  interface LikeableNode {
```
redreceipt, line 44:
```suggestion
  interface LikableNode {
```

packages/apollos-data-schema/src/index.js
```diff
@@ -1,6 +1,61 @@
...
+    likedCount: Int
+  }
+
+  interface ShareableNode {
+    sharing: SharingFields
+  }
+
+  interface ChildNode {
+    parentNode: Node
+    siblingConnection(first: Int, after: String): NodeConnection
```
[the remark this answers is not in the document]
vinnyjth: I would say "yes" 99% of the time. The most common situation where a child doesn't have siblings is if they are the only child (so far), like the first entry in a sermon series.

packages/apollos-data-schema/src/index.js
vinnyjth: Alright, you got me.

packages/apollos-data-schema/src/index.js
```diff
@@ -1,6 +1,61 @@
 import gql from 'graphql-tag';
 import { extendForEachContentItemType } from './utils';
 
+export const interfacesSchema = gql`
+  interface Card {
```
redreceipt, line 5: What about `ContentNode`? those are fields that make sense for anything coming from a CMS. that way the UI is abstracted from the data

redreceipt: Even an `Event` could be a `ContentNode` and it makes sense.

vinnyjth: What would do with the ContentNode I have below? Combine them or rename HTMLNode?

redreceipt: the biggest win in my mind has _one_ `ContentNode` that includes a `body` field. The one below could be a deprecated `HTMLContentNode`

vinnyjth: I'm slowly coming around. One of my goals for this PR is "no renaming though", so I think we would avoid the `HTMLContentNode`

redreceipt: Gotcha, in that case I think it's worth to figure out to have one `ContentNode` with everything.

vinnyjth:
This might come in handy down the road: https://dev.to/mikemarcacci/intermediate-interfaces-generic-utility-types-in-graphql-50e8 (available as of GQL 15)

packages/apollos-data-schema/src/index.js
```diff
@@ -513,6 +589,21 @@ export const sharableSchema = gql`
     sharing: SharableContentItem
 `)}
 
+  interface ShareableNode {
+    sharing: SharableContentItem
+  }
+
+  ${addInterfacesForEachContentItemType(
```
vinnyjth, line 596: This helper is intended to save space, and avoiding having 100s of lines of schema config. It performs no magic.

packages/apollos-data-schema/src/index.js
```diff
@@ -1,5 +1,82 @@
 import gql from 'graphql-tag';
-import { extendForEachContentItemType } from './utils';
+import {
+  extendForEachContentItemType,
+  addInterfacesForEachContentItemType,
+} from './utils';
+
+export const interfacesSchema = gql`
+  interface ContentNode {
```
conrad-vanl, line 8: I know there was some chatter above about combining `Card` and `ContentNode` into one. That only feels like a stretch to me when we consider how we were intending on using the previous `Card` interface.... the intention of the `Card` interface was to have a small interface set aside for rendering nodes inside of our card UI. `htmlContent` is not a requirement for cards, and I don't know how I feel about making all content require `html`.... what if we have other non-html content types down the road? There was some talk about having an `HTMLContentNode` ....what would we feel about splitting that field into that interface?

Or...is this all premature optimization?

conrad-vanl: Another question: Can we deprecate fields inside of interfaces? For example, later if we want to pull `HTMLContent` as an attribute from `ContentNode` and put it inside of a different interface, can we do that in a non-breaking way?

Or I guess types can always define their own interfaces, so removing attributes from interfaces aren't really a problem cuz that's non-breaking

vinnyjth: > Another question: Can we deprecate fields inside of interfaces? 

I think the answer for this is "as long as we deprecate it on all types that implement it". So I think it would be a bad idea to do so :) 

 > The intention of the Card interface was to have a small interface set aside for rendering nodes inside of our card UI.

I'm really struggling with the naming/fields of this specific type. `Card` is more useful, but less appropriate (and strands HTMLContent off in no mans land. `ContentNode` is less useful, but matches the rest of the schema better. 


My proposal! Let's keep `Card`. It's foundational enough to the app, that I think it's fine. Move HTMLContent into ContentNode.

```
 fragment Card {
    title(hyphenated: Boolean): String
    coverImage: ImageMedia
    summary: String
  }

fragment ContentNode {
  htmlContent: String
}

```

conrad-vanl: Actually, I think it should be?

```
 fragment Card {
    title(hyphenated: Boolean): String
    coverImage: ImageMedia
    summary: String
  }

fragment ContentNode {
    title(hyphenated: Boolean): String
    coverImage: ImageMedia
    htmlContent: String
}

`title` and `coverImage` are needed for both content and cards
```

vinnyjth: Good call, that schema looks as close to perfect as we are going to get!

vinnyjth: I'll make the change and then we'll merge this.

PR #1545, opened 2020-09-03
feat: Connected Content Node

vinnyjth:

This PR accomplishes two major goals 

1. Adds a `ConnectedNodeComponent` that can render the major features of a `ContentSingle` from a `ContentNode`. 
2. Updates the `MediaControlsConnected` to accept a NodeID, and render a MediaControl given that NodeId. 

### How do I test this PR?

Same as last time, I added some stories to storybook 👍 

Simulator Screen Shot - iPhone 11 - 2020-09-03 at 13 12 09
Simulator Screen Shot - iPhone 11 - 2020-09-03 at 13 11 55

Tests of all new components. 

## TODO:

- [x] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-ui-connected/src/ContentNodeConnected/ContentNodeConnected.js
```diff
@@ -0,0 +1,97 @@
...
+        data: { node: { htmlContent, title, coverImage } = {} } = {},
+        loading,
+        error,
+      }) => {
+        if (!htmlContent && error) return <ErrorCard error={error} />;
+        const coverImageSources = coverImage?.sources || [];
+        return (
+          <>
+            {coverImageSources.length || loading ? (
+              <ImageWrapperComponent>
```
vinnyjth, line 55: This is primarily to add a StretchyView in.

packages/apollos-ui-connected/src/ContentNodeConnected/ContentNodeConnected.js
```diff
@@ -0,0 +1,97 @@
...
+              <H2 padded isLoading={!title && loading}>
+                {title}
+              </H2>
+              <HtmlComponent
+                isLoading={!htmlContent && loading}
+                onPressAnchor={onPressAnchor}
+              >
+                {htmlContent}
+              </HtmlComponent>
+            </PaddedView>
```
conrad-vanl, line 74: Will Features, Continue button, Child/Sibling Content, etc end up here eventually?

vinnyjth:
Taking the ready for review label off for a second because there's a tweak I want to make

packages/apollos-ui-connected/src/ContentNodeConnected/ContentNodeConnected.js
vinnyjth: Current thinking is that they will end up in separate components. 

Then, there will be a `ConnectedNode` component that includes all of the above (probably will live in UI Connected also).

vinnyjth:
@conrad-vanl Ready again. I wanted to stay true to my promise to keep the new interfaces optional.

vinnyjth:
Backed out my changes, all queries now include the new interfaces and reference the old ContentItem interface.

packages/apollos-ui-connected/src/ContentNodeConnected/ContentNodeConnected.js
conrad-vanl: Ah got it. I think I'm seeing now what you're suggesting. Have you considered splitting this component up more?

This could just export everything in the `<PaddedView>`, then separate components for `MediaControls` and `ContentNodeImage` ?? I'm okay with this through, just thinking through the API exposed to client apps.

conrad-vanl (approved):
Overall, I think this is great. My only 🤔 is the `ContentNodeConnected` component, and whether it's the right level to expose. This is kind of a medium-sized component - doesn't contain everything you need for rendering a content-single (you'll also need a scrollview at minimum), but it also contains a lot of things you'll likely want to customize (`ImageWrapperComponent`, `MediaControlsComponent`, etc). Makes me wonder if we should go further in one direction or the other.

vinnyjth:
@conrad-vanl This was definitely trick to compose, because I couldn't include the ScrollView (there will be other components, in general, that belong in the scroll view) but I also wanted to include as much as possible to make rendering basic content, simple. the "NodeConnected" component will provide the "All in one" component, and I am cleaning up a lot of the smaller pieces as we go, so I'm feeling okay about this :)

PR #1577, opened 2020-09-18
feat: Connected Theme Mixin

vinnyjth:
## DESCRIPTION

This PR is pretty simple, adds a ThemeMixin that can take a NodeId, pull in the theme for that node, and passes it down to it's children.

I had to make a change to the ThemeMixin in order to update properly when the `mixin` prop changes. It's a tiny bit of a hack, but I think it makes sense given the context and usage of the component.  

Screen Shot 2020-09-18 at 1 15 24 PM

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


ApollosStorybook/package.json
```diff
@@ -7,7 +7,7 @@
     "ios": "react-native run-ios --simulator=\"iPhone 11\"",
     "generate-stories": "rnstl --searchDir ./src --pattern **/*.stories.js",
     "prestart": "yarn generate-stories",
-    "postinstall": "jetify && yarn pods",
+    "postinstall": "jetify && yarn pods && (cd .. && yarn lerna run generate-stories)",
```
vinnyjth, line 10: This is how I fixed the storybook problem in my PR.

redreceipt: still need this?

vinnyjth: Nope!

vinnyjth, line 10:
```suggestion
    "postinstall": "jetify && yarn pods",
```

packages/apollos-ui-connected/src/ThemeMixinConnected/index.js
```diff
@@ -0,0 +1,28 @@
...
+        {...props}
+      />
+    )}
+  </Query>
+);
+
+ThemeMixinConnected.propTypes = {
+  nodeId: PropTypes.string,
+  children: PropTypes.oneOf([PropTypes.func, PropTypes.node]),
+  defaultMixin: PropTypes.shape({}),
```
redreceipt, line 25: these should be handled on whatever is down below

packages/apollos-ui-connected/src/ThemeMixinConnected/ThemeMixinConnected.stories.js
```diff
@@ -0,0 +1,57 @@
...
+  .addDecorator((story) => (
+    <BackgroundView>
+      {/* eslint-disable-next-line react-native/no-inline-styles */}
+      <CenteredView style={{ alignItems: 'center' }}>{story()}</CenteredView>
+    </BackgroundView>
+  ))
+  .addDecorator(ApolloStorybookDecorator({ mocks }))
+  .add('example', () => (
+    <ThemeMixinConnected
+      // mixin={{ colors: { primary: 'rgb(0, 255, 0) ' } }}
```
redreceipt, line 52:
```suggestion
```

packages/apollos-ui-connected/src/ThemeMixinConnected/index.js
vinnyjth: Thanks. At one point i was explicitly pulling them in, but now I spread them down.

packages/apollos-ui-kit/src/theme/mixins.js
```diff
@@ -39,7 +39,8 @@ const withThemeMixin = (themeInput) =>
     withPropsOnChange(
       (props, nextProps) =>
         props.theme !== nextProps.theme ||
-        props.themeInput !== nextProps.themeInput,
+        props.themeInput !== nextProps.themeInput ||
+        props.ownProps.mixin !== nextProps.ownProps.mixin,
```
redreceipt, line 43: care to take this opportunity to explain this HOC to me 😂 I learned web dev after mixins and HOCs were a thing so this is all catch up

vinnyjth: Gah I can try and walk you through it, but this HOC always takes me like half a day to unparse. 

The functions (hocs) you see lined up are called one after each other, each one affects the Wrapped component or lifecycle differently. 

There's an HOC above that maps the passed props as `ownProps`, and then this `withPropsOnChange` consumes the `theme` and `themeInput` (which is the user's theme input) and only runs the below HOCs when either one of them changed. 

Because of that, the Mixin HOC doesn't update if it get's a new set of Mixin values. The mixin theme values aren't in either the theme or themeInput. I added a check against the mixin value (which is part of the components theme that was mapped into `ownProps`) to make sure that if the mixin value changed (like after a GQL query loaded) the change would be picked up. 


@conrad-vanl This actually fixes something we were seeing in LCBC where the `theme` wouldn't load until you refreshed, navigated and came back, or remounted the ThemeMixin some other way.

redreceipt (approved):

PR #1590, opened 2020-09-25
fix: Generate feature IDs in the resolvers, rather than in the datasources. 

vinnyjth:

Taking a step back ....

IDs in the context of the app mean one of two things. **Global IDs** (which we expose to the client) and internal **IDs** that we pass around our app, but never show to the client. 

In general, all IDs that enter our app (through a mutation or query variable) should be **Global IDs**. All IDs that exit our app, via GraphQL fields, should be **Global IDs**. Inside of a Global ID is enough information to determine **what** something is, and enough information to determine **where** to find it. Sometimes this is the object itself (for things like links) and sometimes this is a pointer, like a Rock.ID so you can use the were and the what to find it. 

I ran into a bug the other day where our Features, under a specific set of circumstances, were returning their local Ids in GraphQL. A big no no. I made the mistake of changing the `Node.get` method, which consumes an internal ID and returns the object itself, to return the ID of the object as a GlobalID. This fixed my circumstance with the `Features`,  because the `Features` **did not have a id resolver**.  Whatever was in the `id` field of the resolver was passed straight to the client. My **fix** broke other areas of the app (mainly search) that relied on `Node.get`, because in those cases where `id` resolvers existed, like `ContentItem`, the ID was being passed through `createGlobalID` twice. Once in `Node.get` and once in the `ContentItem.resolvers.id()` 

So the fix in this case is to do what I should have done all along and make the ID field of the `Feature` objects we create the **actual id** instead of an immediately obfuscated (createGlobalId) version of the actual id. I then moved the `createGlobalId` logic to the resolvers, and now the id will be properly obfuscated in the special set of circumstances that sent me down this rabbit hole to begin with. 

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt (approved):
nice. good change. linked app boots and home feed refreshes. good enough for me!

PR #1600, opened 2020-10-05
fix: Checks For Cast Devices

redreceipt:
## DESCRIPTION

Screen Shot 2020-10-06 at 8 17 36 AM

(This image has a random icon where airplay should be because it doesn't show up in a sim. Just to show that Google Cast is not there.)

This will check if any cast devices are available and then set local state accordingly so there's not an empty hole where Google Cast should be.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
just waiting to test this. need to add a video to the CMS and prob boot a real device

packages/apollos-ui-media-player/src/MediaPlayer/GoogleCastController.js
```diff
@@ -40,6 +41,10 @@ class Controller extends React.Component {
   componentDidMount() {
     // get Google Cast state on mount
     GoogleCast.getCastState().then((state) => {
+      const noDevices = state === 'NoDevicesAvailable';
+      this.props.client.mutate({
+        mutation: gql`mutation {mediaPlayerUpdateState(isCastAvailable: ${!noDevices}) @client }`,
```
vinnyjth, line 46: We should probably move the mutation into its own `const` to match how we handle the other mutations in this file.

packages/apollos-ui-media-player/src/MediaPlayer/FullscreenPlayer.js
```diff
@@ -234,7 +238,7 @@ class FullscreenPlayer extends PureComponent {
             showAudioToggleControl={this.props.showAudioToggleControl}
             showVideoToggleControl={this.props.showVideoToggleControl}
             airPlayEnabled={this.props.airPlayEnabled}
-            googleCastEnabled={this.props.googleCastEnabled}
+            googleCastEnabled={this.props.googleCastEnabled && isCastAvailable}
```
vinnyjth, line 241: 👍 👍

vinnyjth (approved):

packages/apollos-ui-media-player/src/MediaPlayer/GoogleCastController.js
redreceipt: the only reason I didn't is that all this is changing soon anyway (hopefully 🤞) plus how we do mutations and queries on this package is kind of an antipattern. logic is spread out and very difficult to find. to match how we did it before, I would need a `CAST_NOT_AVAILABLE` and `CAST_AVAILABLE` mutation and would have made this line and logic longer instead of just swapping the variable.

So I did give it some thought but if you'd like me to match anyway, no problem, I can swap it out.

vinnyjth: Ah, I see what you're saying. Would it be possible to have a `variable` in this mutation to get rid of the need to have two mutations? 

something like 

```
mutation updateIsCastAvailable($isCastAvailable: Boolean){
  mediaPlayerUpdateState(isCastAvailable: $isCastAvailable) @client
}
```

vinnyjth: The resolver would see the same mutation/value with either syntax.

redreceipt: yep I could do that

redreceipt:
alright @vinnyjth check it out now

vinnyjth:
@redreceipt Looks good to me! Fix that linter error and we should be good to merge.

PR #1602, opened 2020-10-07
fix: Fixes Android Notifications Onboarding

redreceipt:
## DESCRIPTION

Screen Shot 2020-10-07 at 10 45 16 AM

Android doesn't have the concept of a notifications prompt ([OneSignal SDK](https://documentation.onesignal.com/docs/sdk-reference#ospermissionsubscriptionstate)) so this just bypasses that for Android.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-onboarding/src/slides/AskNotifications/AskNotificationsConnected.js
+++ b/packages/apollos-ui-onboarding/src/slides/AskNotifications/AskNotificationsConnected.js
@@ -33,7 +33,6 @@ const AskNotificationsConnected = memo(
         const ready =
           Platform.OS === 'ios' ? value.hasPrompted : value.hasPushPermission;
 
-        console.log(value);
         return (
           <Component
             isLoading={value.loading}
--- a/packages/apollos-ui-onboarding/src/slides/AskNotifications/AskNotificationsConnected.js
+++ b/packages/apollos-ui-onboarding/src/slides/AskNotifications/AskNotificationsConnected.js
@@ -31,7 +31,9 @@ const AskNotificationsConnected = memo(
         // Android has no concept of push prompt, notifications enabled by default.
         // so we'll just show them it's enabled and allow to proceed
         const ready =
-          Platform.OS === 'ios' ? value.hasPrompted : value.hasPushPermission;
+          Platform.OS === 'ios'
+            ? value.hasPushPermisson || value.hasPrompted
+            : value.hasPushPermission;
 
         return (
           <Component
--- a/packages/apollos-ui-onboarding/src/slides/AskNotifications/AskNotificationsConnected.js
+++ b/packages/apollos-ui-onboarding/src/slides/AskNotifications/AskNotificationsConnected.js
@@ -1,5 +1,4 @@
 import React, { memo } from 'react';
-import { Platform } from 'react-native';
 import PropTypes from 'prop-types';
 import { NotificationsConsumer } from '@apollosproject/ui-notifications';
 
@@ -31,7 +30,7 @@ const AskNotificationsConnected = memo(
         // Android has no concept of push prompt, notifications enabled by default.
         // so we'll just show them it's enabled and allow to proceed
         const ready =
-          Platform.OS === 'ios'
+          value.hasPrompted === undefined
             ? value.hasPushPermission || value.hasPrompted
             : value.hasPushPermission;
 
--- a/packages/apollos-ui-onboarding/src/slides/AskNotifications/AskNotificationsConnected.js
+++ b/packages/apollos-ui-onboarding/src/slides/AskNotifications/AskNotificationsConnected.js
@@ -32,7 +32,7 @@ const AskNotificationsConnected = memo(
         // so we'll just show them it's enabled and allow to proceed
         const ready =
           Platform.OS === 'ios'
-            ? value.hasPushPermisson || value.hasPrompted
+            ? value.hasPushPermission || value.hasPrompted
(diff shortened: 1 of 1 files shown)


packages/apollos-ui-onboarding/src/slides/AskNotifications/AskNotificationsConnected.js
```diff
@@ -25,25 +27,34 @@ const AskNotificationsConnected = memo(
...
-          }
-          pressPrimaryEventName={'Ask Notifications Completed'}
-          pressSecondaryEventName={'Ask Notifications Skipped'}
-          {...props}
-        />
-      )}
+      {(value) => {
+        // Android has no concept of push prompt, notifications enabled by default.
+        // so we'll just show them it's enabled and allow to proceed
+        const ready =
```
vinnyjth, line 33: I'm on board with the logic here so far, but do you think it would be better placed inside the `notifications` `NotificationsConsumer` that's providing these values in the first place?

redreceipt: reason I put it here is because the UI is deciding what to do with the information coming from OneSignal

redreceipt: `ui-notifications` should report on OneSignal raw data, `ui-onboarding` should decide what to do with that information

also, fwiw I think the levels of abstraction in this flow is a little overkill :) but when don't I think that...

vinnyjth: I'm wondering if we put that logic here: https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-ui-notifications/src/permissionUtils.js we could avoid needing to repeat this logic down the road, and can avoid issues where the user has overridden AskNotificationsConnected (which is more likely than overriding the notifications consumer)

redreceipt: if we put it there, it's misleading what the data means downstream, `ready` would mean something different for different platforms. `hasPrompt` and `hasPermission` are very different and I like the UI being able to make the decisions on what to do with that.

vinnyjth: Okay, I'm tracking more now. Stepping back, are those values `undefined` on android or flat out wrong?

redreceipt: if they've overridden that slide, they will likely want to do something different with prompts and notifications than we are. in that case, they'll want the full picture of onesignal data coming from `ui-notifications`

redreceipt: `undefined`

redreceipt:
cc @christinamegura in case you wanna follow along with a logic flow conversation. good programming topics being discussed.

packages/apollos-ui-onboarding/src/slides/AskNotifications/AskNotificationsConnected.js
vinnyjth: That seems wrong to me, I feel like they should have _some_ value (at least from looking at the OneSignal API docs.). Maybe we should look closer there?

redreceipt: only `hasPrompt` is undefined on Android, `hasPermissions` is always correct from what I've seen

vinnyjth: Context for my inquisitiveness: Something about the current approach feels off. I feel like we're masking another problem by checking for the Platform in this part of the code, I'm trying to discern how we can fix this bug and make the code simpler at the same time. Too many code paths right now (and there were too many before this PR)

Another direction that might  sense: Right now we respect the value of `hasPrompted` before we consider if we have `pushPermissions`. ie: if `hasPrompted` is falsy, it doesn't matter if we actually have pushPermission. Could  we refactor this component to correct for that, and so if `pushPermission` is true we let user's continue, even if we have never prompted?

redreceipt: could you describe what you wanna see perhaps with a screenshot or drawing or something? there's a lot of logic flying around

vinnyjth: yah - I'll draw something but I think this is the clearest way written out: 
```
- On screen mount, check to see if push notifications granted.
  - If they have been granted, show 'Continue'

- If push notifications haven't been Granted.
  - Show the skip button.
  - If the user has been prompted, have the button direct the user at settings. 
  - If the user hasn't been prompted, pushing the button should prompt for permission.
```

redreceipt: ok yea I can do that, no drawing necessary

redreceipt:
@vinnyjth how about that

redreceipt:
ok @vinnyjth ready to test

vinnyjth (approved):
Tested a bit in a variety of conditions (including toggling push notifications off), things are looking good 👍

PR #1608, opened 2020-10-12
chore: Simplifies Actions Schema

redreceipt:
This attempts to simplify the naming and purpose of various types related to feature actions.


redreceipt:
@vinnyjth made a revision to this, same thing in principle but I think "Action" > "Link" really clears up the purpose of this section of schema.

I also removed `primaryAction` completely because I think it's throwing us for a loop in other discussions for good reason and there's probably a better way to achieve this functionality from a step higher. Obviously up for discussion but I almost think the "primary action" hook of a feature could be entirely driven off the feature type. Most of the time we want some form of "Read more".

packages/apollos-data-schema/src/index.js
```diff
@@ -830,14 +834,39 @@ export const featuresSchema = gql`
     action: ACTION_FEATURE_ACTION
   }
 
+  interface Link {
+    title: String
+    subtitle: String
+    image: ImageMedia
+  }
+
+  type NodeLink implements Action & Node {
```
redreceipt, line 843:
```suggestion
  type NodeLink implements Link & Node {
```

packages/apollos-data-schema/src/index.js
```diff
@@ -830,14 +834,39 @@ export const featuresSchema = gql`
...
+    id: ID!
+    title: String
+    subtitle: String
+    image: ImageMedia
+
+    node: Node
+  }
+
+  type UrlLink implements Action & Node {
```
redreceipt, line 852:
```suggestion
  type UrlLink implements Link & Node {
```

packages/apollos-data-schema/src/index.js
```diff
@@ -830,14 +834,39 @@ export const featuresSchema = gql`
...
   type ActionListFeature implements Feature & Node {
     id: ID!
     order: Int
 
     title: String
     subtitle: String
-    actions: [ActionListAction]
-    primaryAction: FeatureAction
+    actions: [ActionListAction] @deprecated(reason: "Use links")
+    primaryAction: FeatureAction @deprecated(reason: "Not supported")
```
redreceipt, line 868: this is going to be "view all" of whatever category the other links are

packages/apollos-data-schema/src/index.js
```diff
@@ -846,23 +875,26 @@ export const featuresSchema = gql`
 
     title: String
     subtitle: String
-    actions: [ActionListAction]
     heroCard: CardListItem
-    primaryAction: FeatureAction
+    actions: [ActionListAction] @deprecated(reason: "Use links")
+    primaryAction: FeatureAction @deprecated(reason: "Not supported")
```
redreceipt, line 880: this is going to be a click-through for the hero content

PR #1610, opened 2020-10-12
feat: Adds Discover Feed Query

redreceipt:
This PR breaks out the `userFeedFeatures` query into `homeFeedFeatures` and `discoverFeedFeatures` and defines a new way to deterministically generate a feed of feeds. Look through the schema for details. Test with `feature-feed` branch in templates.

Screen Shot 2020-10-13 at 1 19 38 PM

### DIFF
--- a/packages/apollos-data-connector-rock/src/feature-feeds/index.js
+++ b/packages/apollos-data-connector-rock/src/feature-feeds/index.js
@@ -1,0 +1,35 @@
+import RockApolloDataSource from '@apollosproject/rock-apollo-data-source';
+import { createGlobalId, parseGlobalId } from '@apollosproject/server-core';
+import ApollosConfig from '@apollosproject/config';
+
+const resolver = {
+  FeatureFeed: {
+    // lazy-loaded
+    features: ({ getFeatures }) => getFeatures(),
+  },
+};
+
+class FeatureFeed extends RockApolloDataSource {
+  getFromId = (id) => {
+    const {
+      id: { type, args },
+    } = parseGlobalId(id);
+    return this.getFeed({ type, args });
+  };
+
+  getFeed = ({ type = '', args = {} }) => {
+    let config = [];
+    if (type === 'apollosConfig' && args.section === 'home')
+      config = ApollosConfig.HOME_FEATURES || [];
+    else if (type === 'apollosConfig' && args.section === 'discover')
+      config = ApollosConfig.DISCOVER_FEATURES || [];
+
+    return {
+      __typename: 'FeatureFeed',
+      id: createGlobalId({ type, args }, 'FeatureFeed'),
+      getFeatures: () => this.getFeatures(config),
+    };
+  };
+}
+
+export default { resolver, dataSource: FeatureFeed };
--- a/packages/apollos-data-connector-rock/src/feature-feeds/index.js
+++ b/packages/apollos-data-connector-rock/src/feature-feeds/index.js
@@ -31,6 +31,8 @@ class FeatureFeed extends RockApolloDataSource {
   };
 
   getFeed = ({ type = '', args = {} }) => {
+    const { Feature } = this.context.dataSources;
+
     let config = [];
     if (type === 'apollosConfig' && args.section === 'home')
       config = ApollosConfig.HOME_FEATURES || [];
@@ -40,7 +42,7 @@ class FeatureFeed extends RockApolloDataSource {
     return {
       __typename: 'FeatureFeed',
       id: createGlobalId({ type, args }, 'FeatureFeed'),
-      getFeatures: () => this.getFeatures(config),
+      getFeatures: () => Feature.getFeatures(config),
     };
   };
 }
--- a/packages/apollos-data-connector-rock/src/feature-feeds/index.js
+++ b/packages/apollos-data-connector-rock/src/feature-feeds/index.js
@@ -24,10 +24,7 @@ const resolver = {
 
 class FeatureFeed extends RockApolloDataSource {
   getFromId = (id) => {
-    const {
-      id: { type, args },
-    } = parseGlobalId(id);
-    return this.getFeed({ type, args });
+    return this.getFeed(JSON.parse(id));
   };
 
   getFeed = ({ type = '', args = {} }) => {
@@ -41,10 +38,12 @@ class FeatureFeed extends RockApolloDataSource {
 
     return {
       __typename: 'FeatureFeed',
-      id: createGlobalId({ type, args }, 'FeatureFeed'),
+      id: createGlobalId(JSON.stringify({ type, args }), 'FeatureFeed'),
+      // Defer parsing of feature feed if not requested in gql.
+      // Useful if the config comes from the network.
       getFeatures: () => Feature.getFeatures(config),
     };
   };
 }
 
-export default { resolver, dataSource: FeatureFeed };
+export { resolver, FeatureFeed as dataSource };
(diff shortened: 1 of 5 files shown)


redreceipt:
@vinnyjth I think there's some overlap in the task list but it made sense for me to combine these. Is this the approach you were thinking?

vinnyjth:
No worries about the overlap!

vinnyjth:
@redreceipt as a heads up - I'm branching off of this PR for some of my work. No worries if you change things around, I'll catch those changes in my PR.

packages/apollos-data-connector-rock/src/features/data-source.js
```diff
@@ -450,9 +450,12 @@ Make sure you structure your algorithm entry as \`{ type: 'CONTENT_CHANNEL', aru
       .join('\n\n');
   }
 
-  async getHomeFeedFeatures() {
+  // deprecated
+  getHomeFeedFeatures = this.getFeedFeatures.bind(this);
+
+  async getFeedFeatures(featuresConfig = []) {
```
redreceipt, line 456: @vinnyjth this is technically a breaking change isn't it? I'm now passing in the config. Only for custom overrides though

vinnyjth: I don't think it is. Overrides of this method will still work, but won't have the new functionality.

packages/apollos-data-connector-rock/src/features/resolver.js
```diff
@@ -47,7 +48,12 @@ export default {
   },
   Query: {
     userFeedFeatures: async (root, args, { dataSources: { Feature } }) =>
-      Feature.getHomeFeedFeatures(),
+      console.warn('getHomeFeedFeatures is deprecated. Use getFeedFeatures.') ||
+      Feature.getHomeFeedFeatures(get(ApollosConfig, 'HOME_FEATURES', [])),
+    homeFeedFeatures: async (root, args, { dataSources: { Feature } }) =>
+      Feature.getFeedFeatures(get(ApollosConfig, 'HOME_FEATURES', [])),
```
vinnyjth, line 54: Something to keep in mind as you work on this PR @redreceipt 

I _think_ the return of these resolvers probably needs to be a `FeatureFeed` (right now it's still an array). 

This FeatureFeed should have a few characteristics. It should have an `id` (which can probably be something like `{ source: 'config', type: 'HomeFeatures' }` and it should have a `featureFeed` property, which is the result of `Feature.getFeedFeature`

It's important that the `featureFeed` property is "lazy loaded", that it is only called if it's actually requested via graphql. That's easy to accomplish by wrapping it inside a function :) See: https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-data-connector-rock/src/features/data-source.js#L68

vinnyjth: I have some ideas on how to implement this, but I trust you also have ideas. Let me know if you want to collaborate though :)

packages/apollos-data-connector-rock/src/features/data-source.js
redreceipt: If a church is counting on `getHomeFeedFeatures` to automatically pull from the config, they will be surprised when it doesn't. Is this something we should take into account? Or are you not concerned that anyone has done that?

vinnyjth: I think people have overridden `getHomeFeedFeatures`, but I don't think anyone is calling `getHomeFeedFeatures` themselves. If you are worried about it, you can pass the correct apollos config key into the method here on line 454 instead of (or in addition to) in the resolver and it would be safer.

vinnyjth:
🙌

redreceipt:
> and `args.section` would become `HOME_FEATURES`

love it

packages/apollos-data-connector-rock/src/feature-feeds/index.js
```diff
@@ -24,10 +24,7 @@ const resolver = {
 
 class FeatureFeed extends RockApolloDataSource {
   getFromId = (id) => {
-    const {
-      id: { type, args },
-    } = parseGlobalId(id);
-    return this.getFeed({ type, args });
+    return this.getFeed(JSON.parse(id));
```
redreceipt, line 27: is this all `parseGlobalId` does?

packages/apollos-data-connector-rock/src/feature-feeds/index.js
```diff
@@ -41,10 +38,12 @@ class FeatureFeed extends RockApolloDataSource {
 
     return {
       __typename: 'FeatureFeed',
-      id: createGlobalId({ type, args }, 'FeatureFeed'),
+      id: createGlobalId(JSON.stringify({ type, args }), 'FeatureFeed'),
```
redreceipt, line 41: why is this step necessary?

packages/apollos-data-connector-rock/src/feature-feeds/index.js
vinnyjth: ID is already parsed at this point (see: https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-server-core/src/node/model.js#L37) 

We do need to turn it from a string back into an object.

packages/apollos-data-connector-rock/src/feature-feeds/index.js
redreceipt: haha oh

vinnyjth: yah I've been bit by that a few times "why is every id the same? (decodes id) oh..... "

vinnyjth:
> > and `args.section` would become `HOME_FEATURES`
> 
> love it

I'll knock it out real quick

packages/apollos-data-connector-rock/src/feature-feeds/index.js
redreceipt: ah

vinnyjth:
@redreceipt Let me know if you care for my changes. I'll approve this and we can merge after that.

packages/apollos-data-connector-rock/src/feature-feeds/index.js
```diff
@@ -5,15 +5,18 @@ import ApollosConfig from '@apollosproject/config';
...
+      }),
     discoverFeedFeatures: async (
       root,
       args,
       { dataSources: { FeatureFeed } }
     ) =>
       FeatureFeed.getFeed({
         type: 'apollosConfig',
-        args: { section: 'discover' },
+        args: { section: 'DISCOVER_FEED' },
```
redreceipt, line 19:
```suggestion
        args: { section: 'DISCOVER_FEATURES' },
```

packages/apollos-data-connector-rock/src/feature-feeds/index.js
```diff
@@ -16,7 +16,7 @@ const resolver = {
     ) =>
       FeatureFeed.getFeed({
         type: 'apollosConfig',
-        args: { section: 'DISCOVER_FEED' },
+        args: { section: 'DISCOVER_FEATURES' },
```
vinnyjth, line 19: whoops, thanks

redreceipt:
> @redreceipt Let me know if you care for my changes. I'll approve this and we can merge after that.

changed to "DISCOVER_FEATURES". looks good. merge away!

vinnyjth (approved):

vinnyjth:
Adding a basecamp todo to write unit tests for this :)

PR #1614, opened 2020-10-14
feat: add date to content items

christinamegura:
This PR adds `publishDate` to the content item schema so that we have the ability to show dates on content items in the future.

### How do I test this PR?
Start the server by running `TZ=UTC yarn start`

In graphql, query to find a content item id
Screen Shot 2020-11-18 at 10 09 58 AM

Use that content item id to query for `publishDate` and 
1. Confirm that the queried `publishDate` returns the same date as running `new Date ("publishDate")` in the console &
2. Confirm `publishDate` is a UTC string from the API (four hours different from the string returned in Rock, as time stamp is assumed Eastern).

Screen Shot 2020-11-24 at 10 41 54 AM

Screen Shot 2020-11-24 at 10 50 47 AM

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-data-schema/src/index.js
```diff
@@ -7,6 +7,7 @@ import {
 export const interfacesSchema = gql`
   interface ContentNode {
     title(hyphenated: Boolean): String
+    publishDate: String
```
christinamegura, line 10: @vinnyjth what is the difference between the `ContentNode` interface and the `ContentItem` interface?

christinamegura (commented):
@vinnyjth tests are passing locally but I'm seeing jest spit out a heuristic fragment matcher warning. Can you point me in the right direction to address this?

packages/apollos-data-schema/src/index.js
vinnyjth: `ContentNode` is the new interface that describes a `Node` that is "viewable" as a piece of "content". Pretty much means that it's something you can tap on and open and read, and it normally has an image, title ,and a bit of text.  

The `publishDate` probably doesn't belong on the `ContentNode`, but it does belong on the `ContentItem`. The `ContentItem` specifically describes `ContentItems` from Rock (or their equiv in other CMSes)

packages/apollos-data-connector-rock/src/content-items/__tests__/resolvers.tests.js
```diff
@@ -113,6 +113,7 @@ const contentItemFragment = `
     id
     __typename
     title
+    publishDate
```
vinnyjth, line 116: Awesome way to get more test coverage! Nice work.

packages/apollos-data-connector-rock/src/content-items/resolver.js
```diff
@@ -53,6 +54,12 @@ export const defaultContentItemResolvers = {
   coverImage: (root, args, { dataSources: { ContentItem } }) =>
     ContentItem.getCoverImage(root),
 
+  publishDate: ({ startDateTime }) =>
+    moment(startDateTime)
+      .tz(ROCK.TIMEZONE)
+      .format()
+      .split(/[-+]\d+:\d+/)[0],
```
vinnyjth, line 61: Okay - this might be a bit funky so please let me know if it doesn't make any sense. 

1. Dates come back from Rock with no timezone attached, but they are in Eastern. 
2. When we parse dates with Moment, they default to UTC
3. If you call `.tz()` it changes the timezone to `Eastern`, moving the time 4 or 5 hours in one direction. 
4. Calling `format()` and then `split()` pops the timezone data back off the string. 

There's a reason we use the dance in some parts of the codebase, and I think it has to do with filtering with UTC dates, or sending dates from Rock back into Rock. 

All that to say - if you look at the API right now it's returning the exact same data that's in Rock. 

image

This won't be a good long term format for this field b/c devices in different time zones (or the server itself) will interpret that date as existing in it's own timezone. So Devices parsing that date in Central, will display the wrong time (off by an hour). If a server method was to try to use that `publishDate` dataSource method, it will parse that time as in UTC (and the time would be 4 hours off)

To fix this, you'll need to make sure two things are true. 

1. The string returned from the API has a timezone attached (look for a Z or a +/-) (`2018-05-08T00:00:00.000Z`)
2. The string returned from the API is about four hours different from the string returned in Rock. That's because the API should be returning UTC strings. 

I would investigate the `moment-timezone` `moment.tz(date, timezone)` methods to start :)

vinnyjth (changes requested):

packages/apollos-data-connector-rock/src/content-items/resolver.js
christinamegura: Screen Shot 2020-11-23 at 10 47 34 AM

vinnyjth:
@christinamegura You're getting closer! 

There's one important tip I forgot to mention - when you run the server locally it runs in Eastern Timezone. When the server runs in the cloud, it's going to run in UTC. That's a frustrating fact of `node`. We can get around that by starting the server like this: `TZ=UTC yarn start` We should probably make that a default, sorry for not mentioning this earlier. 


So running the server in UTC, here's what I currently get 
Screen Shot 2020-11-24 at 8 43 49 AM

It _looks_ correct, but in that screenshot I've pasted the date into the browser's `new Date()` function and you can see I end up with the 7th instead of the 8th, since my browser converts the UTC date into the equivalent eastern time, which is four hours (and thus a day) earlier. 

Going back to the requirements above,  you have a timezone attached do your date which is awesome! Now we just need that four hour offset. You had that with the code above, but it will only work if the server runs in the same time zone as Rock. The TZ=UTC should make it a little easier to test :) 

Thanks for your hard work @christinamegura! This timezone and date stuff is really frustrating, and very difficult to reason about and get right. I spent a lot of yesterday scratching my head and making sure I was thinking things through properly.

vinnyjth:
@christinamegura As for the tests, you'll probably need to either remove that `publishedAt` date from the `ui-fragments` package, or update all the tests that reference that fragment so their mocks include that field. You can probably remove the field from the fragment for now, since we aren't using it on the front end yet.

christinamegura:
@vinnyjth I think I finally got it! I'm not sure why my PR title is failing though 😬

vinnyjth (approved):
Nice work @christinamegura  :D

packages/apollos-data-schema/src/index.js
```diff
@@ -381,6 +381,7 @@ export const contentItemSchema = gql`
   interface ContentItem {
     id: ID!
     title(hyphenated: Boolean): String
+    publishDate: String
```
conrad-vanl, line 384: @vinnyjth should we really use `String` here or should we make our own type? I'm thinking to enforce a certain convention or something

vinnyjth: I have gone back and forth on this. Using `Date` is probably more correct, but I've seen more issues and problems arise from using a custom `scalar` and the `Date` field than I have seen it help.

vinnyjth: @conrad-vanl Adding to that, we are already using `Strings` for the date we have in our Schema, so I'm fine with `String` going forward.

PR #1618, opened 2020-10-15
feat: Adds Content Feature Feeds

redreceipt:
Screen Shot 2020-10-15 at 4 28 48 PM

partner branch: https://github.com/ApollosProject/apollos-templates/pull/200


redreceipt:
ok so @vinnyjth I'm sure you had some scope requirements in mind but technically it works as is, what am I missing?

vinnyjth:
We should be using the `FeatureFeedConnected`, not making a new feed for the content features :) 

That should work out of the box, but we need to update the `FeatureFeedConnected/FeatureFeed` mapper to include the `Text` and `Scripture` features.

redreceipt:
ah, so is it safe to assume we can deprecate `ContentSingleFeaturesConnected` and `NodeFeaturesConnected`? This library is getting very confusing...

vinnyjth:
Yes :)

packages/apollos-ui-connected/src/FeaturesFeedConnected/FeaturesFeedConnected.js
```diff
@@ -114,11 +114,12 @@ class FeaturesFeedConnected extends PureComponent {
       >
         {({ error, data, loading, refetch }) => {
           const features = get(data, 'node.features', []);
+          const featureFeed = data.node?.featureFeed?.features || [];
```
vinnyjth, line 117: Rather than doing this, I there's a simpler solution: 

1. On the API, make sure that `contentItem.featureFeed.id` returns a valid id (contains the type contentItem and the args { id }(
2. On the front end, add the featureFeed.id to the content item query (or have ContentSingleFeatures connected fetch it, and then we can still make a place for the ContentSingleFeatures) 
3. Pass that featureFeed.id into the <FeaturesFeedComponent>
4. No need to make any other changes, besides Adding features the featuresFeedComponentMapper. Each component can be hyperfocused on their purpose and not have to worry about pulling data from potentially different parts of a quer.y

redreceipt:
@vinnyjth ok I think it's in a good spot

packages/apollos-ui-connected/src/FeaturesFeedConnected/FeaturesFeedConnected.js
```diff
@@ -93,9 +82,14 @@ class FeaturesFeedConnected extends PureComponent {
   };
 
   render() {
-    const { Component, onPressActionItem, nodeId, ...props } = this.props;
-    // Early return if we don't have a nodeId.
-    if (!nodeId) {
+    const {
+      Component,
+      onPressActionItem,
+      featureFeedId,
```
vinnyjth, line 88: Nice naming change. It doesn't match our other connected components, but I like the clarity.

packages/apollos-ui-connected/src/FeaturesFeedConnected/featuresFeedComponentMapper.js
```diff
@@ -16,6 +19,9 @@ const MAPPINGS = {
   // like Prayer, don't all get stuck here, forcing all church apps to have
   // them all installed?
   PrayerListFeature: PrayerListFeatureConnected,
+  ScriptureFeature,
```
vinnyjth, line 22: 💥

packages/apollos-ui-connected/src/FeaturesFeedConnected/__snapshots__/FeaturesFeedConnected.tests.js.snap
```diff
@@ -0,0 +1,341 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`FeaturesFeedConnected should render 1`] = `
```
vinnyjth, line 3: Hm, the snap doesn't match for some reason 🤔

redreceipt:
Oh good catch, not sure what happened...

packages/apollos-ui-connected/src/NodeFeaturesConnected/NodeFeaturesConnected.js
```diff
@@ -24,7 +24,11 @@ const NodeFeaturesConnected = ({ Component, nodeId, ...props }) => {
         const features = get(node, 'features', []);
         if (!features || !features.length) return null;
 
-        return <Component nodeId={nodeId} features={features} {...props} />;
+        return (
```
vinnyjth, line 27: This Component I think we can keep actually. In another PR, I'll refactor this so that it fetches a node's featureFeed, and then renders FeatureFeedConnected. Kinda weird, but it let's us go from a nodeId -> the featureFeed -> the features

vinnyjth:
Yah, it looks fine from here. I’ll merge it as-is and tackle it eventually. A half working test is better than none.

vinnyjth (approved):
Very cool to have this. Should make explaining Content Features and Feature Feeds way easier (because they are now the same thing!)

PR #1620, opened 2020-10-16
chore: Reorganizes Algorithms

redreceipt:
This moves action algorithms into its own data source, it was starting to get unruly


redreceipt:
@vinnyjth this what you were thinking?

vinnyjth:
@redreceipt yah, looks great so far :D 

I'll throw two additional requirements your way which should fit within what you have already done: 

1. Inside those action algorithms, `this` should refer to the Feature `dataSource` OR there should be a good way to reference the `Feature dataSource`. Lines like this https://github.com/ApollosProject/apollos-apps/pull/1620/files#diff-237992155337966aa17aca015273928af7972013592015790968383023ddb434R40 aren't going to work otherwise. 

2. Backwards compatible in this case means that algorithms that a user has added by extending the `Feature` dataSource. We should probably check both dataSources for feature algorithms so users have some middle ground before they have to move them all over into the structure.

packages/apollos-data-connector-rock/src/action-algorithms/index.js
```diff
@@ -0,0 +1,229 @@
...
+                'Please move action algorithms from Feature to ActionAlgorithm data source.'
+              );
+              return Feature.ACTION_ALGORITHIMS[algorithm.type](
+                algorithm.arguments
+              );
+            }
+
+            return this.ACTION_ALGORITHMS[algorithm.type](algorithm.arguments);
+          }
+          return this.ACTION_ALGORITHMS[algorithm]();
```
vinnyjth, line 45: I think you need a test right here for non-object algorithms that are found on the `Feature.ACTION_ALGORITHMS` object. 

Not sure if this is obvious, but the `typeof algorithm === 'object'` is there so we can refer to algorithms by name alone in the `config.yml` or the more explicit `{ type: ...., arguments:.... }`

redreceipt: done

redreceipt:
@vinnyjth trying to make tests work in the simplest way possible but the features tests are written differently than most libs, is this close to working?

Screen Shot 2020-10-19 at 9 54 32 AM

vinnyjth:
@redreceipt yah, I think that's exactly what I would try and do

vinnyjth:
> @vinnyjth trying to make tests work in the simplest way possible but the features tests are written differently than most libs, is this close to working?

You might need to set the `context` of the `new FeatureAlgo` to be the context you're creating (so it has access to all those mocks)

vinnyjth (approved):
Nice work! This will make a lot of customization on client apps so much more compartmentalized, which I appreciate.

packages/apollos-data-connector-rock/src/action-algorithms/index.js
```diff
@@ -4,7 +4,7 @@ import ApollosConfig from '@apollosproject/config';
 
 class ActionAlgorithm extends RockApolloDataSource {
   // Names of Action Algoritms mapping to the functions that create the actions.
-  ACTION_ALGORITHMS = Object.entries({
+  ACTION_ALGORITHIMS = Object.entries({
```
redreceipt, line 7: you added back in the spelling error here?

vinnyjth: I was getting confused with the multiples spellings in the file, and one of them was wrong (I was getting a `cannot read property ACTION_ALGORITHIMS of undefined error`). I'm fine if you revert! I couldn't figure out which one was technically correct so I went back and made them all what we had before

vinnyjth: Looking at your last commit, I think the second reference to the misspelling was accidentally spelled correctly, if that makes sense.

redreceipt:
yea I don't know the easy way to make these tests work. I don't know enough about the underlying test tooling we have available

vinnyjth:
@redreceipt I'll take a stab at it

PR #1630, opened 2020-10-23
fix: Fixes Actions with Defined related node

redreceipt:
If a related node was defined already, the function was stripping it off. this fixes that

### DIFF
--- a/packages/apollos-data-connector-rock/src/features/data-source.js
+++ b/packages/apollos-data-connector-rock/src/features/data-source.js
@@ -27,6 +27,8 @@ export default class Feature extends RockApolloDataSource {
 
   // eslint-disable-next-line class-methods-use-this
   attachRelatedNodeId({ relatedNode, ...action } = {}) {
+    // if we already have an ID, return as is
+    if (relatedNode?.id) return { relatedNode, ...action };
     if (relatedNode && !relatedNode.id) {
       return {
         ...action,


packages/apollos-data-connector-rock/src/features/data-source.js
```diff
@@ -27,6 +27,8 @@ export default class Feature extends RockApolloDataSource {
 
   // eslint-disable-next-line class-methods-use-this
   attachRelatedNodeId({ relatedNode, ...action } = {}) {
+    // if we already have an ID, return as is
+    if (relatedNode?.id) return { relatedNode, ...action };
```
vinnyjth, line 31: Great catch :D I'll suggest one change though: We currently have three different `return` statements, and I think two of them can be combined. 

On line 45, could we return `{ relatedNode, ...action }` instead of the return here and also squash this bug?

redreceipt: totally, I don't fully understand all scenarios you have in mind for this one. this would mean that if `relatedNode` is `null` or `undefined`, it's gonna make it through. good with that?

vinnyjth: Yah, that should be totally fine :)

vinnyjth:
@redreceipt Looking at the snaps, it looks like your changes are generating the correct data while the committed snaps demonstrate the bug. Nice work.

vinnyjth (approved):

PR #1631, opened 2020-10-26
chore: Removes Redundant Fragment

redreceipt:
Not sure of a good way to warn about the deprecation...


packages/apollos-ui-connected/src/HeroListFeatureConnected/HeroListFeature/index.js
```diff
@@ -174,7 +174,9 @@ const HeroListFeature = memo(
           }
           actions={isLoading && !actions.length ? loadingStateObject : actions}
           onPressActionItem={onPressItem}
-          onPressActionListButton={() => onPressActionListButton(primaryAction)}
+          onPressActionListButton={() =>
+            console.log(primaryAction) || onPressActionListButton(primaryAction)
```
vinnyjth, line 178:
```suggestion
            onPressActionListButton(primaryAction)
```

packages/apollos-ui-fragments/src/features.js
```diff
@@ -33,38 +33,6 @@ const PRAYER_LIST_FEATURE_FRAGMENT = gql`
...
-      title
-      subtitle
-      isCard
-    }
-    ...TextFeatureFragment
-    ...ScriptureFeatureFragment
-    ...WebviewFeatureFragment
-  }
-`;
-
```
vinnyjth, line 67: I think we do actually want these fragments. The reasoning as follows: 

1. We want to show the user loading data as quickly as possible. 
2. We can't show them generic loading data because it won't match the actual feed.
3. Thus, in the feed query we load in a super slim subset of the `Feature` data . All of the fields fetched are defined in the config and don't require a rock/other call, so they load almost instantly and all fields are catchable. 
4. Then, each feature goes and fetches it's own data. Some of these are catchable, so super fast, while some aren't catchable, so their performance hit is limited to just their data. 

For those reasons, I think we do actually need both sets of fragments (a feed fragment, and fragments for each feature's full set of data).

redreceipt: ah. very clever. what if we just queried IDs and call it `FEATURES_LITE_FRAGMENT` or something to speak to its purpose?

vinnyjth: I think `titles` and `subtitles` are nice because they are free and enable a progressive loading state. We could look at making a fragment that captures those fields though :) 

I'm all for renaming this query. maybe INTIAL_FEATURE_FEED_FRAGMENT ?

redreceipt: what about `LOADING_FEATURES_FRAGMENT`

vinnyjth: `LOADING` seems confusing here, especially since we'll see `loading` scattered around those components. What do you think about language that describes intent here, rather than what it's used for? that's why I like `INITIAL` but I'm not attached to it.

redreceipt: I like it better than doing nothing. this file is just getting a bit confusing. I'll make the change.

redreceipt:
@vinnyjth I'm just gonna close this. not sold that really any change I make will make it better. I understand the purpose now, may revisit later

packages/apollos-ui-fragments/src/features.js
```diff
@@ -283,13 +294,13 @@ export {
   TEXT_FEATURE_FRAGMENT,
   SCRIPTURE_FEATURE_FRAGMENT,
   CARD_FEATURES_FRAGMENT,
-  FEATURES_FRAGMENT,
```
vinnyjth, line 286: We should probably keep this export for backwards compat.

redreceipt: yea...I don't think anyone uses it directly but I agree. There's just no way to warn about the deprecation so I guess with 2.0 we just remove it.

vinnyjth: Yah :/ I did a search in The Porch and found a usage of it, but it's only because The Porch needs to refactor some custom code to go back inline with core.

redreceipt: ah

vinnyjth:
Tested, seems to work great 👍

vinnyjth (approved):

PR #1632, opened 2020-10-26
feat: CampusID support in home feed

vinnyjth:
## DESCRIPTION

Time for this to belong in core. 

This code adds the support for passing in a campusId to the default `getHomeFeed` query. Or not, you don't have to pass an ID. It's optional. 

This PR enables this functionality by passing `arguments` (a second param to getFeatures) all the way down to the algorithms. These args are also included in the `IDs` for the feed, and for each of the features. This functionality is used for `campusId`, but it could also be used for similar contextual data that's relevant to some or all features in a feed. 

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-data-connector-rock/src/action-algorithms/index.js
```diff
@@ -230,6 +239,30 @@ Make sure you structure your algorithm entry as \`{ type: 'CONTENT_CHANNEL', aru
       summary: ContentItem.createSummary(item),
     }));
   }
+
+  async currentCampusAlgorithm({ campusId }) {
```
vinnyjth, line 243: This is here more for demonstration purposes than anything else.

vinnyjth:
https://github.com/ApollosProject/apollos-templates/pull/206

packages/apollos-data-connector-rock/src/action-algorithms/index.js
redreceipt: if no one is going to use this algorithm, I'd almost rather leave this out instead of having dead code in there. I don't think anyone loves the state of features/actions/args code right now anyway so introducing more that we'd have to refactor later seems like a bad idea.

I do think this PR solves a valid problem but I'm almost wondering if we just do it via context as we did with other apps? that way new campus stuff will just work out of the box and will work for content features as well

vinnyjth: The issue with storing in context is that it's really easy to accidentally cause issues down the road if you're not careful. Context is a temporal global variable, and it's discouraged to store data there. We would run into issues with it down the road if we ever turned on features like query batching or similar. 

I think this approach will be pretty easy to implement for pretty much all our client apps, and I think the passing of `args` is flexible enough to solve other similar issues we haven't run into yet :)

redreceipt: I don't disagree, the whole thing just seems short sighted to me. Like we're trying to MVP something before it's ready.

That being said `args` is probably the best path forward on our current trajectory so I'll test and approve. You still feel strongly about leaving this current campus algorithm in?

vinnyjth: > That being said args is probably the best path forward on our current trajectory so I'll test and approve. You still feel strongly about leaving this current campus algorithm in?

Not at all, rip it out after you test :)

vinnyjth: Oh, another critique of the context approach is that it's very difficult to debug if you didn't write it. It just shows up in the context, and if you're trying to work backwards from a value, or forwards from the value is set in context, it's easy to lose the trail (it kinda disappears and then reappears, since it's shared through that global state). While I don't think `args` is amazing, and it's a bit more code, I really like that it leaves a very solid paper trail through the function calls.

redreceipt: agreed

redreceipt:
@vinnyjth is there a reason you opted to not do arguments like we already do them?

```
   algorithms:
      - type: USER_FEED
        arguments:
          campusId: 6
```

no gql schema changes that way

vinnyjth:
@vinnyjth is there a reason you opted to not do arguments like we already do them?

not sure they are applicable. 

arguments in this part of the config.yml are static and algorithm specific. 

arguments like campusId are more broad (should change the feed itself) and dynamic per user/request.

redreceipt:
tested. i'm gonna strip out campus algo and approve

redreceipt (approved):

PR #1635, opened 2020-10-27
chore: Simplflies Root Package

redreceipt:
Removes `ci-test` command and removes some unnecessary root level packages


redreceipt:
wait I forgot about tests needing apollo...

redreceipt:
not exactly sure why snaps are failing here. `react-apollo` is a dep in each package

PR #1638, opened 2020-10-28
fix: Exports Correct Feed Features Query

redreceipt:
Forgot to export this last time 🤦‍♂️


vinnyjth (approved):

vinnyjth:
I always forget this export also.

redreceipt:
the only reason we need it (from what I can tell) is for test mocks. technically we could just mock the whole `FeaturesFeedConnected` from the dependency, no point in testing things twice...

PR #1639, opened 2020-11-03
feat: UI Media Player rewrite

conrad-vanl:
## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


conrad-vanl:
Alright @vinnyjth - here's what I'd like your advice on.

The media player in master requires you to render a `<Player>` component that wraps your entire app, and effectively does something like this:

```
<PlayerContainer>
  <App />
  <PlayerView />
</PlayerContainer>
```

This rewrite effectively moves the `<PlayerContainer />` component to wrapping just an individual screen. This assures that on both Android and iOS we are able to effectively "overlay" the screen with the video player in both minimized and fullscreen modes. It ends up looking something like this:

```
<PlayerContainer>
  <PlayerView />
  <ScrollView>
     {children}
  </ScrollView>
</PlayerContainer>
```

Now, the hard part here is that the Player is currently rendering a ScrollView internally. Technically, I could probably rewrite this component _not_ to do that, however I have future design plans for having scroll events adjust the video layout, and so I think it might make sense to go ahead and move the scrollview out of the `-templates` repo and into `-apps`.

Also, `<PlayerContainer/>` needs to be rendered very specifically for fullscreen mode to work. Fullscreen mode could break if you were to try and wrap a `<ScrollView />` around the player, or rendered things above or below the player, so I think we'd be making the component less error prone by forcing a pretty specific convention. 

So... I'd like your advice on how we update `ContentNodeConnected` to support the new video component.

Essentially, there are probably two versions of `ContentNode`s now - if the `ContentNode` has media, then it should look like this:

```
# ContentNode with Media
<PlayerContainer {...mediaProps}>
  <PaddedView>
    <H2>{title}</H2>
    <HtmlComponent>{htmlContent}</HtmlComponent>
    <ContentFeatures />
  </PaddedView>
</PlayerContainer>
```

And without media...

```
# ContentNode sans Media
<StretchyView>
  <ScrollView>
    <Stretchy><ImageHeader></Stretchy>
    <PaddedView>
      <H2>{title}</H2>
      <HtmlComponent>{htmlContent}</HtmlComponent>
      <ContentFeatures />
    </PaddedView>
  </ScrollView>
</StretchyView>
```

I have ideas for a few options...we talked about one option over slack of just having two ContentNode templates we switch between based on the existence of content media.

Another idea I had is I could tweak the `<PlayerContainer />` component to have a similar API as `<Stretchy />` where it takes a function as the children prop, with `scrollViewProps` and such that you pass to your `<ScrollView />`. This might allow us to keep `-templates` closer to how it is today, except it would add a layer of complexity I'm not loving.

For a visual aid, here's what the expected end states are (or close representations):

Screen Shot 2020-11-04 at 3 37 26 PM

conrad-vanl:
Another idea would be to bring the ScrollView and such into the `ContentNodeConnected` template. I know you were originally against that idea....

conrad-vanl:
On my first pass through I missed the `NodeSingleConnected` component and paradigm. I think I can interject the video functionality there. Let me take a pass at that.

conrad-vanl:
@vinnyjth check out changes in this commit - https://github.com/ApollosProject/apollos-apps/pull/1639/commits/05a1890f98d58680e99962e695e34eda65ebb15b - obviously we'd want to organize this better, split into multiple files or whatever, but this would atleast work - and should be 100% backwards compatible if you render `ContentNodeConnected` directly since it only touches `NodeConnected`

redreceipt:
@conrad-vanl trying to make heads or tails of this PR to help move it forward. I've had a couple of questions along the way as I've gone through and figured the best place to put them would be here. I'll take it one at a time. 

Which of the `presentations` did you say we were not using and were safe to delete?

redreceipt:
Is there a reason you didn't scope it `@apollosproject/react-native-apollos-player`?

packages/react-native-apollos-player/src/Container.tsx
```diff
@@ -0,0 +1,163 @@
...
+  );
+  const [
+    isControlVisibilityLocked,
+    setIsControlVisibilityLocked,
+  ] = React.useState<boolean>(false);
+  const [progressHandlers, setProgressHandlers] = React.useState<
+    Array<(props: IProgressProp) => void>
+  >([]);
+
+  const reset = React.useMemo(
```
vinnyjth, line 53: @conrad-vanl Sell me on this :) I'm not convinced it's actually an optimization. Since creating a new function isn't an expensive calculation, I think it's probably a draw which is more performant (since now React has to keep track of change deps to the memo, etc, etc) 

Is it enforce a static value and avoid potential re-renders in the context below?

vinnyjth: Hm, maybe `useCallback` would be better here?

conrad-vanl: Yeah, since these values are passed into context state objects, I err'd on the side of `useMemo` and keeping things pure, since `useContext` and that state gets passed around a lot.

vinnyjth (approved):

PR #1656, opened 2020-12-01
chore: Next > Master

redreceipt:
***Don't Squash!***


redreceipt:
Ok so what _I think_ should happen here is when it's merged in (not squashed) the tags should follow, preserving the canary commit number. Then when we're ready to push a beta, the changelogs will take into account the "!" commits for breaking changes. Haven't actually seen this work yet. 

I already pushed a v1 branch off of master. On that branch, we should be able to just make commits and run `yarn release` manually when we want to push a hotfix, assuming the tags followed that branch as well. This is all speculative but it seems right.

PR #1661, opened 2020-12-07
chore: Simplifies Story Generation

redreceipt:
I always try to run yarn generate-stories from the root and then realize it does nothing :) this fixes that


vinnyjth:
I think this change is already in `next` also :)

vinnyjth (approved):

PR #1662, opened 2020-12-07
chore: Extends Watch to Include UI Pkgs

redreceipt:
- chore: Simplifies Story Generation
- chore: Extends Watch Script to Include UI Packages

### DIFF
--- a/scripts/watch.js
+++ b/scripts/watch.js
@@ -3,16 +3,9 @@ const chokidar = require('chokidar');
 
 console.log('Watching for changes');
 chokidar
-  .watch(
-    [
-      'packages/apollos-data-*/**',
-      'packages/apollos-rock-apollo-data-source/**',
-      'packages/apollos-server-core/**',
-    ],
-    {
-      ignored: ['**/node_modules/**', '**/lib/**', '*.html', '*.snap'],
-    }
-  )
+  .watch(['packages/**'], {
+    ignored: ['**/node_modules/**', '**/lib/**', '*.html', '*.snap'],
+  })
   .on('change', (fullPath) => {
     const pkg = fullPath.split('/')[1].replace('apollos-', '@apollosproject/');
     exec(`yarn lerna run build --scope ${pkg}`, (err, stdout, stderr) => {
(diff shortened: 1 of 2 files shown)


scripts/watch.js
```diff
@@ -3,16 +3,9 @@ const chokidar = require('chokidar');
...
-    {
-      ignored: ['**/node_modules/**', '**/lib/**', '*.html', '*.snap'],
-    }
-  )
+  .watch(['packages/**'], {
+    ignored: ['**/node_modules/**', '**/lib/**', '*.html', '*.snap'],
+  })
   .on('change', (fullPath) => {
     const pkg = fullPath.split('/')[1].replace('apollos-', '@apollosproject/');
     exec(`yarn lerna run build --scope ${pkg}`, (err, stdout, stderr) => {
```
vinnyjth, line 11: What do you think about amending this line so it doesn't run `build` on the `ui` packages? It will technically be a `noop`, but might save some time/computational overhead.

vinnyjth: Beyond that, this seems to work well!

redreceipt: if the packages don't have a build script, does `lerna` attempt to run anything? we may be good here

vinnyjth: It attempts, and then fails. I think we could avoid a useless `exec` which involves some io if we add a check. If you think it doesn't matter, or that we'll move away from this script, I'm fine ignoring!

redreceipt: only takes 0.4s

```
apollos-apps % yarn lerna run build --scope @apollosproject/ui-kit                                                                      (master) apollos-apps 
yarn run v1.22.10
$ /Users/michael/Documents/Projects/apollos/apollos-apps/node_modules/.bin/lerna run build --scope @apollosproject/ui-kit
lerna notice cli v3.14.1
lerna info filter [ '@apollosproject/ui-kit' ]
lerna success run No packages found with the lifecycle script 'build'
✨  Done in 0.43s.
```

redreceipt: plus if package names ever get more diverse than `ui`, it will be complex to skip

vinnyjth: 👍

vinnyjth (approved):

PR #1663, opened 2020-12-09
fix: Splits Interaction Requests

redreceipt:
Breaking up the request to the interaction custom endpoint because something is breaking (maybe ASP.NET)

### DIFF
--- a/packages/apollos-data-connector-rock/src/interactions/data-source.js
+++ b/packages/apollos-data-connector-rock/src/interactions/data-source.js
@@ -1,7 +1,7 @@
 import { parseGlobalId, createGlobalId } from '@apollosproject/server-core';
 import RockApolloDataSource from '@apollosproject/rock-apollo-data-source';
 import ApollosConfig from '@apollosproject/config';
-import { flatten, get } from 'lodash';
+import { flatten, get, chunk } from 'lodash';
 
 export default class Interactions extends RockApolloDataSource {
   resource = 'Interactions';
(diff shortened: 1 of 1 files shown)


redreceipt:
@vinnyjth this is up and _ I think_ ready, I just have to test it

packages/apollos-data-connector-rock/src/interactions/data-source.js
```diff
@@ -115,12 +115,23 @@ export default class Interactions extends RockApolloDataSource {
...
 
     if (ApollosConfig.ROCK.USE_PLUGIN) {
-      return this.request(
-        `/Apollos/GetInteractionsByForeignKeys?keys=${nodeIds.join(',')}`
-      )
-        .filterOneOf(actions.map((a) => `Operation eq '${a}'`))
-        .andFilter(`PersonAliasId eq ${currentUser.primaryAliasId}`)
-        .get();
+      // need to split up the request for ASP.NET
+      const splitNodeIds = new Array(Math.ceil(nodeIds.length / 10))
```
vinnyjth, line 119: I admire your `native` aproach, but also consider `chunk` 
https://lodash.com/docs/#chunk

vinnyjth:
@redreceipt I think your code is covered by existing tests, so you're good on that count. I left you a comment about code style, but good to merge after you decide one way or the other on that

redreceipt:
Did you happen to test at all? I wasn't able to before I had to leave for the day￼

redreceipt:
Oh I see what you're saying about the tests. OK I'll mark this ready for review and you can give me the green check. I'll always opt for a native implementations :-)￼

redreceipt:
what the heck is wrong with this title action?!

vinnyjth:
@redreceipt Would you consider moving your `chunking` code outside the body of the dataSource and calling `chunk(nodeIds)` or `group(nodeIds)`

It's not immediately intuitive what's being done right now - given that `splice` is weird and mutates the array and that the `new Array()` syntax is a little foreign. I think making the code more self documenting (by isolating the complexity of the chunk from the complexity of making multiple calls to rock) would help with that.

redreceipt:
@vinnyjth this what you're looking for?

vinnyjth (approved):
bingo! thanks

PR #1668, opened 2020-12-14
feat: Adds Whitelist to Series Progress Algorithm

redreceipt:
This will let us specify which series channels we want to look through for progress items. So instead of including all sermon series _and_ devotional series by default, we can pick one.

### DIFF
--- a/packages/apollos-data-connector-rock/src/content-items/data-source.js
+++ b/packages/apollos-data-connector-rock/src/content-items/data-source.js
@@ -547,25 +547,18 @@ export default class ContentItem extends RockApolloDataSource {
       .map(({ id }) => id);
 
     const inProgressIds = ids.filter((id) => ![...completedIds].includes(id));
+    let cursor = this.getFromIds(inProgressIds);
 
-    // whitelist only channels we've specified
-    if (channelIds) {
-      const whitelist = (await this.byContentChannelIds(channelIds).get()).map(
-        ({ id }) => `${id}`
-      );
-      const finalIds = inProgressIds.filter((id) =>
-        [...whitelist].includes(id)
-      );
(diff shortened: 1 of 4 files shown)


redreceipt:
@vinnyjth not sure what is going on with tests here but what do you think about this approach? it uses the blacklist/whitelist paradigm but I didn't know if you wanted to get away from that as stated in one of the comments.

vinnyjth:
@redreceipt Those two new queries scare me a lil bit, do you think we could consolidate that `eq channel id` / `ne channel id` logic into the call to `getFromIds`? It returns a RequestBuilder object so you might be able to chain off of that.

redreceipt:
let me see. I like that better too

redreceipt:
something like this?

packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -553,11 +546,19 @@ export default class ContentItem extends RockApolloDataSource {
...
+    // only search through allowed channels
+    channelIds.forEach((id) => {
+      cursor = cursor.andFilter(`ContentChannelId eq ${id}`);
+    });
+
+    // exclude campaign channels
+    ROCK_MAPPINGS.CAMPAIGN_CHANNEL_IDS.forEach((id) => {
+      cursor = cursor.andFilter(`ContentChannelId ne ${id}`);
+    });
+    return cursor;
```
redreceipt, line 561:
```suggestion

    return cursor;
```

redreceipt:
alright ready now

vinnyjth:
Well done 👏

vinnyjth (approved):

PR #1669, opened 2020-12-14
feat: Allows Action URLs to be Authenticated

redreceipt:
This will use the Rock token to authenticate URLs clicked through a feature

### DIFF
--- a/packages/apollos-data-schema/src/index.js
+++ b/packages/apollos-data-schema/src/index.js
@@ -811,6 +811,7 @@ export const featuresSchema = gql`
     READ_CONTENT
     READ_EVENT
     OPEN_URL
+    OPEN_AUTHENTICATED_URL
     OPEN_NODE
     OPEN_CHANNEL
   }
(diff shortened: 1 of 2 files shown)


redreceipt:
@vinnyjth this is what @camrun91 needs to get done what he's trying to. I don't love that Authed Web Browser API to begin with so if there's a better rewrite we should do instead of continuing to hack on it, I'm all for it.

vinnyjth (changes requested):
Looks good! You'll want to add `OPEN_AUTHENTICATED_URL` to the correct `enum` in the schema as well.

vinnyjth:
yah - what I've done for other apps is this: https://github.com/lcbc-digital/ephesus/blob/master/apolloschurchapp/src/tabs/home/Home.js#L52-L56

It's always made sense b/c we have already had other custom actions. It would be great if the API of the Authed Web Browser was better, so you could `{ ...relatedNode }` into the passed props, and then add `withRockUser` to the `Url` type

redreceipt:
alright @vinnyjth added

vinnyjth (approved):

PR #1673, opened 2020-12-17
fix: Fixes Horizontal Feature Loading States

redreceipt:
Fixes: https://public.3.basecamp.com/p/RJrwBjV4x1SnUypig3fMhQr6

Explicitly says that if title or subtitle is present in the horizontal list, hide loading state

### DIFF
--- a/packages/apollos-ui-kit/src/Placeholder/withPlaceholder.js
+++ b/packages/apollos-ui-kit/src/Placeholder/withPlaceholder.js
@@ -12,7 +12,7 @@ const withPlaceholder = (PlaceholderComponent = Line, getProps = {}) => (
       typeof getProps === 'function' ? getProps(props) : getProps;
     return (
       <PlaceholderComponent
-        onReady={!isLoading}
+        onReady={!isLoading || props.children}
         style={style}
         lineNumber={lineNumber}
         {...propInput}
--- a/packages/apollos-ui-kit/src/Placeholder/withPlaceholder.js
+++ b/packages/apollos-ui-kit/src/Placeholder/withPlaceholder.js
@@ -12,7 +12,7 @@ const withPlaceholder = (PlaceholderComponent = Line, getProps = {}) => (
       typeof getProps === 'function' ? getProps(props) : getProps;
     return (
       <PlaceholderComponent
-        onReady={!isLoading || props.children}
+        onReady={!isLoading}
         style={style}
         lineNumber={lineNumber}
         {...propInput}
--- a/ApollosStorybook/yarn.lock
+++ b/ApollosStorybook/yarn.lock
@@ -54,10 +54,10 @@
   version "0.0.0"
   uid ""
 
-"@apollosproject/ui-kit@^1.7.0-beta.3":
-  version "1.7.0-beta.3"
-  resolved "https://registry.yarnpkg.com/@apollosproject/ui-kit/-/ui-kit-1.7.0-beta.3.tgz#202a7f1a105f32d0bca4958550a1f31b8e0c8747"
-  integrity sha512-pIwdbl2G55iI57jfgMqn6A7ckMJJqzvzkDk7q1pcCMoxHdDOS+GmUVLViHXuJPWFgKeQPIG02H6A77f+DaTBbg==
+"@apollosproject/ui-kit@^1.7.0":
+  version "1.7.0"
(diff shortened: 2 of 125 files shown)


packages/apollos-ui-kit/src/Placeholder/withPlaceholder.js
```diff
@@ -12,7 +12,7 @@ const withPlaceholder = (PlaceholderComponent = Line, getProps = {}) => (
       typeof getProps === 'function' ? getProps(props) : getProps;
     return (
       <PlaceholderComponent
-        onReady={!isLoading}
+        onReady={!isLoading || props.children}
```
redreceipt, line 15: @vinnyjth I mention this in the description, this is optional here but I think could save us so much trouble down the road. And would probably allow us to take out A TON of edge case code for other components' loading states.

redreceipt: it almost fixes this PRs problem completely except I think if we have subtitle, we don't need to show title loading, and vice versa, reason for the other file fix. but this isn't always the case for components loading states, most of the time, this placeholder fix would suffice

redreceipt:
ok so all the snap changes are making me second guess the `withPlaceholder` change 😂 what am I missing here? how are children sneaking into the component?

packages/apollos-ui-kit/src/Placeholder/withPlaceholder.js
vinnyjth: Yah - this change is probably a little much. Our best bet is probably to always control the visibility of placeholders by testing for `isLoading` and finessing it further by changing the value of `isLoading`

redreceipt:
ok @vinnyjth I put it all back :)

vinnyjth:
Okay! So the only reason I _wouldn't_ want to merge this PR, is a UX issue. 

As a user, currently, when I pull to refresh, there are two indications that new content is loading in: 

1. I get placeholder loading states in my content items
2. My little pull to refresh indicator spins. 

The spinner logic is driven by the feed, so there's a period of time where a specific feature might still be reloading but the spinner is gone. For that reason, Think it might be valuable to continue to show placeholders _somewhere_ in the Feature if the feature is technically still loading. What do you think about keeping the logic above, by tweaking the logic of the rendered cards so that they are replaced with placeholder cards, while the feature is loading?

redreceipt:
> What do you think about keeping the logic above, by tweaking the logic of the rendered cards so that they are replaced with placeholder cards, while the feature is loading?

I like that. So show cards loading but not titles.

redreceipt:
wait haven't we seen weirdness with this before? I think it jumps in and out of loading states, something we've tried to get away from.

vinnyjth:
@redreceipt We have, yah. That's likely due to any number of reasons, some of which we may have fixed and other we have not fixed. At one point it was due to how the loading state was passed down from the feed to each of the features, but I think that's a bit more stable now. Either way, we'll probably have to make sure it's something we can test for.

redreceipt:
I’ll throw in the change and see if it’s a problem

redreceipt:
so we could handle it two ways:

1. from this component, something like `content={isLoading ? null : cards}`

Screen Shot 2020-12-17 at 3 41 46 PM

or deeper in `TileFeed` itself but that will likely affect lots of components

vinnyjth:
Handling it in this component sounds like the right approach. I'm not convinced this is the perfect final solution, but I think it's worth trying out and seeing how it feels :)

redreceipt:
alright try it now

vinnyjth:
@redreceipt Looking good - with one issue. 

What you'll see here is that when you pull to refetch - you can see "Sermons" temporarily. What's happening is that there's a sermon feed feature that's technically in the feed, but is hidden because it doesn't have any cards. 



In a production app - like newspring - that behavior won't check out fly. Some feeds (like the staff feeds) are in a similar state where they technically exist in the feed but most user's wont see them. 

I think the best case for this PR is to keep your `cards` loading logic (it looks really good!) and take out your tweaks to `title`/`subtitle` loading. If you have time, you could also apply that same `cards` logic to the other Features :)

redreceipt:
try it now

redreceipt:
Double-check you agree with this change on the Newspring app too. I don't really have an opinion either way.

vinnyjth (approved):

PR #1677, opened 2021-01-06
chore: Fixes Release script

redreceipt:
This is fixes the edge case where NPM job would finish first and upload new version numbers and then GPR would publish one above that


.github/workflows/release-canary.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches: [ master ]
 
 jobs:
-  build:
+  realease:
```
vinnyjth, line 8:
```suggestion
  release:
```

vinnyjth (approved):

PR #1679, opened 2021-01-08
fix: Fixes Avatar Cloud with Empty Avatar

redreceipt:


packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
```diff
@@ -194,7 +194,7 @@ class AvatarCloud extends PureComponent {
           typeof this.props.avatars[i] === 'string'
             ? this.props.avatars[i]
             : this.props.avatars[i]?.uri
-        }${this.props.avatars[i].id ? this.props.avatars[i].id : i}`}
+        }${this.props.avatars[i]?.id ? this.props.avatars[i].id : i}`}
```
burkeshartsis, line 197: 💥

burkeshartsis:
@redreceipt is there a test to cover this condition/situation?

packages/apollos-ui-kit/src/Avatar/AvatarCloud/index.js
vinnyjth, line 197: Passing in an array of avatars, with some of them `null`, seems like a recipe for other bugs. What do you think about filtering out `null` avatars or making our `propTypes` more strict to reject `null` avatars? null `ids` or `uris` should be fine, but it seems like a code smell to try and correct for entire null values.

redreceipt: yea I agree, I could default to empty objects coming in instead?

vinnyjth: yah - or something we have done in the past is use add a `get avatars` method that looks something like 
```
get avatars(){ return this.props.avatars.filter(a => a != null )}
```

redreceipt: if all the avatars were `null`, the cloud would show nothing

vinnyjth: That seems correct to me :) I wouldn't want to show 10 grey circles for the avatar cloud.

redreceipt: that would still be the case if URLs are empty strings or invalid. I think a better fix is a default avatar if `null` with initials or something.

redreceipt: here's an example of grey circles with no error, would be the same even if we filtered out (still not sure how this didn't error out)

https://user-images.githubusercontent.com/2659478/104044937-d33a4180-51ab-11eb-9d19-a412a84c16b0.MP4

vinnyjth: Yah - so I think in that case if we peeked at the underlying data you would see that almost empty objects with `ids` but no images are being passed. Caleb is in a weird state where the avatars themselves are null, I'm not sure how that would happen in core. 

In regard to you empty object fix, I think that's a decent idea - but somehow in the future we'll have to pass the initials into the component, and in Caleb's case I bet the initials wouldn't make it in.

vinnyjth: Restated - I think the `null` avatars represent users that don't exist, rather than users without `avatars`

redreceipt:
ok @vinnyjth, null avatars are filtered out

vinnyjth (approved):

PR #1680, opened 2021-01-11
feat: Postgres Data Connector Package

vinnyjth:
## DESCRIPTION

Most of the details are in the Readme, I would start there :) I'll list some goals of the PR below though for context.

- Easy "as much as you want integration" 
- Very little assumptions made about what possible schemas can look like
- Easy and enforceable setup for common fields, like "apollosId, apollosType, and originId/originType" 
- Painless migrations. Just `sync`. 
- Making "joins" easy without needing to keep all the models in one file.


packages/apollos-data-connector-postgres/README.md
```diff
@@ -0,0 +1,87 @@
...
+  initialize(...args) {
+    super.initialize(...args);
+    this.model = this.sequelize.models.content_item;
+  }
+}
+```
+
+## Usage
+
+You can use the postgres datasource like you would any other Datasource. In order for your schema to updated to match the config you have created in your models, you'll need to sync. The easiest way to do that is import `sync` from `@apollosproject/data-connector-postgres` and call it on app boot.
```
jimmyhogoboom, line 87: Should we provide some information here warning about the potentially destructive nature of `sync`?

vinnyjth: That's a good call :)

packages/apollos-data-connector-postgres/src/index.js
```diff
@@ -0,0 +1,7 @@
+export {
+  defineModel,
+  configureModel,
+  sequelize,
```
redreceipt, line 4: why are we exporting `sequelize` here?

vinnyjth: The sequelize we export is our specific instance of `sequelize` rather than the global `Sequelize`. It's nice to export this, because we can do "advanced things" with it, like write migrations, add new models outside the normal flow, access our db models in our bull jobs, etc.

packages/apollos-data-connector-postgres/src/postgres/index.js
```diff
@@ -0,0 +1,100 @@
+/* eslint-disable no-param-reassign */
+
+import { Sequelize, DataTypes } from 'sequelize';
+import ApollosConfig from '@apollosproject/config';
+import { createGlobalId } from '@apollosproject/server-core';
+
+// Use the DB url from the apollos config if provided.
+// Otherwise, use a SQlite database in the server root directory.
+const sequelize = new Sequelize(
```
redreceipt, line 9: Is this the prescribed method? to overwrite the variable like that?

vinnyjth: Straight out of their docs :) 
https://sequelize.org/master/

packages/apollos-data-connector-postgres/src/index.js
redreceipt: the answer to my second question answered this one.

packages/apollos-data-connector-postgres/src/postgres/index.js
redreceipt: weird...

vinnyjth: I don't think that's super uncommon in JS. 

`Capital Model` -> `lower case instance` to pass around

redreceipt: the weird thing to me is `lower case lib name` -> `Capital Class` -> `lower case class instance` all matching

vinnyjth: oh totally - yah that does get weird. I think named exports have helped with that, since you can't call your import whatever you want anymore.

redreceipt: like the import should be `sequelizejs` or class should be `SequelizeCxn` or something

redreceipt (approved):

vinnyjth:
Already merged in another branch!

PR #1682, opened 2021-01-12
chore: Deprecates CHOP Base URL

redreceipt:
This is valuable because it allows us to remove a variable from the config and I cant find anything that uses it


vinnyjth:
It's used implicitly on line `61` and `71` 

https://www.apollographql.com/docs/apollo-server/data/data-sources/#resolving-urls-dynamically

redreceipt:
Oh shoot good call. What about this latest change? default the web view url to the base URL, which is probably most common

vinnyjth:
Sounds good to me

vinnyjth (approved):

PR #1683, opened 2021-01-14
fix: Support loading auth token even if cache is partially loaded/unloaded

vinnyjth:
I managed to reproduce a "logout" on Newspring that matches the following conditions (which match what we are seeing in wild)

- Spontaneous
- Happens when opening the app
- Takes you to the splash screen

By adding a wait to the cache hydration method 
Screen Shot 2021-01-14 at 10 46 53 AM

The flow when opening the app looks like this

1. App is opened, cache is un-hydrated 
2. Cache is hydrated
3. Cache loaded mutation fires
4. ProtectedRoute navigates user to the splash screen if logged out, and the Home Screen if logged in. 

However, if the cache takes a long time to hydrate, it looks more like this 

1. App is opened, cache is un-hydrated 
2. Cache is hydrated
3. Cache loaded mutation fires
4. ProtectedRoute sees the cache has been hydrated, but data still isn't in the cache and navigates user to Home Screen. 
 - I admittedly can't explain why this is true, the `cacheLoaded` values seems to become `true` a tick or two before the data actually is in the cache. 

An easy fix for this, is to make sure that the `isLoggedIn` method is actually going to check for the `authToken`. I opted to continue to use the Apollo client store to make this check, but changed `readQuery` to `query` which will actually perform the check in the cache if the data doesn't exist.

## Testing

Read over the code, then boot the app in `templates` and test logging in and logging out. You can test that the error is fixed via adding a `wait` to the templates code like depicted in the screenshot above.


redreceipt:
what does "cache hydrated" mean?

vinnyjth:
> what does "cache hydrated" mean?

Means that the cache has been filled with the data in "cold storage". More constructively; that the apollo cache has been copied back into memory from AsyncStorage

packages/apollos-ui-auth/src/Provider.js
```diff
@@ -24,12 +24,12 @@ export const GET_AUTH_TOKEN = gql`
...
   Query: {
     authToken: () => AsyncStorage.getItem('authToken'),
-    isLoggedIn: (_root, _args, { cache }) => {
+    isLoggedIn: async (_root, _args, { client }) => {
       // When logging out, this query returns an error.
       // Rescue the error, and return false.
       try {
-        const { authToken } = cache.readQuery({ query: GET_AUTH_TOKEN });
-        return !!authToken;
+        const { data } = await client.query({ query: GET_AUTH_TOKEN });
```
conrad-vanl, line 31: What's the difference between using `cache.readQuery` and `client.query`?

vinnyjth: `cache.readQuery` will always read the cached data and error if it doesn't exist. `client.query` will trigger the client resolver, in this case, if data isn't already in the cache.

redreceipt: Screen Shot 2021-01-15 at 10 22 00 AM

redreceipt:
when does this `isLoggedIn` query run?

vinnyjth:
> when does this isLoggedIn query run? 

I believe a few places across the app, but the most important place for the sake of this PR is here: 

https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-ui-auth/src/ProtectedRoute.js#L84

redreceipt:
@vinnyjth is this a problem? on the `templates` > `client/index.js` file, if there's an error with cache hydration, we're never actually running the `finally` here 

Screen Shot 2021-01-15 at 10 53 53 AM

redreceipt:
like what's the point of `catch` and `throw`?

vinnyjth:
@redreceipt That's a good question. I don't think that's the issue (because we would have seen that in bugsnag) but your question is valid. We can probably change that code not to `throw` safely. 

It's also odd because the `ensureCacheHydration` is already wrapped in `try`/`catch` so we are double safe.

redreceipt:
sorry for questions, trying to understand the flow. why is `ensureCacheHydration` an IFFE? why wouldn't we just call it like normal on the component mount?

trying to explain why `cacheLoaded` is outrunning the persistor

redreceipt:
actually that could be why? the await isn't waiting on an async funtion, it's waiting on an async function inside an IFFE

vinnyjth:
> why is ensureCacheHydration an IFFE?

Per my understanding (this code predates me 😄) So that the cache is hydrated as soon as possible 

>actually that could be why? the await isn't waiting on an async funtion

It is though, because the `IFFE` returns a promise which is then awaited. You can test this by adding the `wait` in my screenshot in the description. You'll see everything resolve in the order you expect. The `cacheLoaded` will only fire after the `wait` time has elapsed.

vinnyjth:
no worries about the questions! glad you're asking them. I don't 100% understand _why_ my fix works, my runing theory has to do with how apollo handles `readQuery`. B/c `readQuery` can be called in `ProtectedRoute` called before the cache is hydrated, we don't get a token from the cache and so "false" is returned to the `ProtectedRoute`. When the hydration is finished, that `false` isn't updated for whatever reason. Maybe because it doesn't get overwritten by the loaded cache? Not sure. 

My fix works because it stops us relying on the apollo cache for knowing if someone is logged in.


Side note: I think that the reason that users like frank get this error a lot more than users like me, is that they have a lot more data in their cache. Thus it will take longer to hydrate, thus they run the risk of this race issue occurring, and seeing the wrong screen when loading the app. Also might explain why it might seem to happen more for Android users.

redreceipt:
ok so how's this for a summary:

`cache.readQuery` is not dependent on `ClientProvider` being mounted (which calls `ensureCacheHydration`) even though it's a client query. `client.query` is.

vinnyjth:
Not quite - I don't think either depend on ApolloProvider being mounted. ApolloProvider only changes your ability to use `<Query>`, `useQuery`, and other React specific bits of code

I would tweak your summary to read like this: 

---

cache.readQuery (InMemoryCache.readQuery) is dependent on the cache being hydrated from storage, while client.query is not.

redreceipt:
and that's because `client.query` has `ensureCacheHydration` in its flow while `cache.readQuery` does not?

vinnyjth:
that's because client.query will call the local resolver to get its data, while cache.readQuery will only return data in the cache. Before the cache is hydrated, there is no data in the cache (it's just the default data). 

This is made slightly more confusing by the fact that calling client.query will only call the resolver once, and after that it pull data from the cache. That doesn't matter for our case though, and is the desired behavior.

redreceipt:
ok I think that makes sense. `cache.readQuery` is much more primitive, just uses `AsyncStorage` to get stuff from disk. It never needs the `ensureCacheHydration` to run first. The `client` in `client.query` is actually a prop coming from `ApolloProvider` which means `ApolloProvider` has already mounted, which means `ensureCacheHydration` has already run.

vinnyjth:
@redreciept you are so close! Forget about AsnycStorage for a second - 

cache.readQuery is much more primitive, just gets stuff from Apollo Cache. client.query is much less primitive, it checks for resolvers, and calls those resolvers to get data. it will normally go to the network to get data, but that's not true in this case. 

Both client and cache are available as soon as the app boots. `ApolloProvider` gives GraphQL components access to client. `cache` is actually a property of `client`, but they are exposed separately in the resolvers for clarity sake. (but I could have called `client.cache.readQuery` or even `client.readQuery`

Data doesn't exist in the Apollo Cache until we put it there. We put it there by 
1. Running a query (either local or network) that gets data
2. calling `writeQuery` which dumps data straight to the local cache
3. Using a cache persister that fetches from `AsyncStorage` 

Our persistor/hydrater/AsyncStrorage exists so that when you close the app, the apollo cache still exists. Normally it's stored only in memory so it would vanish when the app's allocated memory is garbage collected.

redreceipt (approved):

PR #1690, opened 2021-01-21
feat: Adds User Prayer List

redreceipt:
This will give us the ability to pull back a list of prayers for a specific user. It also adds a field on the `Person` type, useful for social features down the road

Screen Shot 2021-01-21 at 11 47 42 AM


vinnyjth (approved):
Good PR, good tests. Nice work!

PR #1691, opened 2021-01-21
feat: Adds User Prayers Feature (API)

redreceipt:
This adds a `UsersPrayers` feature option to the config, returning the prayer list of a specific user

Screen Shot 2021-01-21 at 2 16 34 PM

### DIFF
--- a/packages/apollos-data-connector-rock/src/action-algorithms/index.js
+++ b/packages/apollos-data-connector-rock/src/action-algorithms/index.js
@@ -60,11 +60,16 @@ class ActionAlgorithm extends RockApolloDataSource {
     );
   }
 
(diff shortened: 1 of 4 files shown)


redreceipt:
blocked on #1690

redreceipt:
@vinnyjth before you approve this, would it be a better approach to not use a separate feature, but instead just use the `PrayerList` feature for a different component and simply use the data to construct the UI?

redreceipt:
well on second thought, there would be no way to specify which UI component to use from the config unless we had a separate feature type. just seems very redundant to build almost the same thing on the backend to make the front-end look different.

redreceipt:
for clarity, this is what we're building. could easily be done with the existing `PrayerListFeature`

vinnyjth (approved):

PR #1693, opened 2021-01-21
feat: Comment visibility and flagging

jimmyhogoboom:
## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


packages/apollos-data-connector-postgres/src/postgres/index.js
```diff
@@ -5,11 +5,14 @@ import { Sequelize, DataTypes } from 'sequelize';
 import ApollosConfig from '@apollosproject/config';
 import { createGlobalId } from '@apollosproject/server-core';
 
+const options = process.env.NODE_ENV === 'test' ? { logging: false } : {};
```
vinnyjth: good call ;)

packages/apollos-data-connector-postgres/src/comments/model.js
```diff
@@ -1,6 +1,12 @@
 import { DataTypes } from 'sequelize';
 import { defineModel } from '../postgres';
 
+const Visibility = {
```
vinnyjth, line 4: I like how you are passing this Enum around

packages/apollos-data-connector-postgres/src/user-flags/model.js
```diff
@@ -0,0 +1,17 @@
+import { DataTypes } from 'sequelize';
+import { configureModel, defineModel } from '../postgres';
+
+const createModel = defineModel({
+  modelName: 'user_flags',
```
vinnyjth, line 5: should this be `user_comment_flags`? Just in case we ever support flagging for something like `prayers`. The other option is keep this as `user_comment_flags`, but change `commentId` to `nodeId` and `nodeType` so we can flag models polymorphicly.

packages/apollos-data-schema/src/index.js
```diff
@@ -978,7 +978,13 @@ export const featuresSchema = gql`
...
 export const commentSchema = gql`
   extend type Mutation {
-    addComment(parentId: ID!, text: String!): Comment
+    addComment(
+      parentId: ID!
+      text: String!
+      visibility: CommentVisibility
+    ): Comment
+
+    flagComment(commentId: ID!): UserFlag
```
vinnyjth, line 987: Do you think we should return the `flag` here? I like how It's consistent with the other mutation signatures, but it might be more useful to return the `Comment`, the UI probably won't care about the specifics of the User Flag, and rather only care about current status of the FlaggedComment. 

If we have mutation return a UserFlag, we can also get rid of the `UserFlag` type and bring down the schema size a bit

packages/apollos-data-connector-postgres/src/user-flags/__tests__/dataSource.js
```diff
@@ -0,0 +1,130 @@
...
+    Auth: {
+      getCurrentPerson: () => ({ id: personId }),
+    },
+    Person: {
+      getFromId: () => ({ id: personId }),
+    },
+  },
+};
+
+describe('Apollos Postgres Comment Flags DataSource', () => {
```
vinnyjth, line 21: Great tests!

packages/apollos-data-connector-postgres/src/postgres/index.js
```diff
@@ -5,11 +5,15 @@ import { Sequelize, DataTypes } from 'sequelize';
 import ApollosConfig from '@apollosproject/config';
 import { createGlobalId } from '@apollosproject/server-core';
 
+const sequelizeConfigOptions =
+  process.env.NODE_ENV === 'test' ? { logging: false } : {};
```
vinnyjth, line 9: THANK YOU for adding this. Makes running the tests in this module much nicer.

vinnyjth (changes requested):
Looking really good @jimmyhogoboom! I have two separate pieces of feedback around `user_flags` we'll need to figure out, but besides that this PR is ready to go!

packages/apollos-data-connector-postgres/src/user-flags/model.js
jimmyhogoboom: Good call! I think I like `user_flags` with `nodeId` and `nodeType` better, as then we won't have to implement any new models/dataSources when we want to enable flagging for other objects. I'll make that change.

vinnyjth: I think I like that idea the most as well, sounds good :D

packages/apollos-data-schema/src/index.js
jimmyhogoboom: Do you mean we can get rid of the `UserFlag` type if the mutation returns a Comment?

If so, then I agree, that would be more useful. I'll update this as well and remove that unnecessary type.

vinnyjth: yep! typoed that sentence, I'll fix it so it reads in a way that makes sense

packages/apollos-data-connector-postgres/src/postgres/index.js
jimmyhogoboom: Yeah it was getting pretty old scrolling through all those messages to figure out why a test failed 😅

jimmyhogoboom:
@vinnyjth I can't manage to get these tests to work, and I'm not sure where to go from here. They all work just fine locally.

vinnyjth (approved):

PR #1702, opened 2021-01-28
feat: Add Comment Feature

jimmyhogoboom:

Returns an Add Comment Feature when a Content Item has Comments enabled.

Prompts are returned in the feature, with default values defined in the content-item data-source.

### How do I test this PR?

With this query, anywhere you see a CommentListFeature, you should also see an AddCommentFeature:

```
query {
  contentChannels {
    childContentItemsConnection {
      edges {
        node {
          id
          title
          ... on WeekendContentItem {
            featureFeed {
              features {
                id
                ... on CommentListFeature {
                  id
                }
                ... on AddCommentFeature {
                  id
                  addPrompt
                  initialPrompt
                  relatedNode {
                    id
                    __typename
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

I moved the assertions against the CommentListFeature into a new test along with the AddCommentFeature assertions.

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


vinnyjth:
Hey @jimmyhogoboom :D You'r PR is 90% there, but I'm going to throw another acceptance criteria your way: 

I should be able to run the following query and get the same data on the AddCommentFeature I would get by running the query in your PR description. You'll have to add in the AddCommentFeature

```
query getFeature {
  node(id: (ADD COMMENT FEATURE ID GOES HERE) ){
    ... on AddCommentFeature {
      id
      addPrompt
      initialPrompt
      relatedNode {
        id
        __typename
        ... on ContentItem {
          title # this probably won't be happening in production, but someone _might_ want to do this and we want to make sure people could if they wanted to, if only to preserve the validity of our schema. Aka: If the schema says something is valid, we should make sure it works. 
        }
      }
    }
  }
}```

This mimics how we'll be using the feature in production. If you need tips on how to make that happen, I can try and point you in the right direction. You can also look at some of the other parts of the `Feature` dataSource that deal with relatedNodes and how the Feature IDs are generated.

**Comment by jimmyhogoboom** [comment]
@vinnyjth I'm almost there, but I can't manage to find a good example of how to resolve the relatedNode. When running the query you provided, I get a different relatedNode than the one in my description, and I'm not entirely sure why. 

So I guess I'll take that hint 😅

**Comment by vinnyjth** [comment]
@jimmyhogoboom 

Try something like this :) 

```
    relatedNode: async (
      { id, __type },
      args,
      { models: { Node }, dataSources },
     info
    ) => {
      const item = const item = await Node.get(createGlobalId(id, __type), dataSources, info);
      console.log(item);
      return item;
    },
```

vinnyjth:
And looking over our codebase now, I realize that I steered you wrong about `relatedNode` 😬 Looks like we do something a little different with most uses of `relatedNode`.

jimmyhogoboom:
@vinnyjth Ok I think I finally got it working the way we want it!

vinnyjth:
Looking really good @jimmyhogoboom :D 

Only one thing left to change, I think that your logic for `initialPrompt` and `addPrompt` is a bit off. When you resolve it, I don't think `attributeValues` will ever have any value, since the `AddCommentFeature` doesn't have `attributeValues`. Thus, you'll aways end up with the default value. 

What we have done in the past for prompts/titles/subtitles like this, and this might seem kinda crazy, is we embed them in the ID :) That way you already have them when deconstructing the ID and so don't need to go fetch the content item to resolve them again. 

This has a flaw, in that if you change the prompts on the ContentItem, any client that has a reference to the old prompts won't have the new prompts. That happens to be okay though, because any new visit to the content item means they will get a new AddCommentFeature ID, and thus a new prompt :)

jimmyhogoboom:
@vinnyjth How's that look?

vinnyjth (approved):
Awesome work @jimmyhogoboom  :D

PR #1708, opened 2021-02-01
fix: Adds support for @react-navigation onReady callback

vinnyjth:

We need a way to interface with this new `onReady` callback, otherwise we fire events from deep links before navigation has fully mounted. 

Usage example: 

image

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-ui-kit/src/NavigationService/index.js
+++ b/packages/apollos-ui-kit/src/NavigationService/index.js
@@ -31,7 +31,6 @@ const setIsReady = () => {
 };
 
 const navigate = performWhenReady((...args) => {
-  console.log(...args, 'deese args');
   _navigator.navigate(...args);
 });
 
--- a/packages/apollos-ui-kit/src/NavigationService/index.js
+++ b/packages/apollos-ui-kit/src/NavigationService/index.js
@@ -1,12 +1,13 @@
 import { CommonActions } from '@react-navigation/native';
 
 let _navigator;
+let _ready = false;
 let _pendingActions = [];
 
 const getNavigator = () => _navigator;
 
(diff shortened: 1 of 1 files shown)


packages/apollos-ui-kit/src/NavigationService/index.js
```diff
@@ -15,13 +16,22 @@ const performWhenReady = (func) => (...args) => {
...
+  _ready = true;
+  if (_pendingActions.length > 0 && _navigator) {
+    _pendingActions.forEach((action) => action());
+    _pendingActions = [];
   }
-  _pendingActions = [];
 };
 
 const navigate = performWhenReady((...args) => {
+  console.log(...args, 'deese args');
```
vinnyjth, line 34:
```suggestion
```

redreceipt:
Are we any closer to being able to remove some of this stuff? Seems like a smell to even use the `NavigationService`. Wasn't some of it an effort to fix the auth bug?

vinnyjth:
> Are we any closer to being able to remove some of this stuff? Seems like a smell to even use the `NavigationService`. Wasn't some of it an effort to fix the auth bug?

I wish :/ I dug into this a bit when implementing this fix. The issue is that the providers that **need** this service exist outside the navigation context. I tried moving them inside the navigation context, but navigation depends on some of the providers, so it turns into a chicken and egg problem. 

The final solution is likely to separate the providers into two groups, and put half the providers above and half the providers below, but that is something we can do in the future on `templates`. For now I believe that getting the `NavigationService` working correctly again is probs the biggest bang for our buck.

redreceipt (approved):

PR #1716, opened 2021-02-03
chore: Bumps Packages to Node 12

redreceipt:
Very odd behavior with some of our tests here. I removed some of the problematic ones that weren't getting us that much anyway.


vinnyjth:
Yah, looks like we are getting loading states for a lot of those tests 🤔 I can take a look at fixing at some point, but it might be during cool down.

PR #1727, opened 2021-02-10
fix: Use the status bar height to avoid keyboard correctly

vinnyjth:
For whatever reason, the padding looks wrong in the sim. Be sure to test on an actual device. Tested on Android, and the button is hidden behind the keyboard both before and after this PR :/ It's okay though, because on Android there's a "next" button on the keyboard that advances. 

EA57CFC0-F647-4807-B350-7AA7B83AF986_1_105_c
7B48934A-309F-4990-94FF-8F80F19CCD9B_1_105_c

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


redreceipt:
the diff is hard to read, I see two main things that changed:

- bring `BackgroundView` above `KeyboardAvoidingView`
- use the new ios Native Module status bar thing

is that right? what changed to need the new iOS thing?

vinnyjth:
> the diff is hard to read, I see two main things that changed:
> 
> * bring `BackgroundView` above `KeyboardAvoidingView`
> * use the new ios Native Module status bar thing
> 
> is that right? what changed to need the new iOS thing?

The BackgroundView being moved above the KeyboardView fixes a white flash that would happen when the keyboard was opening, as the dark background was being pushed upward. 

The new ios Native module is used to determine the height of the status bar, so we can properly move the Keyboard to the correct location. I'm not 100% sure why it's the status bar height that's important (I think it has to do with how the padding is calculated for the keyboarding avoiding view) but previously on iOS it was always 0, while on Android it was set properly.

redreceipt:
don't feel great about merging until we understand why the padding is so jacked up on the sim. That's going to make future UI dev to this portion of the app a pain if we _always_ have to test on a real device.

redreceipt:
definitely, some odd behavior happening, padding changes when I go back and forth.

Kapture 2021-02-10 at 16 09 04

vinnyjth:
I assuming it has something to do with the hardware vs software keyboard. The padding on the sim seems to correct itself for me after I toggle the keyboard a few times

redreceipt:
it doesn't correct itself on the first screen

vinnyjth:
How does it look on an actual device? I couldn’t reproduce the wonky padding there.

redreceipt:
haven't tested there yet. but I still don't have warm and fuzzies. we went from only one screen not working to now both screens showing erratic behavior on the sim. gonna make future UI a pain having to test all PRs on a real device. We haven't seen keyboard behavior mismatches anywhere else. I'd love to at least understand why the problem is there if we're going to have to deal with it forever. 

testing on a real device now.

redreceipt:
wonky on a real device for me too

IMG_0761

vinnyjth:
Thanks for your testing @redreceipt 

I think I have a better solution, I dug a bit more into the correct "stack" for `KeyboardAvoidingView`, `ScrollView` and `SafeAreaView` and I think I found something that works more consistently. I haven't tested on an actual device yet, but it works well on the sim.

vinnyjth:
You'll notice the padding is different between both of those screens. I'm still working on that. The "fix" I found is have the keyboard autofocus turned off, which is a really crappy solution. The padding seems to be different depending on if the keyboard is already open or not when the screen mounts. It's incredibly frustrating.

redreceipt:
Check the react navigation header true or false for the difference in padding. I added it to “fix” this problem before on the first screen.

vinnyjth:
@redreceipt Thanks for the header tip. The headers were already off on the first screen, and adding them in made the problem worse (the button disappeared under the keyboard). Also, we had an ugly header which wan't needed. I went the other direction instead, and made sure that the screen with the header (it has a header because of the back button) was taking the header height into account properly. 

I also realized that a lot of the reason these screens are so hard, and the traditional fixes aren't working, are because of `react-native-screens`. Elements like the Header aren't being calculated properly because they exist in a different context. If we use the non-native navigator, things would probably work better out of the box. 

The latest code I pushed up seems to work really well on an actual device (and I'll post a video to prove it) but I'm still seeing wonky stuff in the sim. I THINK it's because the actual iPhone has that grey area above the keyboard with suggestions, but the Sim doesn't have that box. The spacing makes sense

888FD2BE-9C2D-4080-B8C1-D975858A6320_1_102_o

vinnyjth:
https://user-images.githubusercontent.com/1637694/107599829-dc299300-6bef-11eb-8928-9b75bd8b672a.mp4

vinnyjth:
One more note - we'll have to edit the `ProfileEntryFieldContainer` to match our final set of changes before we merge this PR. I'll let you review this implementation first though.

redreceipt:
is the weirdness you're referring to? after showing the keyboard once in the sim, it looks right. but not at first.

Kapture 2021-02-11 at 14 10 09

vinnyjth:
@redreceipt Yep. On the actual device, the keyboard is already open when you hit that screen.

redreceipt:
I suppose minor confusion in the sim is better than an actual bug 😂 good enough for me.

redreceipt (approved):

PR #1728, opened 2021-02-10
feat: Hide unnecessary components 

jimmyhogoboom:

Checks the new TYPEMAP config value to determine if Up Next button, Parent feed, or Child feed should be displayed on a node. 

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**


jimmyhogoboom:
@vinnyjth I think this is what we're going for, but I'm not sure how to go about testing. Everything I can see in the app and find in Rock seems to be a WeekendContentItem, which has all three features enabled. I think we'd want to see what an item looks like if it doesn't.

vinnyjth:
@jimmyhogoboom You're so close here, the goal is to hide the loading states of each of the components, rather than the components themselves. Your strategy looks spot on besides that 👍

jimmyhogoboom:
@vinnyjth what would we gain from having the component exist on a node where it wouldn't be used? I might not understand exactly what's happening here

jimmyhogoboom:
@vinnyjth Ok I think I've got it! Not entirely sure about the UpNext button though. Hopefully I got it right?

vinnyjth:
@jimmyhogoboom This looks really good at a glance! I think your `UpNext` button is actually _more_ correct than the other two components, we might want to go ahead and `return null` early there as well. 

One thought - can you move the `loadingEnabled` prop to a variable inside each of the components? It will help those components be more self contained, although you'll have to duplicate or somehow modularize the `nodeImplementsType` function. It's great when these components will work given only a `nodeId`.

vinnyjth (commented):
@jimmyhogoboom this PR is FANTASTIC! I love the elegance of the `nodeImplementsType` function.

I'm going to test real quick in the sim, and then we can approve this.

jimmyhogoboom:
@vinnyjth woah, thanks man! Do you know what the deal is with the tests, though? Can we ignore that?

vinnyjth:
@jimmyhogoboom It looks like we'll need to add the Typemap into the testing environment. I'll give that a quick shot when I have the PR pulled down :)

vinnyjth (approved):

jimmyhogoboom:
Now a lint error??

vinnyjth:
@jimmyhogoboom Beat me to the linter fix :)

PR #1729, opened 2021-02-10
feat: add scripture features to devotional content items

redreceipt:
This is just to match the functionality of `WeekendContentItem`. We can deprecate content features later if we feel they are not the best method moving forward.


vinnyjth (approved):
Thanks for doing this. I made this change in NS as well. We can probably do the same for `UniversalContentItem` (if we haven't already)

PR #1735, opened 2021-02-12
fix: make features queryable by the sharable interface

redreceipt:
Our in app queries are only using `SharableNode` so we need this for the data to come through. Figured this was better than including more types in the query

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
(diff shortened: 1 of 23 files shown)


vinnyjth (approved):

redreceipt:
This isn't going to work. The schema is too locked in. I'm gonna change the query instead and circle back around to make improvements to the sharable schema in v3.

PR #1738, opened 2021-02-13
fix: Missing cast when querying for comments

vinnyjth:
## DESCRIPTION

This is now the second time I've made this mistake 🙃  It's tricky because there's no good way to test for this, it only shows up when using Postgres. 

I dug into options a little bit so that sqlite will reject invalid values in the same manner that Postgres does, but it doesn't look like there are any easy solutions. We might be able to find a way to do this casting in the Sequelize layer. 

## TODO:

- [ ] Set a relevant reviewer

## REVIEW:

**Manual QA**

### DIFF
--- a/packages/apollos-data-connector-postgres/src/comments/dataSource.js
+++ b/packages/apollos-data-connector-postgres/src/comments/dataSource.js
@@ -47,7 +47,12 @@ class CommentDataSource extends PostgresDataSource {
       [Op.or]: [
         { visibility: Visibility.PUBLIC }, // Show public journals
         ...(currentUser
-          ? [{ externalPersonId: currentUser.id, externalParentSource: 'rock' }]
+          ? [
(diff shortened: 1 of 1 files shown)


jimmyhogoboom (approved):
I don't know if the effort would be worth it, but maybe we should consider a set of separate integration tests that spin up an actual Postgres database for checking stuff like this? Keep it separate so we don't slow down the development flow locally and only run them in CI?

Although adding the casting to the Sequelize layer will probably be less effort.

vinnyjth:
@jimmyhogoboom I poked around a bit, and if we added the casting to sequelize, I think we would have to do it manually in pretty much the same way we are doing it now. Since I'm forgetting to cast now, I'm not sure if would consistently cast then 😂  


I really like the idea of running our integration tests on Postgres.

PR #1740, opened 2021-02-15
fix: add sharing to text and scripture features

redreceipt:


vinnyjth:
@redreceipt Why not have Text/Scripture feature `implement ShareableNode`

redreceipt:
> @redreceipt Why not have Text/Scripture feature `implement ShareableNode`

It is, I just threw it in the function right above that extends content items

redreceipt:
oh sike wrong PR

vinnyjth:
> oh sike wrong PR

😂 been there

redreceipt:
that was the other PR. unfortunately it's not that simple because `SharableNode` uses `SharableContentItem` as its field so it didn't work like I'd hoped.

redreceipt:
which explains the v3 PR I just sent 😂

vinnyjth:
> unfortunately it's not that simple because `SharableNode` uses `SharableContentItem`

Well that's super frustrating. It would all be perfect if ShareableContentItem was just `ShareableContent` or something

vinnyjth (approved):

redreceipt:
> > unfortunately it's not that simple because `SharableNode` uses `SharableContentItem`
> 
> Well that's super frustrating. It would all be perfect if ShareableContentItem was just `ShareableContent` or something

exactly, pretty much what I did in the breaking change PR

PR #1745, opened 2021-02-15
chore!: remove resolver merge function

redreceipt:
On all the recent apps we've done, we don't use this function. We just spread out the base resolver manually.


redreceipt:
blocked by the title checker being fixed in #1743

PR #1746, opened 2021-02-15
feat: add theming to server package

redreceipt:
This lets us remove the theme resolvers from the API template

### DIFF
--- a/packages/apollos-data-schema/src/index.js
+++ b/packages/apollos-data-schema/src/index.js
@@ -275,9 +275,9 @@ export const themeSchema = gql`
   type ThemeColors {
     primary: Color
     secondary: Color
-    screen: Color
-    paper: Color
-    alert: Color
+    screen: Color @deprecated(reason: "Not supported on content items")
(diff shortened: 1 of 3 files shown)


vinnyjth:
@redreceipt Let's hold off on this, we can likely bundle this into the new `postgres` datasource.

redreceipt:
I thought about that but I didn't think it belongs there because it has nothing to do with SQL. Figured the server core spot would be best because it affects content from any source.

vinnyjth:
@redreceipt I actually think the postgres spot is perfect for it, since it's the first time that we'll be able to store themes sanely in postgres. 

Also, I see that you deprecated the `paper`/`screen`/`alert` colors. I think we can leave them undeprecated, although they currently aren't used, they technically work if you define them and add them to the theme fragment on the front end. That said, I don't know if they will ever end up being super useful since most use cases end up setting a "type" to get close enough.

redreceipt:
Isn’t the usefulness of this schema to define theme on a per content item basis? It should probably live with the CMS connector. Type should probably be inferred but color should be edited by the content creator.

PR #1748, opened 2021-02-16
fix: correctly pass reset password URL to auth

redreceipt:
Turns out this was just the first fix in passing params to auth. `screenProps` are no longer a thing in react nav 5 and we're using a lot of them


vinnyjth (approved):

redreceipt:
Our tests are being weird. Fails a different way running from root vs running from `ui-connected`. Also failures seem to have nothing to do with changes.

packages/apollos-ui-onboarding/src/slides/AskNotifications/AskNotificationsConnected.js
```diff
@@ -55,9 +55,9 @@ const AskNotificationsWithStatus = ({
 
 // eslint-disable-next-line react/display-name
 const AskNotificationsConnected = memo((props) => (
-  <NotificationsConsumer>
+  <PushContext.Consumer>
```
vinnyjth, line 58: Are you swapping this component so you can test/mock better?

packages/apollos-ui-auth/src/Password/PasswordEntryConnected.js
```diff
@@ -8,9 +8,11 @@ import { LoginConsumer } from '../LoginProvider';
 import PasswordEntry from './PasswordEntry';
 
 const PasswordEntryConnected = ({
-  handleForgotPassword,
   screenProps,
```
vinnyjth, line 11: We can probs ditch `screenProps`, since you are now correctly using route.params.

packages/apollos-ui-onboarding/src/slides/AskNotifications/AskNotificationsConnected.js
redreceipt: Swapping because I don't think we need the former at all and I'm testing my hypothesis :)

packages/apollos-ui-auth/src/Password/PasswordEntryConnected.js
redreceipt: Totally. was planning on making other improvements, this one included, in another PR.

vinnyjth (approved):

PR #1749, opened 2021-02-16
fix: insure to generate stories for storybook

redreceipt:
This wasn't happening automatically


vinnyjth (approved):
I was wondering about this. Thanks!

PR #1752, opened 2021-02-18
feat: Postgres instead of SQLite

jimmyhogoboom:
Reconfigure the `apollos-data-connector-postgres` package to use a locally running Postgres instance instead of SQLite. This is slower and slightly less convenient, but will ensure our test capture the actual behavior of Postgres in situations where it and SQLite diverge. 

Getting tests to run required having them run in band instead of in parallel, which does take longer.


vinnyjth:
Hey Jimmy! I haven't looked at your code at all here, but you might be able to keep the parallel using a special "JEST_WORKER_ID" that's set by Jest. If you search the repo for the phrase, tou can see how I use it for sqllite

jimmyhogoboom:
Thanks for the heads up! I'm currently working on just such a system, and getting close!

jimmyhogoboom:
@vinnyjth I'm at a bit of a fork in the road here, and I'd like your input please.

Currently I have most of the tests working by injecting a sequelize instance which is created during the test (using the Jest worker id)

This is kinda clunky. You have to specify the instance you want to use every time you use one of the postgres methods. And now it is causing an issue with the Resolve tests because that context is created automatically from the test helpers.

So I think I need to do this a different way, where somehow the postgres module can detect if we're in testing and use the right database based on the Jest Worker ID? I'm wondering if a path forward jumps out to you.

vinnyjth (approved):
Looking great @jimmyhogoboom! Thanks for getting this setup :)

redreceipt:
I love this PR

packages/apollos-data-connector-postgres/__mocks__/@apollosproject/config.js
```diff
@@ -1,9 +1,9 @@
 const ApolloServer = require.requireActual('@apollosproject/config').default;
 
 ApolloServer.loadJs({
-  DATABASE: {
-    URL: `sqlite:${process.env.PWD}/testing-${process.env.JEST_WORKER_ID}.db`,
-  },
+  // DATABASE: {
+  //   URL: `sqlite:${process.env.PWD}/testing-${process.env.JEST_WORKER_ID}.db`,
+  // },
```
redreceipt, line 6:
```suggestion
```

PR #1754, opened 2021-02-19
feat: Follow List Search

lmommaerts:
Creates modal component for searching for people to follow. There are some things that may need adjusted, e.g. should we allow the implementer to specify things like how the search triggers their search function (right now, it triggers after a 500ms timeout after the search value has changed).

### How do I test this PR?
Test FollowListSearch on Storybook

#### Regular
https://user-images.githubusercontent.com/3879073/108903823-c0f55500-75eb-11eb-81ae-6cf33e841936.mov

#### Modal
https://user-images.githubusercontent.com/3879073/108904234-3f51f700-75ec-11eb-8fc0-321f7f0adcc9.mov

### Modal triggered from Follower List
https://user-images.githubusercontent.com/3879073/108904668-cacb8800-75ec-11eb-9a98-5eb79822bce5.mov


ApollosStorybook/App.js
```diff
@@ -1,6 +1,7 @@
-import React from 'react';
 import { LogBox } from 'react-native';
-import ApollosConfig from '@apollosproject/config';
+import { enableScreens } from 'react-native-screens';
+
+enableScreens();
```
redreceipt, line 4: still need this if you're not using react nav?

lmommaerts: nope, good catch

redreceipt (approved):

packages/apollos-ui-kit/src/FollowList/FollowListSearch/FollowListSearchModal.js
```diff
@@ -0,0 +1,91 @@
...
+  alignItems: 'center',
+  justifyContent: 'space-between',
+  padding: theme.sizing.baseUnit,
+}))(View);
+
+const Button = styled(({ theme }) => ({
+  width: theme.sizing.baseUnit * 3,
+}))(TouchableOpacity);
+
+const ButtonText = styled(({ theme }) => ({
```
vinnyjth, line 27: I'm late looking at these, which is my bad :) 

Your code looks fantastic, and pretty much 100% inline with our project. The one thing we'll need to do is make sure we add `overide` names to your styled components. It's pretty simple, you add a second arg to the `styled`  call that's the name of the component, normally `module-name.component-name.sub-componenent-name`, so in this case it would be `ui-kit.FollowListSearchModal.ButtonText`. This allows client churches to tweak components in our ui-kit without needing to redefine them.

lmommaerts: Putting this in the onboarding PR https://github.com/ApollosProject/apollos-apps/pull/1765

PR #1755, opened 2021-02-22
feat: add new generic content query

redreceipt:
I've thought about this a lot. What we need for the web (and for 90% of app feeds) is a simple list of content from a specific channel. For the web specifically, we don't want a "feed" because the arrangement on screen is not as simple as horizontal or vertical. And we also will want to decouple the data and UI because the UI will change depending on the media query.

This is a simple, scalable way to get content using the API. The `CONTENT_CHANNEL` enum map can exist in the `config.yml` if it doesn't already.

### DIFF
--- a/packages/apollos-data-schema/src/index.js
+++ b/packages/apollos-data-schema/src/index.js
@@ -561,7 +561,7 @@ export const contentChannelSchema = gql`
     content(channel: CONTENT_CHANNEL_TYPE): ContentItemsConnection
   }
 
-  enum CONTENT_CHANNEL_TYPE {
+  enum CONTENT_CHANNEL {
     Sermons
     Devotionals
     Articles
--- a/packages/apollos-data-schema/src/index.js
+++ b/packages/apollos-data-schema/src/index.js
@@ -558,6 +558,14 @@ export const contentChannelSchema = gql`
   extend type Query {
     contentChannels: [ContentChannel]
       @deprecated(reason: "No longer supported.")
+    content(channel: CONTENT_CHANNEL_TYPE): ContentItemsConnection
+  }
+
+  enum CONTENT_CHANNEL_TYPE {
+    Sermons
+    Devotionals
+    Articles
+    Events
   }
 `;
 


redreceipt:
@vinnyjth @conrad-vanl would love your thoughts here. I think this could set us up well for the future. I think if we're not careful, the features/algorithms schema and extendable data-source could get out of hand. This gives us a method that gets us the basics if it's all that's called for.

vinnyjth:
@redreceipt The only thing I'm not sold on is the `CONTENT_CHANNEL` enum. It's a bit of a departure from how the rest of the schema operates, and I feel like that points to a slightly different solution 

Example: Most of our queries either take an ID as an arg, or take no arguments. You are either asking for what's on a tab/screen, or you have one of those pieces and you're asking for details. 

This is a different level, you're not asking for something specific, but you also are also expected to provide some sort of initial filter. I'm not saying this is _wrong_, just that it's different from the way we have thought about these problems in the past. 

So - you've mentioned that on web things are never as simple as "feeds". Is there perhaps another way to get a page "feed" that would better support content that's not strictly vertical or horizontal? For example, Feature's currently have an unused "order" field. Could we change that to be "location" and then an enum representing where/how the feed should display?

redreceipt:
> Example: Most of our queries either take an ID as an arg, or take no arguments. You are either asking for what's on a tab/screen, or you have one of those pieces and you're asking for details.
> 
> This is a different level, you're not asking for something specific, but you also are also expected to provide some sort of initial filter. I'm not saying this is _wrong_, just that it's different from the way we have thought about these problems in the past.

This isn't really true. We're already expected to provide the filter you're referring to, just from the `config.yml`. I don't think it's really much better that we keep our schema simple but make algorithms and the config increasingly more complicated. Same control, just harder to find. At least with the args in the schema, we can use already understood GraphQL principles like type safety and the playground docs. 

We are only going to keep adding different features and algorithms to do almost the same thing unless we find a way to scale it. And it's no secret we'd like a better way than hard coding strings of text in the config with no visibility as to what it's doing without looking through source code.

> So - you've mentioned that on web things are never as simple as "feeds". Is there perhaps another way to get a page "feed" that would better support content that's not strictly vertical or horizontal? For example, Feature's currently have an unused "order" field. Could we change that to be "location" and then an enum representing where/how the feed should display?

For the web, this will require three different "feature feeds" for desktop, tablet, and mobile. That's so much infrastructure just to pull a single content item from a channel.

vinnyjth:
> For the web, this will require three different "feature feeds" for desktop, tablet, and mobile. That's so much infrastructure just to pull a single content item from a channel.

So will the web be pulling different content for each of the sizes, or will the web be showing the same content in different ways for those three different sizes?

redreceipt:
> > For the web, this will require three different "feature feeds" for desktop, tablet, and mobile. That's so much infrastructure just to pull a single content item from a channel.
> 
> So will the web be pulling different content for each of the sizes, or will the web be showing the same content in different ways for those three different sizes?

Same content, shown in different orders and arrangements. I suppose it could be handled with flexbox order in CSS but that still will require finding the right content set from the feature feed and pumping that into a UI element on the front end. Either that or we make a `WebStorySingleFeature` and a `WebStorySingleMobileFeature`. 

Seems like this is why it's good to decouple data from UI. It worked fine when our UI set was relatively small in RN, but web opens up to infinitely different UI designs. In my head, it seems like a win to have a react component on the front-end run `content(channel: Story) { body }` and then style it for various screen types right there in the client web code.

vinnyjth:
Gotcha. So what you need is a way to query for content, but a specific set of content, in a way that doesn't make any assumptions about what the possible fields would be. (So instead of a HorizontalCardList, or a VerticalActionList, a GenericGenericList). 

I'm trying to come up with a solution that gets you what you need, with the tools that we have, without needing to circle back to the "query + keyword gives you content" strategy. While I think I see why your solution would be perfect for the work you are doing on the web, I'm think there's a broader paradigm about web that we need to solve and I want to explore more how FeatureFeeds fit into that. I know FeatureFeeds are complex and undocumented, but they have a the benefit of being a single system that drives more and more of the mobile app, granting a lot of server driven configuration and flexibility. (Maybe @conrad-vanl has input here as well.)

redreceipt:
It just seems to me that adding more algorithms with locked in parameters is the same as adding top level queries, only more obscure.

conrad-vanl:
We have a way to do this today. You use the `node` query and the `id` of the content channel

redreceipt:
How do I know the ID?

conrad-vanl:
You figure out the ID then store it on the client.

conrad-vanl:
once we have the shovel in place, I'm not opposed to adding more query functionality to `node`.... to be able to query by other parameters then just `id`, but we need to think through that some

redreceipt:
I can do that. But we'll definitely want a better long term solution as there's no easy way to "figure out" the ID

conrad-vanl:
yeah, if you're mainly only storing these IDs on the website, then I have zero problem with that and we should just do that.

PR #1761, opened 2021-02-24
chore: Run on linux

jimmyhogoboom:
These changes allow running the Android app locally on Linux operating systems while (hopefully) leaving everything working as normal on OSX. 

This required a new package which handles detecting the current OS and running different Yarn commands for different OS's. With this in place, we can avoid trying to run Apple-specific commands on systems that don't support them.


jimmyhogoboom:
Current issue: storybook fails to bundle properly. Getting `Error: Cannot use t "__Schema" from another module or realm.` and `Invariant Violation: Module AppRegistry is not a registered callable module (calling runApplication)`.

redreceipt:
Honestly not a huge fan of this PR. Making the bundle bigger and adding complexity to scripts seems like a mistake when you still have to have a Mac to work with the iOS app.

jimmyhogoboom:
@redreceipt At this point I'm inclined to agree. Going into this I didn't think the differences would be so pronounced, but here we are.

redreceipt:
good with closing this one?

PR #1763, opened 2021-02-24
fix: allow theme HOCs to accept custom theme prop

redreceipt:
WOW. That was enlightening.

This will allow us to pass a `theme` prop when using `withTheme` or `withThemeMixin`. I think there's definitely room to overhaul our theming system to remove the vast assortment of HOCs we have available, they mostly do the same thing in a slightly different way.


vinnyjth:
I need to check this PR on a device, but the code looks great 👍

vinnyjth (approved):
Looks good to me! Thorough solution!

PR #1764, opened 2021-02-25
feat: add notification settings

redreceipt:


vinnyjth:
I like this - I think probs need some designs for this screen. Changing the background to `paper`(?) might be a good start

redreceipt:
> I like this - I think probs need some designs for this screen. Changing the background to `paper`(?) might be a good start

Since that deals with what's behind the component I made here, what do you think about handling that client-side? I like the small footprint of this pkg component but if I prescribe what the nav background looks like, that means I'll need a `BackgroundView` and `FlexedView`, but of which I don't think belong inside this component.

redreceipt:
ok @vinnyjth I slapped a background view and I think it looks right

vinnyjth (approved):

PR #1767, opened 2021-02-25
feat: Journal Confirmation Screen

lmommaerts:
Adds a confirmation modal for journal entry submission.
Also adds a generic Modal component that doesn't use/require a navigator.

### How do I test this PR?
Test the AddCommentInput stories. A modal should pop up upon pressing Submit for the journal entry.

https://user-images.githubusercontent.com/3879073/109359898-3d8a6c80-7854-11eb-9da0-ecd8fc64526d.mov

https://user-images.githubusercontent.com/3879073/109359957-5abf3b00-7854-11eb-8831-06250e95e086.mov


lmommaerts:
I don't know why the test check is failing :( Tests are passing fine locally.

vinnyjth:
@lmommaerts My only guess would be a difference in node versions causing different snapshots(12vs anything else). I can take a look in a bit though

vinnyjth (dismissed):
Looking great! I added in two names to components that I think were moved around and didn't have names.

vinnyjth (approved):

PR #1768, opened 2021-02-26
feat: Postgres Person datasource

vinnyjth:

Phew - this is a big PR that does what feels like very little. The goal of the PR was to enable the API to run with users fetched from either Rock or from Postgres, depending on what DataSource you use. 

If you use the Rock Datasource, you don't need to make any changes to your app (as long as you aren't extending or calling certain internal dataSource functions)

If you want to use Postgres to drive People, you now can! You'll have to use a bit of bridging code for now though, which is covered in the templates PR. https://github.com/ApollosProject/apollos-templates/pull/275/

Some major highlights of this PR: 

- Added a "Person" datasource
- Added a "Campus" datasource (really minimal right now) 
- Added a bit more boilerplate to the postgres/postgres code to provide some better out of the box defaults for those datasources.
- Updated all references to the Person dataSource to make sure they would work with the new postgres datasource.
- Updated some existing rock code so it would be easier to interface with new Person dataSource 
- Split personas out of the Person dataSource, so they could be handled differently long term. 
 

### How do I test this PR?

No great silver bullets here. Your best bet is to pull down the companion PR https://github.com/ApollosProject/apollos-templates/pull/275/ and run through the app. Register, change your first/last name, change your campus, etc.


packages/apollos-data-connector-postgres/src/postgres/index.js
```diff
@@ -25,6 +25,15 @@ class PostgresDataSource {
     this.sequelize = sequelize;
     this.model = sequelize.models[this.modelName];
   }
+
+  getFromId(id, apollosId, { originType = null } = {}) {
```
jimmyhogoboom, line 29: What's the `apollosId` used for here?

vinnyjth: It's the second argument to that function, the "encrypted" id passed from the client. We're not using it, but because our third argument is positional I have to reference it. 


https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-server-core/src/node/model.js#L71-L73

We could rename to `encodedId` :D

vinnyjth, line 29:
```suggestion
  getFromId(id, encodedId, { originType = null } = {}) {
```

jimmyhogoboom: Gotcha, thanks!

jimmyhogoboom (approved):
Man, this is incredible. I can't believe how quickly you made so many deep changes. Nice work here solving a difficult problem!

PR #1772, opened 2021-03-02
feat: automatically add Rock tokens to links

redreceipt:
This will scrape through rock content and add Rock tokens to the link so the user will be automatically logged in when clicking links inside content


redreceipt:
@conrad-vanl @vinnyjth Before moving forward on this, I wanted to confirm the approach. this came up because Long Hollow asked to be able to click authenticated links inside content. Honestly, I'm not sure how I feel about this approach and want other opinions...

vinnyjth:
LCBC asked for pretty much the same thing. It's a bit dangerous for two reasons - 

1. Clicking third party links with the auth token in the query params would leak that users token to that third party site. 
2. Third party tokens are less secure that cookies are. Even on a church's website page, any third party JS can read that token, and if a user clicks out to a third party site, and doesn't set a `noreferrer` policy, they can read that token. 

All that said, I think something like this PR is probably okay  to have in core if....

1. We restrict adding the token to links from a specific domain. Looks like you have already done that :) 
2. We enable this as an "opt in" that doesn't happen by default.

packages/apollos-data-connector-rock/src/content-items/resolver.js
```diff
@@ -15,7 +15,17 @@ const { ROCK, ROCK_MAPPINGS } = ApollosConfig;
 export const defaultContentItemResolvers = {
   id: ({ id }, args, context, { parentType }) =>
     createGlobalId(id, parentType.name),
-  htmlContent: ({ content }) => sanitizeHtml(content),
+  htmlContent: ({ content }, args, { dataSources }) => {
+    const token = dataSources.Auth.getAuthToken();
+    return sanitizeHtml(
```
vinnyjth, line 20: You might be slightly better off with https://github.com/apostrophecms/sanitize-html#transformations

redreceipt:
I'm liking this PR less and less. Seems like a pretty small edge case. Long Hollow only wants it to authenticate into giving, which we could use a button for instead outside the HTML content. Just not a fan of the complexity this adds. And the other features we have added as opt-in end up being hardly used. Not worth it in my opinion. @conrad-vanl I know this is something we talked about doing though, what are your thoughts?

PR #1774, opened 2021-03-03
fix: Infinite loading horizontal feeds

vinnyjth:

Fixes a situation where scrolling to the end of a horizontal list would fire the `fetchMore` resolver repeatedly. This issue was caused on:  https://github.com/apollographql/apollo-client/issues/6354#issuecomment-664879778

### How do I test this PR?

I tested by copying these changes over to Newspring. It's tricky to reproduce in core, since we don't have any items with children that paginate.


packages/apollos-ui-connected/src/HorizontalFeedConnected.js
```diff
@@ -122,6 +121,7 @@ class HorizontalFeedConnected extends Component {
         query={this.props.query}
         variables={this.props.variables}
         fetchPolicy={'cache-and-network'}
+        notifyOnNetworkStatusChange
```
redreceipt, line 124: what is this doing? from the docs, it seems this is only useful if you are using `networkStatus` from the query prop bag, like if you wanted to do something during a refetch

vinnyjth: It also lets you use the `loading` state during a refetch. Normally the loading doesn't flip from `false` -> `true` during a refetch as of apollos 3

vinnyjth: Clue here: https://github.com/apollographql/apollo-client/issues/6354#issuecomment-663584372

redreceipt: oh interesting

redreceipt (approved):

PR #1775, opened 2021-03-03
feat: add prayer component prop handles

redreceipt:
This gives us the ability to customize the text in the prayer experience

Screen Shot 2021-03-03 at 12 37 29 PM


packages/apollos-ui-prayer/src/PrayerExperienceConnected/PrayerExperience.js
```diff
@@ -21,7 +21,7 @@ const PrayerExperience = ({
   track,
   AddPrayerComponent = AddPrayerScreenConnected,
   PrayerScreenComponent = PrayerScreen,
-  OnboardingComponent = PrayerDialogScreen,
+  OnboardingComponent = named('ui-prayer.Onboarding')(PrayerDialogScreen),
```
vinnyjth, line 24: Fascinating! Never would have thought to do this. If you move this to a `const` outside the arguments of the component, I'll merge this (function arguments are reevaluated on every function call)

redreceipt: Good thinking, I'll split it out. the only reason I did this is that `PrayerDialogScreen` is used for multiple purposes so overriding at that level didn't help.

vinnyjth: Yah makes a ton of sense. I've run into this problem before and it's been frustrating to solve, I like your solution.

vinnyjth (approved):

PR #1779, opened 2021-03-04
feat: add button link feature to content

redreceipt:
Test with the `give-content` branch on templates

Screen Shot 2021-03-05 at 8 53 44 AM


packages/apollos-data-schema/src/index.js
```diff
@@ -948,6 +948,14 @@ export const featuresSchema = gql`
     url: String
   }
 
+  type ButtonLinkFeature implements Feature & Node {
+    id: ID!
+    order: Int
+
+    title: String
+    url: String
```
vinnyjth, line 956: Love this. One nitpick - can this be `relatedNode: Node` instead of url? that way you could use this to link to either external urls (by passing a `Url` node) or event other parts of the app, like Content or Channels?

redreceipt: How would you specify another content item from the CMS? Technically you already could by using a deep link

vinnyjth: You might not normally in a CMS, but we've gotten requests before like 

- can you add a button on the bottom of all our sermons which takes you to the baptism content item 

We would have to hardcode that logic, so it wouldn't be CMS driven per say. 


You're right that we could do that with a deep link, but doing that kinda sucks.

redreceipt: How would we "hardcode" the logic to navigate to a content item? I just fear overcomplicating core functionality to accommodate for edge cases. Won't changing it to `relatedNode` require still picking the default of a URL string on core?

vinnyjth: > How would we "hardcode" the logic to navigate to a content item? 

we set relatedNode to `{ id: (content item id), __typename: ContentItem }` and the frontend is already equipped to handle how to navigate to that. 

Look at how we handle `primaryAction` on `HeroListFeature`

> Won't changing it to relatedNode require still picking the default of a URL string on core?

Yep! That part isn't easily solvable right now given our CMS is rock.

redreceipt: so for a specific content item, let's say we want to go to another content item. In the resolver we would have to do this, correct?

```
relatedNode: ({id}) => {
  if (id === ContentItem:123)
    return { id: ContentItem: 456, typename}
  return null
}
```

vinnyjth: here's how we do it in in `HeroListFeature`. 

In the config.yml
```
  - type: ActionBar
    title: Welcome to Apollos
    actions:
      [
        {
          title: Check In,
          icon: check,
          action: OPEN_URL,
          relatedNode: { __typename: Url, url: 'https://www.google.com' },
        },
      ]
```
In the datasource: 
```
    if (primaryAction) {
      // eslint-disable-next-line no-param-reassign
      primaryAction = this.attachRelatedNodeId(primaryAction);
    }
```
and a peek inside attachRelatedNodeId
```
  attachRelatedNodeId({ relatedNode, ...action } = {}) {
    if (relatedNode && !relatedNode.id) {
      return {
        ...action,
        relatedNode: {
          ...relatedNode,
          id: createGlobalId(
            JSON.stringify(relatedNode),
            relatedNode.__typename
          ),
        },
      };
    }
```
and if you want to be able to get more than id/typename from relatedNode
in the resolvers
```
{
   relatedNode: ({ id }, args, { models: { Node } }, info) => Node.getFromId(id, (can't remember), info)
}
```

Has worked well for us so far. Not crazy clean, but we have a bit of helper code already written to try and make this possible. After looking at your PR, and looking at what code we have, you might have a implementation that meshes really well with prior art if you add a `primaryAction: FeatureAction` type instead of URL. I know it's more complex, but I think it's ultimately less of a burden since we already have a system to manage that schema shape on the front end and back end.

redreceipt: We're defining this from CMS, not config.yml. content creators need to be able to edit the destination of that button.

vinnyjth: I don't think that changes anything except how you call the `createButtonFeature` function

redreceipt: Ok so possible flow, in the CMS, it would be `https://url.com` or `123` and then the DS function would determine which action and node based on the presence of "http"?

vinnyjth: You'll need to add a `createButtonFeature` function to this PR, but I think that would be needed anyway to match how we have implemented all of our other features.

redreceipt: are you opposed to me not creating that function to keep the footprint small? helps with future dev and debug. it's contained inside the content items DS right now, jumping to another pkg DS seems unnecessary.

vinnyjth: Yah for a few reasons - 

- We already have the create{featureName} for all other features. it's one way we standardize how features work
- it allows the feature to work with `getFromId`. The feature feed in your PR  works right now because you've got `url` and `title` inside the query, which is fine, but if you removed those it would stop working because `getFromId` wouldn't work from that feature
- makes it more difficult for other users who want to implement that feature to implement. Yah they could copy over your implementation, but that requires finding it :)

redreceipt:
ok @vinnyjth check this

packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -194,6 +194,27 @@ export default class ContentItem extends RockApolloDataSource {
       );
     }
 
+    const buttonLink = attributeValues?.buttonLink?.value;
+    if (buttonLink) {
+      const buttonText = attributeValues.buttonText?.value || '';
+      features.push(
+        Feature.createButtonFeature({
+          id: attributeValues.buttonLink.id,
+          __typename: 'ButtonFeature',
```
vinnyjth, line 203: You shouldn't need this anymore thanks to `createButtonFeature
` :)

packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -194,6 +194,27 @@ export default class ContentItem extends RockApolloDataSource {
...
+          // NOTE: right now only URLs are implemented
+          // could optionally parse Rock IDs for OPEN_NODE action?
+          action: Feature.attachRelatedNodeId({
```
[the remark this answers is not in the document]
redreceipt: keeps the other function tiny and insignificant and this one more where all the logic lives, similar to scripture and text features.

packages/apollos-data-connector-rock/src/content-items/data-source.js
redreceipt: true

packages/apollos-data-connector-rock/src/content-items/data-source.js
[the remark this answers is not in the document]
vinnyjth: Cool - we can leave it like this for now. 

The text and scripture honestly need to be revisited at some point. They are out of date with how we have been doing features since "figuring features out".

vinnyjth (approved):

PR #1780, opened 2021-03-05
feat: add feature sources to schema

redreceipt:
I took a step back in trying to solve the problem of making the front-end look different based on the algorithm type and I think I came up with good way. By simply exposing the algorithms to the front end, we can change how features are rendered without needing to create new feature types. This accomplishes a couple of other things too. Algorithms have always been cryptic and hidden in source code, this exposes them to the GraphQL schema so users can know the different ways features can be rendered simply by using the API. I also renamed the concept from Algorithms to Sources. We can slowly migrate, no need to change everything now. But I feel it's much easier to understand what they actually are and their purpose.

### DIFF
--- a/packages/apollos-data-schema/src/index.js
+++ b/packages/apollos-data-schema/src/index.js
@@ -805,6 +805,11 @@ export const featuresSchema = gql`
   interface Feature {
     id: ID!
     order: Int # 0 is the "Main Content". If order is < 0 than this comes before the body content.
+    sources: [FeatureSource]
+  }
+
+  enum FeatureSource {
+    SERIES_IN_PROGRESS
   }
 
   enum ACTION_FEATURE_ACTION {
@@ -844,6 +849,7 @@ export const featuresSchema = gql`
   type ActionListFeature implements Feature & Node {
     id: ID!
     order: Int
+    sources: [FeatureSource]
 
     title: String
     subtitle: String
@@ -865,6 +871,7 @@ export const featuresSchema = gql`
   type ActionBarFeature implements Feature & Node {
     id: ID!
     order: Int
+    sources: [FeatureSource]
 
     title: String
     actions: [ActionBarAction]
@@ -875,6 +882,7 @@ export const featuresSchema = gql`
   type HeroListFeature implements Feature & Node {
     id: ID!
     order: Int
+    sources: [FeatureSource]
 
     title: String
     subtitle: String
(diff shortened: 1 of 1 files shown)


redreceipt:
@vinnyjth ok using our convo earlier about the empty state nodes and feature types, I came up with a solution I think is simpler and more scalable. Read the description above and then tell me what you think.

vinnyjth:
@redreceipt I'm on the fence about this. We like to think that our API doesn't expose any implementation details, and this is a definite gray area. Stated differently: The API should express "what to render" more than it should express "here's my context". 

One of the biggest advantages of this, is that it allows us to ship new features without app updates, if our API is good at returning building blocks rather than providing raw data and asking the client to guess. It's definitely a compromise, similar complexity either way, but if we avoid adding details like this to the schema we can isolate that complexity to areas where the iteration cycle can be really fast :)

vinnyjth:
Here are some additional questions I think are worth exploring before we go down a route like this 

- If I have two Sources, both add custom UI, which one do I respect? 
- How do you tell a client to use a source with custom UI, without that custom UI?

redreceipt:
So client side, I was assuming it would be something like this:

```
{data.feature.source === "SERIES_IN_PROGRESS" ? <EmptyStateComponent /> : null}
```

Does that answer either of your questions? Currently, through the API, the client is getting "what is the data" and "how to display the data". I guess just don't see the disadvantage of telling the client "where the data came from" as well. And considering our source code is open source, anyone can already see the source of the data in the algorithm map.

vinnyjth:
> So client side, I was assuming it would be something like this:
> 
> ```
> {data.feature.source === "SERIES_IN_PROGRESS" && !data.feature.cards.length ? <EmptyStateComponent /> : null}
> ```
> 
> Does that answer either of your questions?

Except that `feature.sources` is an array, not a single source. 

I think, stepping back a bit, our client will be better off rendering what it's told to render, rather than being given context and told to figure it out. The solutions we talked about in slack, like sending a `EmptyState` card are in this vein, they tell the client exactly what the client is expected to present, versus making assumptions based on other data.

redreceipt:
I just feel like this is a very overloaded UI, passing a `MessageNode` or something similar

image

redreceipt:
I have an idea:

```
type: HorizontalCardListFeature
args:
  emptyMessage: "All caught up!"
```

PR #1783, opened 2021-03-09
feat: add empty message to horizontal card list

redreceipt:
This will allow us to pass a string to show if the card list is empty. Front end will be done in another PR


vinnyjth:
@redreceipt I'm struggling to see the advantages of this over an "EmptyCard" / "TextCard" / "ActionCard" approach. 

Sell me here :) The advantages of an EmptyCard approach is that we can use it in situations where we have only one or two cards, and the third card is "Interested in more?..." It is certainly more complex on the back end, but is a big with with regards to front end flexibility and requires less new schema.

redreceipt:
> @redreceipt I'm struggling to see the advantages of this over an "EmptyCard" / "TextCard" / "ActionCard" approach.
> 
> Sell me here :)

For this use case, I felt this was better because having a card implies being able to click it. Having an "empty" card that you can't click that just says "Nothing new!" doesn't feel like a good UI to me. Embedded text on the screen similar to the title feels more intentional.

vinnyjth:
> For this use case, I felt this was better because having a card implies being able to click it. Having an "empty" card that you can't click that just says "Nothing new!" doesn't feel like a good UI to me. Embedded text on the screen similar to the title feels more intentional.

I think part of the trick to the EmptyCard is that it wouldn't have the traditional card borders or shadow. Looking more like a set of text elements that's inline with a list. @conrad-vanl has the deets.

redreceipt:
ah, ok I'm gonna give this another shot

PR #1784, opened 2021-03-09
chore: remove feature feed component prop

redreceipt:
Not sure when, but we stopped using the custom component prop.


redreceipt:
Totally, but for this component, would we really allow people to use a component other than `FeedView` (if that's the one this should be pointing too)?

vinnyjth:
Oh, I see what you're saying. Strange that there's not an unused const linter error. And you're right, not really anything for `Component` to be.

vinnyjth (approved):

PR #1791, opened 2021-03-10
feat: Adds Postgres migrations

vinnyjth:

Removes the use of `sync` when running the app, which was causing issues

### How do I test this PR?

Remove any tables in your local database. Then get the [companion PR](https://github.com/ApollosProject/apollos-templates/pull/284) for templates and run apps and templates. You should get all tables with the correct structures.


jimmyhogoboom:
One downside to creating migrations by hand: tests don't catch those mistakes. There was a relationship field `campusId` missing from the migration causing errors, but the tests pass because when making the model, Sequelize automatically adds that column.

vinnyjth:
> One downside to creating migrations by hand: tests don't catch those mistakes. There was a relationship field `campusId` missing from the migration causing errors, but the tests pass because when making the model, Sequelize automatically adds that column.

We could fix that by avoiding `sync` all together and using the migrations in the tests. It might make testy a bit more tricky, but I'm all for it to avoid errors like that.

jimmyhogoboom:
@vinnyjth Wanna leave migrating during tests to another PR? In the meantime we can just be really careful about including all of our columns.

jimmyhogoboom:
Before merging, we might want to set up the changes to the comments model in https://github.com/ApollosProject/apollos-apps/pull/1796

vinnyjth:
> @vinnyjth Wanna leave migrating during tests to another PR? In the meantime we can just be really careful about including all of our columns.

Yep! That's fine by me.

packages/apollos-data-connector-postgres/src/follows/migrations/001_create_follows.js
```diff
@@ -0,0 +1,59 @@
...
+      primaryKey: true,
+      type: Sequelize.UUID,
+      defaultValue: Sequelize.literal('uuid_generate_v4()'),
+    },
+    requestPersonId: {
+      type: Sequelize.UUID,
+      references: {
+        model: 'people',
+        key: 'id',
+        as: 'follower',
```
vinnyjth, line 21: Should both of these `as` fields be the same? I' wondering if one should be `as: follower` and the other `as: followed` :)

vinnyjth: We might not even need the `as` in the migration since we define the `key`, which would be great :D

vinnyjth (commented):
Looking good! Only one comment which might need some experimentation to figure out for sure. We can hold off merging this until the `externalPersonId` PR is merged and we can backport those changes to this PR.

packages/apollos-data-connector-postgres/src/follows/migrations/001_create_follows.js
jimmyhogoboom: Whoops, you're right! Too much copy/paste I guess.

vinnyjth: I hear yah! whenever I write PR's like this I inevitably do the same thing about 3000000 times. Everything just blurs together!

jimmyhogoboom:
@vinnyjth How's this look? Is there anything I'm missing?

vinnyjth:
@jimmyhogoboom I think this is good! Feel free to merge whenever you are ready. (I can't, since I started the PR)

jimmyhogoboom (approved):

PR #1796, opened 2021-03-11
chore: Refactor Comments model to point directly at the postgres people table

vinnyjth:

Gets rid of the "external user id" notion, and uses the postgres primitives for a table join. One of those things that we couldn't easily do if we were in production, but thankfully we're technically not :) 

Here's my set of test mutations, but you can also test by clicking around and adding comments in the app. No change in actual functionality. 
```
query content {
  contentChannels {
    childContentItemsConnection {
      edges {
        node {
          id
          ... on FeaturesNode {
            featureFeed {
              features {
                ... on CommentListFeature {
                  id
                }
              }
            }
          }
        }
      }
    }
  }
}

query commentListFeature {
  node(id:"CommentListFeature:f5f25daa33b6f865b2f0d38f199b8de9dbaac83c6efefefeb49da137258038333a3794da043a4ca891c2f3ff182f33ca9dc2869613cc5bae845ecbe13347e517"){
    id
    ... on CommentListFeature {
      id
      comments {
        id
        person {
          firstName
          lastName
        }
        text
      }
    }
  }
}

mutation addComment {
	addComment(parentId:"WeekendContentItem:f8460101368b38c5aa9bbea8b2b8a625", text:"testing this stuff out :) today", visibility:PUBLIC){
    id
  }  
}


mutation flagComment {
  flagComment(commentId:"Comment:46af1b5bbd9b1a1e0a174ec3d955e91f902221e3705c6926055872475ec7ff796d7b52d5be6cfea5bc7f785800a9fa96"){
id}
}
```

### DIFF
--- a/packages/apollos-data-connector-postgres/src/comments/dataSource.js
+++ b/packages/apollos-data-connector-postgres/src/comments/dataSource.js
@@ -7,7 +7,7 @@ class CommentDataSource extends PostgresDataSource {
   modelName = 'comments';
 
   async addComment({ text, parentId, visibility = Visibility.PUBLIC }) {
-    const currentUser = await this.context.dataSources.Auth.getCurrentPerson();
+    const currentPersonId = await this.context.dataSources.Person.getCurrentPersonId();
 
     const { id, __type } = parseGlobalId(parentId);
 
@@ -17,7 +17,7 @@ class CommentDataSource extends PostgresDataSource {
         externalParentId: String(id),
         externalParentType: __type,
         externalParentSource: 'rock',
-        externalPersonId: String(currentUser.id),
+        personId: currentPersonId,
       },
       defaults: {
         visibility,
@@ -34,9 +34,9 @@ class CommentDataSource extends PostgresDataSource {
   }
 
   async getForNode({ nodeId, nodeType, flagLimit = 0 }) {
-    let currentUser;
+    let currentPersonId;
     try {
-      currentUser = await this.context.dataSources.Auth.getCurrentPerson();
+      currentPersonId = await this.context.dataSources.Person.getCurrentPersonId();
     } catch {
       // no user signed in, that's fine.
     }
@@ -46,11 +46,10 @@ class CommentDataSource extends PostgresDataSource {
       externalParentType: nodeType,
       [Op.or]: [
         { visibility: Visibility.PUBLIC }, // Show public journals
-        ...(currentUser
+        ...(currentPersonId
           ? [
               {
-                externalPersonId: String(currentUser.id),
-                externalParentSource: 'rock',
+                personId: currentPersonId,
               },
             ]
           : []), // Or show journals that belong to you.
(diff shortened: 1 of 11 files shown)


packages/apollos-data-connector-postgres/src/follows/dataSource.js
```diff
@@ -11,7 +11,22 @@ class Follow extends PostgresDataSource {
...
+
+    return this.model.findOne({
+      where: {
+        requestPersonId: currentPersonId,
+        followedPersonId: id,
+      },
+    });
+  }
+
+  async getPersonFollowingCurrentUser({ id }) {
```
jimmyhogoboom, line 25: Is this supposed to be named something else?

vinnyjth: woops! I think I added this method in twice while resolving one of the many merge conflicts ;)

jimmyhogoboom:
I'm getting an error when I run the example commentListFeature query: "invalid input syntax for type uuid: \"11\"". Seems like something somewhere is trying to use my Rock id as a postgres id.

vinnyjth:
@jimmyhogoboom do you have any steps to reproduce? I can't seem to trigger the error. What database are you pointed at? 

Screen Shot 2021-03-12 at 9 47 36 PM

jimmyhogoboom:
@vinnyjth Sorry, it's working now. Not sure what I did to fix it. I guess I had to yarn and link-packages one more time, but I swear I did that before. 🤷‍♂️

jimmyhogoboom (commented):
Looks good! Just one question: should we remove the person field from the `Comment` type in the GraphQL schema? That's not returning anything anymore, right?

vinnyjth:
@jimmyhogoboom It should be :) The logic is now all in the resolver, I call `getPerson` on the comment object and lean on the defined relationships.

packages/apollos-data-connector-postgres/src/comments/dataSource.js
```diff
@@ -65,14 +64,6 @@ class CommentDataSource extends PostgresDataSource {
 
     return comments;
   }
-
-  async getPerson({ id }) {
```
jimmyhogoboom, line 69: @vinnyjth OK I see where `getPerson` is called, but it appears to have been removed? Did we lose it in a merge?

vinnyjth: It's a "magic method" on the `comment` model. Notice how I don't call `dataSource.getPerson()` but `root.getPerson()`

jimmyhogoboom: Ahh ok, thank you! Got it working I think. All the test comments you had in there were showing `null` for the person, and so were my own until recently. But now my name is showing up. Not sure what changed... again...

jimmyhogoboom (approved):
Despite all the trouble I inflicted on myself, this does actually work! 😅

vinnyjth:
I think one of our production syncs wiped out peopleIds, I saw the same thing, but upon checking saw that the peopleIds were null. Adding comments seemed to be adding personIds as expected. Likely caused by my other comment pr schema changes :)

PR #1798, opened 2021-03-15
chore: Use Postgres uuids wherever possible

vinnyjth:

We opted to make the apollos ids `${typename}:{database_uuid}`. The shovel does this by default, but your API still was using a different strategy. 

### How do I test this PR?

Easiest way to test is add some comments and check their ID. 

```
mutation addComment {
	addComment(parentId:"WeekendContentItem:f8460101368b38c5aa9bbea8b2b8a625", text:"testing this stuff out :) today 2", visibility:PUBLIC){
    id
  }  
}
```


jimmyhogoboom (approved):
Good call!

PR #1799, opened 2021-03-15
chore: Added simple / inline api setup

vinnyjth:
## DESCRIPTION

@jimmyhogoboom if you test this, don't spend too much time on it if it gives you problems. This is a quick stab on a way to do API development in only one repo. Follow the instructions in the readme.md and let me know what you think.


redreceipt:
I do like this idea a lot. correct me if I'm wrong, taking this philosophy to its end, we could eliminate the need for linking all together in the templates repo?

vinnyjth:
If we also do this for the App, yes!

vinnyjth:
This has some unique differences to linking, if you make any changes to the Api/App they are lost by default. You need to copy them over to `templates`, or make them in `templates` first. I think that's okay as long as linking is technically possible (even if not the default)

redreceipt:
I think if we do this, we remove linking completely. the advantage of this in my mind is the infrastructure simplicity it brings to the templates repo. 

I don't think it's a bad thing that if there's specific template work that needs to be done, you do it there first with a PR, then work on your apps PR with the updated template.

redreceipt:
said another way, if the only reason we do this is to make our lives easier on development, I don't think it's worth it

vinnyjth:
👍 I don't 100% agree, but I tend to be more pessimistic with our future needs. I'm fine ditching linking and giving just this strategy a shot.

redreceipt:
we can slow play it. leave our templates repo unlinked and pretend as if we don't have it and try it for a month

scripts/setup-api.sh
```diff
@@ -0,0 +1,35 @@
...
+shopt -s dotglob
+mv apollos-templates/apollos-church-api/* ../ApollosApi
+
+
+
+cd ..
+
+node scripts/swap-package-json-to-links.js ./ApollosApi
+
+rm -rf tmp
```
redreceipt, line 35: this bash script makes my heart happy

scripts/setup-api.sh
```diff
@@ -0,0 +1,35 @@
...
+mkdir -p tmp
+
+echo $2
+BRANCH=$2
+
+cd tmp
+git clone git@github.com:ApollosProject/apollos-templates.git
+git checkout ${BRANCH:-master}
+
+cd apollos-templates
```
redreceipt, line 14:
```suggestion
cd apollos-templates
git checkout ${BRANCH:-master}
```

scripts/setup-api.sh
```diff
@@ -0,0 +1,35 @@
...
+
+mkdir -p ApollosApi
+mkdir -p tmp
+
+echo $2
+BRANCH=$2
+
+cd tmp
+git clone git@github.com:ApollosProject/apollos-templates.git
+git checkout ${BRANCH:-master}
```
redreceipt, line 12:
```suggestion
```

scripts/setup-api.sh
vinnyjth: good catch

redreceipt (approved):

PR #1804, opened 2021-03-17
fix: correctly build content sharing URL

redreceipt:
Not sure what we were using the `channel.itemUrl` for but it was blank on core Rock so we can assume it wasn't working for other folks

Screen Shot 2021-03-17 at 10 35 47 AM


packages/apollos-server-core/src/index.js
```diff
@@ -106,6 +106,10 @@ export const createContext = (data) => ({ req = {} } = {}) => {
   // For all non-datasource connectors. Right now only `Node`.
   const models = getModels(data);
   let context = {
+    requestUrl:
+      req.protocal && req.headers
```
vinnyjth, line 110: I'm stuck on this. Most of the time the content won't be live on the `host` of the API. Should this be a config.yml url (but maybe named better than before).

I don't have good ideas around a sane default since we always customize this pretty heavily for each client. Should we maybe return `""` to make that clear? I also like the idea of returning a deep  / universal link.

redreceipt: Valid, I'm assuming universal links here. Essentially we need to answer the question "what is the Apollos default for `ContentItem.sharing.url`. @conrad-vanl do you have any thoughts here?

We need to do something because currently it's broken, most of the time now, it just returns the API URL which isn't right.

conrad-vanl: > what is the Apollos default for ContentItem.sharing.url

I think the default is a universal link using the api url that redirects you to the app or App Store. That is the only thing that _can_ work out of the box with zero configuration, so I think we should make it work out of the box.

redreceipt: Ok that's exactly what I was hoping to accomplish with this PR

packages/apollos-data-connector-rock/src/content-items/__tests__/__snapshots__/data-source.tests.js.snap
```diff
@@ -10,6 +10,8 @@ Array [
 ]
 `;
 
+exports[`ContentItemsModel creates a sharing URL 1`] = `"undefined/app-link/content/WeekendContentItem:ae01c3f623e7f125f60cfcabd06e65cf"`;
```
redreceipt, line 13: i can't figure out why the config loader isn't working here

redreceipt:
alright @vinnyjth take a look, front end PR will come later, we're able to take liberties here because the functionality is already broken and everyone is already overwriting the `getShareUrl` function

packages/apollos-data-connector-rock/src/content-items/__tests__/__snapshots__/data-source.tests.js.snap
redreceipt: it works properly on the CI, causing the test failure

vinnyjth (approved):

PR #1805, opened 2021-03-17
chore: Add GIN name index to optimize searching by name

jimmyhogoboom:
## DESCRIPTION
Speed up searching by full name in the people table.

This adds an index on `lower("firstName" || ' ' || "lastName")` to allow efficiently searching via full name like "Jimmy Hogoboom". Note that the input search term will need to be made lowercase ("jimmy hogoboom").

### How do I test this PR?
Try running `EXPLAIN ANALYZE SELECT "firstName" || ' ' || "lastName" FROM people WHERE lower("firstName" || ' ' || "lastName") LIKE '%vincent wilson%'` before and after applying this PR to see the speed difference.

### DIFF
--- a/packages/apollos-data-connector-postgres/src/people/migrations/003_add_names_index.js
+++ b/packages/apollos-data-connector-postgres/src/people/migrations/003_add_names_index.js
@@ -4,8 +4,11 @@ async function up({ context: queryInterface }) {
   );
 }
 
-async function down() {
-  // no
+async function down({ context: queryInterface }) {
+  await queryInterface.sequelize.removeIndex(
+    'people',
+    'in_people_firstName_and_lastName'
+  );
 }
 
 const name = '003-add-name-index';
--- a/packages/apollos-data-connector-postgres/src/people/migrations/003_add_names_index.js
+++ b/packages/apollos-data-connector-postgres/src/people/migrations/003_add_names_index.js
@@ -1,7 +1,15 @@
+import Sequelize from 'sequelize';
+
 async function up({ context: queryInterface }) {
-  await queryInterface.sequelize.query(
-    `CREATE index in_people_firstName_and_lastName ON people USING gin (lower("firstName" || ' ' || "lastName") gin_trgm_ops)`
-  );
+  await queryInterface.addIndex('people', {
+    name: 'in_people_firstName_and_lastName',
+    using: 'gin',
+    fields: [
+      Sequelize.literal(
+        'lower("firstName" || \' \' || "lastName") gin_trgm_ops'
+      ),
+    ],
+  });
 }
 
 async function down({ context: queryInterface }) {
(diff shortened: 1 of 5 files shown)


packages/apollos-data-connector-postgres/src/people/migrations/003_add_names_index.js
```diff
@@ -0,0 +1,13 @@
+async function up({ context: queryInterface }) {
+  await queryInterface.sequelize.query(
+    `CREATE index in_people_firstName_and_lastName ON people USING gin (lower("firstName" || ' ' || "lastName") gin_trgm_ops)`
+  );
+}
+
+async function down() {
+  // no
```
vinnyjth, line 8: 😂 I love the idea of having this as the `down` for non-column migrations. However, we should probably play it safe and use `removeIndex` :) 

https://sequelize.org/master/class/lib/dialects/abstract/query-interface.js~QueryInterface.html#instance-method-removeIndex

packages/apollos-data-connector-postgres/src/people/migrations/003_add_names_index.js
```diff
@@ -0,0 +1,13 @@
+async function up({ context: queryInterface }) {
+  await queryInterface.sequelize.query(
```
vinnyjth, line 2: We have the option of using `addIndex` here. Which method do you prefer? 

https://sequelize.org/master/class/lib/dialects/abstract/query-interface.js~QueryInterface.html#instance-method-addIndex

packages/apollos-data-connector-postgres/src/postgres/migrations/001_add_pg_trgm_extension.js
```diff
@@ -0,0 +1,13 @@
+async function up({ context: queryInterface }) {
+  await queryInterface.sequelize.query(
+    'CREATE EXTENSION IF NOT EXISTS "pg_trgm";'
```
vinnyjth, line 3: Awesome :D Crazy how these `pg` extensions work

vinnyjth (changes requested):
image

Fantastic work @jimmyhogoboom! Elegant and speedy solution. a few nitpicks with the migration and then I think we're off to the races!

packages/apollos-data-connector-postgres/src/people/migrations/003_add_names_index.js
jimmyhogoboom: I honestly think it would be better to use `addIndex`, but I opted for this method purely in the interest of time. Maybe I'll work on the follow auto-accepting first, and if I get that done quickly enough I'll come back and update this.

vinnyjth:
Awesome, thanks for calling that out :) good prioritizing and I appreciate the compromise. I’ll take a quick stab at this tomorrow, and if I can’t solve it in 5 mins, we can merge it as is :D

vinnyjth (approved):

jimmyhogoboom:
Nice!

PR #1806, opened 2021-03-18
feat: Push notifications on follow request

vinnyjth:

Companion PR here: https://github.com/ApollosProject/apollos-templates/pull/296

This is a fun one ... 

Adds in a check for the `actions` that OneSignal sends over, and then lets users provide a custom function to fire when an action matches a passed prop. You can see that being used here: 
https://github.com/ApollosProject/apollos-templates/pull/296/files#diff-9b1509d1d4517465bef7cbb0733f3cf179f1e9b555e50041214d2103218846e6R21-R28

Also adds a call to onesignal to send a notification when a user requests to follow another user. That part is straightforward.

### How do I test this PR?

This is what I did.

1. Clone this PR and it's companion. you need to have a database_url and run postgres.
2. Triple check your env variables. You should have an OneSignal api key and id in your api, and a matching one signal id in your client. 
3. Boot android locally. Check in onesignal that your device showed up. If it doesn't, double check the names of your API keys. 
4. Using graphql, send a follow request to whoever is logged into your android sim. You'll have to use the postgres server to find your apollos id. 
5. Accept that follow request in the sim! 

Note! If you view the notification and then dismiss it, or pull down poorly on the top bar and half view the notification, the button goes away. It's part of the android ui, and it's less of a problem on a real device where it's easy to pull the top bar up and down.

### DIFF
--- a/packages/apollos-ui-notifications/src/Provider.js
+++ b/packages/apollos-ui-notifications/src/Provider.js
@@ -92,7 +92,6 @@ class NotificationsInit extends Component {
   };
 
   onOpened = (openResult) => {
-    console.log(openResult);
     console.log('Message: ', openResult.notification.payload.body);
     console.log('Data: ', openResult.notification.payload.additionalData);
     console.log('isActive: ', openResult.notification.isAppInFocus);
--- a/packages/apollos-ui-notifications/src/Provider.js
+++ b/packages/apollos-ui-notifications/src/Provider.js
@@ -101,15 +101,15 @@ class NotificationsInit extends Component {
     // apolloschurchapp://SomethingElse/Connect
     // apolloschurchapp://SomethingElse/ContentSingle?itemId=SomeItemId:blablalba
     const url = get(openResult, 'notification.payload.additionalData.url');
-    if (url) {
-      this.navigate(url);
-    } else if (
+    if (
       openResult?.action?.actionID &&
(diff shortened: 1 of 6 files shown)


redreceipt:
This is awesome! We're blowing past Rock's capabilities with this PR. I'll test soon.

packages/apollos-ui-notifications/src/Provider.js
```diff
@@ -98,6 +103,13 @@ class NotificationsInit extends Component {
     const url = get(openResult, 'notification.payload.additionalData.url');
     if (url) {
       this.navigate(url);
+    } else if (
```
conrad-vanl, line 106: Should this be an `else if` or just a separate `if` ? Is there a scenario where we'd have a push notification with an action that also opens a URL?

conrad-vanl (approved):
Flow makes sense and is pretty simple. One question but good to merge

packages/apollos-ui-notifications/src/Provider.js
vinnyjth: Good callout. I think we actually want the logic as it is, but reversed. Look for an action first, if we don't have one, check for a URL.

vinnyjth: That way we can use `urls` as fallbacks for buttons with no associated logic deployed (like older apps)

PR #1809, opened 2021-03-18
chore: deprecate ContentSingle

redreceipt:
Moving to `NodeSingle`

### DIFF
--- a/packages/apollos-ui-connected/src/ContentNodeConnected/getContentNode.js
+++ b/packages/apollos-ui-connected/src/ContentNodeConnected/getContentNode.js
@@ -6,6 +6,7 @@ export default gql`
     node(id: $nodeId) {
       id
       ...ContentNodeFragment
+      # TODO: ContentSingleFragment deprecated
       ...ContentSingleFragment
     }
   }
--- a/packages/apollos-ui-connected/src/ContentSingleFeaturesConnected/ContentSingleFeaturesConnected.js
+++ b/packages/apollos-ui-connected/src/ContentSingleFeaturesConnected/ContentSingleFeaturesConnected.js
@@ -29,16 +29,16 @@ const ContentSingleFeaturesConnected = ({
         const features = get(node, 'features', []);
         if (!features || !features.length) return null;
 
+        // TODO deprecated
+        console.warn(
+          'ContentSingleFeaturesConnected is deprecated. Please use ContentFeatureFeedConnected.'
+        );
         return (
-          console.warn(
-            'ContentSingleFeaturesConnected is deprecated. Please use ContentFeatureFeedConnected.'
-          ) || (
-            <Component
-              contentId={contentId || nodeId}
-              features={features}
-              {...props}
-            />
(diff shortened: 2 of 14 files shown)


vinnyjth:
@redreceipt My bad, I misread your question. `ContentSingle` is still being used actively (and should still be used). It includes things like the MediaPlayer that aren't included in `NodeSingle`. I thought you were talking about `ContentChannelItem` (which is deprecated, and I think even deleted) 

Before we merge this PR (which I think is a great goal) we need to consolidate NodeSingle and ContentSingle


https://github.com/ApollosProject/apollos-templates/blob/5f9848476fbaf84a64f8b2735c692b694da15fd2/apolloschurchapp/src/node-single/NodeSingle.js#L14
https://github.com/ApollosProject/apollos-templates/blob/5f9848476fbaf84a64f8b2735c692b694da15fd2/apolloschurchapp/src/content-single/ContentSingle.js#L20

redreceipt:
@vinnyjth those files are client and can be consolidated after this. Are there features in the apps repo where the two components don't match? This PR doesn't delete anything, just throws warnings. So the client shouldn't break.

vinnyjth:
> This PR doesn't delete anything, just throws warnings. So the client shouldn't break.

Feels weird to throw warnings for something that we haven't correct on the client / templates. "Hey you should be doing this, but this isn't up to feature parity with current approach yet".  Given how similar NodeSingle and ContentSingle are, I'm wondering if we should actually ditch `NodeSingle` and just have ContentSingle in the template. Since it looks like ContentSingle is just NodeSingle + theme mixin, analytics, and media player support.

redreceipt:
so you're suggesting ditching `NodeSingle` completely? Deprecating here as well?

vinnyjth:
> so you're suggesting ditching `NodeSingle` completely? Deprecating here as well?

yah, instead of ditching ContentSingle. Given that we actually use ContentSingle, is NodeSingle is just a less built out version of ContentSingle. In theory anything that you need to render with NodeSingle you can instead render with ContentSingle, but the inverse is not true.

redreceipt:
it was created for the `OPEN_NODE` feature action I believe which is more general than `READ_CONTENT`

vinnyjth:
> it was created for the `OPEN_NODE` feature action I believe which is more general than `READ_CONTENT`

Yah that's true, but that was before `ContentNode` when the only other option was `ContentItemSingle`

redreceipt:
Ok, I'll play around with that idea. The reason I have to start in this repo and not templates is that I have to ensure nothing is trying to navigate there before removing it from templates.

vinnyjth:
That's fair! And I'm sorry for the confusion I caused this morning, I should have ready your slack message closer 😬

redreceipt:
easily misunderstood, we have like 100 components with "Content" and "Node" in them. I'm just trying to make that better :)

vinnyjth:
amen! so many nodes and contents and features 🤯

PR #1814, opened 2021-03-18
feat: useTheme hook

redreceipt:
I've been wanting to try this for awhile and it turned out way simpler than I thought. It has _many_ advantages over our HOC system:

1. _MUCH_ simpler backend code to acheive the same thing
2. Greater transparency when writing components that use theme
3. Smaller front end code footprint
4. Move from antiquated HOC paradigm to React core context
5. Smaller render trees
6. Plus...hooks

I made it opt-in, it can be tested using the `use-theme` branch in templates. The search button is using it. Let me know your thoughts.

Screen Shot 2021-03-19 at 7 57 49 AM


vinnyjth:
I really like this, but I don't think it replaces `withTheme`. Main reason is override support, and it also encourages less separation of concerns when building out components. I think you're on board with that, given how you're using it on templates (where overrides don't matter and there's good reason to have good component separation). I would like @conrad-vanl's feedback on this PR

redreceipt:
I LOVE your `named` function which is all we need for overrides. Theme could be handled by this completely. I'm a fan of having one way to do things and do it well.

redreceipt:
Does `const SearchIcon = named("SearchIcon")(Icon)` not work for what you're describing?

redreceipt:
It's confusing for newcomers to the platform that we use the `withTheme` HOC to do deep prop overrides. We lucked into that functionality and it works but I think it's reasonable to use the `named` function as the sole method to do that.

vinnyjth:
> Does `const SearchIcon = named("SearchIcon")(Icon)` not work for what you're describing?

It could do that, but then I think we're back to having more tools than needed. One tool to name a component, one tool to theme it. It's easy to get wrong, though, since if you `name` higher in the stack than you `useTheme`, your `name` won't apply. 

Going back to first principles, I like this PR! I want this hook. I want to make sure we can agree on when it should be used and when it shouldn't. be used. It seems nice for making changes on templates, but we probably should avoid it for the most part in `apps` for now.

conrad-vanl:
Fwiw, I think separating theme and name is probably a good thing. One tool for theming one tool for naming and replacing components.

vinnyjth:
> Fwiw, I think separating theme and name is probably a good thing. One tool for theming one tool for naming and replacing components.

That's helpful! 

So @redreceipt - I guess my ask would be if there's a way we could help idiot proof uses of `useTheme` inside a `named` component? I don't have any good ideas of what that would look like off the top of my head, but here's an example to work with. 

```
const HomeSearchButton = ({ onPress }) => {
  const theme = useTheme();
  return (
    <Animated.View style={styles.container(theme)}>
      <Touchable onPress={onPress}>
        <Icon
          name={'search'}
          size={theme.sizing.baseUnit * 2}
          fill={theme.colors.primary}
        />
      </Touchable>
    </Animated.View>
  );
};
```

vinnyjth:
this is the naive solve that's functionally equiv to how this would have worked in the past.

```
const HomeSearchButton = ({ onPress, name, size, fill, theme: customTheme }) => { // or ...props
  const theme = useTheme();
  return (
    <Animated.View style={styles.container({ ...theme, ...customTheme })}> // this won't actually work because prop spreading isn't deep but ignore that for now
      <Touchable onPress={onPress}>
        <Icon
          name={name || 'search'}
          size={size || theme.sizing.baseUnit * 2}
          fill={fill || theme.colors.primary}
           // or {...props}
        />
      </Touchable>
    </Animated.View>
  );
};

export default named('HomeSearchButton')(HomeSearchButton);

```

conrad-vanl:
This is using two separate themes now right?

Is there a way we can migrate the old theme HOC to use the new useTheme hook under the hood?

redreceipt:
@vinnyjth would this not work?

```

const SearchIcon = named('SearchIcon')(Icon)

const HomeSearchButton = ({ onPress }) => {
  const theme = useTheme();
  return (
    <Animated.View style={styles.container(theme)}>
      <Touchable onPress={onPress}>
        <SearchIcon
          name={'search'}
          size={theme.sizing.baseUnit * 2}
          fill={theme.colors.primary}
        />
      </Touchable>
    </Animated.View>
  );
};

export default named('HomeSearchButton')(HomeSearchButton);
```

conrad-vanl:
And @vinnyjth it might be simpler then that....named() is still wrapping the component that is using the hook, so if you override theme for that named() component, shouldn't it be passed down?

vinnyjth:
> And @vinnyjth it might be simpler then that....named() is still wrapping the component that is using the hook, so if you override theme for that named() component, shouldn't it be passed down?

If I understand `named` right, it doesn't change the theme values, it merges in the override with the outcome of the `themeMapper` function.

conrad-vanl:
Then what is your concern? You’re saying you want people to be able to overrride theme when name is used?

packages/apollos-ui-kit/src/Providers.js
```diff
@@ -1,6 +1,15 @@
...
 
-import { ThemeProvider } from './theme';
+import { ThemeProvider, ThemeContext, createTheme } from './theme';
 import { LayoutProvider } from './LayoutContext';
 
-export default nest(ThemeProvider, LayoutProvider);
+// TODO: probably don't need to nest here, we should just export LayoutProvider after
+// deprecating the old ThemeProvider for more transparency with props
+export default nest(
+  ThemeProvider,
```
conrad-vanl, line 10: I really think there’s a way we can remove the old ThemeProvider component and make the old HOCs use the new theme. We’d need to update styled

redreceipt: yes agree, there's a lot of work that can be done later to deprecate. this PR just gives us an opt-in way to play around with the new method

vinnyjth:
> would this not work?

It would, yah! And I guess the `AnimatedView` in the example doesn't really need it's values overridden.

> Then what is your concern? You’re saying you want people to be able to overrride theme when name is used?

I think @redreceipt is suggesting we replace/migrate using `withTheme` with `named` and `useTheme`. If we do that and aren't careful to nest `named` underneath whatever is passing theme values down, then `named` won't be able to override the values it used to be able to override when it was `withTheme`.

packages/apollos-ui-kit/src/Providers.js
conrad-vanl: I think keeping both around could cause weird bugs

conrad-vanl: Since we have two separate theme state objects

redreceipt:
I think it's important to think of `named` as a prop override HOC and nothing more. and `useTheme` as a way to access the theme context, nothing more. Those two functions should get us everything we want.

packages/apollos-ui-kit/src/Providers.js
conrad-vanl: Like, styled still uses the old theme provider. Theme can be changed within child components, but that would only effect on one theme object. 

Theme isn't just global, it's mutable and with two theme objects that's going to introduce bugs.

vinnyjth:
> I think it's important to think of `named` as a prop override HOC and nothing more. and `useTheme` as a way to access the theme context, nothing more. Those two functions should get us everything we want.

That's very helpful, I like where this is going.

packages/apollos-ui-kit/src/Providers.js
vinnyjth: Yah that's a good point @conrad-vanl. We could test this really easily by using `useTheme` inside a <ThemeMixin>. Migrating the context to use the new context might be easier than you expect.

conrad-vanl (changes requested):
We need a test case like this in the code base, and it needs to pass before we can safely merge this PR, or I think there could be breaking changes with using the new `withTheme` component and not getting proper values.

```jsx
import { Providers, withTheme, ThemeConsumer, ThemeMixin } from '@apollosproject/ui-kit';

const TestCase = () => {
  const newTheme = useTheme();
    return(
      <ThemeConsumer>
        {(oldTheme) => (
          newTheme.colors.paper === oldTheme.colors.paper
        )}
      </ThemeConsumer>
    );
}

const App = () => (
  <Providers>
    <ThemeMixin mixin={{ colors: { paper: 'RED' } }}>
      <TestCase />
    </ThemeMixin>
  </Providers>
);
```

packages/apollos-ui-kit/src/Providers.js
redreceipt: they use two separate inputs, `themeInput` and `customTheme`, with the new one being a higher priority in the nest. Not sure if this is an issue. I could test to make sure.

conrad-vanl: The problem is that components that use the new `withTheme` hook could get different values for theme then components that use the old theme provider. That is bad, as it will cause visual inconsistencies.

redreceipt:
I'll add that test

redreceipt:
@conrad-vanl that test comes out false but I'm not sure what it accomplishes. In the case of the `TestCase` component, we would switch from `Mixin` to `useTheme`, not leave in both.

redreceipt:
Are you saying, `useTheme` should also inherit from `ThemeMixin`?

conrad-vanl:
@redreceipt there are instances in the code base that rely on that functionality today. The way cards and content items are themed with data coming in from GraphQL. By mixing theme stores, that creates a high chance of having issues.

Say we built a component in storybook using the `useTheme` hook. It works great. Then, later it starts getting used in the app. We add it to a node single view, still looks great on core. But then when it rolls out to newspring which uses themed content singles, it breaks.

I'd like to avoid that.

conrad-vanl:
> Are you saying, `useTheme` should also inherit from `ThemeMixin`?

Not necessarily. There are two HOCs - `withTheme` and `themeMixin` - that I think we need to rewrite to use your new theme store. That level of difficulty may be high or low.

conrad-vanl:
`withTheme` will be easy to replace. Just need to write a wrapper component that uses `useTheme` internally.

`themeMixin` will be a little harder. It'll need to render a new `Provider` component that merges the parent theme with the values from the mixin. 

Good news is there are pretty good tests written that can confirm compatibility :)

conrad-vanl (commented):
@redreceipt here's another thought:

The use case of the theme mixin is to be able to alter theme values for the current component, as well as all children of that component. Use cases vary, but they are used in cases like:

- Buttons: a primary button uses a different subset of theme (ie...dark theme) so that text and icons inside of it automatically render as "white" instead of "black"
- Themed experiences, such as Newspring content items

I think if you added this capability to your current `useTheme` hook, I think it then becomes really easy to deprecate the old components and make them point at the new context.

The API could be as simple as whatever object you pass in as the argument to `useTheme`, could be merged in and set as the new theme for that component and all children:

`useTheme(newThemeValues)` (@vinnyjth @redreceipt would like feedback on this API)

So, here's a test case to see if you can get working:

```jsx
const MyThemedComponent() => {
   // ability to pass theme values into the hook, for the purpose of mutating the theme store
   // This should only change theme for this component and children
  const theme = useTheme({ colors: { paper: 'green' }});

  return (
    <>
      {theme.colors.paper === 'green'}
      <MyInnerComponent />
    </>  
  );
};

MyInnerComponent = () => {
  const theme = useTheme();
  return (
    {theme.colors.paper === 'green'}
  );
};

App = () => (
  <Providers theme={{ colors: { paper: 'red' } }>
    {theme.colors.paper === 'red'}
    <MyThemedComponent />
  </Providers>
);
```

redreceipt:
Interesting idea. Maybe it overloads the `useTheme` hook a bit but not bad. Still doesn't solve the backward compatibility issue though right? I feel like you would just pass in `green` as a prop here, the child you need to change wouldn't be that far from the source.

redreceipt:
I think a good start is simply figuring out how to make the `Mixin` HOC override the new hook first.

conrad-vanl:
> I think a good start is simply figuring out how to make the `Mixin` HOC override the new hook first.

Go for it. If you run into challenges, the above might be simpler 😉

redreceipt:
ok @conrad-vanl @vinnyjth I'm very proud of where this landed. I think the API is super clear and covers all the bases. What are y'alls thoughts?

```jsx
import React from 'react';
import { Text } from 'react-native';
import renderer from 'react-test-renderer';
import Themer, { useTheme, Theme } from './Themer';

describe('Themer', () => {
  it('should pass theme down through useTheme hook', () => {
    const Paper = () => {
      const theme = useTheme();
      return <Text>{theme.colors.paper}</Text>;
    };
    const App = () => (
      <Themer>
        <Paper />
      </Themer>
    );
    const tree = renderer.create(<App />);
  });

  it('has a backwards compatible consumer for class components', () => {
    class Paper extends React.PureComponent {
      render() {
        return <Theme>{(theme) => <Text>{theme.colors.paper}</Text>}</Theme>;
      }
    }
    const App = () => (
      <Themer>
        <Paper />
      </Themer>
    );
    const tree = renderer.create(<App />);
  });

  it('should allow for updating the theme provider', () => {
    const Paper = () => {
      const theme = useTheme();
      return <Text>{theme.colors.paper}</Text>;
    };
    const PaperParent = () => {
      const theme = { colors: { paper: 'red' } };
      return (
        <Themer theme={theme}>
          <Paper />
        </Themer>
      );
    };
    const App = () => {
      return (
        <Themer>
          <PaperParent />
        </Themer>
      );
    };
    const tree = renderer.create(<App />);
  });
});
```

packages/apollos-ui-kit/src/theme/Themer.js
```diff
@@ -0,0 +1,57 @@
+import React, { createContext, useContext } from 'react';
```
vinnyjth, line 1: Could per-chance use this component inside the old `ThemeMixin` component so that `useTheme` would be mostly backwards compatible with old theme api? I think it would be an easy win without much added complexity or overhead.

redreceipt:
As long as we leave the old `themeInput` prop in the providers, the two systems should work fine independently. I don’t see doing extra work to retrofit a deprecated system that still works. Is there some functionality that will break that I’m not seeing?

vinnyjth:
> As long as we leave the old `themeInput` prop in the providers, the two systems should work fine independently. I don’t see doing extra work to retrofit a deprecated system that still works. Is there some functionality that will break that I’m not seeing?

Conrad's example of a content item with a custom theme. We'll have to either use Themer and ThemeMixin and migrate all child components to use the old or new theme system. If we keep the themes in sync by having `ThemeMixin` pass it's theme to `Themer` I think we can avoid that scenario.

redreceipt:
that will be a breaking change though as they will have to make sure they add the new prop, we can migrate all components slowly on core with no breaking changes if we leave them separate.

packages/apollos-ui-kit/src/theme/Themer.js
redreceipt: @conrad-vanl @vinnyjth revisiting this:

I used a separate prop and context because if I forced `ThemeMixin` to pull from the new context, it increases the surface area of this code immediately for everything. Having a separate context allows us (and partners) to slowly move everything over by using the new prop: `<Providers theme={theme} />` and then in child components, we can replace `withTheme` with the new context tools. What am I missing?

redreceipt:
I'd like to move this forward but I'm a little unclear on what is being asked? So we want `ThemeMixin` to use my new context as a priority over the old context?

redreceipt:
Would simply passing in the new prop to override the old prop in the old `ThemeProvider` get us what we want?

redreceipt:
like this? 
Screen Shot 2021-05-04 at 4 39 47 PM

redreceipt:
I'm going to strip out the guts of the old functions and replace them with the logic and context of the new stuff. And then make sure it all still works.

redreceipt:
ok so now the old functions are using the new context, but I can't get all tests to pass so something isn't quite right...

redreceipt:
@vinnyjth @conrad-vanl ... I think it's ready

packages/apollos-ui-kit/src/Avatar/AvatarList/__snapshots__/AvatarList.tests.js.snap
```diff
@@ -21,15 +21,15 @@ Array [
   />,
   <RCTScrollView
     ListHeaderComponent={
-      <mapProps(getContext(mapProps(withPropsOnChange(mapProps(View)))))>
```
vinnyjth, line 24: yes, yes YES! Love these snapshots changes. One of the biggest advantages of moving away from `recompose`

packages/apollos-ui-kit/src/theme/OLD_theme.js
```diff
@@ -0,0 +1,55 @@
...
+      : {};
+
+  return (
+    <Component {...props} {...themeProps} {...overrideProps} {...styleProps} />
+  );
+};
+
+/* eslint-disable-next-line */
+export const withThemeMixin = (input) => (Component) => (props) => {
+  const theme = useTheme();
```
vinnyjth, line 39: really love the way this integrated.

vinnyjth:
I think we're really close here. The code looks excellent, but I'm running into a few bugs in the app. 

The first one is a crash bug when the HomeFeed loads. 

image

The second (and I can't really screenshot it yet) is that I noticed a grey line above the tab bar before the crash from above. I will get a pic as soon as I can glimpse behind this crash.

vinnyjth:
Simulator Screen Shot - iPhone 11 - 2021-06-15 at 15 23 14

vinnyjth:
@redreceipt We're getting really close. I fixed the red screen error above. I've noticed two interrelated issues: 

1. Dark mode doesn't work (I can't get the app into dark mode)
2. The line above the tab bar shows up, but only when in dark mode. It's like that single border is in dark mode ;)

redreceipt:
@vinnyjth got it

packages/apollos-ui-kit/src/GradientOverlayImage/index.js
```diff
@@ -5,7 +5,7 @@ import PropTypes from 'prop-types';
 import { pure } from 'recompose';
 
 import styled from '../styled';
-import withTheme from '../theme/withTheme';
+import { withTheme } from '../theme/OLD_theme';
```
conrad-vanl, line 8: Why does this use `OLD_theme` now?

packages/apollos-ui-kit/src/theme/OLD_theme.js
conrad-vanl: Agreed.

conrad-vanl (approved):
This is like ⚡️in a 🍾

conrad-vanl:
Holy moly you merged that quickly

PR #1820, opened 2021-03-23
fix: Don't search non-apollosUsers and trim whitespace

vinnyjth:

### DIFF
--- a/packages/apollos-data-connector-postgres/src/people/__tests__/dataSource.js
+++ b/packages/apollos-data-connector-postgres/src/people/__tests__/dataSource.js
@@ -124,7 +124,7 @@ describe('Apollos Postgres People DataSource', () => {
       apollosUser: true,
     });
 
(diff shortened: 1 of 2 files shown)


redreceipt:
Just for my own understanding, why do we have people in the DB that aren't searchable?

vinnyjth:
> Just for my own understanding, why do we have people in the DB that aren't searchable?

Our rock people puller is setup to pull all user profiles from rock. We could pull _just_ apollos user profiles but that has a few issues
1. We would have to add code that would go fetch an existing user's profile from rock (or create it) when they login for the first time. Not fun. 
2. It's actually not that easy to tell if a person is an apollos user. best bet is to look at person -> personal alias -> personal device. It would slow down our initial people dump to do that filtering
3. in the future we plan on pulling groups into postgres. that gets tricky if groups are full of mixes of apollos and non-apollos people.

redreceipt (approved):

PR #1821, opened 2021-03-23
fix: Pull to refresh not working consistently

vinnyjth:

This functionality was behaving oddly because of my invalid use of `hooks`. The `refetchFunctions` state was `{}` for both calls to `refetchRef`  most of the time the component was rendered, and so the last child to call `refetchRef` would be the only `refetch` set

The functionality below more closely mimics what we do in `FeaturesFeedConnected`


redreceipt:
How can I test to confirm this is working before and after the change?

vinnyjth:
> How can I test to confirm this is working before and after the change?

Good Q: Two good methods 

- Linked setup - Visit connect tab, clear API console, pull to refresh and make sure you see two queries firing. 
- Also Linked setup - Visit connect tab. Edit the code to `console.warn` `refetchFunctions`. Make sure you see an object with two key/values when you pull to refetch.

redreceipt (approved):

PR #1826, opened 2021-03-24
chore: do canary on stable releases

redreceipt:


vinnyjth (approved):

vinnyjth:
@vinnyjth If this works well we can close my PR instead of merging it.

PR #1827, opened 2021-03-24
feat: Onboard only seen slides

vinnyjth:

Gives the templates the ability to define what slides to show to users by defining versions. Users will have a `version` stored in async storage. If there are new slides with a higher version than your stored, version you are shown that slide after auth / when opening the app. 

### How do I test this PR?

Testing this is a pain, so I added some helpers

1. Login / Open the app. You should see the new suggested followers slide. 
2. Close / open the app, you should see no new slides.
3. Go to the test utilities tab and reset your onboarding seen status (click the second button). Now close and open the app. You should see the full onboarding.

You'll need this companion PR: https://github.com/ApollosProject/apollos-templates/pull/312

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,6 +55,9 @@ jobs:
     - name: Install Dependencies
       run: yarn
 
+    - name: Force publish canary with latest code
+      run: yarn release:canary --yes --force-publish
+
     - name: NPM Release
       run: |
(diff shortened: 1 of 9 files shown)


vinnyjth:
@conrad-vanl I don't need your review on this per-say, but I would love to hear any thoughts on the approach.

conrad-vanl (approved):
I think this makes sense based on what we discussed

PR #1828, opened 2021-03-24
chore: Login when releasing canary

vinnyjth:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,20 +55,12 @@ jobs:
     - name: Install Dependencies
       run: yarn
 
-    - name: Force publish canary with latest code
-      run: |
-        git config --global user.email "apps@apollos.app"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -56,7 +56,13 @@ jobs:
       run: yarn
 
     - name: Force publish canary with latest code
-      run: yarn release:canary --yes --force-publish
+      run: |
+        git config --global user.email "apps@apollos.app"
+        git config --global user.name "Apollos Admin"
+        yarn release:canary --yes --force-publish
```
redreceipt, line 62: drop this line down below line 70 and remove this CI step altogether and we should be good. i would but github won't let me suggest lines off the PR

vinnyjth: Done 👍

redreceipt (approved):

PR #1831, opened 2021-03-25
fix: find correct person when updating push notifications

redreceipt:


packages/apollos-data-connector-onesignal/src/data-source.js
```diff
@@ -48,7 +48,7 @@ export default class OneSignal extends RESTDataSource {
     if (pushProviderUserId != null) {
       await this.updateExternalUserId({
         playerId: pushProviderUserId,
-        userId: currentUser.primaryAliasId,
+        userId: primaryAliasId,
```
redreceipt, line 51: @vinnyjth my only question, why are we using alias id here? I still don't understand the difference

packages/apollos-data-connector-onesignal/src/data-source.js
```diff
@@ -36,8 +36,8 @@ export default class OneSignal extends RESTDataSource {
 
   async updatePushSettings({ enabled, pushProviderUserId }) {
     const { Auth, PersonalDevice, Person } = this.context.dataSources;
-    const currentUser = await Auth.getCurrentPerson();
-    const currentPerson = await Person.getFromId(currentUser.personId);
+    const { id, primaryAliasId } = await Auth.getCurrentPerson();
+    const currentPerson = await Person.getFromId(id);
```
vinnyjth: why was `currentPerson` null here? that seems like a red flag?

redreceipt: it wasn't, `personId` was the wrong field

redreceipt: you changed it on march 3rd, do you remember why?

packages/apollos-data-connector-onesignal/src/data-source.js
vinnyjth: my understanding: a user (UserLogin) can be attached to multiple people. those people are likely the same person, but have been de-duped. The "alias" is the person that represents and points to the "real" person. 

It's also the attribute we have to store in order to send push notifications from Rock. The rock plugin for whatever reason can't use `personId`

vinnyjth: I would love to change that and use `ApollosId`. /`guid` once Postgres is stable.

packages/apollos-data-connector-onesignal/src/data-source.js
vinnyjth: oh, that makes more sense. I needed to update teh code to return the `Person` instead of the `Auth.getCurrentPerson`. since they are/can be different. Auth.getCurrentPerson returns an "AuthPerson" which right now is a rock person, and `Person.getFromId` returns a postgres person OR a rock person, depending on the data connectors

packages/apollos-data-connector-onesignal/src/data-source.js
```diff
@@ -48,9 +47,9 @@ export default class OneSignal extends RESTDataSource {
     if (pushProviderUserId != null) {
       await this.updateExternalUserId({
         playerId: pushProviderUserId,
-        userId: currentUser.primaryAliasId,
+        userId: primaryAliasId,
       });
     }
-    return currentPerson;
+    return Person.getFromId(id);
```
vinnyjth, line 53:
```suggestion
    return Person.getFromId(id, null, { originType: 'rock' });
```

packages/apollos-data-connector-onesignal/src/data-source.js
vinnyjth: Good detective work 👍

packages/apollos-data-connector-onesignal/src/data-source.js
redreceipt: fill in the gaps for me. `Person.getFromId` doesn't have three args and the postgres DS doesn't even have that function

vinnyjth: Postgres ds: https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-data-connector-postgres/src/postgres/index.js#L73
Rock Person safely ignores the second two args :) 
`Node.get()` sets the precedent for three args: https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-server-core/src/node/model.js#L85

redreceipt: ah nice

vinnyjth (approved):

PR #1834, opened 2021-03-25
fix: allow adding images to Rock CMS

redreceipt:
Resolving images on the Rock CMS is actually much simpler than I thought it would be.


redreceipt:
use `rock-images` branch in `templates`

packages/apollos-data-connector-rock/src/content-items/resolver.js
```diff
@@ -2,20 +2,57 @@
...
+      allowedAttributes: {
+        a: ['href', 'target'],
+        img: ['src'],
+      },
+      transformTags: {
+        img: (tagName, attribs) => {
+          return {
+            tagName,
+            attribs: {
+              src: `${ROCK.URL || ''}${attribs.src}`,
```
vinnyjth, line 50: we should probably handle the situation where users are using already valid non-relative links. I think Newspring and LCBC have both trained their staff to swap in the correct full urls.

packages/apollos-data-connector-rock/src/content-items/resolver.js
```diff
@@ -2,20 +2,57 @@
...
-
 const { ROCK, ROCK_MAPPINGS } = ApollosConfig;
 
 export const defaultContentItemResolvers = {
   id: ({ id }, args, context, { parentType }) =>
     createGlobalId(id, parentType.name),
-  htmlContent: ({ content }) => sanitizeHtml(content),
+  htmlContent: ({ content }) =>
+    sanitizeHtml(content, {
+      allowedTags: [
```
vinnyjth, line 19: Awesome! glad to have this

redreceipt:
Shoot I meant to cover that case you’re right

redreceipt:
ok @vinnyjth check in place

packages/apollos-data-connector-rock/src/content-items/resolver.js
redreceipt: done

packages/apollos-data-connector-rock/src/sanitize-html.js
```diff
@@ -1,39 +0,0 @@
-import sanitizeHtml from 'sanitize-html';
-
-const allowedTags = [
```
vinnyjth, line 3: Any reason you moved this function out of its own file? The reason we moved it out of the resolvers initially was due to clients wanting to use it for other fields, but needing to copy / paste from a resolver to get its functionality. Seems nice to be able to export it and use it other places.

redreceipt: It wasn't being exported unless I completely missed it?

Besides it was confusing working in the resolver that we were essentially using a function by the same name as from the library except we were cutting out all its extra functionality. It took me a minute to realize that `sanitizeHtml` was not the right one and it just seemed like an unnecessary abstraction.

vinnyjth: Hah! my bad, looks like we just kept our copy/paste :D 

https://github.com/lcbc-digital/ephesus/blob/master/apollos-church-api/src/data/craft/ContentItem.js#L35

packages/apollos-data-connector-rock/src/content-items/resolver.js
```diff
@@ -2,20 +2,58 @@
...
+        a: ['href', 'target'],
+        img: ['src'],
+      },
+      transformTags: {
+        img: (tagName, { src }) => {
+          return {
+            tagName,
+            attribs: {
+              // adds Rock URL in the case of local image references in the CMS
+              src: src.startsWith('http') ? src : `${ROCK.URL || ''}${src}`,
```
vinnyjth, line 51: lgtm. The one thing we'll have to watch for is [protocol-less](https://tosbourn.com/serve-assets-without-protocol/) links. they aren't very common from what I've seen but will throw a wrench in this code if encountered.

vinnyjth (approved):

packages/apollos-data-connector-rock/src/content-items/resolver.js
redreceipt: I definitely wouldn't put it past Rock to change how images are referenced in the HTML in v13 or something. we can't get out fast enough

PR #1838, opened 2021-03-26
feat: <ContentTitles /> and other new components

conrad-vanl:

- Adds `<ContentTitles />` component
- Adds`<SocialBar />` component (to replace ActionBar and other usages of like/sharing buttons)
- Adds experimental Reanimated-based `<StretchyView />` component
- Adds `useLike` and `useShare` hooks
- Updates `NodeSingleConnected`, `ContentNodeConnected`, `DefaultCard`, and `FeaturedCard` to use new components, which starts the process of migrating some of our most commonly used card components to the designs we've been using for almost a year in Figma.

### How do I test this PR?

Stories:

- `NodeSingleConnected`
- `DefaultCard`
- `FeaturedCard`


conrad-vanl:
@vinnyjth yeah changes to files in the `theme/` directory could be moved out to a separate PR.

I do want to start migrating theme to fall in line with UI updates that are in Figma, so hopefully that PR can still be merged and not just put off.

conrad-vanl:
I could also try and .... adjust the theme colors some so that there's not a ton of snap changes while still updating theme

conrad-vanl:
Note: Going to close this in favor of a different approach

conrad-vanl:
@redreceipt this one kind of got away from me, but I think it's a net positive change:

- Introduces new components that match in both naming and props to our Figma UI-Kit
- Uses those components across Cards and Node views
- Reduces overall code footprint some 

This also leaves a couple of components that we can deprecate and remove:

- ContentCard (already deprecated but we can remove it at any time I think)
- LikeButtonConnected (should no longer be used on templates)
- ShareButtonConnected (should no longer be used on templates)

This is ready for review.

conrad-vanl:
Also throwing @rkdavidson on here as a potential reviewer

packages/apollos-ui-connected/src/LikeButtonConnected/useLike.js
```diff
@@ -0,0 +1,68 @@
...
+
+import { useQuery, useMutation } from '@apollo/client';
+
+import { useTrack } from '@apollosproject/ui-analytics';
+
+import UPDATE_LIKE_NODE from './updateLikeNode';
+import GET_LIKED_NODE from './getLikedNode';
+import updateLikedContent from './updateLikedContent';
+
+const useLiked = (nodeId) => {
```
redreceipt, line 11:
```suggestion
const useLike = (nodeId) => {
```

packages/apollos-ui-connected/src/LikeButtonConnected/useLike.js
```diff
@@ -0,0 +1,68 @@
...
+      });
+    } catch (e) {
+      throw e.message;
+    }
+  }, [likeInteraction, track, nodeId, isLiked, isLikeable]);
+
+  return [isLiked, like];
+};
+
+export default useLiked;
```
redreceipt, line 68:
```suggestion
export default useLike;
```

redreceipt:
this is awesome. couple of things:

the actionsheet is now showing up over tabs

Screen Shot 2021-06-30 at 10 39 54 AM

redreceipt:
content is now a bit washed out because I'm guessing there's no gradient

Screen Shot 2021-06-30 at 10 41 00 AM

redreceipt:
an accompanying templates PR to remove the like and share bar should come with this so we have an option for people to upgrade too if they want to use these new components

Screen Shot 2021-06-30 at 10 42 45 AM

conrad-vanl:
> the actionsheet is now showing up over tabs

This PR didn't change this.... at least to my knowledge.

> content is now a bit washed out because I'm guessing there's no gradient

Hm....yeah. I'll look into it. I know highlight cards we are either going to drop or change design of. I didn't mean to change those in this PR, so I'll dig in and see what I did.

BTW - how did you run the templates app while pointing at this branch?

redreceipt:
> BTW - how did you run the templates app while pointing at this branch?

`yarn app`

redreceipt:
> > the actionsheet is now showing up over tabs
> 
> This PR didn't change this.... at least to my knowledge.

this may have been a fluke, force closing out, I can't reproduce

redreceipt:
also, drop some `console.warns` in the deprecated components so we can remove at some point and say we warned people

packages/apollos-ui-kit/src/BackgroundView/index.js
```diff
@@ -1,17 +1,38 @@
...
+const BackgroundView = ({
+  material = 'screen',
+  flexed = true,
+  style: styleProp,
+  ...viewProps
+}) => {
+  const theme = useTheme();
+  const backgroundColor =
+    theme?.colors?.background[material] || theme?.colors?.background?.screen;
+  const style = useMemo(
```
redreceipt, line 17: there a reason you wouldn't just use a style sheet here? if someone wanted to change the `View` component, we should prob make it a `Component` prop and do `named("ui-kit.BackgroundView")(BackgroundView) to match our convention. 

then you don't need that `PropTypes.any` thing

conrad-vanl:
Alright @redreceipt check it out now...

I updated `DefaultCard` some more so that it supports some of the overlay image effects (so that we don't have to pass theme into cards for custom colored cards). They'll look like this now:

Screen Shot 2021-06-30 at 12 50 22 PM

Also, `DefaultCard` supports all of the features of `FeaturedCard` now, so I made `FeaturedCard` component really simple by just having it render a `<DefaultCard featured {...props} />`

I added `console.warn`'s to `HighlightCard`s and fixed the overlay effect on them.

Lastly, I changed our card mappers to no longer return `HighlightCard`s

packages/apollos-ui-connected/src/ContentCardConnected/ContentCardComponentMapper/cardMapper.js
```diff
@@ -1,19 +1,9 @@
 import React from 'react';
-import { get } from 'lodash';
 
-import { DefaultCard, HighlightCard } from '@apollosproject/ui-kit';
+import { DefaultCard } from '@apollosproject/ui-kit';
 
 const cardMapper = (props) => {
```
redreceipt, line 5: in this case we don't need a mapper, just a component they can override

packages/apollos-ui-connected/src/ContentCardConnected/ContentCardComponentMapper/contentCardComponentMapper.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
-import { get } from 'lodash';
 
-import { DefaultCard, HighlightCard } from '@apollosproject/ui-kit';
+import { DefaultCard } from '@apollosproject/ui-kit';
 
 const contentCardComponentMapper = (props) => {
```
redreceipt, line 5: same comment as above. don't need mappers at this point which is awesome because it's unconventional and confusing

packages/apollos-ui-kit/src/BackgroundView/index.js
conrad-vanl: Not using a style sheet since the styles are based off of props. Not passing an object into a style property because that changes every render, `useMemo` is the "correct" way to do that now.

You're right, the alternative is to use `styled`. Then, the component would become:

```jsx
const BackgroundView = styled(({ theme, material, flexed }) => ({
  backgroundColor: theme?.colors?.background[material] || theme?.colors?.background?.screen,
  flex: flexed ? 1 : 0,
  height: flexed ? '100%' : undefined,
}, 'ui-kit.BackgroundView')(View);
```

I'm guessing that would be preferred. fixing now...

packages/apollos-ui-connected/src/ContentCardConnected/ContentCardComponentMapper/cardMapper.js
conrad-vanl: Mapper needs to stay for backwards compatibility I think

conrad-vanl: Component override is already supported, so I think we just need to throw a.... well I don't want to throw a `console.warn` cuz we'll get blasted with that since it's still ran, but we shouldn't remove it altogether yet? Just don't encourage its use anymore.....

packages/apollos-ui-kit/src/FeaturedCard/index.js
```diff
@@ -1,226 +1,9 @@
...
-    colors: PropTypes.shape({}),
-  }),
-};
-
-FeaturedCard.defaultProps = {
-  actionIcon: 'play-solid',
-};
-
-FeaturedCard.displayName = 'FeaturedCard';
+const FeaturedCard = (props) => <DefaultCard featured {...props} />;
```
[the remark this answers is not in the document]
conrad-vanl: I think there's value to keeping it for now - the way we use it today is as a component override - `CardComponent={FeaturedCard}` is a lot better then say....

```
CardComponent={(props) => <DefaultCard featured {...props})}
```

packages/apollos-ui-connected/src/ContentCardConnected/ContentCardComponentMapper/cardMapper.js
redreceipt: I'm suggested we _don't_ use it anywhere and then warn people it's deprecated

conrad-vanl: It's an override though....how would we warn people it's deprecated?

conrad-vanl: Like if we remove it from our code, that overrides won't work anymore

redreceipt: right before the return statement

packages/apollos-ui-kit/src/BackgroundView/index.js
conrad-vanl: Did this. committed.

packages/apollos-ui-connected/src/ContentCardConnected/ContentCardComponentMapper/cardMapper.js
conrad-vanl: So leave it in the codebase, just add `console.warn`?

That means the console.warn will get triggered every time a `ContentCardConnected` component is rendered

conrad-vanl: Let me think of a solution.... @redreceipt can you link me to a repo where this override is used somewhere?

redreceipt: right, `ContentCardConnected` doesn't need mapper by default, just `DefaultCard`

packages/apollos-ui-kit/src/FeaturedCard/index.js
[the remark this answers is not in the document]
redreceipt: I agree it's better, but that's our system. We tell people that's how you do it. And not removing it just leaves us maintaining 9 lines of dead code.

Chances are they won't override anyway, it's used internally through features.

redreceipt:
> Alright @redreceipt check it out now...
> 
> I updated `DefaultCard` some more so that it supports some of the overlay image effects (so that we don't have to pass theme into cards for custom colored cards). They'll look like this now:
> 
> Screen Shot 2021-06-30 at 12 50 22 PM
> 
> Also, `DefaultCard` supports all of the features of `FeaturedCard` now, so I made `FeaturedCard` component really simple by just having it render a `<DefaultCard featured {...props} />`
> 
> I added `console.warn`'s to `HighlightCard`s and fixed the overlay effect on them.
> 
> Lastly, I changed our card mappers to no longer return `HighlightCard`s

this is freaking cool

packages/apollos-ui-kit/src/FeaturedCard/index.js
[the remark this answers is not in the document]
redreceipt: fair

packages/apollos-ui-connected/src/ContentCardConnected/ContentCardComponentMapper/contentCardComponentMapper.js
redreceipt: actually, this is already a dead component

redreceipt (dismissed):
looks good on the sim. needs an accompanying templates PR to pull out the old like/share bar

conrad-vanl:
Okay @redreceipt this should be g2g now, added a pr on templates https://github.com/ApollosProject/apollos-templates/pull/418

redreceipt:
@conrad-vanl here's some content with no picture

Screen Shot 2021-06-30 at 3 06 34 PM

conrad-vanl:
> @conrad-vanl here's some content with no picture

What are you trying to show with this? Cards with no images show up pretty much like that on core now, I don't think that's a regression

conrad-vanl:
@redreceipt anything else you want to see on this before approving?

redreceipt:
Only "breaking" thing is you can't click the share button when there's no image. Good if you are?

conrad-vanl:
Oh, okay I'll fix that I see now.

Also, fixing the contentcardmapper deprecation thing forgot about that. You can approve if you want and I'll get those two changes pushed up and merged :)

redreceipt (approved):

PR #1839, opened 2021-03-26
feat: Optimistic response for requesting follows

jimmyhogoboom:


jimmyhogoboom:
@vinnyjth This worked great for requests! 

However I'm not sure the accept/hide is actually doing anything. It still seems to take the same amount of time with or without the optimistic responses. Am I missing something?

vinnyjth:
@jimmyhogoboom My best guess as-to your problem, was that you needed to pass `requestIds` through the update functions to they could be used in the optimistic responses. I've done that for us here :)

vinnyjth (approved):

PR #1845, opened 2021-03-29
chore: add manual trigger for indexing content

redreceipt:


packages/apollos-data-connector-algolia-search/src/jobs.js
```diff
@@ -65,9 +65,9 @@ const createJobs = ({ getContext, queues }) => {
 
   FullIndexQueue.add(null, { repeat: { cron: '15 3 * * 1' } });
   DeltaIndexQueue.add(null, { repeat: { cron: '15 3 * * *' } });
-  // Uncomment this to trigger an index right now.
-  // FullIndexQueue.add(null);
-  // DeltaIndexQueue.add(null);
```
conrad-vanl, line 70: Should this be protected under a password, or something?

conrad-vanl: There's already a password within `ApollosConfig.APP.JOBS_PASSWORD`....might make sense to use that password too

redreceipt: it is, see the server-core file that runs the trigger

conrad-vanl (approved):

PR #1846, opened 2021-03-30
chore: publish canary before stable release

redreceipt:
Currently, the first canary after the release fails because there's already a `.0`. This should be fine even if the version number doesn't reflect the conventional commits version, we only to need to make sure the canary is the latest possible code, version number is irrelevant


redreceipt:
oh shoot, this will fail too, I need a commit for the canary to work right. just a sec...

vinnyjth (approved):

PR #1849, opened 2021-03-31
chore: decouple search channels from user feed

redreceipt:
This PR accomplishes two things:
1. Decouples the search indexed content from the user feed. Sometimes you want to be able to search for more than just what's in the user feed.
2. Replaces two algorithms with one.

Test using the `feeds` branch on templates

### DIFF
--- a/packages/apollos-data-connector-rock/src/content-items/data-source.js
+++ b/packages/apollos-data-connector-rock/src/content-items/data-source.js
@@ -18,6 +18,10 @@ const { APP, ROCK, ROCK_MAPPINGS, ROCK_CONSTANTS } = ApollosConfig;
 export default class ContentItem extends RockApolloDataSource {
   resource = 'ContentChannelItems';
 
+  activeChannelIds =
+    ROCK_MAPPINGS.ACTIVE_CONTENT_CHANNEL_IDS ||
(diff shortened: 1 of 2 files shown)


packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -18,6 +18,10 @@ const { APP, ROCK, ROCK_MAPPINGS, ROCK_CONSTANTS } = ApollosConfig;
 export default class ContentItem extends RockApolloDataSource {
   resource = 'ContentChannelItems';
 
+  activeChannelIds =
+    ROCK_MAPPINGS.ACTIVE_CONTENT_CHANNEL_IDS ||
```
vinnyjth, line 22: what do you think about `ACTIVE` vs `APP`?

redreceipt: it's not just app, web will use it to index content too. it's essentially just a list of all channels with relevant content. I did think of changing the config constant to `SEARCH_CHANNEL_IDS` since that will be the only thing it will be used for but that can come later when we fully remove the deprecated name.

vinnyjth (approved):

PR #1850, opened 2021-03-31
chore: allow multiple campaign features

redreceipt:
This simplifies the config by removing the need for a constant and also gives us the ability to have multiple campaign features throughout the app

### DIFF
--- a/packages/apollos-data-connector-rock/src/action-algorithms/index.js
+++ b/packages/apollos-data-connector-rock/src/action-algorithms/index.js
@@ -192,11 +192,11 @@ Make sure you structure your algorithm entry as \`{ type: 'CONTENT_CHANNEL', aru
   }
 
   // Gets a configurable amount of content items from each of the configured campaigns
-  async campaignItemsAlgorithm({ limit = 1 } = {}) {
+  async campaignItemsAlgorithm({ channelIds = [], limit = 1 } = {}) {
     const { ContentItem } = this.context.dataSources;
 
     const channels = await ContentItem.byContentChannelIds(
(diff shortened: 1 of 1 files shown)


redreceipt:
test using the `campaigns` branch on templates

packages/apollos-data-connector-rock/src/action-algorithms/index.js
```diff
@@ -192,11 +192,11 @@ Make sure you structure your algorithm entry as \`{ type: 'CONTENT_CHANNEL', aru
   }
 
   // Gets a configurable amount of content items from each of the configured campaigns
-  async campaignItemsAlgorithm({ limit = 1 } = {}) {
+  async campaignItemsAlgorithm({ channelIds = [], limit = 1 } = {}) {
     const { ContentItem } = this.context.dataSources;
 
     const channels = await ContentItem.byContentChannelIds(
-      ApollosConfig.ROCK_MAPPINGS.CAMPAIGN_CHANNEL_IDS
+      channelIds || ApollosConfig.ROCK_MAPPINGS.CAMPAIGN_CHANNEL_IDS
```
vinnyjth, line 199: love it. doesn't break existing apps but opens the door for way more functionality. would love to find a better name for this whole algo long term, I still struggle to remember how/why this algo exists.

vinnyjth (approved):

packages/apollos-data-connector-rock/src/action-algorithms/index.js
redreceipt: it exists because it's a short cut to pulling the child item of the first item in the channel. so maybe we could call it something along those lines

PR #1851, opened 2021-03-31
feat: Added analytics for new features - journal and following

vinnyjth:
### How do I test this PR?

On tempaltes, change your `AnalyticsProvider` to
```
        <AnalyticsProvider trackFunctions={[console.warn]}>
```

Then like, flag, follow, etc.

### DIFF
--- a/packages/apollos-ui-connected/src/NodeSingleConnected/index.js
+++ b/packages/apollos-ui-connected/src/NodeSingleConnected/index.js
@@ -19,6 +19,7 @@ import NodeFeaturesConnected from '../NodeFeaturesConnected';
 import ScriptureNodeConnected from '../ScriptureNodeConnected';
 
 import GET_MEDIA from './getMedia';
+import GET_TITLE from './getTitle';
 
 const Noop = () => null;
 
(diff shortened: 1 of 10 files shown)


packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -135,6 +136,8 @@ NodeSingleConnectedWithMedia.defaultProps = {
   Component: NodeSingleInner,
 };
 
+export { GET_MEDIA, GET_TITLE };
```
redreceipt, line 139: is there a reason you're exporting these here?

redreceipt:
this is awesome

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
vinnyjth: I noticed that we use a `GET_TITLE` query in templates, and I figure it's always good to export our queries.

vinnyjth: oh doh, lol

vinnyjth: fixed

redreceipt (approved):

PR #1854, opened 2021-04-01
fix: Optimize Cloudinary image settings

conrad-vanl:

The Cloudinary settings we had were pretty liberal and can result in large images. This forces the fetch format to jpg which forces Cloudinary to make images lossy, and makes the base image size smaller. Because of this, we can get rid of the `eco` setting we previously had. 

The same image that is several megabytes on Newspring's instance is now ~150kb.


vinnyjth (approved):
🌉

PR #1858, opened 2021-04-06
fix: component overrides for NodeSingleConnected

conrad-vanl:
Previously you could only set a custom `Component` if your node single had media. This is an edge case regression when the new media player was added


vinnyjth (approved):

vinnyjth:
I had the same fix in a branch that I forgot to turn into a PR ;) 
https://github.com/ApollosProject/apollos-apps/compare/master...allow-node-single-overrides-to-work-for-non-media-items

PR #1860, opened 2021-04-06
feat: add series to series items

redreceipt:
New approach to an old problem. We define the list of series IDs in the config. That will rule out possibly returning a campaign item as the series

test using the `series` branch on templates

Screen Shot 2021-04-06 at 11 20 33 AM


redreceipt:
instead of `ContentSeriesContentItem` I think I need to include this field on all content types.

PR #1861, opened 2021-04-07
feat: Figma / 2.0 Theme Changes

vinnyjth:

This PR begins the process of updating the theme object to better mirror what's been used in Figma for quite some time now.

There are snapshot changes as we are updating some of the default grays to more neutral tones (the current neutrals are NewSpring's neutrals). The updated default theme values for the neutral palette utilizes Apple's neutral palette for a more universal look.

If you want to see the entire new UI Kit and theme you can do so in Figma - https://www.figma.com/file/YHJLj8pdFxWG9npF2YmB3r/UI-Kit-2.0?node-id=0%3A1


conrad-vanl:
@redreceipt I think we'll need you to review this one... I wrote the PR, but @vinnyjth opened it so neither of us should/can review it. LMK what questions you have...I want to get this merged, as well as help with a few other UI related tweaks to help with the new apps that we are going to be rolling

packages/apollos-ui-kit/src/theme/defaultTheme.js
```diff
@@ -30,25 +45,30 @@ export const colors = {
...
   lightTertiary: '#A5A5A5',
-
-  // Statics
-  black: '#000000',
-  white: '#FFFFFF',
-  transparent: 'transparent',
-  wordOfChrist: '#8b0000', // only used in Scripture.
+  // (note: paper and screen are now set within the theme type directly)
+  // screen: '#F8F7F4',
+  // paper: '#FFFFFF',
```
redreceipt, line 71: I don't think we need this note here

packages/apollos-ui-kit/src/theme/defaultTheme.js
```diff
@@ -30,25 +45,30 @@ export const colors = {
...
   tertiary: '#6EC5B8',
-  screen: '#F8F7F4',
-  paper: '#FFFFFF',
   alert: '#c64f55',
+  warning: '#F4B740',
+  success: '#17B569',
+
+  gray: '#8E8E93',
+  darkPrimary: '#27272E',
+  lightPrimary: '#F4F7F8',
```
redreceipt, line 54: not that I really need to know this, not a designer, but what's the difference between primary and light primary in application?

packages/apollos-ui-kit/src/theme/defaultTheme.js
```diff
@@ -30,25 +45,30 @@ export const colors = {
...
 
+  wordOfChrist: '#8b0000', // only used in Scripture.
+
+  black: '#000000',
+  white: '#FFFFFF',
+  transparent: 'transparent',
```
redreceipt, line 60: what's this here for?

redreceipt (approved):
app didn't crash, looks good to me!

packages/apollos-ui-kit/src/theme/defaultTheme.js
conrad-vanl: I want to eventually remove `darkPrimary`, `lightPrimary`, etc... our old theme (that we inherited from NewSpring) had a brand palette with brand colors (`primary`, `secondary`, `tertiary`) and a neutral palette (`darkPrimary`, `darkSecondary`, `lightPrimary`, etc). Problem is they had very similar names, which makes it confusing.

The new theme colors will utilize different naming:

Base colors, that are the same for all themes:
Screen Shot 2021-06-29 at 11 42 50 AM

Then light and dark palettes: 
Screen Shot 2021-06-29 at 11 43 10 AM

these colors will be accessed with nested selectors, like `theme.colors.fill.paper` or `theme.colors.text.primary` to further help avoid confusion.

packages/apollos-ui-kit/src/theme/defaultTheme.js
conrad-vanl: there used to be a linter rule that failed you if you embedded any text strings for colors... os if you tried to do like `backgroundColor: "transparent"` it would fail, so we started doing `backgroundColor: theme.colors.transparent`.

It's probably not needed any more...but is still in here likely for backwards compat

packages/apollos-ui-kit/src/theme/defaultTheme.js
conrad-vanl: I kept this note in here for upgrading ease... I think once we formally deprecate all of the old theme values, we can remove fully?

PR #1864, opened 2021-04-09
fix: Buttons hiding too soon in editing modal on Android

vinnyjth:
## DESCRIPTION

For the editor window, the initial bottomSheetIndex was 0 (since the component was full screen).


redreceipt (approved):
nice

PR #1866, opened 2021-04-13
feat: Better Avatar Placeholder

vinnyjth:
## DESCRIPTION

Screen Shot 2021-04-13 at 11 26 01 AM
Screen Shot 2021-04-13 at 11 00 34 AM
Screen Shot 2021-04-13 at 11 00 31 AM
Screen Shot 2021-04-13 at 9 33 03 AM
Simulator Screen Shot - iPhone 11 - 2021-04-13 at 12 45 35
Simulator Screen Shot - iPhone 11 - 2021-04-13 at 12 45 25


redreceipt:
nice change on the render tree

packages/apollos-ui-kit/src/Avatar/Avatar.js
```diff
@@ -149,6 +157,11 @@ Avatar.propTypes = {
   ...ConnectedImage.propTypes,
 };
 
+Avatar.initials = (...names) =>
```
redreceipt, line 160: I see what you did here but it's not really in line with any other API we have. The only place you'd use initials is inside the avatar, why not just put this function directly on line 118 and just accept a list of names as a prop?

if I'm a current Apollos developer, I would be a little confused as to what you're doing using a component's attribute as a prop on itself.

packages/apollos-ui-kit/src/Avatar/Avatar.stories.js
```diff
@@ -62,6 +62,19 @@ storiesOf('ui-kit/Avatar', module)
       />
     </>
   ))
+  .add('placeholder initals', () => (
+    <>
+      <Avatar isLoading size={'small'} placeholderInitials={'VW'} />
```
redreceipt, line 67: I like this a lot better, was this method first or a later implementation?

redreceipt: well, taking a list of names would be even better I think to control how they turn out

packages/apollos-ui-kit/src/Avatar/Avatar.js
vinnyjth: so you suggest taking a list of names directly? `name={[firstName, lastName]}` that seems weird to pass an array. Maybe `name={{ firstName, lastName }}`.

My reasoning was thus. 

`name` name prop is okay, but doesn't make sense for our use cases because we have a first and last name.
`names` is awkward. What does it mean? What do I pass.
`initials` is super clean and suave`. But now I have to pass initials, and need a standard way to do that. 
Hence, `Avatar.initials. 

What if we compromised by exporting Avatar.initials differently? Moved it to a utils function or something that we exported separately.

redreceipt: what if you just passed the profile in, with names and photo and everything it would need, prop types would show the right shape

redreceipt: `<Avatar size={"m"} profile={profile} />`

vinnyjth: @redreceipt I'm trying not to change the external API :/

redreceipt: no name would show the old floating head thing

redreceipt: `profile.photo || photo` makes it simpler and backwards compatible

vinnyjth: You're convincing me :)

vinnyjth: I'll go that route.

packages/apollos-ui-kit/src/Avatar/Avatar.stories.js
redreceipt: oh I see, this is prop spreading down the line

vinnyjth: correct. I'll fix this story so it's accurate though

vinnyjth:
@redreceipt Ready for your review again

packages/apollos-ui-kit/src/Avatar/AvatarList/index.js
```diff
@@ -63,6 +63,8 @@ const renderItem = (onPressAvatar, isLoading) => ({ item }) => (
     onPress={() => onPressAvatar({ item })}
     /* eslint-disable-next-line */
     source={item?.source}
```
redreceipt, line 65:
```suggestion
```

redreceipt (approved):
love this. love the prop types. love the API. love love love

PR #1867, opened 2021-04-13
chore: add new external link handler

redreceipt:
This allows for passing a more complete external link handler that works well with the API link generation


vinnyjth (approved):
Dig it. This will help us remove a bunch of custom code in NS.

PR #1870, opened 2021-04-15
chore: deprecate PushConsumer

redreceipt:
We are already exporting the `PushContext`


vinnyjth (approved):
It's hard to not approve this

PR #1871, opened 2021-04-16
feat: Allow requests to drill in on a specific attribute value

vinnyjth:


packages/apollos-rock-apollo-data-source/src/request-builder.js
```diff
@@ -189,6 +189,23 @@ you can return request.empty()
     return this;
   };
 
+  loadAttributes = (loadAttributeKeys = []) => {
```
redreceipt, line 192: love it. call this just `attributes` so the name doesn't conflict with the rock query field of the same name

redreceipt: and actually, this should be a property on the data source class, not the request. that way it works the same way with `node.get`

redreceipt: oh wait nevermind that wouldn't work. so how do we match functionality with all future requests including `node.get`?

vinnyjth: @redreceipt Here's the uses....

```
// if you want to load all attributes
class.expanded = true

// if you want to load _some_ attributes, takes precedence over expanded = true
instance.attributes()

// doesn't exist, but maybe should? 
class.loadAttributes = [...]
//  or 
class.expanded = [...] || true/false
```

vinnyjth: I thought about going the `class.loadAttributes` route to begin with, but I realized that the instance method would actually be more helpful. I think it's more common that we only need a specific attribute when doing specific things (like fetching ministries) and want more general attribute access or no attribute access most of the time.

packages/apollos-rock-apollo-data-source/src/__tests__/request-builder.tests.js
```diff
@@ -88,6 +88,16 @@ describe('RequestBuilder', () => {
     ).resolves.toMatchSnapshot();
   });
 
+  it('loads specific attribute keys', () => {
+    expect(request.attributes(['Puppies']).get()).resolves.toMatchSnapshot();
+  });
+
+  it('chains multiple load attribute keys', () => {
+    expect(
+      request.attributes(['Puppies']).loadAttributes(['Cats', 'Dogs']).get()
```
vinnyjth, line 97:
```suggestion
      request.attributes(['Puppies']).attributes(['Cats', 'Dogs']).get()
```

packages/apollos-rock-apollo-data-source/src/request-builder.js
redreceipt: in this instance, if we get a person id back, then later I want to get their social media stuff outside the `getStaff`, it wouldn't be there so I think we need a more top level approach. maybe from the class you set `expanded=true` to turn it on and `attributeKeys=[...]` to drill down

redreceipt: or by specifying `class.attributes = [...]` it just knows to turn on `expanded=true` also from the class. that would probably get us what we need. I probably would use it from the request for the previous mentioned reason, it would come back to bite later

vinnyjth:
@redreceipt Took your feedback into consideration, moved the attribute settings into class properties.

packages/apollos-rock-apollo-data-source/src/index.js
```diff
@@ -26,6 +26,9 @@ export default class RockApolloDataSource extends RESTDataSource {
   // Subclasses can set this to true to force all requests to turn extended responses.
   expanded = false;
 
+  // Sublasses can use this to load specific attributes.
+  rockAttributes = [];
```
redreceipt, line 30: doesn't need to be called `rockAttributes` since this is a rock data source class, `attributes` is cleaner

stepping back though, is this feature still needed since we just optimized the query in a different way?

vinnyjth: > stepping back though, is this feature still needed since we just optimized the query in a different way?

Yes :) I noticed a noticeable difference in load speeds when loading a small subset of attributes on people vs all attributes. 

`attributes` is for sure cleaner, and i like it more, but it's too vague. I'm open to other names, like `attributesLoaded` or `attributeValuesToLoad` or something like that.

redreceipt: good call, I like `attributesLoaded`

packages/apollos-rock-apollo-data-source/src/index.js
```diff
@@ -72,11 +75,22 @@ export default class RockApolloDataSource extends RESTDataSource {
     return mapKeys(normalizedValues, (value, key) => camelCase(key));
   };
 
+  buildDefaultOptions() {
+    const defaultOptions = {};
+    if (this.rockAttributes?.length)
```
vinnyjth, line 80:
```suggestion
    if (this.attributesLoaded?.length)
```

packages/apollos-rock-apollo-data-source/src/index.js
```diff
@@ -72,11 +75,22 @@ export default class RockApolloDataSource extends RESTDataSource {
...
+      defaultOptions.attributeKeys = this.rockAttributes.join(',');
+    if (this.expanded || this.rockAttributes?.length)
```
vinnyjth, line 82:
```suggestion
    if (this.expanded || this.attributesLoaded?.length)
```

packages/apollos-rock-apollo-data-source/src/index.js
```diff
@@ -72,11 +75,22 @@ export default class RockApolloDataSource extends RESTDataSource {
     return mapKeys(normalizedValues, (value, key) => camelCase(key));
   };
 
+  buildDefaultOptions() {
+    const defaultOptions = {};
+    if (this.rockAttributes?.length)
+      defaultOptions.attributeKeys = this.rockAttributes.join(',');
```
vinnyjth, line 81:
```suggestion
      defaultOptions.attributeKeys = this.attributesLoaded.join(',');
```

packages/apollos-rock-apollo-data-source/src/__tests__/data-source.tests.js
```diff
@@ -81,5 +81,25 @@ describe('RestDataSource', () => {
       expect(result).toBeTruthy();
       expect(getFn.mock.calls[0][0]).toBe('TestResource?%24top=2&%24skip=0');
     });
+
+    it('sets default request options w/ no expansion', () => {
+      const defaultOptions = dataSource.buildDefaultOptions();
+      expect(defaultOptions).toEqual({});
+    });
+    it('sets default request options w/ rockAttributes', () => {
+      dataSource.rockAttributes = ['Dog', 'Cat'];
```
vinnyjth, line 90:
```suggestion
      dataSource.attributesLoaded = ['Dog', 'Cat'];
```

redreceipt (approved):

PR #1874, opened 2021-04-19
feat: add details field to content items

redreceipt:
This will give us the ability to add arbritrary information to a content item and render it distinctly on the front end \(think 'Cost: '\)


redreceipt:
I'm gonna go with a fully custom solution for Long Hollow, will maybe dig this up if that ends up not being what's best

PR #1877, opened 2021-04-20
chore: remove issue and PR templates

redreceipt:
I think we take a minimal approach to templates and only create them if it gets bad, I appreciate repos where issues and PRs are blank, lets the convo happen naturally. plus no one ever uses them right anyway


vinnyjth (approved):

vinnyjth:
This is great for now - would love to create a super short issue template in the future. Maybe it's just a comment you delete, reminding people that we need the "what's going on? what do you have custom? how can I reproduce?"

PR #1914, opened 2021-05-04
chore: add livestream safety check

redreceipt:
prevents error before the URL is defined


packages/apollos-data-connector-church-online/src/data-source.js
```diff
@@ -99,6 +99,7 @@ export default class LiveStream extends RESTDataSource {
 
   async getLiveStreams() {
     const { ContentItem } = this.context.dataSources;
+    if (!ApollosConfig.CHURCH_ONLINE.URL) return [];
```
vinnyjth, line 102:
```suggestion
    if (!ApollosConfig?.CHURCH_ONLINE?.URL) return [];
```

vinnyjth (approved):

PR #1922, opened 2021-05-04
chore: add group type defaults

redreceipt:
also removes the HUGE rock constants section that we've never had to touch, only 5 of the items are used anyway and in the extremely unlikely chance there's a change required, they could just overwrite the DS function.

### DIFF
--- a/packages/apollos-data-connector-rock/src/personal-devices/data-source.js
+++ b/packages/apollos-data-connector-rock/src/personal-devices/data-source.js
@@ -22,7 +22,7 @@ export default class PersonalDevices extends RockApolloDataSource {
       PersonAliasId: currentUser.primaryAliasId,
       DeviceRegistrationId: pushId,
       PersonalDeviceTypeValueId:
-        ApollosConfig.ROCK_MAPPINGS.MOBILE_DEVICE_TYPE_ID,
+        ApollosConfig.ROCK_MAPPINGS.MOBILE_DEVICE_TYPE_ID || 671,
       NotificationsEnabled: 1,
       IsActive: 1,
     });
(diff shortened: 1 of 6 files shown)


packages/apollos-data-connector-rock/src/personal-devices/data-source.js
```diff
@@ -22,7 +22,7 @@ export default class PersonalDevices extends RockApolloDataSource {
       PersonAliasId: currentUser.primaryAliasId,
       DeviceRegistrationId: pushId,
       PersonalDeviceTypeValueId:
-        ApollosConfig.ROCK_MAPPINGS.MOBILE_DEVICE_TYPE_ID,
+        ApollosConfig.ROCK_MAPPINGS.MOBILE_DEVICE_TYPE_ID || 671,
```
vinnyjth, line 25: This code seems good to me, but I believe this ID does change from rock instance to instance.

https://github.com/NewSpring/crete/blob/master/config.yml#L264 
https://github.com/Differential/willow-creek/blob/master/packages/apollos-api/config.yml#L347
https://github.com/Differential/grace-apollos-app/blob/master/apollos-api/config.yml#L168

What do you think about doing a follow up PR where this value is looked up and then cached in redis?

vinnyjth (approved):

packages/apollos-data-connector-rock/src/personal-devices/data-source.js
redreceipt: yep, I'll make an issue

redreceipt: #1923

PR #1925, opened 2021-05-06
fix: prevents error when logging in for the first time

redreceipt:
This fixes three things:
1. makes cached cookie auth error more explicit
2. narrows overbearing try-catch
3. sets cookie in cache before lookup

Was showing itself as the attached screenshot and was very hard to track down because of all the nested try/catch statements.

Screen Shot 2021-05-06 at 4 32 13 PM

### DIFF
--- a/packages/apollos-data-connector-rock/src/auth/data-source.js
+++ b/packages/apollos-data-connector-rock/src/auth/data-source.js
@@ -54,7 +54,7 @@ export default class AuthDataSource extends RockApolloDataSource {
     });
     if (!cachedUserName) {
       throw new AuthenticationError(
-        'Invalid user cookie; no eligble user login found'
+        'Invalid user cookie; no eligble user cached'
       );
     }
     const login = await this.request('/UserLogins')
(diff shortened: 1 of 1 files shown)


vinnyjth (approved):

vinnyjth:
This looks good. We can honestly most likely get rid of the cookie cache, since we determined that cookies weren't the eventual cause of the auth issues.

redreceipt:
the code definitely needs some TLC but I don't understand it enough to make sweeping changes. how do you think this snuck by so long? I don't know why it wouldn't always happen with someone new.

redreceipt:
I don't wanna merge this yet, after I made the fix, it gave me a big long token but now I'm thinking it's bad because the `GetCurrentPerson` was actually what was failing

vinnyjth:
Closing for same reasons as https://github.com/ApollosProject/apollos-apps/pull/1930. Would prefer not to touch this code at this point in time. Reopen if you think it's important :D

PR #1930, opened 2021-05-07
chore: throw a better authentication error

redreceipt:
Makes the 404 stand out letting us know the auth endpoint is not configured properly


vinnyjth:
@redreceipt can we keep this code on the backburner for now? I think removing the cache is a good choice, but I think `try` / `catch` interceptions might be invalid. We will want to throw an AuthError if almost anything happens in this method, because it will force a logout. Maybe we need better logging to help compensate for that. 

Either way, am I correct that we discovered that the core code is valid, overall, and it was a firewall issue on CO?

redreceipt:
@vinnyjth That's fair, I put back the try/catch in the `getCurrentPerson` call. This code is good because it was much harder to find the error with so many exception layers. I just wonder if there's a difference between the 404 for the endpoint being blocked and other reasons why the cookie would fail, like expired.

I left the try/catch out of the `authenticate` method because they are already logged out at that point, a deeper exception is better than generic. And I left the cache stuff out.

Technically we do need something because the code is broken with the cache lookup in the wrong place, that was why it was so hard in the first place. That was the original bug.

vinnyjth:
> Technically we do need something because the code is broken with the cache lookup in the wrong place, that was why it was so hard in the first place. That was the original bug.

I don't think this is true, we probally shouldn't cache the cookie unless we know it returns a valid user. 

> I left the try/catch out of the authenticate method because they are already logged out at that point, a deeper exception is better than generic. And I left the cache stuff out.

I'll take another look at this to confirm, but sounds good to me.

redreceipt:
I see what you're saying now. It's wrapped in multiple try/catches for that reason. Regardless, It adds layers to debug that probably aren't necessary now that we've solved the actual auth bug and it's not a cookie issue.

vinnyjth:
@redreceipt I'm going to close this, but feel free to reopen. I would rather not touch this code right now, since I believe it's currently correct and working fine. I understand that it's tricky to debug, but hopefully we will have swapped it with `Postgres` by the time we need to debug it again.

PR #1935, opened 2021-05-10
feat: Update OneSignal to v4

vinnyjth:
Your best bet for testing this is pulling the companion PR and building an app for the phone, then sending yourself a push notification with a button on iOS, and confirming you see the button. 

You can add a button when sending a push through OneSignal here. 
image


vinnyjth:
Companion PR: https://github.com/ApollosProject/apollos-templates/pull/351

redreceipt:
@vinnyjth explain to me why this wouldn't be a breaking change. I want to say "no biggie" but if clients point to this version, their app will break, correct?

vinnyjth:
It is a breaking change :/ you also need to update the client with the template changes. You can't update templates without this change, and you can't update your package versions without having the code in templates. 

We could potentially "fix" that dependency by adding a bunch of try / catches around the existing code so it will safely try both the old and new code. What do you think?

packages/apollos-ui-notifications/src/permissionUtils.js
```diff
@@ -8,19 +8,10 @@ import {
 } from 'react-native-permissions';
 
 const getPushPermissions = async () =>
-  new Promise((resolve) =>
-    OneSignal.getPermissionSubscriptionState((status) =>
-      // Ensure the client (notificationsEnabled) && OneSignal (subscriptionEnabled) are boolean values
-      resolve(!!status.notificationsEnabled)
-    )
-  );
+  (await OneSignal.getDeviceState()).hasNotificationPermission;
```
redreceipt, line 11: since these are one-liners that simply use the OS SDK, I say we push this line to whatever is calling the `getPushPermissions` function and then deprecate and `console.warn` this and `getHasPrompted` function. Not helpful to abstract.

vinnyjth: 👍  sounds good to me.

redreceipt:
> It is a breaking change :/ you also need to update the client with the template changes. You can't update templates without this change, and you can't update your package versions without having the code in templates.
> 
> We could potentially "fix" that dependency by adding a bunch of try / catches around the existing code so it will safely try both the old and new code. What do you think?

Unfortunately, I think it's the right thing to do. It will ensure that folks can upgrade past 2.19 to get some bug fix without having to fool with OS v4 if they don't need it.

redreceipt (changes requested):
- make it non-breaking
- deprecate utility functions

packages/apollos-ui-notifications/src/permissionUtils.js
```diff
@@ -7,20 +9,45 @@ import {
...
-    )
-  );
-
-const getHasPrompted = async () =>
-  new Promise((resolve) =>
-    OneSignal.getPermissionSubscriptionState((status) =>
-      resolve(status.hasPrompted)
-    )
-  );
+const getPushPermissions = async () => {
```
redreceipt, line 12: same question with this file. seems smelly to keep catch with nothing all over the place.

packages/apollos-ui-notifications/src/Provider.js
```diff
@@ -56,14 +58,33 @@ class NotificationsInit extends Component {
...
+    } catch (e) {
+      console.warn(
+        'Core has been updated to use OneSignal 4.x.x. It is recommended that you upgrade as well to access the latest functionality.'
+      );
+      console.warn(e);
+    }
+
+    // backup, for OneSignal 3.x
+    try {
+      OneSignal.init(this.props.oneSignalKey, {
```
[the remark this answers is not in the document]
vinnyjth: I wanted to separate the old/new code more clearly, and I liked the idea that an exception raised in the inital block would trigger our deprecation message.

packages/apollos-ui-notifications/src/permissionUtils.js
vinnyjth: Same as above, it also felt messy calling two different implementations of OneSignal side by side. Seemed like this has an easy deprecation path moving forward, delete the second block and remove try/catch.

redreceipt: the only risk is if there's a legit problem with the old commands, we won't be properly notified.

packages/apollos-ui-notifications/src/Provider.js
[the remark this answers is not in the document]
vinnyjth: oh I feel dumb, I see what you're saying. So 

```
try {
 v4
 } catch {
   warn
   v3
 }
``` 

The only thing I don't like about this, is that if there's a legit issue with v4, we end up raising two warnings. v4 will throw an exception, and then v3 will throw an exception (since it's code doesn't exist). You'll be left with a `console.warn()` of the real error, and then an actual redscreen of v3 code failing to be called. I went this route to avoid that, and that's also why I let v3 fail silently.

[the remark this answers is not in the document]
vinnyjth: what I would prefer is 
```
if OneSIgnal.v3
{
code
} 

if OneSignal.v4
{
code
}
```

I could probably do that by checking for a property or function that only exists in v3.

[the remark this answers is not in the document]
redreceipt: I do like that the best by far. I don't like using try/catch blocks for things that you have already fully anticipated.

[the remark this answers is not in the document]
vinnyjth: same. I'll do that instead.

vinnyjth:
Thanks for your feedback @redreceipt, feeling way better about this PR than where it started.

packages/apollos-ui-notifications/src/Provider.js
```diff
@@ -56,14 +58,26 @@ class NotificationsInit extends Component {
...
-    OneSignal.init(this.props.oneSignalKey, {
-      kOSSettingsKeyAutoPrompt: false,
-    });
-    OneSignal.addEventListener('received', this.onReceived);
-    OneSignal.addEventListener('opened', this.onOpened);
-    OneSignal.addEventListener('ids', this.onIds);
-    OneSignal.setSubscription(true);
+  async componentDidMount() {
+    // One Signal 4.x
+    if (OneSignal.setAppId) {
```
redreceipt, line 63:
```suggestion
    if (OneSignal.setAppId && this.props.oneSignalKey) {
```

to avoid this and to make the prop optional

Screen Shot 2021-06-10 at 3 59 49 PM

redreceipt: actually it's gonna need to go one level up so when there's no prop, v3 code doesn't try to run

vinnyjth: v3 doesn't error, so I think your change is good

redreceipt:
trying to test this but I may be missing something obvious, I can't even get a test notification to send to my sim. I've got `ONE_SIGNAL_KEY` defined and am just sending a new push to all devices on Apollos-Prototype

packages/apollos-ui-notifications/src/Provider.js
redreceipt: well, if you're on v4 and don't pass the prop, it's gonna drop to v3 code and try to run `Onesignal.init` which will error because it doesn't exist.

redreceipt: nevermind, it just comes through as an unresolved promise.

Screen Shot 2021-06-10 at 4 58 51 PM

vinnyjth:
> trying to test this but I may be missing something obvious, I can't even get a test notification to send to my sim. I've got `ONE_SIGNAL_KEY` defined and am just sending a new push to all devices on Apollos-Prototype

iOS Simulators can't receive push notifications :/

redreceipt (approved):

PR #1936, opened 2021-05-11
feat: add new tab feed query

redreceipt:
Gives us a simple way to create more queries for more tabs without polluting the schema

test using the `tab-component` branch on templates

Screen Shot 2021-05-12 at 9 35 41 AM

### DIFF
--- a/packages/apollos-data-connector-rock/src/feature-feeds/index.js
+++ b/packages/apollos-data-connector-rock/src/feature-feeds/index.js
@@ -4,6 +4,11 @@ import ApollosConfig from '@apollosproject/config';
 
 const resolver = {
   Query: {
+    tabFeedFeatures: (root, args, { dataSources: { FeatureFeed } }) =>
+      FeatureFeed.getFeed({
+        type: 'tab',
+        args,
+      }),
     homeFeedFeatures: (root, args, { dataSources: { FeatureFeed } }) =>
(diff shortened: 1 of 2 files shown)


vinnyjth:
I'm good with this! what do you think @conrad-vanl? I know we bounced back and forth a while when initially trying to design this schema.

redreceipt:
Same thing is hard coded, just in cleaner way. And makes the client much cleaner as well.

conrad-vanl:
It's still not my favorite, but I'm good with it. I'm going to have more commentary on the -templates side.

Essentially, I wouldn't render the "header" within the `Tab` component, I'd do that within the react-navigation config. On the RV app, all of the tab headers (both the home page header and the large title headers on like the Read tab) should be using the native react-navigation header component. 

Also, generally we render "Tabs" inside of a react-navigation Native Stack component because some navigation transitions happen "within the tab" (keeps the tab bar fixed to bottom of screen) and some navigation transitions occur over the tab (such as opening a NodeSingle). Might want to think through that some too.

redreceipt:
love it. yea drop some comments over there if there's anything else and I'll get it cleaned up. now someone gimme a green check!

conrad-vanl (approved):

conrad-vanl:
> yea drop some comments over there
I didn't see an opened PR yet?

PR #1940, opened 2021-05-14
chore: easy app development

redreceipt:
let's us spin up a template app with no linking headaches


redreceipt:
@vinnyjth if you want a fun little challenge, it started this similar to your API script, would be awesome. Currently, it's getting hung on the splash screen so I put it down. But i know you had to do some other stuff to get storybook working?

vinnyjth:
@redreceipt you'll likely need to add this :) https://github.com/ApollosProject/apollos-apps/blob/master/ApollosStorybook/metro.config.js

redreceipt:
is that different than the symlink config we already have in the templates repo?

redreceipt:
so strange. by adding "index" to file imports, even works when changing it back

https://www.dropbox.com/s/80wi52n1ptrk0wo/Kapture%202021-05-17%20at%2016.03.17.gif?dl=0

redreceipt:
maybe fixed by #1942

vinnyjth (approved):

redreceipt:
the only thing that would make this amazing is if I could figure out how to get it all in one command

redreceipt:
or figure out why metro isn't auto-opening like it does for storybook

vinnyjth:
Yah - we can get there :) I like how we have walked through this iteratively so we are building the right tools

PR #1941, opened 2021-05-17
feat: add link table feature

redreceipt:
dynamic link tables on the connect tab, this is the first step

test with the `link-table` branch on templates

Screen Shot 2021-05-17 at 2 21 24 PM


packages/apollos-data-schema/src/index.js
```diff
@@ -918,6 +918,19 @@ export const featuresSchema = gql`
     actions: [ActionBarAction]
   }
 
+  type Link {
```
vinnyjth, line 921: Only one piece of feedback! I think this needs be an `Action`. Some apps navigate to content using the action link table. We could use deep links.. but I think we'll be better off supporting this in a more robust way by creating a new `Action` type or reusing an existing type.

All the handling pieces for the existing `Actions` (which have a `relatedNode` and some other bits) already have front end code, which is a bonus.

vinnyjth (changes requested):

packages/apollos-data-schema/src/index.js
redreceipt: Only reason I didn’t is because the developer experience and config is so much more complicated and opaque. Look at `ActionBar` currently vs what this is and they do the same thing. 

Screen Shot 2021-05-18 at 10 57 59 AM

What if I simply added a `contentId` field to the config and the front end knew that if you supplied the `contentId`, to navigate there, otherwise use the URL? Way better since there's no easy way to get the `relatedNode` ID

vinnyjth: @redreceipt what if went the other direction and added some shorthand to the `action` field? That way we can keep our existing schema and use the same `handleOnPressAction` on the frontend, but also simplify the configuration for everyone on the backend?

redreceipt: We could, but I feel like that just leaves in more complicated code that we don't really need? Wouldn't it be better to retrofit with a less complicated schema?

vinnyjth: In this case, I think the less complicated schema is also less flexible, and given that we have taken advantage of that flexibility in the past I think it's worth committing to the more flexible schema. And while It's complicated, I would hope that we could make it less so through documentation or shorthand :D

vinnyjth: for example `contentId` won't let you point at a `TAB`, or a ConnectPage, or open an AuthedUrl vs a normal url, etc.

redreceipt: I hear you, I’ll switch back to the old system. We can improve later if we’re so inclined.

vinnyjth: Totally. And when/if we do switch to a new system we can do that across the board

redreceipt:
@vinnyjth it pains me but here you go 
Screen Shot 2021-05-18 at 1 50 04 PM

vinnyjth (approved):

vinnyjth:
@redreceipt how was using the quick and easy api for this PR?

redreceipt:
> @redreceipt how was using the quick and easy api for this PR?

great with one pain point, it doesn't accept the branch name like we intend, does go through, a bit of work needs to be done to make it 100%

PR #1943, opened 2021-05-18
fix: sends birthdates with the correct date

redreceipt:
The datetime picker defaults to zero time but we were overwriting time when we did our own subtraction. Dates going to the API had a time offset included. Technically they were right but it causes problems because Rock ignores TZs when patching birthdays


vinnyjth (approved):

vinnyjth:
@redreceipt You'll likely need to update a `jest.setup` file to fix a `mock` that references the date time picker

PR #1944, opened 2021-05-18
fix: Select the first media with a source in the NodeSingleConnected

vinnyjth:
This code fixes a bug where the item would register media existing, but wouldn't select it properly. 
(https://github.com/ApollosProject/apollos-apps/blob/fix-support-null-media-sources/packages/apollos-ui-connected/src/NodeSingleConnected/index.js#L90)


redreceipt (approved):
nice

PR #1945, opened 2021-05-19
chore: allow default DB options

redreceipt:
allows us to remove the options section of the config and take the guess work out of what it should be in different situations


vinnyjth (approved):
Good compromise.

PR #1947, opened 2021-05-19
feat: add action table feature to client

redreceipt:
Screen Shot 2021-07-02 at 2 03 04 PM

to test:
- start server with `action-table` branch
- start app and switch the server to localhost


redreceipt:
need to wait for testing in the app until action table feature is released to the production API

redreceipt:
blocked by https://github.com/ApollosProject/apollos-templates/pull/375

conrad-vanl:
Looks like no longer blocked?

redreceipt:
I think this one is probably good but I can't test it because `yarn api` is broken... https://github.com/ApollosProject/apollos-apps/pull/2046

redreceipt:
ok this one is ready

@vinnyjth in testing, I found and fixed a bug with action IDs that I'm surprised hasn't caused us problems. they were being encoded with `undefined` making them all the same on the client when cached.

redreceipt:
@vinnyjth also take a look at those snap updates for my sanity. I added IDs to all actions but then realized that the generic ones don't even have them. this should only help with caching on the front end but it's touching more than I intended at first...

vinnyjth (approved):

PR #1951, opened 2021-05-19
fix: Don't check postgres server certs

vinnyjth:
Heroku SSL certs are self signed.


vinnyjth:
@redreceipt looking at what _was_ in templates, you'll see that this is a closer match.

redreceipt:
oh nice, yea I had no idea what that meant

redreceipt (approved):

PR #1954, opened 2021-05-19
feat: Notification Preferences Model

vinnyjth:
Screen Shot 2021-06-02 at 9 48 26 AM

1. Pull this branch.
2. `BRANCH=notifications yarn api`
3. Use this mutation to test (or peek at the tests)
```
mutation updatePushPref {
  updateUserPushSettings(input:{ enabled:false, pushProviderUserId:"abc-123"}) {
    id
  }
}
```


redreceipt:
Unrelated question, why do we filter mutation args through an "input" arg?

vinnyjth:
The exact reason eludes me, but I think it's something like this
- It can be easier to manage on the client.  You can add fields to a mutation without needing to rewrite the query (this is probally the biggest reason) - you can see that explained a bit more here: https://blog.logrocket.com/robust-graphql-mutations-the-relay-way/
- Relay does it 🤷  and at one point we did a lot of things that relay did.

vinnyjth:
> Moving a large parameter list into an input object like this is known as the Parameter Object Pattern, and can help simplify your code.

> For example, for our CreatePost mutation above, we can rewrite the input using a CreatePostInput type like this:
```
mutation CreatePost(input: CreatePostInput!): CreatePostPayload!

type CreatePostInput {
  title: String!
  body: String!
}
```
>This lets us use our mutation like this:
```
mutation createPost($input: CreatePostInput!) {
  createPost(input: $input) {
    post { ... }
  }
}
```
>Now, no matter how many more fields we add to the createPost mutation, the above GraphQL is still correct.

redreceipt:
@vinnyjth is this the branch to test with? https://github.com/ApollosProject/apollos-templates/pull/371

redreceipt (approved):

PR #1956, opened 2021-05-20
chore: allows templates branches and app in one command

redreceipt:
Run this to test the search templates branch: `BRANCH=search-to-apps yarn app`

Everything should happen with that one command.


ApollosStorybook/ios/Podfile
```diff
@@ -22,9 +22,9 @@ target 'ApollosStorybook' do
   #
   # Note that if you have use_frameworks! enabled, Flipper will not work and
   # you should disable these next few lines.
-  use_flipper!
```
vinnyjth, line 25: Can we either fix this or remove this? The fix in `templates` seems to be working well :)

vinnyjth (approved):

PR #1960, opened 2021-05-25
feat: get comment feature from parents

vinnyjth:
Looks at the parent content item to determine if comments should be enabled, as well as the item itself 

Code copied over from Newspring, which will be able to eliminate custom code after this is merged.

### DIFF
--- a/packages/apollos-data-connector-rock/src/content-items/__tests__/__snapshots__/data-source.tests.js.snap
+++ b/packages/apollos-data-connector-rock/src/content-items/__tests__/__snapshots__/data-source.tests.js.snap
@@ -857,6 +857,52 @@ Array [
 ]
 `;
 
+exports[`ContentItemsModel returns comment features when a contentItem has a parent with a Comments field set to true 1`] = `
+Array [
+  Object {
+    "__typename": "AddCommentFeature",
+    "addPrompt": "What stands out to you?",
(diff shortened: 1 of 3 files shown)


redreceipt:
Would it be better to just enable this for all `DevotionalContentItem` types?

vinnyjth:
@redreceipt 🤷  I don't think so - It wouldn't be accurate for Newspring and we are also moving away from type based features.

redreceipt:
That's fair, I just hate adding more hardcoded attribute value names than are necessary. For any new app to use this feature, we have to go look up what the attribute value is called. It doesn't work out of the box. I know it's the case with all the features but I really wish there was a better way to enable these features by default.

redreceipt (approved):

vinnyjth:
> For any new app to use this feature, we have to go look up what the attribute value is called. It doesn't work out of the box . . . I really wish there was a better way to enable these features by default.

amen. amen. amen.

vinnyjth:
> > For any new app to use this feature, we have to go look up what the attribute value is called. It doesn't work out of the box . . . I really wish there was a better way to enable these features by default.
> 
> amen. amen. amen.

If we had more time, maybe that could be a feature of the rock plugin? Essentially adding an easy way to toggle feature support in the Rock CMS

PR #1961, opened 2021-05-25
feat: Add skip to content feed algo

vinnyjth:


redreceipt (approved):
I know what this is for :) I was thinking the same thing

PR #1962, opened 2021-05-26
chore: stops assuming local DB

redreceipt:
As a new postgres developer, I have been very confused as to what is happening with our local database setup and configuration. Assuming a specific db called "postgres" has made it difficult. Not to mention a database was being created, even though we are specifying (or worse, not specifying) a URL. This PR attempts to clear up what is actually happening when spinning up a local instance by expecting the developer to create their own db if they are choosing to use our DB functionality, something they should already understand, or be forced to learn about. I'll follow up with simple instructions in the template repo.


packages/apollos-data-connector-postgres/package.json
```diff
@@ -47,6 +46,7 @@
   "devDependencies": {
     "@apollosproject/babel-preset-apollos": "^2.16.0",
     "jest": "^25.1.0",
+    "pg": "^8.5.1",
```
vinnyjth, line 49: Good change, but we should probably list in in `peerDeps` as well

packages/apollos-data-connector-postgres/src/postgres/index.js
```diff
@@ -1,77 +1,30 @@
...
+                ssl: { require: true, rejectUnauthorized: false },
+              },
+            }),
+      })
+    : connectJest();
 
 class PostgresDataSource {
   initialize(config) {
+    if (!ApollosConfig?.DATABASE?.URL && process.env.NODE_ENV !== 'test')
+      throw new Error('Must specify DATABASE_URL variable!');
```
vinnyjth, line 27: good change!

packages/apollos-data-connector-postgres/src/postgres/index.js
```diff
@@ -1,77 +1,30 @@
 /* eslint-disable no-param-reassign */
-/* eslint-disable no-console */
-
 import './pgEnum-fix';
 import { Sequelize, DataTypes } from 'sequelize';
-import { Client } from 'pg';
 import ApollosConfig from '@apollosproject/config';
 import connectJest from './test-connect';
-import { ensureLocalDb } from './local-db';
```
vinnyjth, line 9: 👍

packages/apollos-data-connector-postgres/package.json
redreceipt: I don't think we need that package, It's only used in testing, not sure the purpose of peer deps on the API side?

vinnyjth (approved):
Looks good to me. I think we need to move `pg` to the peerDeps, but other than that we're looking good.

packages/apollos-data-connector-postgres/package.json
vinnyjth: We do need it, `sequelize` will throw an error without it. If we include it in the peerDeps, then anyone who uses the postgres package will be notified that they need to install it. More a courtesy than a requirement.

redreceipt: gotcha, I guess I was assuming `sequelize` would have it in its own deps, but it wouldn't install all different types of DBs

vinnyjth: yah - you can see that they list as an optional peerDep :) 

https://github.com/sequelize/sequelize/blob/main/package.json#L95

redreceipt:
> Also I think there's something off with tests in the CI, looks like postgres isn't starting or the connection string isn't right.

oh yea I forgot, I'll have to run `createdb <whatever the default is>` just like you would locally.

redreceipt:
this is only because pg needs a db to create another db. wish it didn't, then we wouldn't even need to worry about creating the default user one

vinnyjth:
> oh yea I forgot, I'll have to run `createdb <whatever the default is>` just like you would locally.

sweet. make sure you include the worker id in the db name, so you don't run into concurrency problems when testing

redreceipt:
not sure how it was ever working if `pg_isready` is failing. on master too: https://github.com/ApollosProject/apollos-apps/pull/1963

.github/workflows/ci.yml
```diff
@@ -32,6 +32,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - run: pg_isready
```
redreceipt, line 35:
```suggestion
```

PR #1963, opened 2021-05-26
debug PR

redreceipt:


redreceipt:
@vinnyjth so this is weird, doesn't look like postgres is ready on the CI (reason why my other PR is failing) but I'm not sure how master is passing...

vinnyjth:
That's so strange. I'm def not clear on how the postgres setup works, after looking at our actions. Nor do I see how your changes would break anything. I can dig in later, but I would probably start by figuring out exactly what is failing to connect. Is that happening in the `jest.setup.js` or somewhere else.

redreceipt:
yea, it's the first `new Client` in the `jest.setup.js`, I've tried a couple of "solutions" from googling around but that did nothing. it would make sense though that it can't connect if `pg_isready` fails. but not sure how it ever worked...

vinnyjth:
wild thought, try moving `pg` back to the deps?

redreceipt:
yea i did...

vinnyjth:
Maybe reach out to @jimmyhogoboom? I believe he set this up originally, and he might have some ideas.

PR #1964, opened 2021-05-26
feat: Add support for feeds using campusIds

vinnyjth:
Here's what usage looks like: 

Screen Shot 2021-05-26 at 3 57 54 PM

### DIFF
--- a/packages/apollos-ui-connected/src/FeaturesFeedTabConnected/FeaturesFeedTabConnected.js
+++ b/packages/apollos-ui-connected/src/FeaturesFeedTabConnected/FeaturesFeedTabConnected.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { gql, useQuery } from '@apollo/client';
 
-import { BackgroundView } from '@apollosproject/ui-kit';
+import { BackgroundView, named } from '@apollosproject/ui-kit';
 
 import { createNativeStackNavigator } from 'react-native-screens/native-stack';
 import RockAuthedWebBrowser from '../RockAuthedWebBrowser';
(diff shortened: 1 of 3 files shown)


redreceipt:
I like this but I have a suggestion. I'm a fan of providing one way to do things, the best way. Overrides seem superior to component props. So what do you think about:
- `createFeatureFeedTab` accepting a `byCampus` option that passes the `campusId` if true
- Simply naming the internal `TabComponent` so it can be overridden if they really need to
- At this point, we don't need the other two components, `Default` and `Campus`

vinnyjth:
> I like this but I have a suggestion. I'm a fan of providing one way to do things, the best way. Overrides seem superior to component props. So what do you think about: ...

This was actually my first idea, but there's a single problem with the override approach. I want to be able override an instance of `createFeatureFeedTab`, rather than all feature feed tabs. For example - on LCBC Home takes into account `campusId`, but their Discover tab doesn't (and shouldn't) 

Given that constraint, I thought about adding a `byCampus: True` argument that we could set, but that felt too limiting, given that we anticipate potentially other changes to what types of queries/feeds we could render. Figured it's better to roll with something flexible with a few prebuilt options, rather than shoehorn us into an arg.

redreceipt:
But wouldn't you rather add a single argument versus an entire component to the API? With this, we'd have to expose more component types for core supported tab feature query params anyway.

redreceipt:
like it's much harder to add (or take away) an exported component vs a single param

vinnyjth:
It's definitely a give and take. With a component prop we can keep the main component simple. No need to have if statements breaking up the logic of the component. If a user wants to add additional complexity, they can write a new component and use it, rather than needing to copy/paste everything.

On the other hand, you're right that it's a bit more complex in other ways. We'll have to solve the complexity if someone wanted a feed that used their campusId AND their groups for example. In this case, I'm optimizing for most likely use case. Most likely this prop never gets touched. It's always the default, the core component is pretty simple. When we need to use the `campusId`, we have a an alternative component available to use.

vinnyjth:
When it comes to deprecating, I actually like removing components more than changing them. Mostly because if you remove a component, you can add a console.warn() or even just remove it. It becomes very clear _where_ the deprecation is most of the time. Vs when you remove or change arguments, you have to track down the differences between old and new and the existing component has to safely handle old args being passed (in some way)

redreceipt (approved):
I see the tradeoffs. Ultimately the goal which we're both shooting for is as simple of an exposed API as possible. Which really starts with simplifying the API schema. This is just the result of having a complex backend.

vinnyjth:
All that to say... if this could be an override that would be enough reason for me to change this to an arg and move on 😂

vinnyjth:
@redreceipt right! And we have so many things to trade off.

Simple API vs Flexible API (not always opposed) 
Complexity weighted towards User vs Complexity weighted towards Core
Backend Complexity vs Frontend Complexity

PR #1965, opened 2021-05-28
fix: remove forced location permissions

conrad-vanl:
CF got this denial on trying to ship an app update:

image
> We noticed your app encourages or directs users to allow the app to access the location. Specifically, your app directs the user to grant permission in the following way(s):
> - A message asks the user to go to the Settings app and grant access after they deny permission.
Permission requests give users control of their personal information. It is important to respect their decision about how their data is used

It appears that if a user selects "No" to permissions we show an alert dialog letting them know that we won't be able to use their location. I believe simply turning these alerts to _off_ will fix this denial, and the app still fundamentally works.

@vinnyjth why did we add these alerts int he first place? do you remember what we were trying to fix?

### DIFF
--- a/packages/apollos-ui-mapview/src/utils/getUserLocation.js
+++ b/packages/apollos-ui-mapview/src/utils/getUserLocation.js
@@ -1,9 +1,5 @@
 /* eslint-disable react-native/split-platform-components */
-import {
-  PermissionsAndroid,
-  ToastAndroid,
-  Platform,
-} from 'react-native';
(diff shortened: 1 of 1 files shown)


vinnyjth:
I'm surprised we even ask again if we've been denied.

vinnyjth (approved):

conrad-vanl:
cc @asleepingpanda

asleepingpanda:
This is great! I've applied the patch to our node_modules locally to get us out the door with this solution. It's working great!

PR #1968, opened 2021-05-28
fix: Bug causing new comments to show up only after a refresh

vinnyjth:


vinnyjth:
This was the warning being thrown. Fixed by standardizing the fields in addComment / commentList via a fragment. 

Screen Shot 2021-05-28 at 1 40 59 PM

redreceipt (approved):

PR #1973, opened 2021-06-02
chore: give ui-connected a peer dep

redreceipt:


redreceipt:
@vinnyjth the codecov report here has to be our testing with apollo snapshot inconsistencies right?

vinnyjth (approved):

vinnyjth:
@redreceipt It likely failed to run for some reason on `master`, so what you're seeing is the change vs an older commit?

PR #1974, opened 2021-06-02
chore: makes app developement faster

redreceipt:
uses the default URL for the API so there's no need to type in the encryption password

### DIFF
--- a/scripts/setup-api.sh
+++ b/scripts/setup-api.sh
@@ -19,6 +19,7 @@ cd ..
 
 shopt -s dotglob
 mv apollos-templates/apollos-church-api/* ../ApollosApi
+cp apollos-templates/yarn.lock ../ApollosApi
 
 cd ..
 
--- a/scripts/setup-app.sh
+++ b/scripts/setup-app.sh
@@ -8,13 +8,7 @@ git clone git@github.com:ApollosProject/apollos-templates.git
 
 cd apollos-templates
 git checkout ${BRANCH:-master}
-tput setaf 1
-echo "Please enter your .env password (typically stored in 1Password)"
-tput sgr0
-
-read SECRET
-./scripts/secrets.sh -d $SECRET
-mv ./apolloschurchapp/.env.shared ./apolloschurchapp/.env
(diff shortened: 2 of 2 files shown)


vinnyjth:
What do we do for the few PRs that rely on values stored in the `.env`? I'm thinking `OneSignal`. I'm fine merging this, but right now running commands / rebuilding the app after running `yarn app` is a bit of an undefined process.

redreceipt:
That's true. I was thinking if there was a variable a certain needs, it would be specified the same way we do branches.

vinnyjth:
Different idea - can we make a file called `.secret` that would be kept out of git and would store the password? We could even do the same thing on `templates` so clients would only have to create a `.secret` file with their password once, and then `yarn secrets` wouldn't need any flags.

redreceipt:
Not sure that would solve this particular problem, the directory is wiped every run which is probably the right thing.

vinnyjth:
The file would live on the root. Which would be true for both `yarn app` in `apps` and `yarn secrets` in templates.

redreceipt:
Honestly not crazy about it. This is really only a problem for development in apps, don't need the password that often in templates. and it would be one more unconventional file that someone would have to explain how to use on the template. I like the current flow because if you don't know how to use it, it fails and tells you exactly what to do. 

Another reason I like this solution is that it _actually_ allows people to contribute open-source to the core app code. Up until now, unless we give out the password, you can't really contribute. I've been brainstorming how to do the same for the API, whether we have a read-only rock token or something for the API.

vinnyjth:
fair points, how about "push enter to continue". If you didn't type a secret, default to the public api

redreceipt:
I like that

redreceipt:
ok @vinnyjth

scripts/setup-api.sh
```diff
@@ -19,6 +19,7 @@ cd ..
 
 shopt -s dotglob
 mv apollos-templates/apollos-church-api/* ../ApollosApi
+cp apollos-templates/yarn.lock ../ApollosApi
```
vinnyjth, line 22: 👍

vinnyjth (approved):

PR #1976, opened 2021-06-03
chore: Dynamically look up Rock device mobile type ID on user registration

tomthornton:
In GraphQL Playground, run 
`mutation addDevice {
  updateUserPushSettings (input: { pushProviderUserId: "arbitrary id"}) {
    id
  }
}`


vinnyjth (approved):
Looking great! Thanks for knocking this out.

PR #1978, opened 2021-06-04
fix: Close search modal when tapping on an item

vinnyjth:
Need to get a discussion going here. Current approach works well but is a bit of a hack. 

To test, `BRANCH=search-connected yarn app`

### DIFF
--- a/packages/apollos-ui-connected/src/SearchScreenConnected/SearchScreenConnected.js
+++ b/packages/apollos-ui-connected/src/SearchScreenConnected/SearchScreenConnected.js
@@ -3,9 +3,13 @@ import { View } from 'react-native';
 import { SafeAreaView } from 'react-native-safe-area-context';
 import PropTypes from 'prop-types';
 import { throttle } from 'lodash';
+import { useNavigation } from '@react-navigation/core';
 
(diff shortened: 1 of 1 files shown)


redreceipt:
yea don't love this. @conrad-vanl didn't you suggest something about `route.params` or something?

vinnyjth:
@redreceipt @conrad-vanl I tried moving the modal to the root, and then moving the modal inline with the tabs, and then moving the modal inline with the Home tab itself. Same behavior in all three places.

redreceipt:
@vinnyjth what's the branch called? I don't think `search-connected` is right and I don't see any relevant draft PRs

Screen Shot 2021-06-09 at 11 52 54 AM

vinnyjth:
Looks like it was merged 😬 
https://github.com/ApollosProject/apollos-templates/pulls?q=is%3Apr+is%3Aclosed

redreceipt:
Oh cool. Guess the core PR wasn’t necessary so this should work with master.

redreceipt:
We've done goofier things with nav than this. I'm gonna test and approve and we can circle back around later if the UX is too annoying or we get a better idea.

redreceipt (approved):

PR #1984, opened 2021-06-08
chore: remove custom datepicker text color

redreceipt:
Can't figure out how the background color of the box is determined but on a black ground, it looks bad. The defaults seem fine. No snaps changed. Check the core app with `yarn app`

Group 1


redreceipt:
just FYI, the background color is black with dark mode. I think default settings for this picker are the right move instead of trying to theme it.

vinnyjth:
We added those custom colors because the date picker was broken when left to the defaults. wonder when it started working again

vinnyjth (approved):
image

PR #1986, opened 2021-06-08
fix: properly refetch liked content list

redreceipt:
Not really sure if this will fix anything but it's at least a step in the right direction. Currently, the liked content list isn't right on a production app. 

IMG_00B8406DA291-1

This insures that it's being refetched with the rest of it.


vinnyjth (approved):
Nice! great addition, and also gets us closer to this being a `feature`

PR #1987, opened 2021-06-10
chore: move onboarding to core

redreceipt:
test with `onboarding` branch in templates

### DIFF
--- a/packages/apollos-ui-onboarding/__mocks__/react-native-safe-area-context.js
+++ b/packages/apollos-ui-onboarding/__mocks__/react-native-safe-area-context.js
@@ -1,4 +1,8 @@
+import { View } from 'react-native';
+
 module.exports = {
+  SafeAreaConsumer: ({ children }) =>
+    children({ top: 0, bottom: 0, left: 0, right: 0 }),
   SafeAreaProvider: ({ children }) => children,
-  SafeAreaView: jest.requireActual('react-native').View,
+  SafeAreaView: View,
 };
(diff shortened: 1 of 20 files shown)


vinnyjth (commented):
What do I do if I want an onboarding screen _not_ to show up? 

How do I swap out the full screen background view? 

I think those are the two biggest things we'll need to solve here to make this mergeable.

redreceipt:
@vinnyjth I can't tell how this `version` prop is being used? https://github.com/ApollosProject/apollos-apps/blob/617625d89861ea366ae3b9411c2b6f88b88e8cf9/packages/apollos-ui-onboarding/src/Onboarding/index.js#L64

vinnyjth:
> @vinnyjth I can't tell how this `version` prop is being used?
> 
> https://github.com/ApollosProject/apollos-apps/blob/617625d89861ea366ae3b9411c2b6f88b88e8cf9/packages/apollos-ui-onboarding/src/Onboarding/index.js#L64

https://github.com/ApollosProject/apollos-apps/blob/617625d89861ea366ae3b9411c2b6f88b88e8cf9/packages/apollos-ui-onboarding/src/OnboardingSwiper/index.js#L79-L82

redreceipt:
oh. so would setting version to 0 disable them?

redreceipt:
ok @vinnyjth take a look at this

vinnyjth:
> oh. so would setting version to 0 disable them?

Setting user's version to a higher number (like 3) would. I don't think that's what we want though. Some of the apps (I think Willow and Grace) omit some of the slides. I think WIllow omits the gender/age and maybe Grace omits campus selection. I think it's reasonable that we make sure that's still possible

vinnyjth:
or maybe we tell everyone that it's not supported out of the box, and they need to copy this component in? I can see that being reasonable for apps moving forward

redreceipt:
I think I came up with a solution on my latest commit

vinnyjth:
Oh nice! apologies, didn't see it the first time around.

vinnyjth (approved):

PR #1990, opened 2021-06-12
chore: move events to core

redreceipt:
test with `events` branch

Screen Shot 2021-06-11 at 9 54 41 PM


vinnyjth:
I feel great about deleting Events from Core (I don't think any app uses it) but I feel like we might be better off _not_ moving it to `apps`. Instead, how do you feel about updating `Event` (on the server) and `ContentNodeConnected` (on the client) so Event can be rendered by `ContentNodeConnected`?

redreceipt:
Could we make that decision _after_ moving to core? If no one uses it, it will be easy to deprecate. I'm just trying to shrink the template before Chase Oaks. Honestly, I don't know why we don't invest in `Events` as they are because I'm pretty sure every church uses them in Rock and you get things like campus support out of the box.

vinnyjth:
I feel great about investing in events, but feel like we should do it in the context of how we've structured content items. should be way less surface area and easier integration that way. 

How about we delete in templates now? Any church app that _is_ using Events (Willow _MIGHT_ be, but I'll check before updating them) can choose not to delete those files and then we can tackle porting them over to `apps` when we have time.

redreceipt:
I think they're fundamentally different than content items though, even if we do them on postgres, they should support ical and registration and calendars and things that content items shouldn’t touch. Trying to merge all that functionality into a generic “node” seems like a bad idea. Plus, events are arguably more important for churches than content items. Not all churches have content, all churches have events.

And if I’m correct in assuming, content items in DB will be coming with 3.0 which is sooner than later. It would be easy to move events into core now as they are and deprecate at the same time. And then just delete in 3.0. Reason I don’t like just deleting and not moving is then we’ll have dangling API code that we may never circle back around and get to in all honesty. Plus technically there will be a bug, UPCOMING_EVENTS also will be broken. At least this way, we’ll be motivated to change it.

vinnyjth:
@redreceipt They definitely different in how they act, but I think they are pretty similar in how they render. As of right now, they are an image, an html description, and a block describing when/where. The only difference between that and the `ContentNodeConnected` is the when/where block, and that could be implemented as either a connected component with an associated fragment or as a feature.

 On the API, I believe it would stay pretty much identical to what it is today except it would implement `ContentNode` (which I believe it technically already does, but it doesn't have  `implements ContentNode`) There's a chance we would need to rename fields in order to match 100%, but I think that would be appropriate. 

I agree that dangling API code is bad, but moving in code which is contrary to what we want to support seems a bit worse, especially if nobody except Willow is using it right now :)

redreceipt:
I'd rather not do anything now that will complicate upgrade flows. I'll just close both of these and we can circle back around later when we're motivated to fix events as we want.

PR #1991, opened 2021-06-12
chore: remove yalc

redreceipt:


package.json
```diff
@@ -8,11 +8,7 @@
     "nohoist": []
   },
   "scripts": {
-    "start": "lerna run build && yarn push-all && yarn build:watch",
-    "build:watch": "node ./scripts/watch.js",
```
vinnyjth, line 12: We still need `yarn build build:watch` - otherwise changes aren't picked up when making API changes with `yarn api`

package.json
```diff
@@ -47,7 +43,6 @@
     "eslint-plugin-prettier": "^3.1.4",
     "lerna": "^4.0.0",
     "prettier": "^2.1.2",
-    "react-native-storybook-loader": "^1.8.0",
-    "yalc": "^1.0.0-pre.35"
```
vinnyjth, line 51: FINALLY :D

vinnyjth (changes requested):
Need to keep `build:watch`, but otherwise looking great :D

vinnyjth:
> Need to keep `build:watch`, but otherwise looking great :D

I should add - we don't currently have `build:watch` integrated into `yarn api` but we need to :) Right now if you make a change in any of the `data-connector` packages, we end up needing to ether run `yarn lerna run build` to pick up those changes, or run `yarn api` again.

redreceipt:
I think the expectation now is that we run have to run `yarn start` (it does the watch script you're looking for) and then run `yarn api` in a separate window. they can't be sequential because we don't want any node processes to start in the background

vinnyjth (approved):

PR #2001, opened 2021-06-17
fix: correctly sorts latest series children

redreceipt:
I'm ok overriding the default sort like this because the purpose of this algorithm is to pull the most recent


vinnyjth (approved):
I buy it. Good call.

Issue #2002, opened 2021-06-18
IOS Provisional Notifications


camrun91:
as of IOS 12 Onesignal added a checkbox on the ios configuration to all the user to set up provisional Notifications. if this is enabled it makes it so the onboarding process does not work as intended. The "Yes get notified" button kicks you out to settings every time.

camrun91:
The code here https://github.com/ApollosProject/apollos-apps/blob/0de8ca3e013e81d097264f8809abca596ec3ebe1/packages/apollos-ui-notifications/src/permissionUtils.js#L48  

always returns granted with values of false for most of the settings.

redreceipt:
> as of IOS 12 Onesignal added a checkbox on the ios configuration to all the user to set up provisional Notifications

can you show me where this is? I don't fully understand the problem

camrun91:
<img width="1792" alt="Screen Shot 2021-06-18 at 10 27 09 AM" src="https://user-images.githubusercontent.com/16566029/122597513-1210d500-d031-11eb-8ae5-49c880fa2aa2.png">
That is a screenshot of the settings on one signal. If that box is checked when someone set up one signal then here is what happens. 

when a user installs that app notification permission is granted, however, it is never asked for, The incoming notifications will come through silently to notification center, They will have an option to keep or turn off. If you select to keep you get another prompt for how you want to receive them.  I will add a screenshot of the notification as it came through for me yesterday.

camrun91:
![image](https://user-images.githubusercontent.com/16566029/122597879-a2e7b080-d031-11eb-8d4a-715867210288.png)

@redreceipt here is how they come through.

PR #2003, opened 2021-06-18
fix logic for Provisional IOS notifications.

camrun91:
Combines the logic for IOS Permission check with the logic for checking if we have asked for permission or not. This works in my testing. If there is a case I am missing here I will gladly add it. 

Documentation from onesignal 
https://documentation.onesignal.com/docs/ios-customizations

### DIFF
--- a/packages/apollos-ui-notifications/src/permissionUtils.js
+++ b/packages/apollos-ui-notifications/src/permissionUtils.js
@@ -43,9 +43,21 @@ const GET_PUSH_ID = gql`
   }
 `;
 
-const requestPermissions = (updateStatus) => {
+const requestPermissions = (
+  updateStatus,
+  { hasPrompted, hasPushPermission }
+) => {
   checkNotifications().then((checkRes) => {
+    console.log('CheckRes is: ', checkRes);
     if (checkRes.status === RESULTS.DENIED) {
+      requestNotifications(['alert', 'badge', 'sound']).then(() => {
+        updateStatus();
+      });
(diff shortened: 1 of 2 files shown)


camrun91:
Should fix #2002

camrun91:
@redreceipt looks like I may not have the linter set up on my fork right.

redreceipt:
This is great @camrun91 ! we'll get this looked at. I'll fix the linter on my end. Make sure you `yarn` before so you pick up the correct version of `eslint`. If you have already, not sure why it's failing for you.

for future, our PRs need to start with chore, fix, or feat so they magically show up in the changelog, the title CI check gives the details.

redreceipt (commented):
Could you post a video of what happens with and without this fix? This is very edge casey which makes it hard to validate on our end.

camrun91:
@redreceipt will a video from simulator work? I can do one later showing the setting in Onesignal as well as how it affects the app.

redreceipt:
Sure, anything to show a before and after. Also something that confirms this won’t affect the vast majority of normal OneSignal configurations would be very helpful.

camrun91:
@redreceipt it actually looks like this has changed how it works with the recent updates in core to update Onesignal. I am trying to verify on a "New device" as I think the notifications setting is getting saved somewhere.

redreceipt:
@camrun91 still working on this?

camrun91:
I think the latest that I found was that it was working as expected. so it can be closed.

PR #2007, opened 2021-06-22
Updates m1 script fix

tomthornton:
Apparently it's reading my architecture as x86_64 instead of arm64 now.

### DIFF
--- a/packages/apollos-cli/scripts/secrets.sh
+++ b/packages/apollos-cli/scripts/secrets.sh
@@ -5,7 +5,7 @@ if [ $# -ne 2 ]; then
 fi
 
 function encrypt() {
-  if [[ `uname -m` == 'arm64' ]]; 
+  if [[ `uname -m` == 'x86_64' ]]; 
   then
     /opt/homebrew/Cellar/openssl@1.1/1.1.1k/bin/openssl enc -aes-256-cbc -pbkdf2 -iter 20000 -in "$1" -out "$1".enc -k "$2"
   else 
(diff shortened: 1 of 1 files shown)


vinnyjth:
Hmmm - are you using rosetta? 

`x86_64` is my current arch. 

image

tomthornton:
Yes. That's probably why.

vinnyjth:
Maybe we need something like this
```
#!/bin/zsh
 
arch_name="$(uname -m)"
 
if [ "${arch_name}" = "x86_64" ]; then
    if [ "$(sysctl -in sysctl.proc_translated)" = "1" ]; then
        echo "Running on Rosetta 2"
    else
        echo "Running on native Intel"
    fi 
elif [ "${arch_name}" = "arm64" ]; then
    echo "Running on ARM"
else
    echo "Unknown architecture: ${arch_name}"
fi
```
https://cutecoder.org/software/detecting-apple-silicon-shell-script/

rkdavidson:
I forget what exactly I did to work around this, but IIRC it was just a matter of `brew install`ing the openssl package and double checking my shell's `PATH` stuff.

@tomthornton and @vinnyjth , think this PR is still necessary?

I've had other M1 issues like needing to open Xcode in Rosetta to get `apollos-apps` to run, etc.
So I'm sure there's some work to do, but this specific PR may not be necessary.

redreceipt:
@tom we still need this PR?

redreceipt:
@tomthornton we still need this PR?

tomthornton:
> @tomthornton we still need this PR?

Yeah I need to make a change to it though. I will at some point today.

redreceipt:
@tomthornton are you still having trouble on the M1

tomthornton:
@redreceipt nope, no more problems.

PR #2018, opened 2021-06-24
feat: add font weight prop

MatthewKyle22:
Adds fontWeight prop to ActionBarItem component which passes to UIText for style control

### DIFF
--- a/packages/apollos-ui-kit/src/ActionBar/ActionBarItem.js
+++ b/packages/apollos-ui-kit/src/ActionBar/ActionBarItem.js
@@ -21,6 +21,7 @@ const ActionBarItem = ({
   icon = 'empty',
   tint,
   size,
+  fontWeight,
   label = '',
   ...touchableProps
 }) => (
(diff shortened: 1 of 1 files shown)


redreceipt (commented):
I'm good with it, wanna write me a story and test?

packages/apollos-ui-kit/src/ActionBar/ActionBarItem.js
```diff
@@ -29,7 +30,7 @@ const ActionBarItem = ({
       <Icon fill={tint} size={size} name={icon} />
 
       {/* using stlye object here is just as efficient as creating a new styled component with prop, since UIText is already a styled component */}
-      <UIText style={{ color: tint }}>{label}</UIText>
+      <UIText style={{ color: tint, fontWeight }}>{label}</UIText>
```
conrad-vanl, line 33: I honestly think the better solve here is to allow a `style` prop to be passed down

redreceipt: or just "name" that `UIText` component so they can override the style prop? that matches better and doesn't extend the API

conrad-vanl:
Closing in favor of other solutions discussed above. We can re-open when changes are made.

PR #2021, opened 2021-06-25
chore: Update Postgres media model with orginId and originType

tomthornton:


vinnyjth:
@tomthornton Since we have already deployed the initial migration, you'll have to do this in an additional, second migration :( Looking great otherwise :D

tomthornton:
@vinnyjth We can't just drop the table, delete the SequelizeMetadata row,  and restart the server locally?

vinnyjth:
@tomthornton we could, but I think it's probably better that we add the second migration. I don't want to get into the habit of doing that.

packages/apollos-data-connector-postgres/src/media/migrations/002_update_media_primary_keys.js
```diff
@@ -0,0 +1,34 @@
...
+    );
+
+    await queryInterface.addIndex('media', ['originId', 'originType'], {
+      unique: true,
+    });
+  });
+}
+
+async function down({ context: queryInterface }) {
+  await queryInterface.dropTable('media');
```
vinnyjth, line 29: In the future, the `down` should likely be something that removes just those two columns. It's probably okay for now though, since I don't think the `down` will ever be run ;)

packages/apollos-data-connector-postgres/src/media/migrations/002_update_media_primary_keys.js
```diff
@@ -0,0 +1,34 @@
+import { Sequelize } from 'sequelize';
+
+async function up({ context: queryInterface }) {
```
vinnyjth, line 3: 👍

packages/apollos-data-connector-postgres/src/media/migrations/002_update_media_primary_keys.js
```diff
@@ -0,0 +1,34 @@
...
+  });
+}
+
+async function down({ context: queryInterface }) {
+  await queryInterface.dropTable('media');
+}
+
+const name = '002-update_media_primary_keys';
+
+module.exports = { up, down, name, order: 4 };
```
vinnyjth, line 34: So that it's one bigger than `create media` :) 
```suggestion
module.exports = { up, down, name, order: 5 };
```

vinnyjth (approved):

vinnyjth (approved):

PR #2023, opened 2021-06-25
chore: deprecate NavigationService

redreceipt:
test with `tabs` branch for the tabs to not be a modal


packages/apollos-ui-onboarding/src/Onboarding/index.js
```diff
@@ -56,12 +51,7 @@ function Onboarding(props) {
             userId: data?.currentUser?.id,
             version: ONBOARDING_VERSION,
           });
-          navigation.dispatch(
-            NavigationService.resetAction({
-              navigatorName: 'Tabs',
-              routeName: 'Home',
-            })
-          );
+          navigation.navigate('Tabs');
```
vinnyjth, line 54: Have you considered hardcoding the `resetAction` instead? The reason we went with `reset` is that on Android you could press the back button to get back to onboarding, even if you turned off gestures.

redreceipt: oh shoot, yea i forgot about that

redreceipt: ok @vinnyjth good now

PR #2026, opened 2021-06-28
fix: Postgres Media bugfixes

vinnyjth:
1. Migration was't running because there were two primary keys (`originId` can't also be a primary key) 
2. `hasOne` expects there to be a primary key on ContentItem. I don't think we need to setup the media.contentItem relationship at this point
3. Added a few fields that were missing in the model via a migration.


tomthornton (approved):
Looks good. I'll go back and modify the Cover Image DAG to fit this new schema

PR #2035, opened 2021-06-29
feat: Add cover image id

vinnyjth:


tomthornton (approved):
Yep, I had done something similar.

PR #2039, opened 2021-06-30
chore: allow video tags through htmlcontent

redreceipt:


packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -318,13 +318,15 @@ export default class ContentItem extends RockApolloDataSource {
         'br',
         'caption',
         'img',
+        'video',
```
vinnyjth, line 321: Is this for LH?

redreceipt: yea

vinnyjth: What does a video tag look like? As long as there aren't any children this will render fine in the app.

conrad-vanl: I don't like the idea of doing this in core....

conrad-vanl: I'd rather turn off html stripping on LH altogether

redreceipt: NOW we're talking

PR #2043, opened 2021-07-01
feat: Create tags and node tags model

tomthornton:


tomthornton:
@vinnyjth wasn't sure how you wanted to handle relationships with sequelize.

vinnyjth:
@tomthornton I would go ahead and create the `ContentItem.getTags()` and `Person.getTags()` relationship for now :)

vinnyjth:
@tomthornton this might also be a good PR to add tests demonstrating creating tags, and then assigning tags to both content items and people. Something similar to https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-data-connector-postgres/src/content-item-categories/__tests__/model.tests.js#L5

tomthornton:
@vinnyjth two quick questions
- Can you send me an example for how to set up People.getTags() and ContentItems.getTags()? I figured that Sequelize did this automatically or is this part of the data source?
- Do you have examples of tags for content items/people? Was going to use them for the test. I guess I could put anything for the tags.

vinnyjth:
@tomthornton 
> Can you send me an example for how to set up People.getTags() and ContentItems.getTags()? I figured that Sequelize did this automatically or is this part of the data source?

Yah - this happens automatically when you define the relationship. Here's an example for `getParent()`
https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-data-connector-postgres/src/content-items/model.js#L27

> Do you have examples of tags for content items/people? Was going to use them for the test. I guess I could put anything for the tags.
For the tests, I would instantiate the model with the bare minimum set of attributes. You can look at pretty much any of the `model.tests.js` in the `data-connector-postgres` module for examples :)

packages/apollos-data-connector-postgres/src/content-items/migrations/002_add_cover_image_id_column.js
```diff
@@ -0,0 +1,15 @@
+import { Sequelize, DataTypes } from 'sequelize';
```
vinnyjth, line 1: @tomthornton Is this file a duplicate?

packages/apollos-data-connector-postgres/src/tags/model.js
```diff
@@ -0,0 +1,34 @@
...
+import { defineModel, configureModel } from '../postgres';
+
+const createModel = defineModel({
+  modelName: 'tag',
+  resolveType: () => 'Tag', // shouldn't get called, ContentItem will be set on the shovel.
+  external: true,
+  attributes: {
+    name: DataTypes.TEXT,
+    type: DataTypes.TEXT,
+    data: DataTypes.JSONB,
```
vinnyjth, line 11: Awesome! Good use of `JSONB`

packages/apollos-data-connector-postgres/src/tags/model.js
```diff
@@ -0,0 +1,34 @@
...
+  },
+});
+
+const setupModel = configureModel(({ sequelize }) => {
+  sequelize.models.contentItem.belongsToMany(sequelize.models.tag, {
```
vinnyjth, line 16: Looking good!

packages/apollos-data-connector-postgres/src/tags/__tests__/model.test.js
```diff
@@ -0,0 +1,69 @@
+import { sequelize, sync } from '../../postgres/index';
```
vinnyjth, line 1: Awesome! These tests are very helpful when reviewing and will make future additions to this feature a breeze.

packages/apollos-data-connector-postgres/src/tags/migrations/001_create_tags.js
```diff
@@ -0,0 +1,45 @@
+import { Sequelize, DataTypes } from 'sequelize';
```
vinnyjth, line 1: This `migration` looks great! 

However, I think we are missing a migration. The `content_tag` table and the `people_tag` tables will also need to be created. They are created automatically in the tests, since we call `sync` in the test environment, but in production we don't call `sync` and rely on the migrations. 

You'll have to work backwards from the postgres schema changes that `sequelize` makes with the ` belongsTo` calls to determine the tables/fields it needs, and then add a migration for those tables/fields. One way to is to boot up your API, include the `tag` model in your `src/data/index.postgres` file, and call the `sync` function from the `src/server.js` file. Or you could drop some console.logs in your tests.(but I'm not quite sure what you would log)

vinnyjth (changes requested):
You're really close! We need migrations for the join tables :) Nice work on this PR, I appreciate the tests :D 

https://github.com/ApollosProject/apollos-apps/pull/2043/files#r663004876

tomthornton:
@vinnyjth doesnt sequelize take care of the join tables? Or do we have to actually explicitly create them in another migration?

vinnyjth:
@tomthornton it implicitly creates the join tables if you use `sync`, but we don't use sync in the api. We do use it for tests though, which is why the tests run fine.

tomthornton:
@vinnyjth added some fields to the junction tables to get these tests working. Wasn't sure if the tags model should be external or not, so I just left it false. But at least the tests are running good!

packages/apollos-data-connector-postgres/src/tags/migrations/001_create_tags.js
```diff
@@ -0,0 +1,47 @@
...
+    },
+    updatedAt: {
+      type: Sequelize.DATE,
+      allowNull: false,
+    },
+  });
+}
+
+async function down({ context: queryInterface }) {
+  await queryInterface.dropTable('tag');
```
vinnyjth, line 42:
```suggestion
  await queryInterface.dropTable('tags');
```

vinnyjth:
Looking good! Awesome tests, those will make it so much easier to find issues with migrations before we merge. 

Once you fix the `linter` for this PR we can merge it.

vinnyjth (approved):

PR #2044, opened 2021-07-02
fix: Fix theme merging

conrad-vanl:
#1814 introduced a few regressions to the way themes are merged, and with #1838 using `useTheme` more extensively, it became pretty quickly clear of the issues.

Theming is hard.

I like to think of our theming system as having two parts - a theme input, and a theme object. A theme input is used to generate a theme object.

For example, if I do something like `<Themer theme={{ type: 'dark' }} />` you'd expect `Themer` to create a theme with the dark color scheme. In this case, `{ type: 'dark' }` is the theme input, and the resulting theme object with a dark color scheme is the generated theme.

Currently in master we are merging generated themes with theme inputs, resulting in incorrect themes getting generated. Example:

```jsx
<Themer theme={{ type: 'dark' }}>
  <Themer theme={{ type: 'light' }} />
</Themer>
```

What would you expect the inner theme to look like? I would expect the inner theme to be a "light" theme with "light" colors. However, on master, the inner them will have `dark` theme colors, even though it will have a key `type` of "light". 

In other words, the inner theme is simply merging the new theme input `{ type: 'light' }` with the existing generated theme (which is all dark colors). Something like:

```js
const previouslyGeneratedTheme = { type: 'dark', colors: { /* dark theme colors */ } }
const newThemeInput = { type: 'light' }

const newTheme = { ...previouslyGeneratedTheme, ...newThemeInput }
```

What this PR does is keep track of the previously used `previousThemeInput` to generate the theme, and merges that with the new `newThemeInput`. Then, use the newly merged `mergeThemeInput` to create a new theme (using `createTheme`). Something like:

```js
const previousThemeInput = { type: 'dark' }
const newThemeInput = { type: 'light' }

const newTheme = createTheme({ ...previousThemeInput, ...newThemeInput }
```

I added a new test case showing this condition within our test coverage, and verified this fixes theme issues in the example app.

Furthermore, I removed other instances of `createTheme` and so now `Themer` is the only place that `createTheme` should ever be used.

### DIFF
--- a/packages/apollos-ui-kit/src/theme/OLD_theme.js
+++ b/packages/apollos-ui-kit/src/theme/OLD_theme.js
@@ -1,16 +1,10 @@
 import React from 'react';
-import { useColorScheme } from 'react-native';
 import PropTypes from 'prop-types';
 import mergeStyles from '../styled/mergeStyles';
 import Themer, { useTheme } from './Themer';
-import createTheme from './createTheme';
 
 export const ThemeProvider = ({ themeInput, iconInput, ...props }) => {
-  const type = useColorScheme();
-  const customTheme = { type, ...themeInput };
-  return (
-    <Themer theme={createTheme(customTheme)} icons={iconInput} {...props} />
-  );
+  return <Themer theme={themeInput} icons={iconInput} {...props} />;
 };
 
 ThemeProvider.propTypes = {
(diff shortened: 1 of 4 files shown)


redreceipt (approved):
very nice

PR #2045, opened 2021-07-02
fix: Feature titles spacing

conrad-vanl:
Main things this PR aims to do:

- Makes titles across all features consistent
- Cleans up spacing so we don't have gigantic spacing between some features

New API created:

- Creates a `<FeatureTitles/>` component that matches Figma that can be used across all features for consistent titles
- Adds color-based props to typography components so that we can easily style typography elements. Such as `<H2 secondary />`. These prop names matches the color options in Figma:

Screen Shot 2021-07-01 at 11 11 40 PM

I believe our UI-Kit is becoming mature enough that short-hand props like this can make sense, especially since it's saves us from having to create the same `styled` components over and over again.

### DIFF
--- a/packages/apollos-ui-connected/src/FollowListConnected/FollowingListConnected.js
+++ b/packages/apollos-ui-connected/src/FollowListConnected/FollowingListConnected.js
@@ -1,7 +1,7 @@
 import React, { useEffect } from 'react';
 import { useQuery } from '@apollo/client';
 import PropTypes from 'prop-types';
-import { named, H4 } from '@apollosproject/ui-kit';
+import { named, FeatureTitles } from '@apollosproject/ui-kit';
 import GET_USERS_FOLLOWING from './getFollowing';
 import FollowListConnected from './FollowListConnected';
 
@@ -23,7 +23,7 @@ const FollowingListConnected = ({
 
   return (
     <Component
-      header={Header}
+      header={<Header />}
       loading={loading}
       followers={suggestedFollows}
       {...props}
@@ -48,7 +48,7 @@ FollowingListConnected.propTypes = {
 
 FollowingListConnected.defaultProps = {
   Component: FollowListConnected,
-  Header: <H4>{'Following'}</H4>,
+  Header: () => <FeatureTitles subtitle="Following" />, // eslint-disable-line react/display-name
 };
 
 FollowingListConnected.displayName = 'FollowingListConnected';
--- a/packages/apollos-ui-connected/src/HeroListFeatureConnected/HeroListFeature/index.js
+++ b/packages/apollos-ui-connected/src/HeroListFeatureConnected/HeroListFeature/index.js
@@ -1,36 +1,15 @@
 import React, { memo } from 'react';
 import PropTypes from 'prop-types';
+import { StyleSheet } from 'react-native';
 import { get } from 'lodash';
 import {
   ActionList,
   DefaultCard,
-  H3,
-  H5,
   PaddedView,
-  styled,
   TouchableScale,
+  FeatureTitles,
 } from '@apollosproject/ui-kit';
 import { LiveConsumer } from '../../live';
-
-const Header = styled(
-  ({ theme }) => ({
-    paddingTop: theme.sizing.baseUnit * 3,
-    paddingBottom: 0,
-  }),
-  'ui-connected.HeroListFeatureConnected.HeroListFeature.Header'
-)(PaddedView);
-
-const Title = styled(
-  ({ theme }) => ({
-    color: theme.colors.text.tertiary,
-  }),
-  'ui-connected.HeroListFeatureConnected.HeroListFeature.Title'
-)(H5);
-
-const Subtitle = styled(
-  {},
-  'ui-connected.HeroListFeatureConnected.HeroListFeature.Subtitle'
-)(H3);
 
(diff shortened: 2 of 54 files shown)


packages/apollos-ui-connected/src/FollowListConnected/FollowingListConnected.js
```diff
@@ -23,7 +23,7 @@ const FollowingListConnected = ({
 
   return (
     <Component
-      header={Header}
+      header={<Header />}
```
redreceipt, line 26: was it broken before?

redreceipt: also this is unconventional right? shouldn't it be a function and not an iffe?

conrad-vanl: Not Broken. And I did not change the way it works. Look at like 51 below - previously the `Header` prop was an `iffy`, then passed into this `header` prop.

Usually in React, a prop with a capitalized first letter (`Header`) is expected to be a component, and a lowercase `header` would be an `iffe`.

Also, the prototype validation here for `Header` assumes you can pass a `func` or an `object` (both component types), but previously it only really supported an `iffe` because that's what `FollowListConnected` expects, it later just renders `header` like `{ header }`.

So while not technically broken before, it is definitely a code smell, and this is mildly better....but still to great.

conrad-vanl: Oh, also the real reason I changed it:

```
Component.defaultProps = {
   header: <SomeStaticReact />
}
```

is subject to errors because `<SomeStaticReact />` is creating an element outside of our normal component rendering...which I _think_ makes it subject to issues like when changing themes (context). But I'm not 100% certain. Either way, it is a smell - that's what I was trying to fix.

You're right about passing `header={<Header />}` isn't great either. but we were doing that before already.

redreceipt: ah, maybe to keep with the standard, `Header` should be an overridable prop on `FollowListConnected` instead? and then nothing is passed in from this component. also fine saving that for another PR.

packages/apollos-ui-connected/src/HeroListFeatureConnected/HeroListFeature/index.js
```diff
@@ -99,13 +78,30 @@ const loadingStateArray = [
   },
 ];
 
+// we use a stylesheet instead of styled here so that we can just
+// pass DefaultCard as the HeroComponent below, and then pass style
+// into it - that way it's easier to swap out for other cards.
+const styles = StyleSheet.create({
```
redreceipt, line 84: love this!

packages/apollos-ui-kit/src/Card/Content.js
```diff
@@ -1,11 +1,6 @@
 import PaddedView from '../PaddedView';
 import styled from '../styled';
 
-const Content = styled(
-  ({ theme }) => ({
-    paddingVertical: theme.sizing.baseUnit,
-  }),
-  'ui-kit.Card.Content.Content'
-)(PaddedView);
+const Content = styled({}, 'ui-kit.Card.Content.Content')(PaddedView);
```
redreceipt, line 4:
```suggestion
const Content = named('ui-kit.Card.Content.Content')(PaddedView);
```

packages/apollos-ui-kit/src/FeatureTitles/index.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { View } from 'react-native';
+
+import { H3, H4 } from '../typography';
+import styled from '../styled';
+
+const Container = styled({}, 'ui-kit.FeatureTitles.Container')(View);
```
redreceipt, line 8:
```suggestion
const Container = named('ui-kit.FeatureTitles.Container')(View);
```

packages/apollos-ui-kit/src/FeatureTitles/index.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
```
redreceipt, line 1: think we should write a snap here?

conrad-vanl: Ah yes I forgot to do that

packages/apollos-ui-prayer/src/PrayerFeature/index.js
```diff
@@ -5,23 +5,22 @@ import {
...
+  FeatureTitles,
 } from '@apollosproject/ui-kit';
 
 const AvatarWrapper = styled(
   ({ theme }) => ({
     flexDirection: 'row',
-    paddingBottom: theme.sizing.baseUnit * 2, // TODO: refactor CardContent to have this be the default
+    paddingBottom: theme.sizing.baseUnit * 1.5, // TODO: refactor CardContent to have this be the default
     // paddingHorizontal: theme.sizing.baseUnit * 1.5, // TODO: refactor CardContent to have this be the default
     paddingHorizontal: 0,
```
redreceipt, line 20: couldn't you just remove these padding 0 lines?

redreceipt (approved):
this is awesome

PR #2048, opened 2021-07-06
feat: Feature Schema/Model

vinnyjth:


vinnyjth:
image

packages/apollos-data-connector-postgres/src/features/model.js
```diff
@@ -0,0 +1,25 @@
...
+  },
+});
+
+const setupModel = configureModel(({ sequelize }) => {
+  sequelize.models.feature.belongsTo(sequelize.models.contentItem, {
+    foreignKey: 'parentId',
+  });
+  sequelize.models.contentItem.hasMany(sequelize.models.feature, {
+    foreignKey: 'parentId',
+    scope: { parentType: 'ContentItem' },
```
tomthornton, line 21: I noticed you are using scopes here. What are the purpose of these scopes and how should I think about adding them to other models?

tomthornton (approved):
Cool! Seems simple enough.

packages/apollos-data-connector-postgres/src/features/model.js
vinnyjth: for `polymorphic` relationships the `scopes` help define how to get features belonging to a content item specifically. So to rephrase the code: A content item has many features, and to find them you can look at " feature" rows that have a type of `parentType` and share a `parentId` with my ID. 

It's not quite as useful for us since we use `uuids` which don't collide, but it sequel build the right object if you call `content.addFeature`

PR #2049, opened 2021-07-06
feat: Feature Unique Index migration

vinnyjth:
I realized this index would be really useful when building features with the shovel. 

image


tomthornton (approved):
Yeah! I think this will definitely help

PR #2050, opened 2021-07-07
chore: Use more helpful error messages when package imports are wrong

vinnyjth:
image

@redreceipt this stops the wild goose chase for tracking down packages with bad imports way easier. stops the old `can't call property dataSource of undefined` error


redreceipt (approved):
🎉

redreceipt:
Blind trust

vinnyjth:
> Blind trust

That's why we got tests :)

PR #2053, opened 2021-07-08
feat: adds content channel images

redreceipt:


vinnyjth:
Do you want to also implement this in the shovel for me :)

packages/apollos-data-connector-rock/src/content-channels/data-source.js
```diff
@@ -6,6 +6,8 @@ const { ROCK_MAPPINGS } = ApollosConfig;
 export default class ContentChannel extends RockApolloDataSource {
   resource = 'ContentChannels';
 
+  expanded = true;
```
conrad-vanl, line 9: @vinnyjth will this have bad side-affects?

packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -398,6 +398,14 @@ export default class ContentItem extends RockApolloDataSource {
 
     if (ourImages.length) {
       image = this.pickBestImage({ images: ourImages });
+    } else {
+      // check channel
+      const channel = await ContentChannel.getFromId(root.contentChannelId);
```
vinnyjth, line 403: it might be worth doing `root.channel || ...` here since we often have already expanded channel.


.... shoot, no attributes though :(

packages/apollos-data-connector-rock/src/content-channels/data-source.js
vinnyjth: Not super bad but it slows down all Channel requests by a bit. you're better off if you can identify what attribute keys to fetch, and just fetch those attribute keys.

redreceipt: it could be any with "image" in the name so not sure there's a good way to select

redreceipt: the function relies on being able to look through all attributes

conrad-vanl: you might be able to set expanded on just this one request maybe?

redreceipt: yea i bet you're right, let me take a look

redreceipt: @vinnyjth that's not what this does is it?
Screen Shot 2021-07-08 at 1 42 10 PM

vinnyjth: unfortunately not, your only option is this one: 
https://github.com/ApollosProject/apollos-apps/blob/bdd84f481e1cefc4c8ac9b3a89ea32979b27a7ae/packages/apollos-rock-apollo-data-source/src/index.js#L80


We had a long convo about if you should be able to filter on just a single request
https://github.com/ApollosProject/apollos-apps/pull/1871

redreceipt: this works

Screen Shot 2021-07-08 at 1 50 58 PM

redreceipt: a little hacky but it gets the job done

redreceipt: and I just realized that line you referenced has a bug 😂

conrad-vanl (approved):

redreceipt:
@vinnyjth you good with removing the test that I did here? I fixed one snap that was easy but the second would be very tricky to try and reach inside the DS method given the "hacky" nature of this PR.

packages/apollos-data-connector-rock/src/content-items/__tests__/data-source.tests.js
```diff
@@ -693,6 +693,9 @@ describe('ContentItemsModel', () => {
     dataSource.getCursorByChildContentItemId = () => ({
       get: () => Promise.resolve([{ attributeValues: {}, attributes: {} }]),
     });
+    dataSource.request = () => ({
```
vinnyjth, line 696: maybe something like this to fix the test?

```suggestion
    dataSource.request = (path) => path.include('ContentChannels') ? ({ get:() => null }) : ({
```

redreceipt: niiiice, yea that's smart

redreceipt:
@vinnyjth made the "fix" but take a look at these snap changes, mine actually looks closer to what it should be except for the error? not sure how we were getting an image before on the test with no attribute values.

packages/apollos-data-connector-rock/src/content-items/__tests__/resolvers.tests.js
```diff
@@ -389,6 +389,12 @@ describe('UniversalContentItem', () => {
       }
       ${contentItemFragment}
     `;
+
+    context.dataSources.ContentItem.request = (path) =>
+      path.includes('ContentChannels')
+        ? { get: () => ({ attributeValues: {}, attributes: {} }) }
+        : { ...context.dataSources.ContentItem.request(path) };
```
vinnyjth, line 396: this function will now call itself :/ you either need to save a reference to the old `method` and then use that reference, or mock this differently

redreceipt: o shoot that makes sense

redreceipt:
closer...

redreceipt:
wow, that wasn't pretty

PR #2055, opened 2021-07-09
chore: don't wait on notifications for comments mutation

redreceipt:


packages/apollos-data-connector-postgres/src/comments/dataSource.js
```diff
@@ -31,7 +31,7 @@ class CommentDataSource extends PostgresDataSource {
     }
 
     if (visibility === Visibility.PUBLIC) {
-      await this.sendNotificationForComment({ comment, parentId });
+      this.sendNotificationForComment({ comment, parentId });
```
vinnyjth, line 34: This will likely break our tests, and I would still love to have test coverage over this 🤔 Not saying that this is the wrong solution, but could we come up with a way to make sure we verify that this function is called properly in the test environment?

redreceipt: tests passed 🤷

vinnyjth: 😂

vinnyjth (approved):

packages/apollos-data-connector-postgres/src/comments/dataSource.js
redreceipt: I think we're lucky though, I agree with you. I started to rewire but I can't get these tests to run locally at all for me to test the solution

PR #2058, opened 2021-07-10
chore: warn on content card mapper deprecation

redreceipt:


packages/apollos-ui-connected/src/ContentCardConnected/index.js
```diff
@@ -1,8 +1,12 @@
+import React from 'react';
 import { DefaultCard } from '@apollosproject/ui-kit';
 import ContentCardConnected from './ContentCardConnected';
 
-export const ContentCardComponentMapper = DefaultCard;
+export const ContentCardComponentMapper = (props) => {
+  console.warn('ContentCardComponentMapper deprecated');
```
rkdavidson, line 6: Were these components deprecated in favor of something else?
Is there anything worth mentioning in the warning message like "See XYZ instead"?

rkdavidson (approved):

packages/apollos-ui-connected/src/ContentCardConnected/index.js
redreceipt: Nah they're just ded ded

PR #2059, opened 2021-07-10
chore: converted ProtectedRoute to functional component

redreceipt:


conrad-vanl (commented):
@redreceipt do these snapshot changes look right to you? Seems like pretty big snap changes for just converting to functional component

redreceipt:
> @redreceipt do these snapshot changes look right to you? Seems like pretty big snap changes for just converting to functional component

It changed a lot because I'm using the new navigation test utility. It's wrapped in a navigator screen now.

redreceipt:
@conrad-vanl I think the snap changes here are fine, see my previous comment

vinnyjth:
I did something similar in the `ui-authentication` branch, so I likely won't review this unless we see an urgent need for it (the code it's pointing at will soon be deprecated)

PR #2062, opened 2021-07-14
chore: add action tables to content item feature feed

redreceipt:


redreceipt:
@vinnyjth  before you approve this... looks like it's already coming through on fellowship somehow???

image￼

redreceipt:
Yep this component is deprecated 😑

PR #2063, opened 2021-07-14
feat: add event details to content items

redreceipt:
Screen Shot 2021-07-14 at 5 14 58 PM


packages/apollos-ui-connected/src/ContentNodeConnected/ContentNodeConnected.js
```diff
@@ -95,6 +96,7 @@ const ContentNodeConnected = ({
       </View>
       <BackgroundView flex={false}>
         <PaddedView>
+          <EventDetailsConnected nodeId={nodeId} />
```
conrad-vanl, line 99: Should this be a feature instead?

conrad-vanl (approved):

packages/apollos-ui-connected/src/ContentNodeConnected/ContentNodeConnected.js
redreceipt: I don't think so. The component is implied from the type name.

PR #2067, opened 2021-07-17
chore: replaces config variables with env variables

redreceipt:
eliminates a lot of middle man code and is a stepping stone to not needing the config.yml file at all. test with the `env` branch


vinnyjth:
I like some of the ideas in this PR - and nice work with your thoroughness in capturing all the various variables. 

However,  I'm not sure I agree with removing the `config.yml` as a middleman. I like having the `config.yml` as a middleman since it acts as some level of documentation about all the different env variables that are needed (even if it doesn't include the variables themselves). 

Do you think that having the `.env.shared.env` have removed the need for that middleman? Has managing `env` variables proved to be easier without the `config.yml`? We introduced the `config.yml` in part to solve the problem of having a bunch of different variables needed and no one place to see what they were, and I'm worried we will end up right back there without it.

redreceipt:
> Do you think that having the `.env.shared.env` have removed the need for that middleman? Has managing `env` variables proved to be easier without the `config.yml`? We introduced the `config.yml` in part to solve the problem of having a bunch of different variables needed and no one place to see what they were, and I'm worried we will end up right back there without it.

I think fundamentally, the API shouldn't _require_ a bunch of variables to work out of the box. Part of the reason why we needed a "menu" of sorts is that the API used to not run without these various tokens. But extensive work has been done to modularize the API so that you don't _need_ to set the Algolia stuff for example for the server to run. If they want Algolia, they'll read the doc for setting up Algolia. And appropriate defaults have been set for most variables so it's not a problem.

Practically speaking the last few apps we've set up have been done through documentation, not the config.yml. For new developers to the platform, I have to explain what the file even is, it's non-obvious.

Plus I really like how this PR only loads the variables if you want the integration. The developer shouldn't know anything about bugsnag if they don't care about the bugsnag integration.

redreceipt:
Now, I hear that maybe we need some method for knowing what the `env` variables are via the command line. Something like `apollos-cli config bugsnag` might be very useful to spell them out for you

vinnyjth:
> Now, I hear that maybe we need some method for knowing what the env variables are via the command line. Something like apollos-cli config bugsnag might be very useful to spell them out for you

Yah - this is what I'm caught up on. I think we need some sort of mandatory / self documenting code about what those ENV variables are. The `config.yml` is very explicit but also a lot to parse. It also has the benefit of  making clear what env variables are being used. If you remove it, you can end up running very different versions of the app locally and in in production without noticing. That's probably okay, up to a point.  Would love to have some sort of documentation around what variables you need / what variables are being used before we ditch the config.yml 

Also I don't know if we will ever be able to truly ditch it at this point, since the feature feed stuff still relies on it. I love the work we have done though to trim the file down, and I think having a smaller and more modular `config.yml` makes what's inside the `config.yml` easier to understand and parse.

redreceipt:
@vinnyjth so with the Rock connector, we've now warned on startup that `ROCK_URL` and `ROCK_TOKEN` are required if they are not present. What if we were to take this approach with the other integrations? Most likely, the dev will follow the documentation to integrate a plugin but if something is missed, what if we just make it obvious what's going on through the console?

redreceipt:
Having them laid for you was good when every server was the same but that's not really the case anymore so a default subset of variables is misleading at best.

redreceipt:
@vinnyjth I actually think this PR can help with documentation and confusion. It should hard to misuse our API. And the current list in the config is overwhelming and doesn't let you know when you're missing something important. What if part of this PR was to throw an exception for missing variables when trying to use certain data connectors? The server will fail and you'll know exactly why right from the beginning.

PR #2068, opened 2021-07-19
chore: Update parentId and coverImageId foreign keys to set null

tomthornton:
This should allow you to delete parent content items and media without a problem


vinnyjth:
image

vinnyjth (approved):
Awesome work! You're killing it with these migrations

tomthornton:
@vinnyjth thanks!

PR #2072, opened 2021-07-20
fix: ScriptureFeature resolver to pick up Scripture references correctly 

vinnyjth:
Simulator Screen Shot - iPhone 11 - 2021-07-20 at 10 19 14


tomthornton (approved):
Nice!

PR #2076, opened 2021-07-22
fix: correctly build follow notification deep links

redreceipt:
closes https://github.com/ApollosProject/apollos-templates/issues/421


vinnyjth (approved):

vinnyjth:
Well done. If I remember correctly, the API for this code was approved prior to the deep link rewrite, but marged after the rewrite was merged, hence the mixup.

PR #2081, opened 2021-07-22
feat: adds default search sort by publish date

redreceipt:
also fixes algolia indexing and universal links 😬

Screen Shot 2021-07-22 at 2 44 13 PM


packages/apollos-data-connector-algolia-search/src/data-source.js
```diff
@@ -22,6 +22,7 @@ if (ApollosConfig.ALGOLIA.APPLICATION_ID && ApollosConfig.ALGOLIA.API_KEY) {
   INDEX.setSettings(
     ApollosConfig.ALGOLIA.CONFIGURATION || {
       searchableAttributes: ['title', 'unordered(summary)'],
+      customRanking: ['desc(publishDate)'],
```
vinnyjth, line 25: shouldn't we sort by relevancy first, then publishDate?

packages/apollos-server-core/src/index.js
```diff
@@ -196,9 +196,11 @@ export const createContextGetter = (serverConfig) => (data) => {
 
 export const createMiddleware = (data) => ({ app, context, dataSources }) => {
   const middlewares = compact(
-    mapValues(
-      { ...builtInData, ...data },
-      safeGetWithWarning('serverMiddleware')
+    Object.values(
```
vinnyjth, line 199: Good catch!

packages/apollos-data-connector-algolia-search/src/data-source.js
redreceipt: yea it does, custom ranking comes after the text input in the event of a tie

vinnyjth: 💥

vinnyjth (approved):

packages/apollos-data-connector-algolia-search/src/data-source.js
redreceipt: still don't understand the difference between using `ranking` and `customRanking` for sorting though. both seemed to have the same effect when testing

Issue #1901, opened 2021-04-28
replace graphql-tag with @apollo/client > gql


redreceipt:
we could swap out all instances of `graphql-tag` with `import { gql } from '@apollo/client';`

nlewis84:
is this as simple as doing a find/replace across the whole repo on `import gql from 'graphql-tag';` with `import { gql } from '@apollo/client';`?

redreceipt:
Maybe :) try it and then see if the app boots

nlewis84:
App booted, guess so! 👍

PR #2083, opened 2021-07-24
fix: replace graphql-tag with @apollo/client > gql

nlewis84:
closes https://github.com/ApollosProject/apollos-apps/issues/1901


nlewis84:
ran `yarn && yarn storybook` and app fired up with no errors.

redreceipt:
This is awesome. Go ahead and remove all the references from the package JSON files too and rerun yarn so it's completely gone from the repo.

nlewis84:
This should be a chore: and not a fix:, huh?

redreceipt:
I'd probably have done it as a "chore" but doesn't really matter. fixes show up in the changelog, chores don't. that's the difference. Also looks like you still have some tests failing.

nlewis84:
I'm not seeing what the problem is. It is with a file that doesn't interact with the gql changes. ui-kit>Card>CardWrapper. The test fail message isn't helping me much.

redreceipt:
Run `yarn test` locally inside the packages until you find the culprit. Try ui kit and ui connected first. But this is a big PR so it could be hiding anywhere. Unfortunately the CI isn't really helpful in finding the failure sometimes.

nlewis84:
Any idea how to change that test in `PrayerExperienceConnected.tests.js`? This passes the test, but I don't think that it should reference graphql-tag and still pass.

redreceipt:
Yea we shouldn't reference graphql tag anywhere. Not sure why we're mocking it out there. Try just importing gql in that file like normal from Apollo client

nlewis84:
I had to import gql with an alias prepended with `mock`, but that seemed to get it working. Jest did not want me to use out-of-scope variables.

redreceipt:
Weird! Why was gql out of scope there?

nlewis84:
I don't know, but that is probably why we were assigning `const gql = jest.requireActual('@apollo/client');` within the mock function.

redreceipt:
this would probably work

`const { gql } = jest.requireActual('@apollo/client');`

nlewis84:
I think that tried `const gql = jest.requireActual('@apollo/client');` and it didn't work. I'll try it your way. Would that be preferred to the mockGql alias?

redreceipt:
yea that way won't work, it's not the default export of `@apollo/client`

redreceipt:
yea I'd say leave it like it was if possible to avoid any other side effects. the preferred way is to understand why it needs to be mocked in the first place and fix that ;)

nlewis84:
That makes sense! This worked.

redreceipt (approved):
nice work!

PR #2084, opened 2021-07-26
chore: wrap CHOP call in try/catch

redreceipt:


redreceipt:
actually don't think we want this, we need to see the network error come through. this won't take down the server anyway

PR #2089, opened 2021-07-27
fix: correctly builds comment feature IDs

redreceipt:
not sure how they were ever working...


redreceipt:
not really a great way to test this either since core is using feature resolvers from postgres

conrad-vanl (approved):

redreceipt:
the ID used to live in `postgres/comments/resolver` now it lives in `postgres/feature/resolver` but all the apps are getting features from `rock/feature/resolver`

PR #2092, opened 2021-08-02
chore: deprecate content item byactive in favor of bycontentchannel

redreceipt:
This removes a redundant rarely used function and allows churches to override the `byContentChannel` function for custom search indexing


redreceipt:
@vinnyjth I think this is good but I cant test it, having trouble hooking up local redis to run the index in our symlinked setup, getting socket timeout errors when using the local and remote redis URLS

redreceipt:
just shows this error when trying to use Redis with `yarn app`. using the same URL from the templates repo works fine.

Screen Shot 2021-08-02 at 12 45 59 PM

vinnyjth:
@redreceipt I think local Redis was broken here: 
https://github.com/ApollosProject/apollos-apps/commit/f72ee33d837bb3b414189ecfe4ee0294b2054b72

packages/apollos-data-connector-algolia-search/src/data-source.js
```diff
@@ -119,27 +127,33 @@ query getItem {
...
-    const args = { after: null, first: 100 };
 
-    while (itemsLeft) {
-      const { edges } = await ContentItem.paginate({
-        cursor: ContentItem.byActive(),
-        args,
-      });
+    // loop through each active channel
+    await Promise.all(
+      ContentItem.activeChannelIds.map(async (channelId) => {
```
vinnyjth, line 133: Why not use `byContentChannelIds`?

redreceipt: that's a good question. I overwrote the single channel function on fellowship because content channel uses it. I tried to customize the multi channel one, but the way Rock REST requests have to work, I can't filter different channels in different ways AND the request be paginated with different channels. it's a mess. I agree overwriting the multi on core makes more sense but then that means I'll need to customize the search datasource as well to use the single just like this.

redreceipt: hopefully soon, they'll both be gone along with the algolia DS anyway

vinnyjth: As long as you have a reason ;) 

This code will less relevant very soon anyway :D

vinnyjth (approved):

PR #2094, opened 2021-08-02
fix: replaces missing function error on button features

redreceipt:


packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -252,7 +252,7 @@ export default class ContentItem extends RockApolloDataSource {
           id: attributeValues.buttonLink.id,
           // NOTE: right now only URLs are implemented
           // could optionally parse Rock IDs for OPEN_NODE action?
-          action: Feature.attachRelatedNodeId({
+          action: Feature.attachActionIds({
```
vinnyjth, line 255: As I understand it, I believe we'll want to handle this within the `createButtonFeature` function rather than inside the `ContentItem.dataSource`

vinnyjth: What you've got is correct, but maybe not ideal :) 

I'll approve this on the grounds that this was in the wrong place to begin with.

vinnyjth (approved):

redreceipt:
@vinnyjth snap is failing in the postgres??? pkg here but passes locally, any idea how that could happen?

vinnyjth:
@redreceipt Hm, maybe re-run? Not sure from here

PR #2098, opened 2021-08-03
fix: adds padding to the last comment item

nlewis84:
The new card styles padding is 24, so that seemed appropriate here.

Before:
Screen Shot 2021-08-03 at 2 33 51 PM
After:
Screen Shot 2021-08-03 at 2 34 15 PM


packages/apollos-ui-connected/src/CommentListFeatureConnected/CommentListFeature/index.js
```diff
@@ -17,6 +17,8 @@ const CommentListFeature = ({ comments, onPressLike, onPressActionMenu }) => (
       />
     )}
     data={comments}
+    // Adds a bit of padding to the last item of the list of comments
+    paddingBottom={comments.length ? 24 : 0}
```
redreceipt, line 21: use `theme.sizing.baseUnit` to fit better with our current pattern. but good find!

nlewis84: Gotcha!

redreceipt (approved):
💪

PR #2099, opened 2021-08-03
fix: added flex to Search views

nlewis84:
Added flex to containers and SafeAreaView so that the last item in the search modal doesn't get clipped off.

Before:
Screen Shot 2021-08-03 at 2 47 24 PM
After:
Screen Shot 2021-08-03 at 2 45 20 PM


redreceipt:
search doesn't seem to be working on the template, I'll test when content is back

redreceipt (approved):
woohoo!

PR #2102, opened 2021-08-04
fix: Adjust FollowSearch ScrollView to fit on screen

nlewis84:
Before:
https://user-images.githubusercontent.com/72768221/128203475-f1accaf3-f2e9-40cf-b3fc-d9ea3934f545.mp4

After: 
https://user-images.githubusercontent.com/72768221/128203514-9c1489d0-3414-4bce-8f62-0a998fa0d4a0.mp4


nlewis84:
When the FullHeightScrollView is set to height: 100% it forces the last row off the bottom of the screen because the StyledSearch above it takes up a percentage of the partial that is rendered. Since there are 9 visible rows, and the StyledSearch takes up roughly the same amount of vertical space as a row, the FullHeightScrollView should take up 90% of the space.

redreceipt:
I think maybe a better fix would be to wrap the scrollview in a paddedview with a padding bottom of baseunit *3 or something, that way if the size of the search bar changes later, bottom spacing will still be ok.

nlewis84:
I had to go up to baseUnit*17 before all the content would stay on the screen and nothing would be cut off 😮

redreceipt:
weird! ok I'm ok with this. I feel like there's a better _flex-y_ fix but we can improve later

redreceipt:
to fix the title, close and reopen the PR

nlewis84:
If I link this to the issue, does it automatically close the issue? Or do I need to go to the issue and just drop the link to this PR there and close the issue manually?

redreceipt (approved):

redreceipt:
link an issue and it will close it

nlewis84:
Is there an automated way to regen the failing snapshots, or do I go into the tests.js.snap file and adjust manually?

redreceipt:
yarn test -u will update snapshots

PR #2103, opened 2021-08-05
fix: added margin for top notch on videos

nlewis84:
Before:

After:
Screen Shot 2021-08-05 at 3 53 14 PM
This fix only effects video media content items. When playing a video in fullscreen mode, this margin is no longer used. The ScrollView below the video is also nudged down appropriately. The amount of margin is determined from React Native Safe Area Context.


conrad-vanl:
Now you can call yourself a typescript developer!

redreceipt (approved):
Good work!

PR #2106, opened 2021-08-06
fix: disappearing cancel search button

nlewis84:
Added logic to treat the cancel button the same as the clear search button (if there is text in the input field). If there is no text, it closes the modal.

Before:
https://user-images.githubusercontent.com/72768221/128533019-998d32b2-98b0-4b91-9ff1-1b26f32447ec.mp4

After:
https://user-images.githubusercontent.com/72768221/128533028-3fa8dea6-52f3-4c55-9c34-99659db6c0ad.mp4


redreceipt:
Why shouldn't the cancel button always close the modal no matter what? that's what it seems like its purpose should be

nlewis84:
Good point

redreceipt:
so now the only change is we're clearing out the text first? i don't understand why that's necessary?

nlewis84:
I'm not sure why, I just know that it works. I'll investigate and get back to you.

nlewis84:
Looks like it has to do with the state of isFocused (lines 55-80)

redreceipt:
that's the bit I feel like we should improve to fix this problem if possible :/ or else we're just pushing our problems around and leaving problematic code in there

nlewis84:
The whole bit about hiding and showing the cancel button and setting focus seems unnecessary.

redreceipt:
I'm always down for removing what's unnecessary. Easy way to fix A LOT of problems and future problems.

nlewis84:
Hmmm okay, I'm spinning my wheels on this. Should I close this since the work will be much different and just open a new PR when I've got something figured out, or just leave it open?

redreceipt:
Up to you. I just don’t feel great approving a seemingly redundant addition that mysteriously fixes problems.

PR #2107, opened 2021-08-10
fix: change updateSeriesProgress to use content channel ids

tomthornton:


packages/apollos-data-connector-postgres/src/content-items/dataSource.js
```diff
@@ -238,7 +238,7 @@ class ContentItemDataSource extends PostgresDataSource {
     return childItems[firstInteractedIndex - 1];
   }
 
-  async getSeriesWithUserProgress() {
+  async getSeriesWithUserProgress({ channelIds = [] } = {}) {
```
vinnyjth, line 241: I think you're really close here. If I run this `algorithm` without passing a channelId, I get 0 items back (I would expect to get items from all categories back). 

Looking at the SQL query
```
Executing (default): SELECT "contentItem"."title", "contentItem"."summary", "contentItem"."html_content" AS "htmlContent", "contentItem"."publish_at" AS "publishAt", "contentItem"."active", "contentItem"."id", "contentItem"."apollos_id" AS "apollosId", "contentItem"."apollos_type" AS "apollosType", "contentItem"."origin_id" AS "originId", "contentItem"."origin_type" AS "originType", "contentItem"."created_at" AS "createdAt", "contentItem"."updated_at" AS "updatedAt", "contentItem"."cover_image_id" AS "coverImageId", "contentItem"."parent_id" AS "parentId", "contentItem"."content_item_category_id" AS "contentItemCategoryId", "coverImage"."type" AS "coverImage.type", "coverImage"."url" AS "coverImage.url", "coverImage"."node_id" AS "coverImage.nodeId", "coverImage"."node_type" AS "coverImage.nodeType", "coverImage"."id" AS "coverImage.id", "coverImage"."apollos_id" AS "coverImage.apollosId", "coverImage"."apollos_type" AS "coverImage.apollosType", "coverImage"."origin_id" AS "coverImage.originId", "coverImage"."origin_type" AS "coverImage.originType", "coverImage"."created_at" AS "coverImage.createdAt", "coverImage"."updated_at" AS "coverImage.updatedAt" FROM "content_item" AS "contentItem" LEFT OUTER JOIN "media" AS "coverImage" ON "contentItem"."cover_image_id" = "coverImage"."id" WHERE "contentItem"."active" = true AND "contentItem"."content_item_category_id" IN (NULL) LIMIT 1;
```

that `"contentItem"."content_item_category_id" IN (NULL)` is always going to return 0 items. maybe we could exclude the filter if the `channelIds` array is empty? 

Also, let's name this argument `categoryId` to avoid confusion :)

tomthornton: Fixed!

vinnyjth:
Looks good! But quick question before I approve, do you think we need to maintain the ability to pass multiple `categoryIds`? I know I said `categoryId` in my comment, which was a mistake, I should have been more clear that I was referring to `category` vs `channel` instead of singular vs plural.

tomthornton:
@vinnyjth sure! that's pretty easy. Just updated

PR #2109, opened 2021-08-10
fix: Added Order to the contentItems

camrun91:
Added order to the Sequelize query for getFromCategoryIds. This makes the most recent items return from Postgres. May need to add the order option to the other queries.

### DIFF
--- a/packages/apollos-data-connector-postgres/src/content-items/dataSource.js
+++ b/packages/apollos-data-connector-postgres/src/content-items/dataSource.js
@@ -173,6 +173,7 @@ class ContentItemDataSource extends PostgresDataSource {
         contentItemCategoryId: { [Op.in]: ids },
         ...args?.where,
       },
+      order: [['publish_at', 'DESC']],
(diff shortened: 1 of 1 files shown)


redreceipt:
I don't have a problem with this PR, although I find it odd that we haven't had anyone complain about the content being out of order. That also may be because no one in production is pulling by category. I don't even know what GQL query that ties to.

camrun91:
@redreceipt this is what is being called by the `contentFeedAlgorithm` in the Postgres data source.  it is here if it helps you track it down. https://github.com/ApollosProject/apollos-apps/blob/7750dc1fd93b3241c2b332b9b4dcfcdd3f757900/packages/apollos-data-connector-postgres/src/action-algorithms/index.js#L220

redreceipt:
yep that one's important 😂 good fix. thank you! i'm gonna close and reopen so the title checker passes

redreceipt (approved):

PR #2113, opened 2021-08-11
chore: moves last canary build

redreceipt:
so this doesn't happen

Screen Shot 2021-08-11 at 3 16 51 PM


redreceipt:
First doesn’t work because it messes up the first commit after a release (that screenshot)

vinnyjth:
@redreceipt It worked prior to this commit I think? 

https://github.com/ApollosProject/apollos-apps/commit/b0749402d272398dfb2475a4a6b0905f4b96ba73

redreceipt:
I know we’ve been back and forth but I don’t remember ever trying to use canary.0 before this. But you may be right. I just feel like this will fix what happened yesterday.

vinnyjth:
The problem from what I can see is that the `canary.0` version that's being released as a part of the `yarn release:canary` doesn't create a `git commit`, so the first commit post release fails to account for it when counting commits. This release will fix that, but it will accomplish that by failing to create a canary.0 release for all packages which will cause problems when running `yarn latest` in church apps.

I can't remember what was wrong with our old strategy of creating an initial commit for each version (which intentionally created a canary release) should we try going back to that? Or even pushing an initial commit after creating that `canary.0` release here?

redreceipt:
Yea maybe we just try to go back to that. I don't remember the problem with it either. This definitely a devops smell though. I'm trying to think of a simpler strategy. I wish canary could always point to either latest or canary depending on which is more recent.

PR #2116, opened 2021-08-12
fix: fixes next up series algorithm

redreceipt:


tomthornton (approved):
Forgot to go back and change the object name. Thanks!

PR #2119, opened 2021-08-13
fix: Made postgres child/sibling item sorting consistent

vinnyjth:


packages/apollos-data-connector-postgres/src/content-items/dataSource.js
```diff
@@ -110,13 +110,22 @@ class ContentItemDataSource extends PostgresDataSource {
 
   // A simple alias at this point.
   async getChildren(model, queryArgs = {}) {
-    return model.getChildren(queryArgs);
+    return model.getChildren({
+      order: [
+        [Sequelize.literal('"contentItemsConnection".order'), 'ASC'],
+        ['publishAt', 'ASC'],
```
conrad-vanl, line 116: is always sorting in `ASC` order going to be right?

conrad-vanl (approved):

packages/apollos-data-connector-postgres/src/content-items/dataSource.js
vinnyjth: The more I think about this, the more I think I should flip this. 

`order` will be correct for things like `devos` and the publishAt fallback should likely be most recent first.

PR #2120, opened 2021-08-13
fix: Revert "chore: safety check on media urls"

vinnyjth:
Reverts ApollosProject/apollos-apps#2117


redreceipt:
i fixed it #2121

PR #2123, opened 2021-08-13
fix: order children by publish at asc

vinnyjth:


conrad-vanl (approved):

packages/apollos-data-connector-postgres/src/content-items/resolver.js
```diff
@@ -21,7 +22,13 @@ export const defaultContentItemResolvers = {
   siblingContentItemsConnection: async (model, args, { dataSources }) =>
```
conrad-vanl, line 22: pr title says children, but this is sibling?

conrad-vanl: do we need to add this to the `childContentItemsConnction` too?

vinnyjth: oh - good catch. needs to be both

conrad-vanl (approved):

PR #2127, opened 2021-08-16
feat: Adds Expires at field

vinnyjth:
Merge https://github.com/ApollosProject/apollos-apps/pull/2126
and https://github.com/ApollosProject/apollos-shovel/pull/53/files first. 

It allows for null values, so it's fully safe to merge without the shovel actually pulling this field.


vinnyjth:
Now that RiverValley has the priority field being set appropriately, this is ready for review :)

packages/apollos-data-connector-postgres/src/content-items/model.js
```diff
@@ -22,10 +23,19 @@ const setupModel = configureModel(({ sequelize }) => {
   sequelize.models.contentItem.addScope('defaultScope', {
     include: [{ model: sequelize.models.media, as: 'coverImage' }],
     where: {
-      [Op.or]: [
+      [Op.and]: [
         // this will need to be an `and`. However, we need to tweak the shovel to not update this field based on publishAt first.
```
redreceipt, line 27: still need this comment?

vinnyjth: nope :)

vinnyjth, line 27:
```suggestion
```

redreceipt:
given, our recent habit of rushing things, I'd like to test this fully. add an item in rock active and expired, create a local db I can migrate, backfill, then check before and after. we have time for that?

vinnyjth:
Yes with a caveat

RiverValley is in a state where if items are Approved OR they are past their publish date, they will go alive. Seeing as their items don't require approval, this might cause issues (all soap items will be live). We could potentially do a PR that flips the `Op.or` to an `Op.and` without the `expiresAt` changes to remedy that.

Also, I wrote tests that mimic real world behavior in a very simplified way, hopefully that will help add confidence :)

redreceipt:
Good point, that means next time we run the backfill, everything is going live :| considering the alternative is worse and they’re the only ones…let’s test in the wild!

redreceipt (approved):

vinnyjth:
https://user-images.githubusercontent.com/1637694/129792599-84f2b8b1-e32d-49e8-80c3-bf644b997f1f.mp4

PR #2130, opened 2021-08-16
fix: android prayer onboarding crash

vinnyjth:


conrad-vanl (approved):
I think this makes a ton of sense...

packages/apollos-ui-prayer/src/screens/PrayerDialogScreen/index.js
```diff
@@ -42,10 +42,13 @@ const PrayerDialogScreen = ({
...
-    toValue: visible,
-    useNativeDriver: true,
-  }).start();
+
+  useEffect(() => {
+    Animated.spring(animation, {
+      toValue: visible ? 1 : 0,
+      useNativeDriver: true,
+    }).start();
+  }, [visible])
```
vinnyjth, line 51:
```suggestion
  }, [visible]);
```

PR #2131, opened 2021-08-17
chore: Add feature priority field for sorting

tomthornton:


vinnyjth:
@tomthornton Should we include the code that uses this field for sorting in this PR?

tomthornton:
@vinnyjth yeah we can do that. Where is this in the code?

vinnyjth:
https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-data-connector-postgres/src/content-items/dataSource.js#L103

tomthornton:
Sorted by ascending. I keep forgetting that `getFeatures` is a sequelize function and not one we created. That's where I was confused.

vinnyjth:
@tomthornton Looks good Tom! Could you add a test for the sorting? Maybe something that features created in an arbitrary order by a set priority will always sort a specific way? I know the changes are simple, but tests like this will help prevent unintentional changes moving forward.

tomthornton:
@vinnyjth added tests

vinnyjth (approved):
Looking great, thanks @tomthornton!

PR #2134, opened 2021-08-18
chore: name the ContentCard and HeroItem

redreceipt:


redreceipt:
actually this won't work...

redreceipt:
actually yes it will :) I'll drop in the `cardMapper` to Fellowship which should show `HighlightCard` on sermons

conrad-vanl (approved):

PR #2138, opened 2021-08-19
feat: Postgres Prayers

tomthornton:
In order to run, import `PrayerRequest` from `apollos-dataconnector-postgres` and remove the import from `apollos-dataconnector-rock` in the `index.postgres.js` file of your ApolloChurchApi


vinnyjth:
At a glance, looking great!

Remember tests, we’ll need them before we merge :) try to get 100% test coverage over the dataSource methods as a goal

tomthornton:
@vinnyjth tests added!

packages/apollos-data-connector-postgres/src/prayers/__tests__/dataSource.test.js
```diff
@@ -0,0 +1,120 @@
...
+
+    const prayer1 = await sequelize.models.prayerRequest.create({
+      text: 'Test Prayer',
+    });
+    await person2.addPrayerRequest(prayer1);
+    await prayerRequestDatasource.incrementPrayed(prayer1.id);
+    await prayer1.reload();
+
+    expect(prayer1.usersPrayed).toContain(currentPerson.id);
+  });
```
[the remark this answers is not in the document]
vinnyjth: Actually I take that back! We don't have flagging :)We probably should though! Since Rock supports it. We can add a field in a follow up PR.

packages/apollos-data-connector-rock/src/interactions/data-source.js
```diff
@@ -198,14 +198,19 @@ export default class Interactions extends RockApolloDataSource {
...
-      InteractionComponentId: interactionComponent.id,
-      Operation: action,
-      InteractionDateTime: new Date().toJSON(),
-      InteractionSummary: `${action}`,
-      ForeignKey: nodeId,
-    });
+
+    // Currently it throws an error for a postgres created prayer, this will be fixed once they are shoveled back
+    // into Rock or on command during the prayer creation
+    if (action !== 'PRAY') {
```
vinnyjth, line 204: We'll likely have to be a bit more careful here :/ if we have this `action` skipping the `interaction` all of the rock folks are going to see their `Prayer` interactions drop off. Maybe we could do something (even if it's hacky) to figure out if it's a Rock or Postgres ID and then skip the `post` then? or even wrap the `post` in a try/catch

packages/apollos-data-connector-postgres/src/prayers/dataSource.js
```diff
@@ -0,0 +1,115 @@
...
+      const followingB = isFollowingPrayerRequestor(prayerB);
+
+      // eslint-disable-next-line no-nested-ternary
+      return followingB > followingA ? 1 : followingB < followingA ? -1 : 0;
+    });
+
+    return sortedPrayerRequests;
+  }
+
+  async getPrayerById(prayerId) {
```
vinnyjth, line 74: You might be able to get away without this method by using the inherited `getFromId` method.

packages/apollos-data-connector-postgres/src/prayers/dataSource.js
```diff
@@ -0,0 +1,115 @@
...
+    return this.model.findOne({
+      where: {
+        id: prayerId,
+      },
+    });
+  }
+
+  async incrementPrayed(prayerId) {
```
vinnyjth, line 82: Is this working? I believe the function signature here (coming from the additional interactions) is an object of the form `{ id, __type, action }`

vinnyjth (commented):
Here's my initial round of feedback @tomthornton  :D Looking pretty good so far, I think the only major thing we need to figure out is how to make the `prayed_for` array not a JSONB field :) 

I'm going to test in the sim, and then I might have more feedback.

packages/apollos-data-connector-postgres/src/prayers/migrations/001_create_prayers.js
```diff
@@ -0,0 +1,55 @@
+import { Sequelize, DataTypes } from 'sequelize';
+
+async function up({ context: queryInterface }) {
+  await queryInterface.createTable('prayer', {
```
vinnyjth, line 4: What do you think about naming this table `prayer_request`?

packages/apollos-data-connector-postgres/src/prayers/dataSource.js
```diff
@@ -0,0 +1,115 @@
...
+      where: {
+        personId: currentPersonId,
+        text,
+      },
+    });
+
+    return newPrayer[0];
+  }
+
+  async byDailyPrayerFeed({ numberDaysSincePrayer = 3, limit = 10 }) {
```
vinnyjth, line 21: I believe this `method` takes an optional `personId` argument that, when provided, returns _only_ prayers from that person. We'll need to implement that functionality to support this "My Prayers" feature. (Right now it's showing the same content as the "Pray for others" feature. 

image

vinnyjth (changes requested):
I think that's all the feedback I have :D thanks for your hard work Tom! PR is looking really good

tomthornton:
@vinnyjth Thank you! I'll be working on those changes

packages/apollos-data-connector-rock/src/interactions/data-source.js
tomthornton: Yeah I figured this would be the case. I like the try/catch fix for now. Once they are shoveled back into rock it should still work.

packages/apollos-data-connector-postgres/src/prayers/migrations/001_create_prayers.js
tomthornton: Sure that makes sense. And the join table `user_prayer` or just `prayer`?

camrun91:
While you all are working on this @vinnyjth @tomthornton it would be awesome to have the ability to mark your own prayers as answered with an answer text Rock supports this currently. Also would be neat if you prayed for a prayer to get notified if it is marked answered. 

If I need to move this to discussions I can do that.

packages/apollos-data-connector-postgres/src/prayers/dataSource.js
tomthornton: Yep it's working. Line 18 of the interactions data source in Rock takes the id out and passes it to the incrementPrayer function.

vinnyjth: Ah, thanks I missed that. 👍

tomthornton:
@camrun91 yeah that sounds like a cool feature! We'll probably have to wait to implement that until after we've connected things back to rock.

tomthornton:
- Added `getCurrentPerson` to the People datasource... not sure if this is already accessible somewhere else
- Added joint table `person_prayed_for` and some foreign key constraints to help with deleting
- Updated tests

vinnyjth:
@tomthornton I think there's a few more review comments that might require changes. Do you want to take a look at those before I re-review?

tomthornton:
@vinnyjth ready for review now. I added the `personId` argument to `dailyPrayerFeed` and wrapped the interaction rock call with a try/catch.

packages/apollos-data-connector-postgres/src/prayers/dataSource.js
```diff
@@ -0,0 +1,120 @@
...
+
+      // eslint-disable-next-line no-nested-ternary
+      return followingB > followingA ? 1 : followingB < followingA ? -1 : 0;
+    });
+
+    return sortedPrayerRequests;
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  async isPrayedFor({ prayerRequest, currentPersonId }) {
```
vinnyjth, line 79: Have you considered using an `includes: [...]` to eager load this relationship? Might help reduce the number of queries significantly in a longer list of prayers :)

vinnyjth: If you add it as a default scope (see `content_items` with cover images`) you can know that the relationship will always be present

tomthornton: Yep I was thinking the same thing, didn't know you could do that. I'll try it out.

packages/apollos-data-connector-postgres/src/people/dataSource.js
```diff
@@ -141,7 +141,7 @@ export default class Person extends PostgresDataSource {
     return where;
   };
 
-  async getCurrentPersonId() {
+  async getCurrentPerson() {
     if (this.context.currentPostgresPerson) {
       return this.context.currentPostgresPerson.id;
```
vinnyjth, line 146: Returning `id` here probably will result in a bug :)

packages/apollos-data-connector-postgres/src/prayers/dataSource.js
```diff
@@ -0,0 +1,121 @@
...
+    return prayerRequest
+      .getPrayedUsers({
```
vinnyjth, line 82: @tomthornton I think calling `.getPrayedUsers` here will re-run the query, rather than using the eager-loading that you added earlier with the `defaultScope`, Is that intentional?

packages/apollos-data-connector-postgres/src/prayers/migrations/001_create_prayers.js
```diff
@@ -0,0 +1,55 @@
...
+      type: DataTypes.TEXT,
+    },
+    person_id: {
+      type: Sequelize.UUID,
+      references: {
+        model: 'people',
+        key: 'id',
+      },
+    },
+    users_prayed: {
```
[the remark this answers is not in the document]
vinnyjth: @tomthornton You can delete `users_prayed` from the migration, I think :) if we had deployed this code, we would have to do another migration, but in this instance it's totally safe to simply remove it from the initial migration.

packages/apollos-data-connector-postgres/src/prayers/dataSource.js
vinnyjth: Actually I can't prove this! So I think we're good.

vinnyjth: 😅 I realized I can't prove it, because this code isn't getting called anymore. You should be safe to remove it

vinnyjth:
@tomthornton 

Just three things to resolve and then we can merge this! Works well in the sim :) 

1. Dead code in the initial migration
2. Dead code in the dataSource `isPrayedFor` method
3. Tiny bug in `getCurrentPerson` logic

tomthornton:
@vinnyjth what is the dead code in the initial migration?

vinnyjth (approved):

PR #2140, opened 2021-08-22
chore: remove notifications local resolvers

redreceipt:
nice little refactor that simplifies and removes local resolvers for notifications package


packages/apollos-ui-notifications/src/Provider.js
```diff
@@ -51,10 +46,19 @@ class NotificationsInit extends Component {
...
-    client.writeQuery({ query: GET_PUSH_ID, data: defaults });
+    client.writeQuery({
+      query: GET_PUSH_ID,
+      data: {
+        pushId: null,
+      },
+    });
     client.onClearStore(() =>
-      client.writeQuery({ query: GET_PUSH_ID, data: defaults })
+      client.writeQuery({
```
vinnyjth, line 56: This might actually be a source of bugs? The `pushId` shouldn't change per-device, so we have no reason to clear it once we have it.

vinnyjth:
I dig this! I want to pull it down and test it before I approve, and I will try and make time for that soon :)

packages/apollos-ui-notifications/src/Provider.js
vinnyjth: @redreceipt  ^

redreceipt: I'm just doing here what we were already doing. if it's a bug, can we fix in another PR?

conrad-vanl:
@redreceipt I don't feel comfortable approving this, as I'm not familiar with this code. Really need @vinnyjth to give it the go ahead

vinnyjth (approved):

PR #2142, opened 2021-08-23
chore: removes need for oneSignalKey prop

redreceipt:
test with `BRANCH=onesignal yarn app` with and without the env variable set


vinnyjth:
Love this :D Shoudl we also take this as a good time to rename the variable to match the API? `ONE_SIGNAL_APP_ID`? we could look for both the old and new name for now, for backwards compat.

redreceipt:
ready

vinnyjth (approved):

PR #2143, opened 2021-08-23
chore: default content component loading states

redreceipt:
this is a simpler, more Apollos-y way to customize content component loading states. We also will no longer need the unconventional `TYPEMAP` loaded into ApollosConfig


vinnyjth:
I like the idea of this - our `TYPEMAP` approach lacks is kludgy - but I worry this is even harder to understand? The code itself is easier to understand, but there's more magic in the wrapper. Where does the string `parent` come from? What about `child`? If we can solve that somehow, maybe by mapping to component names rather than arbitrary strings I think we can rock and roll with this.

redreceipt:
@vinnyjth what about this? much less magical and way easier to follow

vinnyjth:
@redreceipt Sorry to continue to be nitpicky - I want to get this riight. 

I don't want to introduce any new category/naming concepts in this PR. I think your latest change brings us way closer, but now I have learn what `childContentTypes` and `parentContentTypes` are. Can we rewrite this in a way that 

option a - uses a structure we already have? Like `ParentNode` or `SiblingNode` (kinda how this was done oringally). 

option b - ties together the components and typenames directly. No abstractions, map component names to the typenames that should enable loading states.

redreceipt:
ok there you go @vinnyjth, option B, no abstractions, overridable

vinnyjth:
@redreceipt Love it.

vinnyjth (approved):

PR #2144, opened 2021-08-23
feat: Postgres interaction table

vinnyjth:
Adds an `interaction` table / data source. 

Best way to test is to clone the PR and run `BRANCH=interactions yarn api`, then boot the app and click around a few content items. You should see interactions start to fill up in the interaction table.


vinnyjth:
This should be ready for review, I am double checking everything now in my sim.

vinnyjth:
image

tomthornton:
@vinnyjth awesome! I'm currently only getting "COMPLETE" interactions when I view a content item. That's what you had in the screenshot so I'm assuming everything is working?

vinnyjth:
Yep! You'll only get the `SERIES_STARTED` when visiting a a child item of a parent.

tomthornton (approved):
Looks great! Nice work.

PR #2147, opened 2021-08-25
fix: adds openurl and on press handler to node features

redreceipt:


vinnyjth:
I don't understand the context for this PR. Shouldn't whatever needs to open a url use the `onPressActionItem` prop?

redreceipt:
it's unclear in the code because we prop spread all over the place, but the `onPressActionItem` prop requires the `openUrl` prop to also be there

vinnyjth:
Gotcha - so why is the change needed? What is this fixing or replacing.

redreceipt:
The Tab feature feed works but the Node feature feed doesn’t work, on FNWA, I have an action table feature on content

vinnyjth:
@redreceipt Gotcha! Thanks.

PR #2148, opened 2021-08-25
fix: ensures onesignal config error wont causes problems

redreceipt:
this is throwing a 400 bad request


packages/apollos-data-connector-onesignal/src/data-source.js
```diff
@@ -24,6 +24,7 @@ export default class OneSignal extends RESTDataSource {
     subtitle,
     ...args
   }) {
+    if (!ONE_SIGNAL?.APP_ID) return null;
```
vinnyjth, line 27: Should this be one of those areas where we fail harder rather than hiding our errors? As in - if we aren;t using `OneSignal`, shouldn't we opt out of the entire datasource? We've launched an app or two in the past with no `ONE_SIGNAL.APP_ID` accidentally and it's caused a headache tracking it down.

redreceipt: yes I agree, but it wasn't failing gracefully in this case, posting a journal was just spinning and never resolving. I think my "no config yaml" PR can help tremendously with this problems we've seen because it checks for the appropriate env variables on module import

vinnyjth: I guess I'm advocating that we don't want a graceful failing here :) We should bomb hard so we avoid launching without OneSignalID, or remove this datasource entirely.

redreceipt: I agree, what do you think about my "no config yaml" PR given this? I think it eliminates a lot of confusion and makes it the server harder to misuse

vinnyjth: This is a discussion for that PR, but I don't like how much configuration is hidden when we take out the `config.yml`. Even just the three code paths we have right now that are managed by config.yml-less configuration have been confusing to manage while developing, and I'm really worried about making it worse

vinnyjth (approved):

vinnyjth:
I'll approve this PR because I understand why it can be frustrating on integrations, but I think we both know we are also kicking the can down the road in regard to how to solve the plethora of configuration options we need to launch an app and how to avoid launching without missing values.

redreceipt:
the plethora of configuration options is why it makes sense to me to handle it on a per connection basis. Each one of fastlanes commands take their own ENV variables and it yells at you when you misuse one by not supplying one. It doesn't make you have a giant config for all possible commands. which is where we are. Granted it also has a `fastlane actions XXX` command that tells you what config params are. maybe that's the piece we need?

vinnyjth:
>  Granted it also has a fastlane actions XXX command that tells you what config params are. maybe that's the piece we need?

Yah! I can agree with that. 

The problem I see with the YML file is that it has too much info, and you can delete half of it. 

The problem I see with the `ENV` variables is we still have problems with values from the`.shared` file getting wiped accidentally, and there's no way to get a "checklist" of what values you are missing / still need to fill out.

vinnyjth:
Also I will add that I think `fastlane` is a little odd in how they do things :) I like systems like `sublime texts` config management better, where editing the config involves overriding values from a default config. So you consult the default config, then copy over the values you want to change.

redreceipt:
values getting wiped isn't a huge deal because they don't get wiped from heroku. And locally, I think it would be as simple as 
```
Error: Server wont start, missing ONE_SIGNAL_APP_ID
```

and maybe having an `apollos config` command that tells you all the options you need given the modules you import?

PR #2149, opened 2021-08-25
feat: Pull to refetch user avatar

vinnyjth:


vinnyjth:
@redreceipt This is tangentially related to LCBC - I noticed that pulling to refetch wasn't actually pulling in the new avatar/user name.

redreceipt (approved):
i noticed the same, this will likely have to change again when it's not in a feature feed but we can cross that bridge later

PR #2150, opened 2021-08-26
feat: user likes based on interaction table

vinnyjth:
Your best bet for reviewing this PR is to run the api with `BRANCH=likes yarn api` and then like some content in the app. Here are some queries / mutations you could run as well (you may need to swap in your own content item id.)

```
query node{
  node(id:"WeekendContentItem:10d2d2a6-f2ad-4edf-a781-570e76d6452b"){
    ... on LikableNode {
      isLiked
      likedCount
    }
  }
}

mutation likeNode {
  updateLikeNode(input: { nodeId:"WeekendContentItem:10d2d2a6-f2ad-4edf-a781-570e76d6452b", operation: Like }){
    id
  }
}


mutation unLIkeNode {
  updateLikeNode(input: { nodeId:"WeekendContentItem:10d2d2a6-f2ad-4edf-a781-570e76d6452b", operation: Unlike }){
    id
  }
}

query likedContent {
  likedContent {
    edges {
      node {
        id
        __typename
      }
    }
  }
}
```


packages/apollos-data-connector-postgres/src/likes/dataSource.js
```diff
@@ -0,0 +1,137 @@
...
+    await Interactions.createNodeInteraction({ nodeId, action: 'UNLIKE' });
+  }
+
+  async getLikesCountByNodeId({ nodeId }) {
+    const { id } = parseGlobalId(nodeId);
+
+    const result = await this.sequelize.models.interaction.findOne({
+      attributes: [
+        [
+          sequelize.literal(`SUM(
```
tomthornton, line 48: Interesting!

packages/apollos-data-connector-postgres/src/likes/dataSource.js
```diff
@@ -0,0 +1,137 @@
...
+
+    const likedItems = await this.sequelize.query(
+      `
+        SELECT * FROM content_item INNER JOIN
+          (SELECT
+            node_id,
+            node_type,
+            SUM(
+              CASE action
+                WHEN 'LIKE' THEN 1
```
tomthornton, line 115: Wonder if we can consolidate these lines here and above?

tomthornton (commented):
@vinnyjth Looks good! I like the `CASE` functionality. I want to figure out how we can maybe use `sequelize.fn()` to consolidate some of these lines. Maybe it's not worth it here since it's only a few extra lines. What do you think?

vinnyjth:
@tomthornton I don't like the idea of abstracting away the ` CASE` into a function (at least not yet) but I do think it makes sense to keep it all in one place. I'll move it to a class attribute we can reuse.

vinnyjth:
@tomthornton This is ready for review again :)

tomthornton (approved):
Sweet!

PR #2151, opened 2021-08-26
feat: User progress from Postgres

tomthornton:


tomthornton:
@vinnyjth This is ready for review again. Added a test and the big SQL query we did plus the filters

tomthornton:
@vinnyjth this is ready for review again!

packages/apollos-data-connector-postgres/src/content-items/dataSource.js
```diff
@@ -221,73 +236,118 @@ class ContentItemDataSource extends PostgresDataSource {
...
+        model: this.sequelize.models.contentItem,
+        mapToModel: true,
+      }
     );
 
-    return inProgressItems.map(({ model }) => model);
+    return inProgressItems;
   }
 
   async getPercentComplete(model) {
```
vinnyjth, line 345: I initially wondered if this method should be converted to pure SQL as well, but you have a really good approach here 👍

packages/apollos-data-connector-postgres/src/interactions/dataSource.js
```diff
@@ -97,7 +98,7 @@ export default class Interactions extends PostgresDataSource {
     });
   }
 
-  async getInteractionsForCurrentUser({ actions = [] } = {}) {
+  async getInteractionsForCurrentUser({ actions = [], include = [] } = {}) {
```
vinnyjth, line 101: rather than adding an additional arg, what do you think about adding a `queryArgs` argument similar to some our other methods that end up making `findAll` calls? Signature would become `getInteractionsForCurrentUser({ actions = []} = {}, queryArgs = {})` then you would spread `...queryArgs` inside the call to `findAll` :)

packages/apollos-data-connector-postgres/src/utils/removeDuplicates.js
```diff
@@ -0,0 +1,6 @@
+export default function removeDuplicates(array, basedOn) {
```
vinnyjth, line 1: Do you still need this function?

packages/apollos-data-connector-postgres/src/content-items/__tests__/dataSource.tests.js
```diff
@@ -493,4 +502,62 @@ describe('Apollos Postgres ContentItem DataSource', () => {
...
+      additional: false,
+    });
+
+    const seriesWithUserProgress = await ContentItem.getSeriesWithUserProgress({
+      categoryIds: [contentItemCategory.id],
+    });
+
+    expect(seriesWithUserProgress.length).toBe(1);
+    expect(seriesWithUserProgress[0].id).toEqual(seriesContentItem.id);
+  });
```
vinnyjth, line 562: What do you think about adding a test over `getPercentComplete`? And maybe setup the data in such a way that we can be sure situations like duplicate interactions wouldn't cause issues :)

packages/apollos-data-connector-postgres/src/content-items/dataSource.js
```diff
@@ -221,73 +236,118 @@ class ContentItemDataSource extends PostgresDataSource {
...
-      )
-    ).filter(
-      ({ percent, model }) => percent !== 100 && percent != null && model
+    const { Person } = this.context.dataSources;
+
+    const currentPersonId = await Person.getCurrentPersonId();
+
+    const inProgressItems = await this.sequelize.query(
+      `
+      SELECT content_item.*
```
vinnyjth, line 257: Nice work :D

vinnyjth (commented):
Some initial comments after looking over the code :) I'll run through the code in the app tomorrow and then wrap up.

packages/apollos-data-connector-postgres/src/interactions/dataSource.js
tomthornton: Yeah! I think that makes the most sense going forward

packages/apollos-data-connector-postgres/src/utils/removeDuplicates.js
tomthornton: Nope. I'll remove it.

packages/apollos-data-connector-postgres/src/content-items/__tests__/dataSource.tests.js
tomthornton: Sure! That should be pretty easy given the other tests.

vinnyjth (changes requested):
@tomthornton After testing in the app...

1. The "seriesInProgress" algorithm appears to be working perfectly. Nice work! 
2. The "You're all caught up" logic (driven by `upNext`) is working perfectly, again, nice work! 
3. The "Up Next" logic appears to be working through the items in a series backwards. Video to demo. Hopefully this is an easy fix, do you think we could write a test to capture this situation? It's been a pretty common one throughout the lifecycle of this feature.

packages/apollos-data-connector-postgres/src/interactions/dataSource.js
tomthornton: I think we stopped using this somewhere so I believe it's dead code. I'm going to delete it.

packages/apollos-data-connector-postgres/src/content-items/dataSource.js
```diff
@@ -221,73 +220,134 @@ class ContentItemDataSource extends PostgresDataSource {
...
-    const interactions = await Interactions.getInteractionsForCurrentUserAndNodes(
-      {
-        nodeIds: childItems.map(({ apollosId }) => apollosId),
-        actions: ['COMPLETE'],
-      }
-    );
-    const apollosIdsWithInteractions = interactions.map(
-      ({ foreignKey }) => foreignKey
-    );
+    const childItems = await this.sequelize.models.contentItem.findAll({
```
vinnyjth, line 225: Is there any reason you opted to switch from `getChildren`? I think this will be problematic in some cases since the sort order is undefined.

packages/apollos-data-connector-postgres/src/content-items/dataSource.js
```diff
@@ -298,28 +358,33 @@ class ContentItemDataSource extends PostgresDataSource {
       return null;
     }
 
-    const childItems = await model.getChildren();
+    const currentPersonId = await Person.getCurrentPersonId();
```
vinnyjth, line 361: It's a little odd that we call `Auth.getCurrentPerson()` above (and discard then ignore result) and then call it again here. We should problaly combine these calls, and use `await Person.getCurrentPersonId`. (which calls `await Auth.getCurrentPerson();`) under the hood

packages/apollos-data-connector-postgres/src/interactions/dataSource.js
```diff
@@ -111,6 +112,7 @@ export default class Interactions extends PostgresDataSource {
         personId: currentPersonId,
         ...(actions.length ? { action: actions } : {}),
```
vinnyjth, line 113: You'll want to make sure you spread the `...queryArgs` explicetly for the `{ where }` object, eg ` where { other stuff, ...queryArgs.where }`. Otherwise passing a where clause to `getInteractionsForCurrentUser` will override the entire `where` argument for `findAll`

packages/apollos-data-connector-postgres/src/content-items/dataSource.js
vinnyjth: I'll come up with a test case that demonstrates why we need the sort order :)

vinnyjth (changes requested):
Really close @tomthornton :D I mentioned that we probally should try to use `getChildren` or, at least something that provides a sort order. I'll see if I can come up with a test case to demo that.

packages/apollos-data-connector-postgres/src/content-items/dataSource.js
tomthornton: Yep I agree

packages/apollos-data-connector-postgres/src/content-items/dataSource.js
tomthornton: The reason I opted for to not use `getChildren` is because I thought it would be more accurate. `getChildren` goes through content-item-connections and I had thought that we just needed to pull from `parentId`. I had forgot they needed order. I'll just need to add in the content item connections to the tests.

vinnyjth: Sounds good 👍 You could even add `parentId` to the `queryArgs`  for `getChildren` and have the best of both worlds.

tomthornton:
@vinnyjth updated those tests and using `getChildren` now. Also fixed the weird Auth/Person thing. Are those tests covering enough or do we need to check for more things?

packages/apollos-data-connector-postgres/src/content-items/dataSource.js
```diff
@@ -212,84 +211,144 @@ class ContentItemDataSource extends PostgresDataSource {
...
   async getUpNext(model) {
-    const { Auth, Interactions } = this.context.dataSources;
+    const { Auth } = this.context.dataSources;
 
     // Safely exit if we don't have a current user.
+    let currentPersonId;
+
     try {
-      await Auth.getCurrentPerson();
+      currentPersonId = await Auth.getCurrentPerson().then(({ id }) => id);
```
vinnyjth, line 220: I believe that `getCurrentPerson` will return a rock Id. At this point. Also - why the `await` + `.then`?.

tomthornton: it was await to begin with. I guess I could just do `currentPerson.id`

vinnyjth:
@tomthornton Testing with the current data set, with an empty interactions table, I'm getting an "all caught up state" before visiting any of the child items. I'm also seeing this error in the console. 

image


```
Error: WHERE parameter "person_id" has invalid "undefined" value
    at PostgresQueryGenerator.whereItemQuery (/Users/vincent/differential-projects/apollos/apollos-apps/node_modules/sequelize/lib/dialects/abstract/query-generator.js:2258:13)
    at /Users/vincent/differential-projects/apollos/apollos-apps/node_modules/sequelize/lib/dialects/abstract/query-generator.js:2247:25
    at Array.forEach (<anonymous>)
    at PostgresQueryGenerator.whereItemsQuery (/Users/vincent/differential-projects/apollos/apollos-apps/node_modules/sequelize/lib/dialects/abstract/query-generator.js:2245:35)
    at PostgresQueryGenerator.generateJoin (/Users/vincent/differential-projects/apollos/apollos-apps/node_modules/sequelize/lib/dialects/abstract/query-generator.js:1711:24)
    at PostgresQueryGenerator.generateInclude (/Users/vincent/differential-projects/apollos/apollos-apps/node_modules/sequelize/lib/dialects/abstract/query-generator.js:1549:24)
    at PostgresQueryGenerator.selectQuery (/Users/vincent/differential-projects/apollos/apollos-apps/node_modules/sequelize/lib/dialects/abstract/query-generator.js:1183:34)
    at PostgresQueryInterface.select (/Users/vincent/differential-projects/apollos/apollos-apps/node_modules/sequelize/lib/dialects/abstract/query-interface.js:939:27)
    at Function.findAll (/Users/vincent/differential-projects/apollos/apollos-apps/node_modules/sequelize/lib/model.js:1753:47)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    ```

**Comment by tomthornton** [comment]
@vinnyjth Ah, my fault I had assumed Auth was returning the right ID. Switched it to just use Person. It's working for me now.

**Comment by vinnyjth** [comment]
@tomthornton Working for me also! 

Okay one more scenario to capture - if we write another test this would be a good use case. 

Given items 1, 2, 3, 4, 5
When I visit item 1, then item 3. When I hit check up next, 4 should be returned

Right now the logic is like this

Given items 1, 2, 3, 4, 5
When I visit item 1, then item 3. When I hit check up next, 2 is returned.

**Comment by tomthornton** [comment]
@vinnyjth okay maybe this is done finally lol instead of adding multiple tests I did it all in one. I like how it captures the flow of user interaction.

File: packages/apollos-data-connector-postgres/src/content-items/dataSource.js [hunk 10]
```diff
@@ -235,13 +235,26 @@ class ContentItemDataSource extends PostgresDataSource {
...
+    const lastItemInteractedIndex = childItems.findIndex((contentItem) => {
+      return contentItem.interactions.length > 0;
+    });
+
+    console.log({ lastItemInteractedIndex });
 
-    return childItemsWithoutInteractions.length > 0
-      ? childItemsWithoutInteractions[0]
-      : null;
+    // If no item in the series has been interacted with
```
vinnyjth, line 246: Nice! The comments are very helpful.

packages/apollos-data-connector-postgres/src/content-items/__tests__/dataSource.tests.js
```diff
@@ -76,7 +73,7 @@ describe('Apollos Postgres ContentItem DataSource', () => {
     await sequelize.drop({ cascade: true });
     currentPerson = null;
   });
-
+  /*
```
vinnyjth: we should probably comment this back in :)

tomthornton: I remembered right after. Should be good now.

vinnyjth (approved):
Nice work! Speedy and effective.

PR #2156, opened 2021-08-31
chore: remove old ThemeProvider from Providers

redreceipt:


vinnyjth:
Remind me how this impacts all the older components that still use `styled`/`withTheme`?

redreceipt:
Doesn't affect them since the new provider works with those HOCs, it just doesn't double up the context provider. The lack of snap changes maybe could give you confidence?

vinnyjth (approved):

PR #2159, opened 2021-08-31
feat: Add tvFeedFeatures query

rkdavidson:
# About

The Apollos+ project needs a place to query feature feed data from, separate from existing concerns like `tabFeedFeatures`.
This PR creates a parallel mechanism to `tabFeedFeatures` called **`tvFeedFeatures`**.
It queries feature feed data from an expected root level `TV` config key, similar to `TABS`.

Simplified `config.yml` example:
```yml
TABS:
  HOME:
  WATCH:
  READ:
  CONNECT:

TV:
  HOME:
  SERIES:
  SPEAKERS:
  KIDS:
  MY_LIST:
```

Example query:
```graphql
query {
  tvFeedFeatures(tvFeed: HOME) {
    id
    features {
      id
      order
      __typename
    }
  }
}
```

This is a simple proposal that assumes we're happy with the design of `tabFeedFeatures` overall.
Open to feedback on naming, semantics, etc!

This PR also fixes a handful of spelling mistakes in a `README`.

# Testing
A new unit test was added to verify `FeatureFeed.getFeed` is able to process the `TV` section in `config.yml`.
I ran into problems doing more advanced/robust unit testing, but happy to flesh out more scenarios if I can get a second hand from the team.


redreceipt:
you want this to go into `master` right?

rkdavidson:
@redreceipt Yeah wasn't sure whether to do `master` or `next`, haven't heard back on Slack which was better. Switched it to `mater` 🤞 .

FWIW locally when I ran the tests, I was seeing some failures right out of the gate with no changes.
I'm not sure testing is stable/working at the moment?

packages/apollos-data-schema/src/index.js
```diff
@@ -1049,6 +1049,7 @@ export const featuresSchema = gql`
   extend type Query {
     tabFeedFeatures(tab: Tab!, campusId: ID): FeatureFeed
       @cacheControl(scope: PRIVATE)
+    tvFeedFeatures(tvFeed: TvFeed!): FeatureFeed @cacheControl(scope: PRIVATE)
```
redreceipt, line 1052: can we use a better name than `tvFeed` to match `tab`? maybe it's `tab` too? or `section`? And call the enum the same thing

packages/apollos-data-connector-rock/src/feature-feeds/index.js
```diff
@@ -56,6 +61,11 @@ class FeatureFeed extends RockApolloDataSource {
         Feature.getFeatures(ApollosConfig.TABS[args.tab] || [], args);
     }
 
+    if (type === 'tv') {
+      getFeatures = () =>
+        Feature.getFeatures(ApollosConfig.TV_FEEDS[args.tvFeed] || [], args);
```
redreceipt, line 66:
```suggestion
        Feature.getFeatures(ApollosConfig.TV[args.tvFeed] || [], args);
```

conrad-vanl:
@redreceipt what are your plans for how we refactor the tab queries? Don't we want to move away from giving them explicit names in core?

redreceipt:
yea I think long term the schema could be as simple as `featureFeeds: [FeatureFeed]` and dynamically pulls all from a `FEEDS` section in the config. and for individuals would be `getFeed(feed: TV_HOME): FeatureFeed` But don't think we need to wait on that reality for this PR.

packages/apollos-data-connector-rock/src/feature-feeds/index.js
rkdavidson: Thanks, good catch

packages/apollos-data-schema/src/index.js
rkdavidson: Yeah, that bugged me too.

I'm ok with `tab` and reusing/extending the same `Tab` enum, unless you think that'd get confusing?
Should it be `tvTab: TvTab!` and `enum TvTab` to clearly delineate them?

rkdavidson:
@redreceipt and @conrad-vanl , interesting point about a generic `FEATURE_FEEDS` convention.
That thought crossed my mind developing. Felt arbitrarily limited to keep things "TV-ey".

Is there wider scope involved with the generic version?
Should I just drop the `tv` naming and we sketch that approach out in this PR?

I'll leave it up to you guys if it's worth the pivot. Just hope this PR doesn't add cruft or clutter to core.

packages/apollos-data-schema/src/index.js
```diff
@@ -1049,6 +1049,7 @@ export const featuresSchema = gql`
   extend type Query {
     tabFeedFeatures(tab: Tab!, campusId: ID): FeatureFeed
       @cacheControl(scope: PRIVATE)
+    tvFeedFeatures(tvTab: TvTab!): FeatureFeed @cacheControl(scope: PRIVATE)
```
rkdavidson, line 1052: How does this sound/feel? `TvTab` as a parallel to `Tab`?
We could use `tab: Tab!` and share the `enum Tab` just like `tabFeedFeatures`, but I thought doing `TV_HOME` stuff may get awkward.

If we go much further down the route of sharing, we should almost just skip this whole PR and for immediate Apollos+ needs, use `tabFeedFeatures` with `TV_HOME`, `TV_SPEAKERS`, etc.
Which doesn't sound bad, actually...

rkdavidson: So the changes will be isolated to Chase Oaks extending the `enum Tab` schema to give us more options like `TV_HOME`, `TV_SPEAKERS`, etc (as needed)?

Sounds like it's safe to close this PR against core, then.

rkdavidson:
Closing this PR after further discussion with @conrad-vanl .

Our interim plan is to extend Chase Oaks' API `enum Tab` as needed to give us new feature feed options separate from the mobile app, without adding/changing existing queries.

This'll leave room for the new and improved mechanisms on the roadmap.

Thanks for the help on this @redreceipt / @vinnyjth / @conrad-vanl 🙇‍♂️ .

PR #2163, opened 2021-09-01
Fix: fixed a bug in the new user avatar screen

nlewis84:
Currently, if you click on a content item in the liked content feed on the Profile modal it will load on the screen behind the modal and be inaccessible without swiping down (which is not possible on android devices). This solves that while maintaining current functionality on the Connect Tab for current apps.

https://user-images.githubusercontent.com/72768221/131683542-e6fd2995-ff24-40de-96c3-0be589378dc7.mp4


packages/apollos-ui-connected/src/LikedContentFeedConnected/LikedContentFeedConnected.js
```diff
@@ -39,11 +39,16 @@ class LikedContentFeedConnected extends PureComponent {
   /** Function that is called when a card in the feed is pressed.
    * Takes the user to the ContentSingle
    */
-  handleOnPress = (item) =>
+  handleOnPress = (item) => {
+    // These prevent a bug that causes the Content Item to load behind the modal and be inaccessible
+    this.props.navigation.goBack();
+    this.props.navigation.goBack();
+    // This actually navigates to the content
     this.props.navigation.navigate('ContentSingle', {
```
conrad-vanl, line 47: @nlewis84 did you experiment at all with where this "ContentSingle" route is structured in the template repo? Having these double `goBack()` calls in this template feels like a major code smell to me. I see a couple of more sensible options:

1. See if changing the `stackPresentation` for `ContentSingle` from "push" to `fullScreenModal` fixes this, [this line of code](https://github.com/ApollosProject/apollos-templates/blob/867bdaf9e528376a71cb713dad593c13746d63ae/apolloschurchapp/src/index.js#L107)
2. Duplicating the `ContentSingle` route into the `UserSettings` Navigator, [here](https://github.com/ApollosProject/apollos-templates/blob/avatar/apolloschurchapp/src/user-settings/index.js#L35)

I'd prefer option 1 - there should be a way we can tweak that navigator settings so that this new route "opens on top" of the current route, instead of behind it. All of the `stackPresentation: 'push',` should probably become `stackPresentation: 'fullScreenModal',` TBH. It might be an issue that we're mixing modals with pushes in this navigator.

conrad-vanl (changes requested):

packages/apollos-ui-connected/src/LikedContentFeedConnected/LikedContentFeedConnected.js
nlewis84: I had not tried that. Option 1 worked. I can close this and I'll test out changing all the 'push' to 'fullScreenModal'.

conrad-vanl: Great! Make sure to test the journaling UI with the pullover journaling thingy - we used to use plain "Modal" with all of these routes before and had to switch to "Push" to fix that journal editor popover. But I think "fullscreenModal" should work too. Although, it might require another code change in one other spot.....lmk and I can point you to that

conrad-vanl: Now that my memory is jogging, most likely the bottom-sheet journal editor prompt component will be hidden now.

Most likely you'll have to:

```
import { BottomSheetModalProvider } from '@gorhom/bottom-sheet';
```

and render this component inside of `ContentSingleConnected`

conrad-vanl: It'll have to wrap it like `<BottomSheetModalProvider><ContentSingleConnected/></BottomSheetModalProvider>`


"Modals" do something fancy and on the NATIVE thread, the modal is rendered in a separate layer on the native side. So everything else shows up below it....so we just have to move that bottom sheet to be inside of that modal.

conrad-vanl: You could handle that inside of https://github.com/ApollosProject/apollos-templates/blob/master/apolloschurchapp/src/index.js like we do with `ProtectedRouteWithSplashScreen` by creating a `ContentSingleConnectedWithBottomSheet`, or by adding that `BottomSheetProvider` to `ContentSingleConnected` inside of this -apps repo. Your call.

nlewis84: Testing journal in the template is going to be a challenge with all the errors I'm seeing. Looks that way pointing at local and prod. Is that normal and I just need to checkout the changes in a real project?

conrad-vanl: uhhh no I don't think that is normal,....move this to #apollos on slack and see if we can get someone else to help

nlewis84: I could not figure out where to get that BottomModal working. I did find out that containedModal doesn't affect the journal modal though.

PR #2165, opened 2021-09-01
fix: removed UserAvatar and Recently Liked from Connect Tab

nlewis84:
This is to prepare for User Settings moved to an avatar in the header of all tabs.


nlewis84:
Linking here for reference.

https://github.com/ApollosProject/apollos-templates/pull/365

PR #2169, opened 2021-09-07
chore: Update postgres feature unique index to include priority

tomthornton:


packages/apollos-data-connector-postgres/src/features/migrations/002_unique_index_for_features.js
```diff
@@ -5,7 +5,7 @@ async function up({ context: queryInterface }) {
 }
 
 async function down({ context: queryInterface }) {
-  await queryInterface.dropIndex('features_parent_id_type_data');
+  await queryInterface.removeIndex('features_parent_id_type_data');
```
vinnyjth, line 8: Good catch! Thanks for fixing these calls.

packages/apollos-data-connector-postgres/src/features/migrations/005_update_features_unique_index.js
```diff
@@ -0,0 +1,21 @@
+async function up({ context: queryInterface }) {
+  await queryInterface.removeIndex('feature', 'features_parent_id_type_data');
+  await queryInterface.addIndex(
+    'feature',
+    ['parent_id', 'type', 'data', 'priority'],
```
vinnyjth, line 5: What do you think about moving `priority` before `data` and `type`? That way it can be used for sorting when filtering by `parent_id`
```suggestion
    ['parent_id', 'priority', 'type', 'data'],
```

tomthornton: Yeah sure. I didn't know that it worked that way but that sounds good.

vinnyjth: This might help explain the effect/impact https://blog.timescale.com/blog/use-composite-indexes-to-speed-up-time-series-queries-sql-8ca2df6b3aaa/

vinnyjth (approved):

PR #2170, opened 2021-09-08
fix: Confirm users when creating user logins

vinnyjth:
When creating a new user login, set the `IsConfirmed` flag to true. 

I would like to also set the `IsConfirmed` flag to true when logging in an existing user, but we end up in a bit of a catch-22. We can't check a user's credentials  and set `IsConfirmed` without trying to log them in, and we can't log them in without setting is confirmed to true. We could, of course, not verify a user's credentials, but that opens the door to letting anyone confirm any user's account via attempting to login as them with invalid credentials. Is that a security hole? Probably.


vinnyjth:
Don't merge this, as I need to actually test it :)

conrad-vanl:
@vinnyjth what's the game plan on this today? Do we know if we can easily bump all partner churches up to the latest version of data-connector-rock after this gets merged, or will we need to do any patch releases?

vinnyjth:
@conrad-vanl @redreceipt will know better than I do if we need a patch release or if we can bump everyone to the latest `connector-rock` package

vinnyjth:
I'm glad I tested, this is ready for review now.

redreceipt (approved):

conrad-vanl:
Can we get this deployed on newspring and lcbc?

redreceipt:
working on it

PR #2171, opened 2021-09-09
chore: Update rock and postgres prayers for use 

tomthornton:
Changes around a few things so that it works well in `rockWithPostgres` in templates.

This should be merged before the templates PR.

The draft PR for templates is here [https://github.com/ApollosProject/apollos-templates/pull/477](url)


packages/apollos-data-connector-postgres/src/prayers/dataSource.js
```diff
@@ -6,12 +6,14 @@ import { PostgresDataSource } from '../postgres';
 class Prayer extends PostgresDataSource {
   modelName = 'prayerRequest';
 
-  async addPrayer({ text }) {
+  async addPrayer({ text, originId }) {
```
vinnyjth, line 9: Rather than focusing on making this rock specific, what do you think about allowing users to spread in additional args? ala `{  text, ...args }`. That would give us the tools we need while also opening us up for more easier overrides in the future :)

packages/apollos-data-connector-postgres/src/prayers/model.js
```diff
@@ -6,6 +6,8 @@ const createModel = defineModel({
   resolveType: () => 'PrayerRequest',
   attributes: {
     text: DataTypes.TEXT,
+    originId: DataTypes.TEXT,
```
vinnyjth, line 9: Nice!

packages/apollos-data-connector-postgres/src/prayers/resolver.js
```diff
@@ -14,9 +14,9 @@ export default {
       args,
       { currentPostgresPerson: { id: currentPersonId } }
     ) => {
-      return (
```
vinnyjth, line 17: This seems like a suspect to me, do we know why `prayedUsers` would ever be a non-existent (or rather: not an empty array)

packages/apollos-data-connector-rock/src/prayer-request/resolver.js
```diff
@@ -9,8 +9,8 @@ export default {
     id: ({ id }, args, context, { parentType }) =>
       createGlobalId(id, parentType.name),
     isAnonymous: ({ isPublic }) => !isPublic,
-    requestor: ({ requestedByPersonAliasId }, args, { dataSources }) =>
-      dataSources.PrayerRequest.getRequestor({ requestedByPersonAliasId }),
+    requestor: (root, args, { dataSources }) =>
```
vinnyjth, line 12: Can you explain the need for the change here?

tomthornton: When this is run with the extended `PostgresPrayerRequest` data-source, it returns a Postgres `PrayerRequest` instead of a Rock `PrayerRequest`. This allows it so the overwritten `getRequestor` function in `rockWithPostgres` works.

packages/apollos-data-connector-postgres/src/prayers/resolver.js
tomthornton: `prayedUsers` wasn't being returned from the created prayer request because it's association wasn't included. If it's just now created, it's not going to have any users that prayed for it.

packages/apollos-data-connector-postgres/src/prayers/dataSource.js
tomthornton: Yep that makes sense :)

tomthornton: Updated! Also updated templates as well.

packages/apollos-data-connector-rock/src/prayer-request/resolver.js
vinnyjth: Ah okay - so I think there's two things going on here 

1. The Postgres resolvers are "falling through" to the rock resolvers. To avoid this, we like to do a trick in the data connector import in the `template` where we define a dataSource like `RockPrayers: { dataSource: RockPrayers.dataSource }`. That omits the resolvers and schema when importing so we don't end up with accidental fall throughs. That would be a change in `templates`. 
2. We are likely missing the actual resolver for `requestor` in the Postgres PrayerRequest resolver. We should probably add that so `requestor` starts working (I'm assuming it never worked in the first place). That would be a change here in `apps`.

packages/apollos-data-connector-postgres/src/prayers/resolver.js
vinnyjth: 👍

packages/apollos-data-connector-rock/src/prayer-request/resolver.js
tomthornton: For 2. We didn't need a resolver for `requestor` in Postgres Prayer Requests because Sequelize returns it. I'll go back over to templates and see if I can do #1

vinnyjth: That's good to know with 2! If that's working as intended, 1 is still necessary since the lack of resolver will cause a fallback to the Rock resolvers.

tomthornton:
@vinnyjth I've updated both templates and apps PRs. Added resolvers for the prayer request for postgres. I think this means we don't need the extended datasource in templates now?

packages/apollos-data-connector-postgres/src/prayers/resolver.js
```diff
@@ -9,11 +9,19 @@ export default {
     id: ({ id }, args, context, { parentType }) =>
       createGlobalId(id, parentType.name),
     isAnonymous: ({ isPublic }) => !isPublic,
+    requestor: async (root) => {
```
vinnyjth, line 12: We should probaly clean this up before merging :)

packages/apollos-data-connector-postgres/src/prayers/resolver.js
```diff
@@ -9,11 +9,19 @@ export default {
...
+      console.log({ root });
+      const requestor = await root.getRequestor();
+      console.log({ requestor });
+
+      return requestor;
+    },
     isPrayed: async (
```
vinnyjth, line 19: Any reason you switched to `root.getPrayedUsers`?

vinnyjth:
Looks good overall @tomthornton! Once we resolve the last bits of loose ends (I had a question around a resolver, then another resolver needed a bit of cleanup) We can test and merge this.

packages/apollos-data-connector-postgres/src/prayers/resolver.js
tomthornton: This is so we can cover all of our bases. I couldn't get `prayedUsers` to return as an association when creating it. I tried `include` but it didn't work (even though it showed up fine in the options).

Basically I added these resolvers so that no matter what prayerRequest is being resolved, we can get the correct `requestor` and `prayedUsers` instead of worrying about always returning them in the data source.

packages/apollos-data-connector-postgres/src/prayers/resolver.js
tomthornton: Thought I had removed it. Oops. Fixed it

packages/apollos-data-connector-postgres/src/prayers/resolver.js
vinnyjth: Do you think there's a way we could use the provided `prayedUsers` if it exists, and then sub in a call to `getPrayedUsers` if it doesn't? That way we could avoid a likely N+1 query.

tomthornton: Yep! just updated

packages/apollos-data-connector-postgres/src/prayers/dataSource.js
```diff
@@ -6,12 +6,14 @@ import { PostgresDataSource } from '../postgres';
...
     const currentPersonId = await this.context.dataSources.Person.getCurrentPersonId();
     const newPrayer = await this.model.findOrCreate({
       where: {
         personId: currentPersonId,
         text,
+        originId,
+        originType: originId ? 'rock' : null,
```
vinnyjth: @tomthornton Did you back out a change here? At one point you were spreading args `...args`, but now it looks like you're back to using `originId`?

vinnyjth: Agh my bad! I must have been looking at a specific commit by mistake.

vinnyjth (approved):

packages/apollos-data-connector-postgres/src/prayers/dataSource.js
tomthornton: @vinnyjth I think this is just outdated.

PR #2172, opened 2021-09-10
chore: opt-in to automerge

redreceipt:
seems that churches don't mind the risk so I figured this is better than customizing


packages/apollos-data-connector-rock/src/people/data-source.js
```diff
@@ -72,6 +72,8 @@ export default class Person extends RockApolloDataSource {
     const id = await this.post('/People', {
       Gender: 0, // required by Rock. Listed first so it can be overridden.
       IsSystem: false, // required by rock
+      // in case anyone wants to DANGEROUSLY turn on auto-merge
+      ...(process.env.AUTOMERGE === 'true' ? rockUpdateFields : {}),
     });
     await this.patch(`/People/${id}`, {
```
vinnyjth, line 78: Is it bad that we patch afterwards anyway? Probably not...

redreceipt: I don't _think_ so. Figured this was the simplest logic.

vinnyjth (approved):

PR #2173, opened 2021-09-10
chore: removes metarouter

redreceipt:


vinnyjth:
Haven't tested this but it looks good. @conrad-vanl can you confirm we wanted to remove metarouter?

vinnyjth (approved):

redreceipt:
> Haven't tested this but it looks good. @conrad-vanl can you confirm we wanted to remove metarouter?

he said it was fine on slack when i wrote this

PR #2175, opened 2021-09-10
fix: use ContentItem.getChildren instead of root.getChildren()

vinnyjth:
`ContentItem.getChildren()` has sorting backed in. `root.getChildren()` does not.


packages/apollos-data-connector-postgres/src/action-algorithms/index.js
```diff
@@ -165,7 +165,7 @@ class ActionAlgorithm extends PostgresDataSource {
 
     const series = seriesList[0];
 
-    const items = await series.getChildren({ limit });
```
redreceipt, line 168: does this function even exist? I'm confused

vinnyjth: it's autodefined by sequelize since we define a `children` relationship on the model

redreceipt (approved):

PR #2176, opened 2021-09-13
feat: adds interaction button feature to schema

redreceipt:


redreceipt:
It would be a lot more scalable and obviously I love the simplicity

packages/apollos-data-schema/src/index.js
```diff
@@ -1034,6 +1036,14 @@ export const featuresSchema = gql`
     action: FeatureAction
   }
 
+  type InteractionButtonFeature implements Feature & Node {
```
vinnyjth, line 1039: I like this a lot! What does line 1044 return? Would that be the most recent interaction?

redreceipt: That's a good question actually. Maybe yea, for this it would either toggle_on or off if they specified a "ToggleButtonEnabled" attribute in Rock? But that seems too specific. The other route is we let them define the interaction name but that makes the schema more complicated and we have to do some weird <name>_ON that magically matches the enum or something.

redreceipt:
@conrad-vanl would welcome your thoughts here. we get the schema right and everything else will flow from there

packages/apollos-data-schema/src/index.js
conrad-vanl: yeah, I have the same question as @vinnyjth which in my mind means we should at least explore a more explicit attribute name.

Some other thoughts:

1. VCC isn't opposed to _not_ making this a toggle button, and instead something like a counter. The TOGGLE_ON and TOGGLE_OFF interactions is the only bit that feels weird to me. If we felt better about it being a counter I could tweak design.
2. But also fine keeping the toggle on/off bit. However, to get away with a single interaction name, I think the interaction has to be on the feature, and not not he content item. If the interaction is on the content item, I think the interaction name needs to be more unique (in VCC's case, the interaction would be like "completed")

Is the `interaction` attribute really `lastInteraction`?? Or maybe it needs to be more explicit - like `isToggled: Boolean!`

redreceipt: I had it `isToggled` at first, but then we'd have to add a new interaction to the enum, right? or use the `COMPLETE` one for a different purpose which seems weird.

conrad-vanl: yeah I think the problem is the interaction is on the content item, and not on the feature. You could have multiple InteractionButtonFeatures on a content item but they'd all write to the same interaction. that's what we have to solve for, I think

conrad-vanl: @vinnyjth could we interact with a Feature instead of a ContentItem?

vinnyjth: @conrad-vanl We could - but right now features are bit more non-determanistic than items. Changing the order of a feature, or changing the text of one, etc could cause it's ID to change. We could potentially accommodate that in the shovel, and move the interaction references around as well. 

/cc @tomthornton

conrad-vanl: I'm just thinking if the interaction was on the feature instead of the content item, then we can use a super simple interaction name....otherwise I feel like we'll have to have "unique" interaction names for any feature that creates an interaction.

@redreceipt here's another idea....instead of trying to keep this feature more abstract - what if we made this a "Complete Item Toggle" and tapping it creates a "Completed" interaction and tapping it again creates an "Incompleted" interaction.

Then, this feature would only be used when you wanted to have a button that lets a user mark a piece of content as completed. 

This could be re-used at the bottom of scripture devotionals or other reading plans that you progress through, for example. 

That way, we aren't trying to create something that needs a lot of configurability?

conrad-vanl: Other ways we'd re-use a completed interaction:

- When watching most or all of a video
- When scrolling to the bottom of reading plan content, etc

PR #2177, opened 2021-09-13
feat: adds complete button feature

redreceipt:
currently, a Rock attribute called "Complete Button Text" will add this feature to any content item


packages/apollos-data-connector-rock/src/content-items/data-source.js
```diff
@@ -264,6 +262,19 @@ export default class ContentItem extends RockApolloDataSource {
...
     }
 
+    const completeButtonText = attributeValues?.completeButtonText?.value;
+    if (completeButtonText)
+      features.push(
+        Feature.createButtonFeature({
+          id: attributeValues.completeButtonText.id,
+          action: Feature.attachActionIds({
+            relatedNode: item,
+            action: 'COMPLETE_NODE',
```
conrad-vanl, line 272: So! This is the one spot I might say we should consider having a more abstract interaction (although I could be talked out of it). Something like:

```
  action: 'INTERACTION',
  interactionName: 'COMPLETE',
```

or something like that

conrad-vanl: similar to how `OPEN_URL` allows you to open any arbitrary url, instead of something specific like `OPEN_EVENT_PAGE`

vinnyjth:
One idiomatic way to do this (maybe)? Would be for the action to be TRIGGER_INTERACTION and the related node would be an Interaction or InteractionTemplate or something

conrad-vanl:
I'll let @vinnyjth / @redreceipt weigh in more on whether this action should be specific or not, but the more I think about it lets just make it a specific `COMPLETE_NODE` type action like @redreceipt has it. The nice thing about doing that is we can have optimistic loading and cache updates because we know `isComplete` or whatever should switch to true as a result of this mutation, etc....I think it can lead to an overall better UX.

Just generally I think it's nice when we do add SCHEMA like this on the api side, to make it flexible so that we have more control to customize behavior without touching the client. But I think we just need to be cautious/careful when/how we do that.

redreceipt:
So this PR needs to change or not? I'm confused about where it sits. This solution still has my vote, minimal changes and falls in line with current architecture.

conrad-vanl (approved):

PR #2179, opened 2021-09-15
fix: write campuses back to Rock correctly

redreceipt:


packages/apollos-data-connector-rock/src/campuses/data-source.js
```diff
@@ -114,9 +115,9 @@ export default class Campus extends RockApolloDataSource {
     ).first();
 
     if (!personGroup) return null;
-    const { id: rockCampusId } = parseGlobalId(campusId);
-
-    await this.patch(`/Groups/${personGroup.id}`, { CampusId: rockCampusId });
+    await this.patch(`/Groups/${personGroup.id}`, {
+      CampusId: apollosId ? parseGlobalId(apollosId).id : rockId,
```
redreceipt, line 119: actually @vinnyjth look closely at this, not sure this is right. what does `parseGlobalId(apollosId).id` give us?

vinnyjth (approved):

packages/apollos-data-connector-rock/src/campuses/data-source.js
vinnyjth: A rock campus id (assuming you don't pass a Postgres Id)

redreceipt: there i simplified the signature a bit to make it clearer

redreceipt: confirm you like the new params before i merge

PR #2181, opened 2021-09-16
fix: Stop spreading related node

vinnyjth:
Fixes an issue where the related node breaks in the `childrenOfParentsByCategoriesAlgoritm`. Found in Newspring


vinnyjth:
image

(id was `null` before)

nlewis84 (approved):

PR #2183, opened 2021-09-17
fix: Allow campus media to be an optional association

vinnyjth:
This will solve the error we are seeing on non-content postgres apps. We might need to do this in more places

image


tomthornton (approved):
Ah, that makes sense.

PR #2186, opened 2021-09-21
fix: Fix postgres button feature resolver

tomthornton:


redreceipt (changes requested):
Other button features have a related node so it only should be null if it's null in the DB

redreceipt (approved):

PR #2188, opened 2021-09-23
fix: Remove Image Placeholder for No Image

nlewis84:
In instances where a content item does not have an image, this PR removed the placeholder image (grey background) and only displays the text of a content item. This can be customized based on the algorithm used in the project. In the below video there is a content item (Light The Way) that is displayed with no image on the Home Tab, but is displayed with an image on a different tab.

Tested this on Core and on Vineyard (Go Do Good). The video is from GDG as they had more content to easily see the effect of this change.

https://user-images.githubusercontent.com/72768221/134529180-5109701f-5566-4040-8a64-ee59ec25fc9a.mov


packages/apollos-ui-kit/src/DefaultCard/index.js
```diff
@@ -61,9 +61,11 @@ const DefaultCard = withIsLoading(
     style,
   }) => (
     <Card isLoading={isLoading} style={style}>
-      <BackgroundImageBlur source={coverImage} />
-
-      <CoverImage source={coverImage} />
+      {coverImage
+        ? <BackgroundImageBlur source={coverImage} /> && (
```
redreceipt, line 65: these should children wrapped in a fragment:

```
<>
  <BG>
  <Cover>
</>
```

nlewis84: Got it!

redreceipt:
update snaps and then ping me, I'm not sure they should be changing so much

nlewis84:
Oops, my bad. I thought I got them all. Updating now!

nlewis84:
@redreceipt Check out that last failed test in CI. That's unrelated right? I don't see any of my changes that should have affected that.

redreceipt:
> @redreceipt Check out that last failed test in CI. That's unrelated right? I don't see any of my changes that should have affected that.

rerunning

conrad-vanl:
I think we might want to look at padding on cards. In Figma when I remove an image it looks like this:
Screen Shot 2021-09-23 at 11 45 06 AM

Which is close, but I think this has a bit more breathing room. According to Figma horizontal padding is 20 and vertical padding is 16. Looking at the screenshots in this PR it looks like horizontal padding is visually tighter than vertical padding.

nlewis84:
Yeah padding is 16 vertical and 16 horizontal on the current. I can bump horizontal to 20.

nlewis84:
Before
before

After
after

redreceipt (approved):

PR #2191, opened 2021-09-23
fix: remove image placeholder from horizontal default cards 

nlewis84:
In instances where a content item does not have an image, this PR removes the placeholder image (grey background) and only displays the text of a content item. This can be customized based on the algorithm used in the project. In the below video, you will see horizontal default cards that have an image and are unchanged, as well as cards that now only show text. There was also a need to adjust the position of the text if there was no image.

Tested this on Core and on Vineyard (Go Do Good). The video is from GDG as they had more content to easily see the effect of this change. 

https://user-images.githubusercontent.com/72768221/134541368-62d67b67-3a6f-47be-ae7a-b6f58bd61dba.mp4


redreceipt:
what about if a channel has some with and some without? @conrad-vanl you good with this design change?

nlewis84:
It took me a second to get a channel setup, but here's what it looks like if you have some with images and some without

https://user-images.githubusercontent.com/72768221/134544147-3c17bf8b-2213-4f0e-a1ad-7d536a330808.mp4

.

conrad-vanl:
So the only change I'd like to see here is on these cards even without an image they should be the same height. Maybe aspectRatio: 1 or something, just with centered text.

Screen Shot 2021-09-23 at 11 39 38 AM

nlewis84:
@conrad-vanl Something like this?
https://user-images.githubusercontent.com/72768221/134549795-4507c79c-baba-4652-bf71-10c85c61e7e1.mp4

redreceipt (approved):

PR #2192, opened 2021-09-23
chore: Update NotificationPreference/People foreign key constraint

tomthornton:
People deleted in postgres will delete corresponding notification preferences


packages/apollos-data-connector-postgres/src/notification-preferences/migrations/003_update_notification_preferences_foreign_key_constraint.js
```diff
@@ -0,0 +1,36 @@
+async function up({ context: queryInterface }) {
+  await queryInterface.removeConstraint(
+    'notification_preference',
```
vinnyjth, line 3: It looks here like you're deleting two constraints and replacing it with a single constraint. What's the other constraint that you are deleting?

vinnyjth: NVM! I can't read

vinnyjth (approved):

PR #2196, opened 2021-09-27
feat: Selectable text for HTMLView

vinnyjth:
image

This is part of the work needed to support copy/paste scripture. 

Text being selected doesn't change hilight color on iOS. I believe we would have to implement state w/ onSelected handlers in order to enable that. 

I'm open to other ideas on how to handle this.


conrad-vanl:
That's strange you can't see what you're highlighting

vinnyjth:
> That's strange you can't see what you're highlighting

I agree. There's a `selectionColor` prop, but that's android only, and there's also a `suppressHighlighting` prop, but it defaults to false.

vinnyjth:
@conrad-vanl I did a test of a native text component with no styles or props besides `selectable`. It looked like this...

image

conrad-vanl:
This Stack Overflow post, while it's a different question, makes it sound like it should work on iOS? https://stackoverflow.com/questions/54368090/how-to-make-selectioncolor-prop-in-textinput-transparent

vinnyjth:
@conrad-vanl no luck :/ Trying some other options.

vinnyjth:
Ah @conrad-vanl. 

Using `TextInput` gives you drag handles. 
Using `Text` gives you `copy` dialog. 

Is it worth porting the underlying text component of our app/htmlview to TextInput 😅 ?

vinnyjth:
@conrad-vanl context: https://github.com/facebook/react-native/issues/13938

vinnyjth:
@conrad-vanl what are your thoughts here? Is this worth having without grabby handles? I haven't looked at what it would take to swap out the text elements for TextInputs, but I could.

conrad-vanl:
Eh, I think this is probably something to punt at this point. I'll let frank know. it seems high risk to switch to text inputs.

PR #2198, opened 2021-09-28
chore: adds xs size avatar

redreceipt:
test with `BRANCH=xs yarn app`
Screen Shot 2021-09-28 at 10 43 18 AM


conrad-vanl:
@redreceipt why is the avatar still not circular?

conrad-vanl:
Screen Shot 2021-09-28 at 10 07 42 AM

redreceipt:
yea I'm not sure :/ we should fix but maybe in a different PR so it doesn't hold up FNWA launch

nlewis84:
Maybe this is causing it? https://github.com/ApollosProject/apollos-templates/blob/9bcaafbbb4d88eec9f808c959a51b082579345ad/apolloschurchapp/src/tabs/index.js#L54

conrad-vanl (approved):

PR #2200, opened 2021-09-28
feat: adds custom SMS message

redreceipt:
IMG_376A1D515236-1


packages/apollos-data-connector-rock/src/auth-sms/data-source.js
```diff
@@ -169,7 +169,9 @@ export default class AuthSmsDataSource extends RockApolloDataSource {
 
     await this.context.dataSources.Sms.sendSms({
       to: e164,
-      body: `Your login code is ${pin}`,
+      body: `${
+        process.env.SMS_LOGIN_MESSAGE || 'Thanks for logging in!'
```
vinnyjth, line 173: This feels like a config.yml value :)

redreceipt: that adds code to the template when the default doesn't need to be there. leaving it an env variable requires no code changes and seems simpler. config file here feels very redundant.

vinnyjth: I guess I'm still not on board with moving everything to the .env files. They don't merge well in version control, are ignored on Heroku and don't always leave a good trail to finding what the name of a value should be. Especially for non-secrets. Could we do a `Config.VALUE || 'default'` here?

redreceipt: I agree with you. I would argue we don't need the env shared file on the API at all, it's just confusing. My problem with the config file is it's so specific to each API and can vary greatly so it's impossible to have a template so you have to hunt for values anyway. It's confusing that some values _have_ to be environment variables and some _can't_ be. The developer should only have to be put things in one place.

vinnyjth: @redreceipt I agree with what you're saying. Is it helpful if I add that in the future when we move to a multi-tenant architecture we won't be able to use `.env` variables at all on the API? The will likely be stored in postgres at that point though....

vinnyjth, line 173:
```suggestion
        ApollosConfig.SMS_LOGIN_MESSAGE || 'Thanks for logging in!'
```

I don't think we need to touch the template, since we have a default.

vinnyjth: Then at the top of the file, 
```
```

redreceipt: like all secrets in the DB except the database URL?

vinnyjth: Correct. And the postgres database will be universal (same for everyone) but segmented via church.

vinnyjth: At least that's the running thought.

redreceipt: I made the change :)

vinnyjth: Thanks man ❤️  We're slowly moving towards our more-perfect future

vinnyjth (approved):

vinnyjth:
@redreceipt I think you need to update snaps :)

PR #2211, opened 2021-09-30
fix: Fix Feed View for iPad

tomthornton:
Fixes the feed view to be vertical instead of horizontal on iPad
Screen Shot 2021-09-30 at 12 03 11 PM

Also centers the onboarding logo on iPad. It looked weird all the way to the left
Screen Shot 2021-09-30 at 12 11 47 PM


redreceipt:
doesn't this also adjust the onboarding logo for phones?

tomthornton:
@redreceipt nope it just centers the content within the View.

Screen Shot 2021-09-30 at 3 49 43 PM

redreceipt:
nice. update snaps and I'll take a look. I'm curious how pervasive this change will be.

tomthornton:
@redreceipt do you mean storybook snapshots? Haven't done that yet

redreceipt:
Oh no, I was assuming the test failure you have is `ui-kit` snaps.

tomthornton:
@redreceipt the test failed for something unrelated. I think it's that Sequelize error that sometimes happens and sometimes doesn't.

packages/apollos-ui-kit/src/FeedView/index.js
```diff
@@ -143,7 +143,7 @@ const enhance = compose(
     }))
   ),
   mediaQuery(
-    ({ md }) => ({ maxWidth: md }),
+    ({ lg }) => ({ maxWidth: lg }),
     defaultProps({ numColumns: 1 }),
     defaultProps({ numColumns: 2 })
```
vinnyjth, line 148: Should we eliminate the media query and fix `numColumns` at 1? If I'm understanding correctly we should never have `numColumns: 2`  /cc @conrad-vanl

packages/apollos-ui-kit/src/FeedView/index.js
```diff
@@ -143,7 +143,7 @@ const enhance = compose(
     }))
   ),
   mediaQuery(
```
conrad-vanl, line 145: @tomthornton did you try what happens if we move this HOC `mediaQuery` altogether? I don't think we ever want this thing trying to split into multiple columns, regardless of viewport size

conrad-vanl: Removing this seems like this will force this view to look the same on all devices, which I think since we don't have good testing practices for iPad currently, that may be the best option

packages/apollos-ui-kit/src/FeedView/index.js
conrad-vanl: LOL, I just said the same thing below

conrad-vanl:
Marking this as #priority b/c 3 most recent Apollos Apps look like crap due to iPad support.

packages/apollos-ui-kit/src/FeedView/index.js
tomthornton: @conrad-vanl removing the `mediaQuery` HOC has the same effect. Yeah  agree that it should never be two columns.

vinnyjth: Should we remove `numColumns` from the props at this point in time or default the prop to `1`?

tomthornton: @vinnyjth I went ahead and removed the `mediaQuery` HOC altogether. Is there any reason we need it?

vinnyjth: I don't think we need the `numColumns` prop. Especially since props are spread down to the underlying feedView.

tomthornton:
@vinnyjth is this ready? I don't believe we need the `mediaQuery` at all but I could be wrong.

PR #2212, opened 2021-09-30
fix: Add state for loading between fetching

tomthornton:
For some reason, the `loading` prop from the Apollo HOC isn't updating whenever a feed view is refreshed. This adds state to the FeedView itself in order to get the desired effect.  Also, if there is an error in refetching the loading state will be set `false`


redreceipt:
I don't fully understand the problem, could you possibly post a GIF displaying what this is fixing?

tomthornton:
@redreceipt the loading indicator is immediately dismissed when refreshed. This shows and accurate loading indicator.

**Before**

https://user-images.githubusercontent.com/40962668/135520383-31486d96-be83-4a9a-a851-fed48b326826.mp4

**After**
https://user-images.githubusercontent.com/40962668/135520520-8dbef40c-6bfe-4ce7-899d-c731a0e9a07f.mp4

redreceipt:
nice! yea fix tests and then I'll approve

redreceipt:
snapshots are failing here

tomthornton:
@vinnyjth @redreceipt is it okay to remove the `isLoading` test from components who use the `FeedView`? Since the `isLoading` prop is now being handled within `FeedView` its self there doesn't need to be tests outside. Or we could figure out why `isLoading` isn't updating in the first place but I couldn't figure it out.

redreceipt:
I'm cool with it as long as it doesn't make coverage go down. Codecov will tell us.

redreceipt:
I'm seeing strange animation and behavior with this one

Kapture 2021-12-06 at 19 53 12

PR #2215, opened 2021-10-01
fix: Automerge functionality deletes emails

vinnyjth:
The automerge functionality being turned on was deleting folks email address. 

This was for a few reasons. 

1. We historically pass down `email: null` when doing SMS logins (I need to explore why we do that a bit more) 
2. Prior to turning *automerge* off, we ran the code currently inside the if statement. 
3. When enabling automerge again, we choose to run the `patch` code even if we didn't need to automerge. 
4. That `patch` code ends up deleting folks emails. Hence why it's a separate operation inside the Rock UI.


conrad-vanl:
We should prioritize getting this out on apps that currently have auto-merge turned on....that's chase oaks, fellowship, and.... RV?

vinnyjth:
@conrad-vanl do you want to review it? If you review I can merge and start work on deploy.

vinnyjth:
Here's the code prior to removing automerge (and then allowing folks to turn it back on) 

https://github.com/ApollosProject/apollos-apps/blob/a2b62f6b098fe9403f86f3003b8638a6008f2ff7/packages/apollos-data-connector-rock/src/people/data-source.js#L68-L75

redreceipt (approved):
nice. we probably should have done it this way to begin with.

conrad-vanl:
@redreceipt @vinnyjth I need one of you to prioritize getting this deployed on Chase Oaks ASAP.

@redreceipt do we have any other apps in production with this error?

redreceipt:
Anyone with auto merge turned on would have it.

conrad-vanl:
Yeah but who do we have with auto merge that's in production?

redreceipt:
I wish I’d recorded that somewhere :/ let me do some digging through Heroku instances. I’ll patch any one that I find and let you know.

PR #2219, opened 2021-10-04
chore: Add Phosphor Icons

tomthornton:
### **Phosphor Icon Set**

run `yarn install` to add [phosphor-react-native](https://github.com/duongdev/phosphor-react-native)

### **Organize icon sets within `./theme/icons`**

Moved core icons and phosphor icons into their own separate folders so that we may add icon sets in the future if necessary.

### **Phospher Icon HOC**

In order to keep the `Icon` component consistent, I've created a `HOC` that wraps all Phosphor Icons to have the correct props. For example, `fill` must be changed to `color`

### **Icon Set Theme Option**

The default icon set and weight (really the type of icon it is... outlined, bold, filled, etc.) are configurable through the theme. For now the default icon set is `Phosphor` and the default weight is `regular`.

### **Disclaimer**

A lot of Phosphor icon names don't match the names of the current Icon set. We can reconcile this by having a standard name for all icons across the app and changing the names in the `Phosphor Icon HOC`. For example, change the `House` icon of Phosphor to `Home` so that we can switch between icon sets easily and have no problems.


vinnyjth:
@tomthornton Could you add a screenshot or two of the new icons in the app? Thanks!

tomthornton:
Screen Shot 2021-12-07 at 11 48 53 AM
Screen Shot 2021-12-07 at 11 49 20 AM

vinnyjth (approved):

PR #2222, opened 2021-10-07
Update ContentCardConnected.js

conrad-vanl:
**NOT YET TESTED**


conrad-vanl:
probably need to do the same for HorizontalContentCardConnected

PR #2247, opened 2021-11-03
feat: Tag Filter

tomthornton:
@redreceipt still need to add tests but wanted to go ahead and show you the approach.

### Testing
This [Shovel PR](https://github.com/ApollosProject/apollos-shovel/pull/88) will allow you to run the content item backfill locally and have the tags show up. I'm using VCC for this example, but I can recreate it in core if need be.

### Screen Recording
https://user-images.githubusercontent.com/40962668/140102731-13542c46-4ac3-40e1-b466-d6cd9f4b573a.MP4

### Dark Mode
IMG_2850
e


redreceipt:
Nice! I like the approach. Recreate in core please so we can easily spin up the app and test. And also split the PR into API and app please. We'll need to duplicate the api side in the cluster once we decide on a final outcome.

tomthornton:
@redreceipt gotcha. will do!

PR #2250, opened 2021-11-05
fix: Journal prompt floating label can have multiple lines

tomthornton:
https://user-images.githubusercontent.com/40962668/140551285-229468b1-3ee9-427e-932b-fff0e93acac4.mp4


nlewis84:
Yes! Nice finds. I figured that "magic number" was going to be an issue as well.

nlewis84:
@redreceipt Does this need to go into the cluster repo as well?

nlewis84 (approved):
Yeah, this makes sense to me.

nlewis84:
ah, we need to write some tests for labelShrunk

tomthornton:
@nlewis84 any idea how we'd add tests for that? seems like we don't need one.

nlewis84:
@tomthornton I looked at it and I don't know. @redreceipt any guidance on how to adds tests for this new prompt?

redreceipt:
Looks like all this PR does is add a prop to text inputs. So a snap with and without that prop should give you 100% on this I would think.

tomthornton:
Having a huge problem here with `@gorhom/bottom-sheet` in getting this test to pass. I believe it has something to do with jest mocking the component. Can't understand why it works in the `AddCommentInput.tests.js` but not it's own test. I've added some mocks to `react-native-reanimated` but still can't get it to work.

tomthornton:
@redreceipt we might need to either pair or go ahead and merge this because I can't get the tests to pass. I've searched everything and tried a bunch of different solutions but nothing is working.

redreceipt:
at this point the tests are failing so we definitely shouldn't merge

redreceipt:
what i like to do is take out one piece at a time from a component until i figure out the exact piece that's breaking the render. do you know what component is breaking it?

tomthornton:
@redreceipt yeah it's the `@gorham/bottom-sheet` component.

```
TypeError: Cannot destructure property 'snapPointsCount' of '(0 , _useBottomSheetInternal.useBottomSheetInternal)(...)' as it is undefined.

      at useScrollableInternal (../../node_modules/@gorhom/bottom-sheet/lib/commonjs/hooks/useScrollableInternal.ts:24:5)
```

tomthornton:
I've tried mocking it in the `jest.setup.js` file, installing a new version, nothing has worked

redreceipt:
I feel like this one maybe could be simplified. 100 snap changes is a big change when isn't the goal of this just to make a multi-line label possible? I don't understand why we should need props to handle that?

packages/apollos-ui-kit/src/inputs/FloatingLabel.js
```diff
@@ -84,7 +85,12 @@ class FloatingLabel extends PureComponent {
         onLayout={this.handleLayout}
         style={[styles.floatLabelView, wrapperStyles, this.props.style]}
       >
-        <LabelText color={this.props.color}>{this.props.children}</LabelText>
+        <LabelText
+          numberOfLines={this.props.labelShrunk ? 1 : 0}
```
nlewis84, line 89: Could we move the `labelShrunk` logic here, get rid of that prop, and keep our snap changes to a minimum?

`isEditing || (text.value && text.value.length > 0)`

redreceipt: prob shouldn't. `isEditing` is referring to the text input state, not the label

redreceipt:
@tomthornton what's the best way for me to test this?

redreceipt:
nevermind, @nlewis84 already approved

PR #2252, opened 2021-11-08
fix: Error handling for initialPrompt and addPrompt resolver

tomthornton:


redreceipt (commented):
I would rather find out what's going wrong and fix it for real than just a generic try catch

tomthornton:
What's going wrong is that the shovel used to provide defaults for addPrompt and initialPrompt but I removed them. Those defaults caused them to not be null.

This could just as easily be a shovel PR, we'd just be going back and adding deprecated fields.

redreceipt:
So why don't we just set a default at the resolver instead of an error handler?

tomthornton:
Because I still don't know what `JSON.parse(id` does. To me, it looks like we can remove it and set defaults. I did it like this to cover all bases.

redreceipt:
Do it. We'll fix it later if it breaks a "content from Rock" church.

tomthornton:
@redreceipt done

redreceipt:
@tomthornton would you add this to the cluster too

tomthornton:
https://github.com/ApollosProject/apollos-cluster/pull/38

PR #2256, opened 2021-11-15
fix: Button Feature Interactions fix

tomthornton:
This is for chase oaks. `interactions` was still returning null even though it was given destructuring defaults. This ensures that `interactions` will always have an array value.


packages/apollos-ui-connected/src/features/ButtonFeature.js
```diff
@@ -16,6 +16,9 @@ const ButtonFeature = ({ action }) => {
     `,
     { variables: { nodeId: action.relatedNode.id } }
   );
+
+  const interactions = data?.interactions ?? [];
```
nlewis84, line 20: Makes sense, `null` was not triggering the default previously.

nlewis84 (approved):

nlewis84:
Just need some tests updated.

packages/apollos-ui-connected/src/features/ButtonFeature.js
tomthornton: Yeah. It's cause the `interactions` query doesn't exist

PR #2258, opened 2021-11-15
feat: parse audio source as guid or url

nlewis84:
This PR is an change to how we parse audio sources, allowing guids to hit the Binary Files endpoint in Rock and grabbing the appropriate url.


redreceipt (approved):
looks good!
Screen Shot 2021-11-15 at 4 32 02 PM

PR #2261, opened 2021-11-17
feat: adds FollowPeopleFeature to the API

redreceipt:
test with `BRANCH=follow-people yarn api`

Screen Shot 2021-11-16 at 9 58 46 PM


redreceipt:
@vinnyjth one change I'd like to make with this PR is to take away the ability to just list an email address instead of an object. It seems unnecessary and won't serve us well when we move these to the DB anyway.

redreceipt:
@vinnyjth I just realized a problem with this PR (thanks to good test coverage) we can't get rid of the top-level `SUGGESTED_FOLLOW` key because of the auto-accept feature. I could do a deep search through features but that's kinda goofy. Unless you have a better idea, I'll put that back. I was just excited to remove that key and make it a per-feature list.

redreceipt:
> @vinnyjth I just realized a problem with this PR (thanks to good test coverage) we can't get rid of the top-level `SUGGESTED_FOLLOW` key because of the auto-accept feature. I could do a deep search through features but that's kinda goofy. Unless you have a better idea, I'll put that back. I was just excited to remove that key and make it a per-feature list.

we could put that logic on the front end??? `if suggestion, new follow mutation, else requestFollow mutation`

vinnyjth:
@redreceipt The "correct" way in my head to handle auto follows is to have that data live in a sql table or as an attribute on a user. It would still be global data, but we could manage that functionality a lot more naturally.

redreceipt:
The advantage of it per feature is context isolation. If they don't have a follow feature, they don't care about auto follows.

redreceipt:
well, this PR currently doesn't change it. we can iterate later.

redreceipt:
ok, @vinnyjth we can make that decision later. For now, I left it as is. This is ready for review.

vinnyjth (approved):

PR #2264, opened 2021-11-22
feat: Simple Donation Giving Tab

tomthornton:
## Testing

1. Run `BRANCH=simple-donation yarn api` then `BRANCH=simple-donation yarn app`
2. For the client `.env`, you'll need to change `APP_DATA_URL` to `http://localhost:4000` to make the `/simpledonation` route available from the API.
3. You'll then need to add `MERLIN_KEY` to the `.env` for the API. For testing purposes, we'll use Fellowships, so DM me on Slack for it.

## Screenshot

Screen Shot 2021-11-22 at 1 11 37 PM


packages/apollos-server-core/src/giving/simple-donation.js
```diff
@@ -0,0 +1,17 @@
+export default function useSimpleDonationRoute({ app }) {
+  // Give Screen
+  return app.get('/simpledonation', (req, res) => {
```
conrad-vanl, line 3: @tomthornton before creating a custom endpoint on our API, did you attempt to localize this HTML on the frontend side of things? I believe we can basically take your html code below and set it as the body of our web view:

```
<webview source={{ html: "<!DOCTYPE html>" }} /> 
```

Kind of thing. Right? https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#source

packages/apollos-ui-kit/src/GiveScreen/index.js
```diff
@@ -0,0 +1,23 @@
...
+import { SafeAreaView } from 'react-native';
+import { WebView } from 'react-native-webview';
+import ApollosConfig from '@apollosproject/config';
+import styled from '../styled';
+
+const Container = styled(() => ({
+  flex: 1,
+}))(SafeAreaView);
+
+export default function GiveScreen() {
```
conrad-vanl, line 11: Thoughts on building this as a feature and having a normal feed tab with a `WebViewFeature` (or specific `SimpleDonationFeature`)

packages/apollos-server-core/src/giving/simple-donation.js
tomthornton: @conrad-vanl yeah @camrun91 and I tried that. Unfortunately it didn't work for some reason. I believe there might be something in the simple-donation script or the web view in general that wants it to be served from an endpoint.

packages/apollos-ui-kit/src/GiveScreen/index.js
tomthornton: Sure, we could do it that way. You mean have it configurable through the config.yml not tied to a content item as a feature right?

packages/apollos-server-core/src/giving/simple-donation.js
camrun91: @conrad-vanl I think there is currently a bug in webview that it needs to have the page rendered and served.

conrad-vanl:
@vinnyjth @redreceipt would like your thoughts on how we should build this feature. Here's some screenshots of what we might want to do eventually. I think this makes sense as a feed feature, not tied to content, yeah. Something we can enable via config.yaml

image

image

redreceipt:
`WebView` feature makes sense to me, with ability to inject the HTML via config

packages/apollos-server-core/src/giving/simple-donation.js
conrad-vanl: Hmmmm it's probably something on Simple Donation's side. 

Here's another idea (I really don't want to add an endpoint to our API from a maintenance perspective):

The Simple Donation script loads an <iframe>....what if we figured out the URL of that iframe and just set that to the webview source instead?

tomthornton: I can give this a shot and see what happens

tomthornton:
Added `WebView` as a config feature. Here's the `config.yml`

```
TABS:
  HOME:
    - type: WebView
      url: http://localhost:4000/simpledonation
      height: 500
```

Screen Shot 2021-11-29 at 9 54 27 AM

conrad-vanl:
this is looking great! How hard would it be to inject some custom CSS on the embed that removes the "X" close button and the "?" mark button at the top? maybe remove that whole "header" if it's a separate div or something?

tomthornton:
@conrad-vanl yeah I know, the X button sucks 😕 I tried that but unfortunately you can't modify iframe CSS unless it comes from the original source.

vinnyjth:
> @conrad-vanl yeah I know, the X button sucks 😕 I tried that but unfortunately you can't modify iframe CSS unless it comes from the original source. I even tried a little vanilla JS to manipulate the DOM but it's the same deal as the CSS.

I think that gives us even more reason to ditch the new `/simpledonation` endpoint and see if we can find the underlying iframe url format for simple donation. Then we can inject JS (and thus CSS)

camrun91:
@conrad-vanl @vinnyjth I bet if you reach out Josh at simple donation can help out with this. He is the one that built the merlin widget.

tomthornton:
@vinnyjth the problem is there are 5 iframes that all use that initial script, and some are from stripe.
Frame 1

conrad-vanl:
The other path is to make simple donation customize it, they have that capability. I think we can reach out to Taylor

conrad-vanl:
Taylor is on our apollos slack

conrad-vanl:
I reached out to Taylor on slack

conrad-vanl:
If we want to test this flow without it charging our card:

Screen Shot 2021-11-30 at 3 17 02 PM

tomthornton:
@vinnyjth I couldn't find a way for it to work without using the endpoint. Static HTML doesn't work.

vinnyjth:
@tomthornton Have you reached out to SimpleDonation and seen if they had any solutions?

tomthornton:
@vinnyjth I believe @conrad-vanl reached out to them via slack

conrad-vanl:
From Taylor at Simple Donation:

> We expect to release a new version that will allow you to paste in a hosted stylesheet.
> It would be like this:
> 
> https://merlin.simpledonation.com/installScript.js?key=1234&cssFile=https://apollosproject.com/myawesomestylez.css
> 
> Would that work? I’ll nudge you when we get the release out.

conrad-vanl:
This change is going to be rolled out today (he said within next 30 minutes). @tomthornton wanna try to 🛬 on this one?

camrun91:
@conrad-vanl so let me make sure I am following this correctly. We are still adding the /simpledonation endpoint but passing in custom CSS to style it?

tomthornton:
The good news is that it works great when the CSS file is being served at `/giving.css`! The problem is that lerna/babel isn't copying over the CSS file needed to the `lib` folder where `__dirname` is being called. I tried sending the CSS as a string and setting the `Content-Type` header to `css` and `text-css` but no luck. I see 3 possible solutions

- Get lerna/babel to copy over the CSS file to the `lib` folder
- Make this route a custom change for every API that needs it.
- Edit the `__dirname` string by replacing `lib` with `src`

@vinnyjth thoughts?

tomthornton:
I've added what I think is the most simple, which is replacing `lib` with `src` and it works this way
Screen Shot 2021-12-15 at 1 02 48 PM
.

vinnyjth:
@tomthornton The `src`/`lib` swap is a good idea! It will work in a dev environment, but I don't think it will work in our production environment, where we often copy over just the `lib` folder. I would see if you can convince babel to copy the file over, it shouldn't be a huge change.

vinnyjth:
Oh, and the `src` folder isn't sent to `npm`

tomthornton:
Ah okay gotcha thanks @vinnyjth!

tomthornton:
Got it! just needed to add `--copy-files`

tomthornton:
For some reason it's not showing rendering the CSS in the app.  It works perfectly fine in the browser. 🙄

Screen Shot 2021-12-15 at 2 29 42 PM

taylorbrooks:
> For some reason it's not showing rendering the CSS in the app.  It works perfectly fine in the browser. 🙄

Possible race condition? I can explore alternative solutions on our end.

tomthornton:
@taylorbrooks after some research, I believe it had to do with safari and our local environment. I put the css file in a remote S3 bucket and used that link and it works now.

packages/apollos-server-core/src/giving/simple-donation.js
```diff
@@ -0,0 +1,20 @@
...
+  // Give Screen
+  app.get('/simpledonation', (req, res) => {
+    res.send(`
+            <!DOCTYPE html>
+            <html>
+                <head>
+                    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+                </head>
+                <body>
+                    <script async src="https://merlin.simpledonation.com/js/installScript.js?key=${process.env.MERLIN_KEY}&cssFile=https://tom-misc.s3.us-east-2.amazonaws.com/giving.css"></script>
```
vinnyjth, line 11: should this be?
```suggestion
                    <script async src="https://merlin.simpledonation.com/js/installScript.js?key=${ApollosConfig.MERLIN.KEY}&cssFile=https://tom-misc.s3.us-east-2.amazonaws.com/giving.css"></script>
```

We have moved away from storing configuration in the env.

vinnyjth, line 11: Also is the plan to launch w/ the css pointed at EC2? Given that it worked fine in prod?

packages/apollos-ui-kit/src/GiveScreen/GiveScreen.tests.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import Providers from '../Providers';
+import GiveScreen from '.';
+
+describe('the GiveScreen component', () => {
```
vinnyjth, line 6: I really appreciate all the tests :)

packages/apollos-ui-kit/src/index.js
```diff
@@ -36,6 +36,7 @@ export {
   default as FollowListSearch,
 } from './FollowListSearch';
 export GradientOverlayImage from './GradientOverlayImage';
+export GiveScreen from './GiveScreen';
```
vinnyjth, line 39: Were we going to keep both the feature and the `GiveScreen` export? that's more a question for you @conrad-vanl

tomthornton: Probably not i don't think it's needed anymore

packages/apollos-server-core/src/giving/simple-donation.js
tomthornton: Yep I can do that. Is there a certain way to add it to the ApollosConfig? I think that's why I didn't do it because I couldn't figure it out

vinnyjth: Yah :D You can add something like this to the end of the `config.yml` in templates

```
MERLIN: 
  KEY: ""
```

tomthornton: Nvm I figured it out

tomthornton: With this, it should work in prod and locally. But I think longterm we forgo the fact that it doesn't work locally because it's cosmetic, and use the other approach I had that gets the server's css.

conrad-vanl: >tom-misc.s3.us-east-2.amazonaws.com

we'd at least need to put this on an apollos-owned S3 bucket for this to be mergeable. I vote the latter option, what @tomthornton suggested here:

> But I think longterm we forgo the fact that it doesn't work locally because it's cosmetic, and use the other approach I had that gets the server's css.

packages/apollos-ui-kit/src/index.js
vinnyjth: Let's take the `GiveScreen` and related files out, in that case :)

tomthornton:
@vinnyjth I think this is ready :)

vinnyjth (approved):

PR #2269, opened 2021-11-23
feat: Following Prayer Request Notification

tomthornton:
When a user creates a prayer request, a notification is sent to the users following that user.


redreceipt:
love it. could you do a cluster PR as well?

packages/apollos-data-connector-postgres/src/prayers/dataSource.js
```diff
@@ -7,15 +7,38 @@ class Prayer extends PostgresDataSource {
...
-        personId: currentPersonId,
+        personId: id,
         text,
         ...args,
       },
     });
 
+    const { Notification } = this.context.dataSources;
+
+    if (Notification) {
```
redreceipt, line 26: probably don't need this block. if `Notification` is false here, we got bigger problems and we should know about it.

tomthornton: Only reason I had this in here is because it seems like a pattern (like in notifications data source) and it actually helped with the tests because Notification wasn't being initialized correctly.

redreceipt: That's why it should be there. Shouldn't we try and understand why it's not being initialized correctly?

tomthornton: That makes sense. I've updated it.

redreceipt (approved):

PR #2271, opened 2021-11-29
fix: Removes duplicate prayers from same user

tomthornton:
DB of Prayers (I'm logged in as the test prayers so it doesn't show them)

Screen Shot 2021-11-29 at 10 24 40 AM

Query

Screen Shot 2021-11-29 at 10 24 16 AM


nlewis84 (approved):
🎒

PR #2274, opened 2021-11-29
feat: adds feature feeds as a nav destination

redreceipt:
test by switching `APP_DATA_URL=https://apollos-test.staging.apollos.app`

Kapture 2021-11-29 at 18 33 24


vinnyjth (approved):

nlewis84:
Something about this PR broke the Home Tab of VCC. We were utilizing the TabComponent prop to pass in a header component.

https://user-images.githubusercontent.com/72768221/144681480-15dda520-6647-4efb-974e-676d3ea66028.mp4

Screen Shot 2021-12-03 at 4 31 52 PM

camrun91:
@redreceipt @vinnyjth It looks like this may have broken the ability to add additional Features at the tab level?

redreceipt:
Shoot. You're probably right, is that because we took out the prop spreading?

camrun91:
give me one sec and I will try 
I think I found the issue

camrun91:
@redreceipt yeah, it looks like that is the problem. However, just adding prop spreading back causes other issues.

redreceipt:
Ok so we just need to pass additionalFeatures back down through. Wanna test it and throw a PR at me?

camrun91:
Yeah I can try  it out

camrun91:
@redreceipt #2289

PR #2275, opened 2021-11-29
chore: removes deprecation warnings

redreceipt:
this cleans up the logs on a sim SO much


packages/apollos-ui-connected/src/SearchCardConnected/index.js
```diff
@@ -52,7 +51,7 @@ SearchCardConnected.propTypes = {
 };
 
 SearchCardConnected.defaultProps = {
-  Component: ContentCardComponentMapper,
+  Component: DefaultCard,
```
vinnyjth, line 54: If we don't use the `ContentCardComponentMapper` we should probably `name` this component at a minimum so we can swap the `Component`

packages/apollos-ui-connected/src/CampaignItemListFeature/index.js
```diff
@@ -26,8 +24,7 @@ const ListItemComponent = ({ contentId, labelText, ...item }) => (
     {(liveStream) => {
       const isLive = !!(liveStream && liveStream.isLive);
       return (
-        <ContentCardComponentMapper
```
vinnyjth, line 29: This should probably get a `name`, and we should likely have a `Component` prop. 

Also, I think this changes the functionality? Shouldn't we be using `FeaturedCard` instead of `DefaultCard`?

redreceipt: named.

that `Component` prop was actually not doing anything before

Screen Shot 2021-12-03 at 11 03 18 AM

packages/apollos-ui-connected/src/SearchCardConnected/index.js
redreceipt: done

vinnyjth (approved):

PR #2278, opened 2021-12-01
fix: Sermon args not being passed correctly w/ getSermons

vinnyjth:


vinnyjth:
See: https://github.com/ApollosProject/apollos-cluster/pull/72/commits

redreceipt (approved):
gotcha

PR #2280, opened 2021-12-02
feat: Setup apollos devices as their own defined type

vinnyjth:
A lot of churches have been asking about this. This makes apollos dataviews really easy to setup, and will apply to the shovel apps (since we still create personal devices with the shovel app) 
Screen Shot 2021-12-02 at 11 20 13 AM
Screen Shot 2021-12-02 at 11 20 25 AM


redreceipt:
guessing we won't be doing this on 3.0?

packages/apollos-data-connector-rock/src/personal-devices/data-source.js
```diff
@@ -16,17 +35,23 @@ export default class PersonalDevices extends RockApolloDataSource {
     // if we already have a device, shortcut the function;
     const currentUser = await this.context.dataSources.Auth.getCurrentPerson();
 
+    console.log(existing);
```
redreceipt, line 38:
```suggestion
```

vinnyjth:
> guessing we won't be doing this on 3.0?

TBD. We haven't fully ironed out our "Shovel first strategy". There's a chance that this code will move to the shovel as well. (Shoveling from our database into rock)

redreceipt:
so better copy this PR to the cluster too then until we decide what to drop completely.

redreceipt:
if it does make it to 3.0, that adds one more rock requirement. I just don't want to lose track of the feature.

vinnyjth:
> so better copy this PR to the cluster too then until we decide what to drop completely.

Correct!

vinnyjth:
> if it does make it to 3.0, that adds one more rock requirement. I just don't want to lose track of the feature.

Yah. To be clear - we need to port the PersonalDevice functionality one way or the other, this code adds a refinement to that functionality instead of creating an entirely new scope.

redreceipt:
you copy this over to the cluster so we don't forget, I'll approve

vinnyjth:
Here @redreceipt 
https://github.com/ApollosProject/apollos-cluster/pull/76

redreceipt (approved):

PR #2281, opened 2021-12-02
fix: prayer tests

vinnyjth:
These tests are passing 1/5 times since we aren't awaiting the results. 

I refactored the code to look a bit more like our Comment.dataSource notification functionality.


packages/apollos-data-connector-postgres/src/prayers/dataSource.js
```diff
@@ -21,23 +17,38 @@ class Prayer extends PostgresDataSource {
       },
     });
 
+    const sendingNotifications = this.sendNotificationForPrayer({
+      prayer: newPrayer[0],
+    });
+    if (sendNotificationsSync) {
+      // mostly used in the test environment.
+      // waits to send notfications before returning comment
```
vinnyjth, line 25:
```suggestion
      // waits to send notfications before returning prayer
```

packages/apollos-data-connector-postgres/src/prayers/dataSource.js
```diff
@@ -6,12 +6,8 @@ import { PostgresDataSource } from '../postgres';
 class Prayer extends PostgresDataSource {
   modelName = 'prayerRequest';
 
-  async addPrayer({ text, ...args }) {
-    const {
-      id,
-      firstName,
-      lastName,
-    } = await this.context.dataSources.Person.getCurrentPerson();
+  async addPrayer({ text, sendNotificationsSync = false, ...args }) {
```
redreceipt, line 9: this seems like a hack to just have this here for tests. we could probably just mock the right external function, that's what we did in comments section

redreceipt:
and whatever we do, it'll need to be copied to cluster PR

vinnyjth:
@redreceipt we do the exact some trick with Comments, that's where I copied the code from


That said, I agree this is a bit hacky. I'm open to other options, although this is breaking all our builds currently

redreceipt:
alternate proposal https://github.com/ApollosProject/apollos-apps/pull/2282

PR #2282, opened 2021-12-03
fix: writes prayer tests differently

redreceipt:


redreceipt:
@vinnyjth what do you think about this instead as a solution? we don't need to cover the logic of the notification DS from the prayer tests, we just need to make sure the function is getting called.

vinnyjth:
Looks good to me! Are you okay also pulling in the slight reorg I did with the `addPrayer` logic, so the `sendNotitificationForPrayer` functionality lives in a seperate function

redreceipt:
just move the two lines into their own function? so we have `Prayer.sendNotifications(followedPerson, requestedPersonId)`. What does that get us? Seems overkill to move that logic out unless something else would it?

vinnyjth:
yah, move the logic it. It's a cleaner separation of concerns and logic and matches what we do in the `Comments` datasource. 

And it should be `Prayer.sendNotificationForPrayer({ prayer} )`
https://github.com/ApollosProject/apollos-apps/pull/2281/files?diff=split&w=0#diff-41bd8924137ae1fda617cc9d0943b05868b2de11e865c0d924ca499dba2a4bbeR32-R51

redreceipt:
We already have a function for maybe a similar purpose down below? Are you wanting me to merge the two? Their intention seems slightly different.

https://github.com/ApollosProject/apollos-apps/blob/c0c620bb56f28d31d79b61b81b6c56f08ab0d13d/packages/apollos-data-connector-postgres/src/prayers/dataSource.js#L123

vinnyjth:
> We already have a function for maybe a similar purpose down below? Are you wanting me to merge the two? Their intention seems slightly different.

No, this is meant to be a seperate function. We can punt I guess, having the notification code inline with `addPrayer` just seemed like a code smell.

vinnyjth (approved):

PR #2283, opened 2021-12-03
feat: Report Prayer Request

tomthornton:
https://user-images.githubusercontent.com/40962668/145843349-e3a60ba3-e336-4263-b790-c53cd5eeab1b.mp4


redreceipt:
something doesn't seem right here. 184 files changed? lots of snaps that don't seem relevant

tomthornton:
@redreceipt I believe it's because I had to add the `ActionSheetProvider` at the top, within Providers. So pretty much all UI component snaps had to be updated.

redreceipt:
that makes sense. only thing left before we approve is copy over the data pkg code to the cluster in a PR

conrad-vanl:
I like the approach in this PR. @redreceipt what would it take to pick it back up / finish? Maybe something @tristanguckenberger can take?

redreceipt:
I could finish it out. not much left since packages are here now. do we need it to go to 2.X servers or will 3.0 be fine?

conrad-vanl:
> do we need it to go to 2.X servers or will 3.0 be fine?

It'd be nice if it could go to shovel apps like RV, they are having issues with spammy prayers this would help with.

conrad-vanl:
But if that's a tall order / you don't think worth it, that's fine too we can target 3.x/newest

packages/apollos-ui-prayer/src/PrayerExperienceConnected/getPrayerFeature.js
```diff
@@ -3,7 +3,7 @@ import { gql } from '@apollo/client';
 import ApollosConfig from '@apollosproject/config';
 
 export default gql`
-  query($id: ID!) {
+  query getPrayerFeature($id: ID!) {
```
redreceipt:
```suggestion
  query GetPrayerFeature($id: ID!) {
```

packages/apollos-ui-connected/src/PrayerListFeatureConnected/getPrayerFeature.js
```diff
@@ -2,7 +2,7 @@ import { gql } from '@apollo/client';
 import ApollosConfig from '@apollosproject/config';
 
 export default gql`
-  query getActionListFeature($featureId: ID!) {
+  query getPrayerListFeature($featureId: ID!) {
```
redreceipt:
```suggestion
  query GetPrayerListFeature($featureId: ID!) {
```

packages/apollos-ui-prayer/src/PrayerCard/index.js
```diff
@@ -90,31 +91,54 @@ const PrayerCard = ({
...
-  </StyledCard>
-);
+  reportPrayer,
+}) => {
+  const { showActionSheetWithOptions } = useActionSheet();
+
+  return (
+    <StyledCard cardColor={cardColor} isLoading={isLoading}>
+      <TouchableOpacity
+        onLongPress={() => {
```
redreceipt: @conrad-vanl is there a clearer way you have in mind to bring up the action sheet or is this fine?

conrad-vanl: heh it's not bad per-say and it's what the crossroads-anywhere app does. It's kind of hidden functionality though.... Crossroads had a more apparent button previously and they found that it was tapped more often and prayers that shouldn't really be reported were being reported. At minimum, I'd rather use our `TouchableScale` component instead of opacity

conrad-vanl: Let's ask Liz on slack too if they have any design thoughts about reporting prayer functionality

redreceipt (approved):

camrun91:
@redreceipt will this work on 2.x apps? Wondering If I can pull this into ours before we launch?

redreceipt:
yes it should

PR #2287, opened 2021-12-06
Attempted upgrading passkit-generator to v3.1.0

alexandercerutti:
Hello there, I'm Alexander and I'm the creator of passkit-generator. Since you've been using my project as a dependency for your `apollo-data-connector-passes` for so long without updating it, and since (as you might have noticed) I've deprecated versions below `v2.0.6`, I've tried to perform some changes (the ones I was able to understand), to help you in migrating to the latest version of passkit-generator.

Of course, I wasn't able to test the changes and I wasn't able to complete all the changes to be done, but this is definitely a starting point 😄 . For this reason, I'm opening a draft PR, so you can see what I tried to do and complete it.

Passkit-generator v3.0 is way freer than v1.5.9, so I think your hacks won't be needed anymore, so I proceeded in removing them.

I've added a `// @TODO` because passkit-generator doesn't have anymore a way to download files since v2.0 and I didn't know which way you wanted to provide for downloading assets.

One last thing: I was able to see that, under `apollos-data-connector-passes/src/_snapshot_`, there's a plain text file (like a Jest output) with a private key in plain sight. I don't know if this is wanted or not and I don't know if that key is used somehow to issue Apple Wallet passes.

Hope this is appreciated!
Let me know!


alexandercerutti:
@vinnyjth Sad that this has been marked as stale and has been closed automatically.

I'll keep my branch with the changes until you'll be able to upgrade (I'm quite tired to see a lot of weekly downloads on v1.5.9 😄 - up to v2.0.8, passkit-generator kinda sucks eheh ). I think that if I'm going to delete it, you are going to lose the code I wrote (even if I don't think that would it be a big loss lol).

Just a question: how is passkit-generator used here? Just for being aware of the use-case. 😊

PR #2291, opened 2021-12-10
fix: cleans up tests

redreceipt:


packages/apollos-data-connector-rock/src/features/__tests__/index.js
```diff
@@ -125,7 +125,7 @@ describe('features', () => {
         PrayerRequest: {
           byDailyPrayerFeed: () => ({
             top: () => ({
-              get: () => ({}),
+              get: () => ({ then: () => null }),
```
vinnyjth, line 128: another option. more than fine either way 

```suggestion
              get: async () => null,
```

or 

```suggestion
              get: () => Promise.resolve(null),
```

vinnyjth (approved):

packages/apollos-data-connector-rock/src/features/__tests__/index.js
redreceipt: oh, nice. that's just good for my own understanding of JS promises.

PR #2293, opened 2021-12-13
fix: Add all phosphor icons

tomthornton:
This adds all of the phosphor icons with their original name in addition to the core icon mappings.


nlewis84 (approved):
Nice updates here!

PR #2297, opened 2021-12-17
Fix: cookie parsing  🍪

conrad-vanl:
Rock sends multiple `set-cookie` headers, they look something like:

```
Set-Cookie: .ROCK_SESSION_AFFINITYCORS=******; Path=/; SameSite=None; Secure
Set-Cookie: .ROCK_SESSION_AFFINITY=*****; Path=/
.ROCK=*****; path=/;SameSite=Lax;Secure; secure; HttpOnly
```

When we use `node-fetch` to retrieve the `set-cookie` headers, like this:

`response.headers.get('set-cookie');`

`node-fetch` appears to do a simple `.join(',')` on the multiple headers, so we end up with a cookie string that looks something like:

```
.ROCK_SESSION_AFFINITYCORS=******; Path=/; SameSite=None; Secure,
.ROCK_SESSION_AFFINITY=*****; Path=/,
.ROCK=*****; path=/;SameSite=Lax;Secure; secure; HttpOnly,
```

This is actually an invalid cookie, and by HTTP spec I think you'll ignore everything after the first semicolon. I'm not sure why this works pre web-farm in Rock - I'm not sure if they implemented the spec incorrectly OR the cookies were sent in a different order, but regardless, the proper way to join multiple cookies are with a semicolon, like this:

```
.ROCK_SESSION_AFFINITYCORS=******;
.ROCK_SESSION_AFFINITY=*****; 
.ROCK=*****;
```

NOTE: I'm not sure if we have to strip out everything past the first `;`, but this appears to be what POSTMAN does, and what the stack overflow article I link to below. Example from Postman after using the `/Login` endpoint:

Screen Shot 2021-12-17 at 11 56 49 AM

So, that's what this PR does. 

For reference, here are two resources I found to confirm this fix, and I have patched LCBC app locally and confirmed I can now login when pointed to their web farm:

- https://www.npmjs.com/package/node-fetch#extract-set-cookie-header
- https://stackoverflow.com/a/55680330


conrad-vanl:
@vinnyjth if you can help me get this reviewed and merged and released I can work on getting it rolled out for LCBC

packages/apollos-data-connector-rock/src/auth/data-source.js
```diff
@@ -85,7 +85,13 @@ export default class AuthDataSource extends RockApolloDataSource {
         })
       );
       if (response.status >= 400) throw new AuthenticationError();
-      const cookie = response.headers.get('set-cookie');
+
```
conrad-vanl, line 88:
```suggestion

```

packages/apollos-data-connector-rock/src/auth/data-source.js
```diff
@@ -85,7 +85,13 @@ export default class AuthDataSource extends RockApolloDataSource {
...
       );
       if (response.status >= 400) throw new AuthenticationError();
-      const cookie = response.headers.get('set-cookie');
+      
+      const rawCookies = response.headers.raw()['set-cookie'];
+      // mmmmm let's bake the cookies 🍪
+      // Cookies might have extra data in them past the ';', so we need to strip them out.
+      // This is a simplification of the function found in https://stackoverflow.com/a/55680330
+      const cookie = rawCookies.map((c) => c.split(';')[0]).join(';');
+
```
conrad-vanl, line 93:
```suggestion

```

redreceipt (approved):
login worked for me

vinnyjth (approved):

PR #2299, opened 2021-12-17
Fix: throw error if no Impersonation Token found

conrad-vanl:
This fixes generation of Impersonation Tokens on expired Rock cookies


vinnyjth (approved):
💥 💥 💥 💥 💥

PR #2303, opened 2021-12-30
feat: added a border to avatar notification dot

nlewis84:
This PR adds a border to Avatar Notification Dot that matches the background screen color. There is a RN bug with regards to views that have border-radius (it causes bleed around the edges, which is exacerbated when bright colors are used). We are actually also seeing artifacts from this in our buttons as well (Check out FNWA buttons that have backgrounds and borders that are the same color).

Here is an example of this bug:

## Solution
Add a second slightly larger view on top of the Notification Dot that has a transparent center and a border the color of the background screen. The transparent center allows the color of the Notification Dot to come through and the border covers the bleed from the Notification Dot below.

| Examples | Image|
| :---: | :---: |
| Before |  |
| After |  |
| After (w/ Color) |  |


redreceipt:
You double check it looks ok on Android? View layering behaves differently sometimes.

nlewis84:
Good thought! I'll check it out.

nlewis84:
Android looks good!

redreceipt (approved):

PR #2304, opened 2022-01-03
fix: adds TLS redis for bull jobs

redreceipt:


packages/apollos-server-core/src/index.js
```diff
@@ -229,7 +229,15 @@ export const createJobs = (data) => ({ app, context, dataSources }) => {
   };
 
   if (process.env.REDIS_URL) {
-    queues = createQueues(process.env.REDIS_URL);
+    queues = createQueues(process.env.REDIS_URL, {
+      ...(process.env.includes('rediss')
```
redreceipt, line 233:
```suggestion
      ...(process.env.REDIS_URL.includes('rediss')
```

packages/apollos-server-core/src/index.js
```diff
@@ -229,7 +229,15 @@ export const createJobs = (data) => ({ app, context, dataSources }) => {
   };
 
   if (process.env.REDIS_URL) {
-    queues = createQueues(process.env.REDIS_URL);
+    queues = createQueues(process.env.REDIS_URL, {
+      ...(process.env.REDIS_URL.includes('rediss')
```
nlewis84, line 233: Is that a typo on the string? `rediss` or `redis`

redreceipt: no that's right

Screen Shot 2022-01-03 at 12 47 25 PM

redreceipt: no that's right
Screen Shot 2022-01-03 at 12 48 03 PM

nlewis84: Gotcha

Screen Shot 2022-01-03 at 11 52 00 AM

packages/apollos-server-core/src/index.js
```diff
@@ -229,7 +229,15 @@ export const createJobs = (data) => ({ app, context, dataSources }) => {
...
+        ? {
+            tls: {
+              rejectUnauthorized: false,
```
nlewis84, line 236: Why do we want to set this option when that variable is `rediss` instead of `redis`? Looking at a lot of other Apollos apps, this variable usually starts with `redis`. I'm not sure what `tls` is....tried to find it in the Bull docs but don't see it.

redreceipt: It's a workaround we had to do for lots of places in the codebase when Heroku switched to using TLS. We just missed this one.

nlewis84 (approved):

PR #2308, opened 2022-01-03
fix: correctly spreads REDIS options

redreceipt:
I swear...


redreceipt:
@vinnyjth if you have any ideas on how i can test a verify these changes before doing all this, i'm all ears

vinnyjth:
@redreceipt can you grab one of the redis urls from the client apps (or from redis) and swap it into your local env temporarily?

redreceipt:
great idea. I don't speak ES-old so I couldn't paste in exactly but this shows what I'm doing in this PR and verifies it works

Screen Shot 2022-01-03 at 4 53 31 PM

Screen Shot 2022-01-03 at 4 55 07 PM

redreceipt:
so this one is ready

vinnyjth (approved):

PR #2314, opened 2022-01-06
chore: updated webview resolver

nlewis84:
This properly instantiates a webview feature. @vinnyjth I used this successfully on RiverValley. Any suggestions to make this inline with the other features above?


packages/apollos-data-connector-postgres/src/features/resolver.js
```diff
@@ -106,7 +106,9 @@ const resolver = {
     id: id('AddCommentFeature'),
   },
   WebviewFeature: {
+    height: (data) => data.dataValues.data.height || 400,
```
vinnyjth, line 109: I think you can skip the `.dataValues.data`? You normally want to access properties on the instance itself rather than digging into the underlying store.

vinnyjth, line 109:
```suggestion
    height: ({ data }) => data.height || 400,
```

vinnyjth (approved):

redreceipt:
What error are you seeing?

redreceipt:
can you show me a screenshot of a failing query in the graphql playground

camrun91:
If I remove the Height and URL and leave the ID in the Postgres resolver it works.  In the simulator however, the CSS does not work for the simple donation stuff. 

Screen Shot 2022-01-12 at 10 23 23 AM

redreceipt:
here you go :) https://github.com/ApollosProject/apollos-apps/pull/2320

redreceipt:
the CSS is busted and we don't have a fix for it yet

PR #2317, opened 2022-01-11
chore: makes HorizontalContentFeed functional

redreceipt:
a stepping stone to making loading states and performance better when clicking through content items
Kapture 2022-01-11 at 14 34 26


vinnyjth (approved):
Looks good to me :)

PR #2318, opened 2022-01-11
chore: move templates into apps pkg

redreceipt:
This PR represents a line in the sand. We no longer have the `apollos-templates` nor `apollos-upgrade` repo. New APIs will be added to the cluster, new apps will be forked from the templates directory here with the intention of implementing a CLI function that will do both of these things for you.

To test:

```
cd templates/mobile
yarn && yarn ios
```


redreceipt:
shoot, I just remembered why we can't do this...a fork is dependent on NPM releases. Ok, so we can still go the `next` branch route like we used to. Cluster work can use the `next` branch in templates.

redreceipt:
Sorry, I'm indecisive. Actually, as long as we don't push any NPM dependent code in the same PR, It will be ok. We can wait until a release to make changes to the template code. And that is getting rarer and rarer since there is hardly any code in the folder.

I think we could try it for a time low risk and see how it works.

camrun91:
Just seeing this... From the open-source side how will this change how I am currently using the templates repo?

redreceipt:
@camrun91 how are you currently using the templates repo?

camrun91:
@redreceipt our repo was made from the template and I am using the upgrade tool to upgrade core stuff in our repo right now. If this does not change that then that is fine.

redreceipt:
The plan is for the upgrade tool to be sunset. It doesn't work great as is and there are only four files left in the app template with plans to remove even those. The only files that will be left are native files that shouldn't change unless we do a react-native upgrade. Updates in the future will be simply updating package versions. What are your thoughts on this? What concerns do you have?

camrun91:
That could work. If you want to I would be more than willing to do a "Clean start" in my personal org and try and see how the process would work. I guess the question would be if another church wanted to use the API and app "Templates" as a starting point how would they do that?

camrun91:
> API is a question mark right now but we're hoping to open source our new version with multi-church support soon. When we do, both will be spun via the `apollos-cli`.

OK awesome. I probably have some repo clean up that I need to do anyways for things we have changed or modified. I will gladly line up with whatever the new process is.

redreceipt:
@vinnyjth how do you feel about this? I'd love to pull trigger so we can move forward with the next iteration if possible

vinnyjth:
@redreceipt I think this is too adjacent to the work we are tackling next cycle. Let's stick with keeping this repo separate, and then have a `next` branch for work pointing at the cluster.

vinnyjth (approved):

PR #2320, opened 2022-01-12
fix: sets defaults for webview feature

redreceipt:


nlewis84 (approved):
🚢

camrun91:
@redreceipt @nlewis84 There is still something odd going on here. This resolved the error however it is not pulling the URL through from the yml file. 
Screen Shot 2022-01-12 at 10 34 35 AM

packages/apollos-data-connector-postgres/src/features/resolver.js
```diff
@@ -106,9 +106,9 @@ const resolver = {
     id: id('AddCommentFeature'),
   },
   WebviewFeature: {
-    height: ({ data }) => data.height || 400,
+    height: ({ data }) => data?.height || 400,
     id: id('WebviewFeature'),
-    url: ({ data }) => data.url,
+    url: ({ data }) => data?.url ?? '',
```
vinnyjth, line 111:
```suggestion
    url: ({ data, url }) => data?.url || url || '',
```

redreceipt:
> There is still something odd going on here. This resolved the error however it is not pulling the URL through from the yml file.

@vinnyjth's commit probably fixed that

camrun91:
Doesn't look like it (I am making the changes to the node modules as you all are making them to test) 

Just to verify this is what the section of the config file looks like.
Screen Shot 2022-01-12 at 10 35 47 AM

camrun91:
@redreceipt @vinnyjth so if I change data to args it works.

packages/apollos-data-connector-postgres/src/features/resolver.js
```diff
@@ -106,9 +106,9 @@ const resolver = {
     id: id('AddCommentFeature'),
   },
   WebviewFeature: {
-    height: ({ data }) => data.height || 400,
+    height: ({ data }) => data?.height || 400,
     id: id('WebviewFeature'),
-    url: ({ data }) => data.url,
+    url: ({ data, url }) => data?.url || url || '',
```
vinnyjth, line 111:
```suggestion
    url: ({ data, args }) => data?.url || args.?url || '',
```

like this @camrun91 ?

camrun91: ` WebviewFeature: {
    height: ({ args }) => args?.height || 400,
    id: id('WebviewFeature'),
    url: ({ args, url }) => args?.url || url || '',
  },` 

that is what I currently have and it looks to be working. The .css file is not loading in the simulator but I don't think this is related.

camrun91:
So changing it to ` WebviewFeature: {
 height: ({ args, height }) => args?.height || height || 400, id: id('WebviewFeature'), url: ({ args, url }) => args?.url || url || '', 
  },` lets me get the height and the url. @vinnyjth @redreceipt

redreceipt:
If that's different than what we currently have, throw a PR at me

camrun91:
OK I will do it first thing in the morning.

camrun91:
@redreceipt #2321 there is some discussion needed prior to merging possibly.

PR #2321, opened 2022-01-13
fix: Webview props

camrun91:
This PR seems to be working for me. I am unsure why but in testing "data" was undefined but all values were set under args. 

The Title prop is working here as well even though it is not being handled by this resolver. Do we need to add the title prop here as well with a default of null?


redreceipt:
@vinnyjth do we need to support coming from DB and config here?

vinnyjth:
> @vinnyjth do we need to support coming from DB and config here?

Yah, so that would be both `args` and `data`.

`data.key || args.key`

redreceipt (approved):

PR #2324, opened 2022-01-18
chore: Adds create mobile app CLI command

redreceipt:
test by running `npm link` inside `packages/apollos-cli` and then running `apollos create mobile` somewhere else 🙌 

render1642484339985


vinnyjth:
I think only the first space is stripped :) 

image

vinnyjth:
🤔 

image

redreceipt:
yea looks like the three words are messing it up. I'll fix the regex.

redreceipt:
ok @vinnyjth should be good now

redreceipt:
also sorry for the no PR description at first. A GIF and testing instructions are up above now. It was late and I left it in "edit" mode 😬 .

vinnyjth:
All good! You can't fault yourself for not pushing code at 11pm on a national holiday 😂

PR #2327, opened 2022-01-18
fix: Try and improve secrets compatibility

vinnyjth:
image


redreceipt:
The problem is the default version doesn't have `pbkdf2`. So the user has to install the 1.1 version.

redreceipt:
At least that was the problem when we created the command. Or else there would be no to install anything as OpenSSL comes on everyone's machine by default.

vinnyjth:
> The problem is the default version doesn't have `pbkdf2`. So the user has to install the 1.1 version.

And this solution attempts to solve that problem by guiding the user to use the homebrew version (which does have `pbkdf2`)

vinnyjth:
It's `LibreSSL` that's the issue, which MacOS uses as a "drop in" replacement for OpenSSL. Any version of OpenSSL will work (at least any version we have identified)

vinnyjth:
I left in `which openssl` for Github actions, where we do need to use the default openssl.

vinnyjth:
@redreceipt Do you think we should swap `which openssl` for `/usr/local/opt/openssl@1.1/bin/openssl`? I think that's normally a symlink to the homebrew version but I'm open to suggestions

redreceipt:
I may be misunderstanding the code but I believe users are required to type `brew install OpenSSL@1.1` at some point. There's no way around it. So the paths we already have are the only options. We could simply check for both of those and if they don't exist, display `brew install OpenSSL@1.1` to the user?

`Brew install OpenSSL` will also install the wrong version.

vinnyjth:
On my system, I have `brew install openssl@1.1` but both paths listed don't exist.

vinnyjth:
> Brew install OpenSSL will also install the wrong version.

I don't think this is true? OpenSSL 3.0 works fine for me.

vinnyjth:
image

redreceipt:
Interesting. What do you get with `which openssl`?

vinnyjth:
@redreceipt  `/opt/homebrew/opt/openssl@3/bin/openssl`

vinnyjth:
@redreceipt I don't think there's anything wrong with any homebrew (or non homebrew!) installed version of OpenSSL. 

The only issue is `LibreSSL`, which is the default on MacOS and doesn't support the features we need.

redreceipt:
Oh cool! I don’t think v3 was out when when we built this thing.

redreceipt:
Instead of all the path logic, we could grep for LibreSSL and then prompt to install? I su that because `which openssl` for me returns the incorrect version.

conrad-vanl (approved):

vinnyjth:
@redreceipt I don't think that will be enough, because on the M1 MacOS, you still need to fix the path to get `OpenSSL` to be first. This PR will bypass the path and try to use a brew installed OpenSSL

vinnyjth:
@redreceipt I fixed my path by changing my .zshrc, but I think @conrad-vanl didn't so he was stuck on LibreSSL if he runs `openssl`.

conrad-vanl:
If `LibreSSL` is the problem, how do we detect that most easily? `openssl version` and if that points at LibreSSL, have instructions for installing openssl vs brew and make it the default? Is that better than trying to do that magically?

vinnyjth:
@redreceipt That's a good point. I added a catch for LibreSSL.

redreceipt (approved):

PR #2330, opened 2022-01-19
chore: updates bugsnag library to multitenant

redreceipt:


packages/apollos-bugsnag/src/index.js
```diff
@@ -1,18 +1,20 @@
...
-import ApollosConfig from '@apollosproject/config';
+import { fetchChurchConfig } from '@apollosproject/config';
 import { get, fromPairs } from 'lodash';
 import Bugsnag from '@bugsnag/js';
 
 const isDev =
   process.env.NODE_ENV !== 'production' && process.env.NODE_ENV !== 'test';
 
-if (ApollosConfig.BUGSNAG && ApollosConfig.BUGSNAG.API_KEY)
+const globalConfig = fetchChurchConfig({ churchSlug: 'global' });
```
vinnyjth, line 9: I think we can avoid needing to fetch the global config and instead use the `churchSlug`'s config. They will end up with the same variable, but it will be a bit more idiomatic.

vinnyjth: Hm, but this does need to be done on the root level. Ignore me :)

vinnyjth:
We've got data flowing into bugsnag now. I wonder how it's working? This PR looks good though.

vinnyjth (approved):

redreceipt:
I just moved it from server to npm. That's how it's currently working.

PR #2331, opened 2022-01-19
fix: Migrations order and improve logging

rkdavidson:
# About

Fixes the order of migrations, so they can be built up and torn down without errors.
Prior to this PR, starting from an empty database, some migrations would fail half way through.

Along the way, I added some logging to show the final order of migrations that will be run.
Example output:
```
📋 Migrations to be run:

Order #0
•  000-add-uuid-extension
•  001-add-pg_trgm-extension

Order #1
•  001-create-campus

Order #2
•  001-create-people

Order #3
•  001-add-otp-table
•  001-create-comments
•  001-create-content-items
•  001-create-follows
•  001-create-interaction
•  001-create-notification-preferences
•  001-create-notifications
•  001-create-prayers
•  001-create-refresh-tokens
•  001-create-tags
•  001-create-user-flags
•  001-create-user-likes
•  002-add-liked-count
•  003-change-profile-image-url-to-text

Order #4
•  001-create-content-items-connections
•  001-create-media
•  002-add-person-id
•  002-create-tag-junction_tables
•  002-rename-campus-fields
•  002-rename-follow-fields
•  002-rename-notification-fields
•  002-rename-notification-preference-fields
•  002-rename-prayer-table
•  002-update-user-likes-node-id-to-uuid
•  003-add-name-index
•  003-rename-comment-fields
```

Also add a graceful warning, instead of an uncaught `ENOENT` stack trace, when a church does not have a `.env` for the currently requested `ENV_STAGE`.

Those warnings look like this:
```
› DATABASE_URL=postgres://localhost/apollos ENV_STAGE=dev yarn migrator up
yarn run v1.22.17
$ babel-node ./src/migrator.js up
Warning: No environment config found for "crossroads_kids_club.dev.env"
Warning: No environment config found for "crossroads_tv.dev.env"
Warning: No environment config found for "eleven22.dev.env"
Warning: No environment config found for "willow_creek.dev.env"
Warning: No environment config found for "woodmen.dev.env"
```

# Tradeoffs/Caveats
I added a brief 2s pause after the order of migrations is logged, to create an opportunity to interrupt the process for human review.
Ideally, this would be some CLI arg or logging flag, but I didn't want to go down a rabbit hole there.

I also did not write or update any tests re: migrations order change, since I didn't want to go down _that_ rabbit hole, either. Happy to though, if someone doesn't mind helping point me in right direction.

# Testing
You can probably think of a better way, but emptying your local `apollos` database and then running `DATABASE_URL=postgres://localhost/apollos yarn migrator up` should complete successfully. Keep an eye out for the new order logging.

Also run `ENV_STAGE=foo yarn migrator up` to see the new warnings about missing env config file.


rkdavidson:
Couple random failing things that I didn't touch:

**apollos-data-connector-passes**
Looks like there are two copies of `resolvers.tests.js`. Haven't looked closely at any differences, but this feels off?
There's [a long URL property causing a failure](https://github.com/ApollosProject/apollos-apps/runs/4874649414?check_suite_focus=true#step:5:71). Trying to override/ignore via `// prettier-ignore`.



**Random test failures**
Not sure what to do with these [random and inconsistently failing tests](https://github.com/ApollosProject/apollos-apps/runs/4874649326?check_suite_focus=true#step:6:2495) 🤔 .

I'll hit re-run and hope for the best, but could use input from the team on how to proceed.
The failures look very unrelated to the PR changes.

rkdavidson:
Escalating my concern re: these failing tests and lint details — the [latest action jobs yielded even _more_ problems](https://github.com/ApollosProject/apollos-apps/runs/4874746621?check_suite_focus=true).
And you can see the scope of files touched in this PR is much smaller than where these problems are happening.

CleanShot_2022-01-19_17 24 40

Could it be something about how the `next` branch is setup? 🤔

redreceipt:
Yea we likely need to fix lint and tests on next, unrelated your PR

vinnyjth (approved):
Neat visualization! We should find a way to represent our migrations as a DAG, since that's how they end up being structured in practice :) (I am VERY kidding)

packages/apollos-data-connector-postgres/src/postgres/performMigrations.js
```diff
@@ -8,6 +8,20 @@ const createMigrationRunner = async ({ migrations, logger = console }) => {
...
+  );
+  migrationsGroupedByOrder.forEach(([order, migs]) => {
+    const sortedMigrations = migs.sort((a, b) => (a.name < b.name ? -1 : 1));
+    console.log(`\nOrder #${order}`);
+    console.log(sortedMigrations.map(({ name }) => `•  ${name}`).join('\n'));
+  });
+
+  console.log('\nPausing for review (2s)...');
+
+  await new Promise((resolve) => setTimeout(resolve, 2000));
```
vinnyjth, line 23:
```suggestion
  if (process.env.NODE_ENV !== 'test') await new Promise((resolve) => setTimeout(resolve, 2000));
```

packages/apollos-data-connector-postgres/src/postgres/performMigrations.js
```diff
@@ -8,6 +8,22 @@ const createMigrationRunner = async ({ migrations, logger = console }) => {
...
+  );
+  migrationsGroupedByOrder.forEach(([order, migs]) => {
+    const sortedMigrations = migs.sort((a, b) => (a.name < b.name ? -1 : 1));
+    console.log(`\nOrder #${order}`);
+    console.log(sortedMigrations.map(({ name }) => `•  ${name}`).join('\n'));
+  });
+
+  console.log('\nPausing for review (2s)...');
+
+  if (process.env.NODE_ENV !== 'test') {
```
vinnyjth, line 23:
```suggestion
  if (process.env.NODE_ENV !== 'test') {
```

packages/apollos-data-connector-postgres/src/postgres/performMigrations.js
rkdavidson: Ah thanks, should have been more surgical with that mandatory pause.

packages/apollos-data-connector-postgres/src/postgres/performMigrations.js
rkdavidson: I'll check my IDE settings, not sure why the trailing whitespace wasn't trimmed automatically 🤔 .
Thanks for fixing 🙇 .

PR #2332, opened 2022-01-20
chore: fix linter

redreceipt:


vinnyjth:
@redreceipt you might be able to go even farther :) Here are the folders that were in `connector-rock` when you moved the packages over from `cluster`

image


https://github.com/ApollosProject/apollos-cluster/tree/00f5d36d341c8d1666d4a5d34260c4a1e991b013/packages/apollos-data-connector-rock/src

vinnyjth (approved):
Looking great! Let's merge this as soon as it passes the linter.

PR #2333, opened 2022-01-20
chore: fix tests

redreceipt:


vinnyjth:
Anyway I can help with this?

redreceipt:
It’s a slog but I’m making my way through and learning a lot about a lot. If I get stuck though I’ll definitely ping you. I’ll commit as I go.

vinnyjth:
@redreceipt Sounds good. I'm surprised so many changes need to be made, since they were passing in `cluster`

redreceipt:
honestly, the most important change seems to be that `getSchema` in the test utils is now async, not sure how tests were ever passing on the cluster. I added the awaits in that function and made it technically async but should've already been there.

redreceipt:
nvm, that wasn't a big deal

redreceipt:
ok, @vinnyjth that was a fun one. take a peek at the changes and let me know if you have questions.

packages/apollos-data-connector-postgres/src/features/__tests__/dataSource.tests.js
```diff
@@ -1,43 +0,0 @@
-/* eslint-disable import/named, new-cap */
```
vinnyjth, line 1: Why was this test deleted?

vinnyjth (approved):

packages/apollos-data-connector-postgres/src/features/__tests__/dataSource.tests.js
redreceipt: I think because I couldn't figure out how to fix it quickly and we didn't have any other feature tests so I opted to skip for now.

Issue #2329, opened 2022-01-19
Prayer postgres model missing field


camrun91:
We currently allow congregates to submit a confidential prayer via rock. We are using the is public field of prayers for this. The Postgres database model is currently missing this. This is causing confidential prayers to show up in the app.

camrun91:
@conrad-vanl @redreceipt I am working on a PR for this. I will also look at the caching of the prayers as there seems to be an issue with that.

camrun91:
I will also look at a PR for the Shovel as well.

camrun91:
#2334

redreceipt:
Oh, I misread the PR. you just don't want them to show up. we can fix that in the shovel, I'm working in there right now. I'll link you once it's done.

camrun91:
@redreceipt I think for our initial release them not showing up is fine. It could be worth the conversation though, It may be neat in the future to allow our prayer team and or staff to see those in app. 

I also have two issues open in the shovel repo. I think I have them both fixed in our repo. If you want PR's for those let me know.

PR #2334, opened 2022-01-21
fix: Add isPublic column to prayers

camrun91:
The current model does not include ispublic flag from rock. We use this to allow people to submit "confidential" prayers. We don't want these to show up in the app. but having this flag could allow for us to let staff or pastoral staff see these prayers in the app.


camrun91:
@conrad-vanl @redreceipt Let me know if there are more changes needed for this. I have not actually tested this yet as we don't have a test database with prayers that are not being used for beta of our app.

redreceipt:
this would be a new feature as we currently don't support adding private prayers. but it's worth a discussion for sure.

redreceipt:
closing this as I misread the problem. see my comment in https://github.com/ApollosProject/apollos-apps/issues/2329

PR #2335, opened 2022-01-21
fix: removes incorrect prayer index

redreceipt:


packages/apollos-data-connector-postgres/src/multi-tenant/migrations/008_remove_incorrect_prayer_index.js
```diff
@@ -0,0 +1,14 @@
+function up({ context: queryInterface }) {
+  queryInterface.removeIndex('prayer_request', ['origin_id', 'origin_type']);
```
vinnyjth, line 2: will this suffer from the same issue as the last migration? should this be - `prayer` instead of `prayer_request`?

redreceipt: oh not sure. I thought that param was simply the table name. let me go read the docs real quick.

redreceipt: I _think_ this is right...

Screen Shot 2022-01-21 at 1 34 19 PM

redreceipt: I didn't test locally. let me do that first and I'll post here

vinnyjth: sounds good. I was about to, but go for it

vinnyjth: If you look at the previous migration, it was essentially the same as what you have above

```
await queryInterface.removeIndex(`${t}`, ['origin_id', 'origin_type']);
```

redreceipt: Yep. I figured out the right syntax locally. But trying to test the "down" its running one from somewhere else. Is that because its order is higher? If so, do you know how to run a specific "down" for testing purposes?

redreceipt:
UP

Screen Shot 2022-01-21 at 2 45 06 PM

DOWN

Screen Shot 2022-01-21 at 3 00 47 PM

vinnyjth (approved):

packages/apollos-data-connector-postgres/src/multi-tenant/migrations/008_remove_incorrect_prayer_index.js
```diff
@@ -0,0 +1,16 @@
...
+    name: 'prayer_origin_id_origin_type',
+    fields: ['origin_id', 'origin_type'],
+    unique: true,
+  });
+}
+
+const name = '008-remove-incorrect-prayer-index';
+
+module.exports = { up, down, name, order: 15 };
+
```
redreceipt, line 16:
```suggestion
```

PR #2346, opened 2022-01-25
feat: Authenticated TV/Device Linking

rkdavidson:
# 🚧  WIP 🚧  

[Basecamp](https://3.basecamp.com/3926363/buckets/25558802/todolists/4494594853)


redreceipt:
this is exciting

vinnyjth:
@rkdavidson  This PR is fantastic! Very thoughtful, straightforward, and well compartmentalized. Straightforward to process and review :) 

I do think there's some thinking we need to do about the input to `claimLinkCode`. Currently the input takes in a person object and and the attributes for an `openIdIdentity`. I think this approach works perfectly given the constraints of Crossroads, but I don't know if it makes sense outside that context for the following reasons
- When using in an Apollos App (like if an Apollos App is validating a OTP) that app won't have an OpenIDIdentity to provide, and shouldn't need to pass in all of the current user's details
- It feels weird to accept an OpenIDIdentity, given that in our other OpenID Endpoints we accept the OpenID type (which is configured to map types to OpenID providers)

Here's what I propose :) I think it will help reduce the amount of code needed for this feature (not that it's large at this point) and slot in better with the endpoints we provide for apollos. **We should have `claimLinkCode` take `code` as an input and use `Person.getCurrentUser` internally.**

From the perspective of a "linking client", the flow would be as follows. 

- User would be prompted to login via OpenID (Okta / Crossroads) or register. Hopefully In the same way as you were anticipating them providing their credentials
  - They could do this via `requestOpenIdRegister` (which will perform a login, if the user is already registered) or the normal `validateLogin`  endpoints. 
- Once a user has logged in / registered (if they are coming from an Apollos app, they will already be registered) the client can call `claimLinkCode` passing in the code and using the normal `token: Bearer` header. 

Essentially - I'm suggesting we delegate the OpenID connection logic to the code we already have for that purpose, and make the assumption that user claiming a link code will already have authenticated themselves via some mechanism. **It feels a bit out of place for `claimLinkCode` to linking a code to an account, AND registering a user and creating a OTP identity** 

😅 Hopefully that' helpful. I can jump on a call and talk over it more if it would be helpful.

vinnyjth (changes requested):
See comment :)

packages/apollos-data-connector-postgres/src/openid-identity/migrations/002_add_provider_session_id_column.js
```diff
@@ -0,0 +1,19 @@
+import { Sequelize } from 'sequelize';
+
+async function up({ context: queryInterface }) {
+  /*
+  Add a column for storing an external auth provider session ID.
+  This is important for Okta, but potentially others.
+  */
+  await queryInterface.addColumn('open_id_identity', 'provider_session_id', {
```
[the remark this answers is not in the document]
rkdavidson: Hey @liamillard, quick double check here — there's a `session id` value we need to track from Okta, yeah?
Or is it a token that includes the session ID in it?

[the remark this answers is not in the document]
liamillard: We should probably send two tokens from the crossroads API to the cluster - the access token _and_ the session id.

We don't technically need the session id for TV - it's only used for handing off someone from the app to the website and staying logged in.

But we wrote some hacky code to get us around okta rate limits when everyone opens their app at 7:16 in the morning after they get a notification that Brian Tome wrote a scripture journal, and that hack requires the session id to work. 

I don't think it's a deal-breaker if we don't want to deal with the two items, we'll just also have to tweak our refresh token endpoint on the Crossroads API

packages/apollos-data-connector-postgres/src/authentication/data-source.js
```diff
@@ -151,6 +158,143 @@ export default class AuthenticationDataSource extends PostgresDataSource {
...
+      const openIdIdentity = await this.context.dataSources.OpenIdIdentity.getFromId(
+        linkCode.openIdIdentityId
+      );
+
+      if (!openIdIdentity) {
+        result = 'USER_NOT_FOUND';
+      } else {
+        // N+1 🙈
+        const person = await this.context.dataSources.Person.getFromId(
+          openIdIdentity.personId
```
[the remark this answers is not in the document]
rkdavidson: Ok, I had no idea Sequelize was as cool as it is. I didn't know I missed SQL as much as I do until hopping on this cluster/multi-tenant work 😍 .

These magic methods are awesome. This feels Rails-ey-ish.
Thanks for calling these quick wins out, keep 'em coming!

packages/apollos-data-connector-postgres/src/openid-identity/migrations/002_add_provider_session_id_column.js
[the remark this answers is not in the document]
liamillard: Ah you're right - I should be more precise with my language 😅

The session id is an id.
The access token is a jwt.

packages/apollos-data-connector-postgres/src/otp/data-source.js
```diff
@@ -60,4 +65,74 @@ export default class OTPDataSource extends PostgresDataSource {
...
+        expiresAt: {
+          [Op.gt]: new Date(),
+        },
+      },
+    });
+  };
+
+  claimLinkCode = async ({ code, person }) => {
+    const otpRow = await this.model.findOne({
+      code,
```
vinnyjth, line 124: @rkdavidson I think you have the syntax a little goofy for `findOne` here :) what you have is good, but you'll need to wrap it in a `where` clause :)

vinnyjth (changes requested):

packages/apollos-data-connector-postgres/src/authentication/data-source.js
```diff
@@ -206,4 +292,19 @@ export default class AuthenticationDataSource extends PostgresDataSource {
     }
     throw new AuthenticationError('Must be logged in');
   };
+
+  createAuthenticatedPerson = async ({ person }) => {
```
vinnyjth, line 296: 💥 💥 💥

packages/apollos-data-schema/src/index.js
```diff
@@ -228,6 +234,33 @@ export const authenticationSchema = gql`
     email: String
     phone: String
   }
+
+  input AuthenticationRequestLinkCodeInput {
+    clientId: String!
+  }
+
+  input AuthenticationClaimLinkCodeInput {
+    code: String!
```
vinnyjth, line 243: What do do you think about naming this field `otp` so it matches the field `opt` in the `RequestLinkCodeAttempt` type?

packages/apollos-data-connector-postgres/src/otp/data-source.js
rkdavidson: Doh! 🤦‍♂️ Good catch... I hadn't tested yet with other OTP options in my database, so that's definitely on me. Sorry it derailed you a bit. Will fix in the morning, and add at least some coverage to the critical parts like this.

packages/apollos-data-schema/src/index.js
rkdavidson: Good catch. The `code` vs. `otp` naming consistency has been tricky to balance up and down the graphql schema and underlying datasource methods, then to the database — so wouldn't be surprised if I messed this up somewhere else, too.

rkdavidson: I went with `otp` to stay consistent with the other mutations and naming: https://github.com/ApollosProject/apollos-apps/pull/2346/commits/7ca6c82a4477237bef605c05a268a0ca5f60c6b6

vinnyjth (approved):
Great work @rkdavidson! Very excited to see this live.

vinnyjth:
@rkdavidson Good news and bad news! 

Good news - I rewrote the migrations strategy in our app and you don't need to scatter migrations all over anymore. They all belong in `postgres/migrations`. You don't need to add an import, or do anything but export an `up`/`down`. The filename just has to sort lower than the previous migrations filename (so name it `07X_migration_name`)

Bad news - You'll have to move your migrations :)

PR #2348, opened 2022-01-25
feat: adds dynamic tabs

redreceipt:
Screen Shot 2022-01-25 at 12 14 20 PM


redreceipt:
@conrad-vanl @vinnyjth this is close but getting this strange behavior, tabs are "sticking". I knew the tab navigator could present problems like this but my react native knowledge is lacking. Any ideas on what could cause this?

Kapture 2022-01-25 at 12 16 41

redreceipt:
it has to do with this, it's overwriting the home feature feed features in apollo cache with the prayer tab features, just don't know why...

Screen Shot 2022-01-25 at 3 57 38 PM

vinnyjth:
@redreceipt most likely the ids. check that you are .

- fetching the ids for feature feeds
- that those ids are distinct (and aren't just base64ed [object Object] or soemthing)

redreceipt:
I’m logging them and they are unique and look ok

redreceipt:
@vinnyjth is typename potentially also required because of how we store items in the cache?

conrad-vanl:
@redreceipt as much as I really would like to push this PR through, I think we should probably pause it (and not let willow creek hold up because of it). As long as we can easily "Add" tabs on the api layer, having the frontend be a little more static isn't horrible for now. I'd like to dig into it more during cooldown perhaps.

redreceipt:
@conrad-vanl no further work is required, this is ready for review. do you still want to wait to merge?

packages/apollos-ui-connected/src/TabsConnected/index.js
```diff
@@ -0,0 +1,52 @@
...
+      }
+    }
+  }
+`;
+
+const GET_TABS = gql`
+  query GetTabs($campusId: ID) {
+    tabs(campusId: $campusId) {
+      title
+      icon
```
conrad-vanl, line 24: Do we have error handling if the icon doesn't exist?

packages/apollos-ui-connected/src/TabsConnected/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+const TabsConnected = ({ hasCampus }) => {
+  const { data: campusData, loading, error } = useQuery(GET_USER_CAMPUS_ID);
+  const { data } = useQuery(GET_TABS, {
+    variables: {
+      campusId: hasCampus ? campusData?.currentUser?.profile?.campus?.id : null,
+    },
+    fetchPolicy: 'cache-and-network',
+  });
+  if (loading || error) return null;
```
conrad-vanl, line 40: So during loading and error states nothing is rendered? HMW improve that?

packages/apollos-ui-connected/src/TabsConnected/Tabs.js
```diff
@@ -0,0 +1,136 @@
...
+    });
+  }, [client]);
+  const tabNavs = useMemo(
+    () =>
+      tabs.map(({ title, icon, feed }) => ({
+        title,
+        icon,
+        component: createTab({
+          featureFeedId: feed.id,
+          screenOptions:
```
conrad-vanl, line 92: I think this should probably be `options` instead of `screenOptions`

packages/apollos-ui-connected/src/TabsConnected/Tabs.js
```diff
@@ -0,0 +1,136 @@
...
+            title === 'Home'
```
conrad-vanl, line 93: What if the "home" tab isn't called `home`? I think my biggest concern with this PR is losing easy configurability of tab headers/containers and such. 🤔 How would we put a search button on non home tabs?

conrad-vanl (changes requested):
Left some comments/thoughts. I'm still conflicted on this approach. I don't disagree about the idea of letting API tell the app which tabs should be shown, but overall this approach creates quite a bit of abstraction and loses some customizability. 

Part of me wonders if we should define a handful of tab "types" that have pre-configured things like the header bar, tab bar icon, and such. A little more restriction, a little less abstraction. But I also don't know if I'd like that either.

Edit: Answer my questions above and we can go from there, I'm not ruling out this approach yet. Just want to consider some of these angles, particularly how the tabs are configured: title, icon, navigation header, etc.

packages/apollos-ui-connected/src/TabsConnected/index.js
```diff
@@ -0,0 +1,52 @@
...
+      title
+      icon
+      feed {
+        id
+      }
+    }
+  }
+`;
+
+const TabsConnected = ({ hasCampus }) => {
```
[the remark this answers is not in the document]
redreceipt: prop overrides

packages/apollos-ui-connected/src/TabsConnected/index.js
redreceipt: activity indicator?

conrad-vanl: Activity indicator and <ErrorCard /> at minimum I think. It would be nice to show placeholder state though.... placeholder tabs and placeholder feed. This is a pretty big change at the root level that everyone will experience, and I want to make sure the resulting UX is solid.

packages/apollos-ui-connected/src/TabsConnected/index.js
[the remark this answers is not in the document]
conrad-vanl: Another example then I think of my worries on how this is abstracted. What if only some of the tabs are campus specific? What does `hasCampus` do exactly?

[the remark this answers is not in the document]
redreceipt: passes `campus` as an arg to the `tabs` query for caching purposes. but we've pretty much moved that to the back end now so that functionality could probably be deprecated

conrad-vanl:
I think I'd sum up my concerns like this:

Right now, the frontend completely defines _which_ tabs are shown and _how_ they are shown. The backend just defines what data/content is shown on each tab.

This PR attempts to move the _which_ to the API layer, but the _how_ is lost. There's some attempts to keep the _how_, such as a hard code of "home" overrides if the tab is "titled" `Home`. But a title is more of a "what" and not a _how_, and so I think the contract is lost. The "title" and "icon" moving to the API layer indicates to me that there might be a goal to move a lot of the configurability to the API-layer, but I think that can get messy too. 

HMW let the API cleanly define "what" tabs should be shown, and the frontend still has a solid handle on the "how"?

One idea that I mentioned above is creating tab types. Similar to what we do with different content item types.... then the frontend can cleanly control how tabs are presented based on its type. If we were to go this route, I'd want to question how these types would be taxonomied: are they named based on the "type" of content? (Similar to our content types). `Giving`, `Media`/`Watch`, etc. Or are they named based on how the tab is "presented" - such as expressing how the header is setup and such. 

Also, based on pain points we've had from our strict content type naming, I'm not saying I like this approach yet, I'm just trying to call out the potential cons I'm seeing to this PR and would love your thoughts @redreceipt

redreceipt:
@conrad-vanl Yep! All very valid. I think I have an idea that aligns with our current theming process and allows for complete customization while providing sane defaults. I give you `ApollosConfig.TABS` what do you think??

Group 1

packages/apollos-ui-connected/src/TabsConnected/index.js
redreceipt: added activity indicator. would love to work in loading states in a separate PR

packages/apollos-ui-connected/src/TabsConnected/Tabs.js
redreceipt: done

packages/apollos-ui-connected/src/TabsConnected/Tabs.js
redreceipt: solved

packages/apollos-ui-connected/src/TabsConnected/index.js
redreceipt: yes

redreceipt:
@conrad-vanl what do you think about my latest comment and changes here?

redreceipt:
Ping

redreceipt:
@conrad-vanl can you take a look at my latest comment and changes here? I'd really like to move template zero forward to get ready for scale cycle goals and this is an important step

conrad-vanl:
I'm sorry, I don't have the mental capacity / mind space to look at this today. I might be able to dig into it tonight or on Friday. @vinnyjth if you do have capacity, I trust your opinion here too and seeing whether my previous comment is properly addressed .

@redreceipt going forward if I'm slow to respond on GitHub ping me on slack. I don't always read my GitHub notifications very well

vinnyjth (approved):
I'm cool with this. 

I see @conrad-vanl's concerns, and while I don't think you've quite addressed them @redreceipt I think you've left the doors wide open to provide a more "packaged" experience using retool or other higher level configuration. 

I like the idea that our database has some pretty low level tools, but we're able to abstract them out into templated concepts using retool or scripts.

PR #2351, opened 2022-01-27
chore: Make required changes to get m1 building

vinnyjth:


vinnyjth:
@redreceipt can you pull this down and make sure you can still build?

vinnyjth:
@redreceipt can you pull this down and make sure you can still build?

redreceipt:
it built, but it changed some stuff for me which won't work for our deploy process. I'll commit the changes and you can run it again and see if it changes back.

Screen Shot 2022-01-27 at 10 32 01 AM

vinnyjth:
@redreceipt The react/react-core change should be fine (I had made a manual change to that file that is getting wiped out, which is good) 

The excluded archs might be troublesome. We need the excluded arch to always be whatever you're _not_ building for... I'll pull down and see what happens.

vinnyjth:
It runs fine, but I get this diff. 

image

What if we keep what you have commited, and make sure we don't commit the change that I'm seeing on `m1` (to keep fastlane working)? I could even write a test for that.

vinnyjth:
I hate doing that, but using Rosetta is such a nightmare sometimes.

redreceipt:
I’m good with that. The CI machines are x86 so if it does get committed, it’ll be easy to spot to failure. And if it gets annoying enough, I’m sure we’ll come up with a way around.

redreceipt (approved):

PR #2356, opened 2022-01-27
fix: my prayers bug and clean up action algo

vinnyjth:
Moves the "my prayer" logic so it's inside an action algorithm.
Defaults VerticalPrayer list to the new action algorithm (to maintain backwards compat)
Fixes a bug where viewing this feature would log the user out. 

image


redreceipt:
this should go to master right?

vinnyjth:
@redreceipt It relies on the new auth code, so no. We could backport part of this to master (the decoupling of CURRENT_PERSON_PRAYERS) but I don't see a need, unless we want to use the VerticalPrayerList for non-user prayers.

redreceipt:
is there not a bug fix here that would apply to the master branch (and v2 apps)?

vinnyjth:
@redreceipt there is no bug in this PR that applies to `master`. 

The bug was occuring because we were calling `Auth.getCurrentPerson`, which is incorrect in `next` and correct in `master`.

vinnyjth:
I made the slight refactor in this PR as well, because it felt wrong to edit that slice of code without tidying up a little :)

redreceipt:
so is there a reason you couldn't just delete the old algorithm? this will be on 3.0 where breaking changes are ok. plus it's not really a breaking change because everything happens server side

vinnyjth:
> so is there a reason you couldn't just delete the old algorithm? this will be on 3.0 where breaking changes are ok. plus it's not really a breaking change because everything happens server side

Do you mean remove the default behavior and require you manually specify the CURRENT_PERSON_PRAYERS algorithm?

redreceipt:
Makes sense to me to have them both be defaults for their respective features. use a boolean flag on the algorithm specifying currentUser. and then have one algo called `PRAYER_FEED`.

vinnyjth:
> Makes sense to me to have them both be defaults for their respective features. use a boolean flag on the algorithm specifying currentUser. and then have one algo called `PRAYER_FEED`.

I don't think defaults are a good idea for the features, I would rather have explicit definitions for what's happening. So if we move in one direction, I think it would be to remove the default behavior of `VerticalPrayerList` 

I'm okay with combining the algos. What would you call the boolean flag?

redreceipt:
`useCurrentUser`

vinnyjth:
Still working same as before. I did have to update the cluster config. image

redreceipt (approved):
love it. thanks for indulging :)

PR #2357, opened 2022-01-27
fix: add body to prayer notification

### LINKED ISSUE #2355: Postgres prayer request notifications not being sent.
@redreceipt @vinnyjth I am getting a 400 bad response error from onesignal when I submit a prayer request and the notification does not get sent. 

I think I have tracked it down. I am not sure what the best fix would be. I will post a few links in just a second.

camrun91:
This just moves part of the notification from the header to the body parameter. The send notifications was failing with body being empty here. 

closes #2355


redreceipt:
@camrun91 could you post a screenshot of what this notification looks like so we can verify the design?

camrun91:
@redreceipt will one with it in notification center work?

redreceipt:
hard to pull the trigger quickly unless we know what it will look like using the body instead of the title. One inside and outside the app would get this merged real quick :)

camrun91:
@redreceipt Here are 3 images 
Home screen 
IMG_4057

apple watch:
IMG_4054

and the backed of onesignal how it comes through 
Screen Shot 2022-01-27 at 2 06 23 PM

camrun91:
> 

It shouldn't change anything inside the app, should it? I am not changing the prayer request just the notification.

vinnyjth:
This looks good to me. I'm not sure why it was broken. @redreceipt I'll let you do the final review.

camrun91:
@redreceipt I just verified that in the Rock datasource we are passing through a content parameter directly to the onsignal datasource. So that should still work.

redreceipt:
so is this PR still valid?

redreceipt:
> @redreceipt I just verified that in the Rock datasource we are passing through a content parameter directly to the onsignal datasource. So that should still work.

Oh I see. You're saying it should still work w/o the DB too.

redreceipt (approved):
Great work!

PR #2363, opened 2022-01-28
feat: Added "register with openid " endpoint

vinnyjth:
Adds a new `graphql` endpoint, `requestOpenIdRegister`. 

This endpoint is the "connectOpenId" equivalent for _logged out_ users. It consumes an openId token (from the frontend openId flow) and then either creates a new user for a given set of credentials, or updates an existing user with a matching openId with the new credentials. 

It then returns a refresh token and access token for that existing (or new) user. 

You can test by visiting the authorization URL from the first query, and plugging it into the `registerWithCode` in the second query. 

```
query oid {
  openIdProviders {
    authorizationUrl
    providerType
  }
}

mutation registerWithCode {
  requestOpenIdRegister(code: {code here}, providerType:"rock"){
    accessToken
    refreshToken
  }
}
```

I also snuck a tweak into the config datasource so you can specify `ENV_STAGING=staging`  instead of the current `ENV_STAGING=.staging`


redreceipt:
looks good to me. mutation should probably be private scoped right?

redreceipt:
oh wait, nvm

redreceipt (approved):

vinnyjth:
> oh wait, nvm

Yah, we're good since it's a mutation :)

PR #2364, opened 2022-01-28
chore: Lockdown prod a teensy bit more

vinnyjth:
Forces you to use a `ENV_STAGE`, which is already set in heroku, and I will add a default to `staging` in the `cluster` repo. 

This isn't foolproof, but light years ahead of our current default.


vinnyjth:
This has an extra commit I need to cherry pick out.... hold on

vinnyjth:
https://github.com/ApollosProject/apollos-cluster/pull/114

conrad-vanl (approved):

PR #2366, opened 2022-01-28
chore: moves all customization to single file

redreceipt:


redreceipt:
I can't test fully right now because the server seems to be down and I'm not sure how to fix

vinnyjth:
This seems great to me. Keep in mind that you might run into issues updating some apps (like Newspring) without being able to control `typemaps`.

vinnyjth (approved):

redreceipt:
Noted. We wouldn't touch the template file without upgrading them to a new version so I think we're safe there.

vinnyjth:
@redreceipt! I and I believe the fact we were overriding `typemaps` was a code smell. And now that I think of it, you _MIGHT_ have fixed the need to do so in a loading state PR of yours.

redreceipt:
Yes it’s not anywhere in NPM anymore. This was my master plan :)

templates/mobile/theme.js
```diff
@@ -0,0 +1,23 @@
+import ApollosConfig from '@apollosproject/config';
+import Svg, { Path } from 'react-native-svg';
+import FRAGMENTS from '@apollosproject/ui-fragments';
+import { makeIcon } from '@apollosproject/ui-kit';
+
+const THEME = {
+  colors: {},
```
conrad-vanl, line 7: Just to call out I think this PR might be a step backwards in documentation....the old theme.js file at least had a lot of documentation on what you could do here, and now that's call gone with no replacement

redreceipt: I agree we need it somewhere. I think explicitly listing the defaults here serves the developer well but prop overrides documentation should probably be in markdown on the site. Our own developers have had trouble understanding this functionality in spite of the previous code comments.

vinnyjth: I agree with both of you here. 

I think losing the documentation is a step backwards, but I also never found the included documentation helpful for actually changing the theme (was always better of looking at examples in other apps). 

example
```
/* Base overlays. These are used as configuration for LinearGradients across the app */
// export const overlays = () => ({});
```
Tells me nothing about how to actually add an overlay. But I do think showcasing that this was an option was a good start.

PR #2374, opened 2022-02-01
feat: Connect identity

vinnyjth:
Adds endpoints for connecting an `identity` to a person. This replaces the ability to update your email/password via `updateProfileFields`

Mutations to test (you'll want to swap in your own values)
```
mutation requestConnectPhone {
  requestConnectIdentity(identity:{ phone: "5133061126" }) {
    result
  }
}

mutation conirmConnectPhone {
  connectIdentity(identity:{ phone: "5133061126" }, otp:"177092"){
    person {
      email
      phone
    }
  }
}

### CODE REVIEW

File: packages/apollos-data-connector-postgres/src/people/dataSource.js [hunk 1]
```diff
@@ -43,6 +43,10 @@ export default class Person extends PostgresDataSource {
 
     const profileFields = fieldsAsObject(fields);
 
+    // Don't allow users to update their phone number of email
```
redreceipt, line 46: what about for non-rock churches?

vinnyjth: They will use the new `requestConnectIdentity` functionality. This code is unrelated to where a user comes from (rock or non-rock). We don't want users to be able to update their phone or email without first confirming they own it.

redreceipt:
probably doing something wrong but I got a token from the validate login mutation but using it is still giving an auth error here

Screen Shot 2022-02-01 at 3 08 21 PM

vinnyjth:
@redreceipt You need to prefix your token with `Bearer  `

redreceipt:
observation, maybe not a big deal, the OTP I was sent was the same for validating login and validating open ID

redreceipt:
if anything, as a user it kinda seems sketchy to get the same one

vinnyjth:
> observation, maybe not a big deal, the OTP I was sent was the same for validating login and validating open ID

That's a good callout. In reality you wouldn't be doing both (if you logged in via phone number, you wouldn't need to reconfirm that same phone number).

redreceipt:
begs the question then why do we need both mutations? I guess I don't fully understand the flow

vinnyjth:
Because some users will be logging in via OpenID and then attach their phone number or email at a later step. 


Here's this flow for context: https://www.figma.com/proto/8MJZo4CDGHJRw5oryMnzGB/(Apollos)-New-Auth?node-id=234%3A4108&scaling=scale-down&page-id=234%3A4095&starting-point-node-id=234%3A4249

redreceipt:
got it. so the mutations in this PR only happen if they didn't log in via phone the first time and would like to attach their phone number?

redreceipt:
and also, should we not maintain the ability to update the phone number on the rock profile? let's say I sign in using my password via OpenID but then attach my phone number. and it's different than the (incorrect) phone number already on my rock profile? or are we not concerning ourselves with incorrect Rock data?

vinnyjth:
> got it. so the mutations in this PR only happen if they didn't log in via phone the first time and would like to attach their phone number?

Correct.

vinnyjth:
> and also, should we not maintain the ability to update the phone number on the rock profile? let's say I sign in using my password via OpenID but then attach my phone number. and it's different than the (incorrect) phone number already on my rock profile? or are we not concerning ourselves with incorrect Rock data?

We are not concerning ourselves with incorrect Rock data (@conrad-vanl can correct me if I'm wrong). For now ...

- We aren't pulling in Rock Profile via the Shovel
- The only way we will get your rock profile, will be via Rock's OpenID client
- We won't be doing writebacks by default.

redreceipt (approved):

PR #2383, opened 2022-02-03
feat: adds campus tags to content feeds

redreceipt:
simply looks for a tag with the campus name string


redreceipt:
@vinnyjth looks like we dont need the M1 check anymore with rn67. the test was failing for me but building the app didn't update anything

vinnyjth (approved):

PR #2388, opened 2022-02-03
feat: Setup content without tags to include everyone

vinnyjth:
Best way to test is to login on core. As Male under 35, I should _not_ see the "A Daily Reading" content item. 

**NOTE**: To use query below, you'll need to move this block of code into lines 167-169 (to simulate providing a category id) 

```
              [Op.or]: [
                { '$tags.people.id$': { [Op.not]: null } },
                { '$tags.id$': { [Op.is]: null } },
              ],
```
```
query persona {
  personaFeed(first:20000) {
    edges {
      node {
        id
        title
      }
    }
  }
}
```


packages/apollos-data-connector-postgres/src/content-items/dataSource.js
```diff
@@ -39,6 +39,8 @@ class ContentItemDataSource extends PostgresDataSource {
       limit,
     });
 
+    console.log(JSON.stringify(result, null, 2), findFunc);
```
vinnyjth, line 42:
```suggestion
```

vinnyjth:
Note: I would like to sort the items _with_ tags to the surface, but that's going to take either breaking this out into a sql statement or some more digging in sequelize.

nlewis84:
Sorry, it took me a second to get used to the new cluster testing environment for core. That query returns all three items that I would expect for me.

Screen Shot 2022-02-08 at 2 50 30 PM

nlewis84 (approved):

PR #2393, opened 2022-02-07
feat: V2 of ui-authentication Flow

vinnyjth:
As a scaffold, this is pretty complete @conrad-vanl. I think I've hit all the major points of your Figma file (but will need to go back and double check) 

I would liken this PR as being in a similar state as the last major `ui-authentication` PR. most of the major elements are present, but expect UX oddities and bad error states. Call stuff out as you see it so I can make sure I track and fix it.


vinnyjth:
(failure at the end of the video was due to my already having a user with that phone number)

conrad-vanl:
@vinnyjth this is looking great. I plan to get into this at some point today.

conrad-vanl:
I think this is looking good. A couple of thoughts:

1. I think we should get rid of the older OpenID screen for now. There's now this weird flow that if you say you don't have a rock.apollos.app account, then sign in, you are asked again to connect with your online account, which you already said you don't have. 
2. I'd really like tapping "No, let's create one" open up the Rock web page still, but directed at the "Register" screen. If that's not possible, I think I'd like to re-think the copy on the AuthLanding page to only have one button that opens the Rock web login, and a Skip button at the top that skips.

packages/apollos-ui-authentication/src/OpenID/OpenID.js
```diff
@@ -1,41 +1,105 @@
...
-            {`To complete your profile and access all features, please connect with your ${
-              providerFriendlyName || 'online'
-            } account.`}
-          </PromptText>
-          {error && (
-            <ErrorCard
-              message={
-                'Something went wrong authenticating you. You can skip, or try again.'
+    <>
+      <WebView
```
vinnyjth, line 63: @conrad-vanl I wanted to chime in that I really like this strategy for the backdrop. Simple and effective.

vinnyjth:
@conrad-vanl Can I get your approval (or review) on this PR?

vinnyjth:
@conrad-vanl bump :)

PR #2395, opened 2022-02-10
feat: Add totalCount to Search

jordanwade:
Adds totalCount to Algolia search connector for pagination.

```
query search($query: String!, $first: Int, $after: String) {
  search(query: $query, first: $first, after: $after) {
      totalCount
      pageInfo {
        startCursor
        endCursor
      }
      edges {
        title
        summary
        coverImage {
          name
          sources {
            uri
          }
        }
        cursor
        node {
          ... on ContentItem {
            id
            __typename
          }
        }
      }
  }
}
```

image


vinnyjth (approved):
This all makes sense to me. Thanks for throwing this together @jordanwade!

jordanwade:
> I ask because `nbHits` makes it sound like it's just the paginated results for which we wouldn't need a query for.

@redreceipt nbHits is "The number of hits matched by the query." from Algolia. Is that what you were hoping?
image

Also, This is how we handled "loading more" functionality in CF group search. So, hopefully they won't be sending emails 🤞

jordanwade:
30

PR #2400, opened 2022-02-16
feat: Add 'TV' to tabFeedFeatures enum

jordanwade:
Adds `TV` to the `tabFeedFeatures` enum so TV apps can have a custom configured TV feature feed.


redreceipt (approved):

redreceipt:
@jordanwade this PR will just need to be moved to the cluster repo. all API code lives there now.

PR #2405, opened 2022-02-21
feat: added OPEN_AUTHENTICATED_URL_EXTERNALLY as an action

nlewis84:
This PR merges a change that is in master to next.


nlewis84:
Closing this, so that we can just merge Master into Next.

PR #2407, opened 2022-02-21
fix: Allow datasources to pass sources array

vinnyjth:
image


nlewis84 (approved):
Ah, yep....I see the null sources. Good find!

PR #2409, opened 2022-02-21
feat: Better persona feed

vinnyjth:
This persona feed sorts items that you have tags for to the front of the feed, then puts all other items after. If there's a match on the number of matching tags, it sorts based on the `publishAt` date of the content item. 

I always hate breaking queries out of sequelize and into straight sql, but I don't think sequelize could handle this in their query builder.


nlewis84 (approved):
I have less experience with Sequelize, so this at least makes total sense to me. Writing quality SQL queries is a super power 💪

PR #2410, opened 2022-02-21
chore: master > next

redreceipt:


redreceipt:
@nlewis84 periodically this is what we have to do. Unfortunately, there are conflicts so pushing straight to origin could lead to broken tests. So manually, I will have to:

1. create a branch based on next
2. merge in master
3. create a PR pointed at next so we can catch broken tests

nlewis84:
👍 👍 I got onto other things, so I didn't even look at merging these yet.

PR #2415, opened 2022-02-24
chore: tracks media plays

redreceipt:
tracks media plays


redreceipt:
@conrad-vanl what do you think here for a less but better media tracking solution to solve Newspring request

PR #2418, opened 2022-02-25
feat: onPlay, onPause, and onEnd side effects for the media player

liamillard:
This PR adds three new side effect props to the video player: onPlay, onPause, and onEnd. The intent is to use the side effects to trigger analytics events, but they could of course be used for other things as well.

When I first started this PR, I attempted to trigger these functions in the `play` and `pause` functions (the ones that use a `React.useCallback`). But I was having a lot of trouble with passing the `playhead` into these functions and actually getting a correct value for `elapsedTime`. (I did add the `playhead` to the list of dependencies, but it still wasn't working). So I ended up making the new functions all trigger based off the `isPlaying` state variable in a `useEffect`.

I haven't been able to get the template app running locally, so I developed this in storybook exclusively. I added the new props to storybook for that reason. I think this will work in the larger apps, but I'd love for the reviewer to test that out.

A couple things you should go out of your way to test:

- set `autoplay` to true, and verify the `onPlay` callback is called on mount
- Finish a video (seeking to close to the end is totally okay), and verify that the `onEnd` function is called
- play/pause a video multiple times and verify that the `elapsedTime` and `totalDuration` appear to be correct.


packages/apollos-ui-media-player/src/Container.tsx
```diff
@@ -23,11 +23,22 @@ import useSourceComponents from './useSourceComponents';
...
 
   /** The Player Component. Defaults to FullscreenSlidingPlayer */
   PlayerComponent?: React.FunctionComponent;
 }
 
+const usePrevious = (value: any) => {
+  const ref = React.useRef(value);
+  React.useEffect(() => {
+    ref.current = value;
+  });
```
vinnyjth, line 38:
```suggestion
  }, [value]);
```
Should `value` be an argument here?

redreceipt:
Also I'll relay this over from a DM, we shouldn't be using "segment" in variable names or concepts. We should be service agnostic.

packages/apollos-ui-media-player/src/Container.tsx
```diff
@@ -151,6 +166,36 @@ const Container: React.FunctionComponent<ContainerProps> = ({
...
+        onPlay({ sessionId, elapsedTime, totalDuration });
+      }
+      return () => onEnd({ sessionId, elapsedTime, totalDuration });
+    },
+    [] // eslint-disable-line react-hooks/exhaustive-deps
+  );
+
+  React.useEffect(() => {
+    const { elapsedTime, totalDuration } = playhead;
+    const videoEnded = totalDuration - elapsedTime < 0.2;
```
vinnyjth, line 189: I'm a little worried about this logic in the context of livestreams, where the elapsed time is often equal the total duration. 

What do you think instead about passing this method `onEnd` down to the `PlayerComponent` and having the player component fire the `onEnd` callback https://github.com/react-native-video/react-native-video#onend 

Other players (like the webview player) will have to implement `onEnd` as well :)

vinnyjth: /cc @conrad-vanl

liamillard:
@vinnyjth oooh definitely didn't think about livestreams 😬 I'll see what I can do to instead use the onEnd exposed by react-native-video.

@redreceipt should I rename the variable or do you think we shouldn't be doing this in core at all? (determining something that could function as a session id, or player id)

vinnyjth:
> @redreceipt should I rename the variable or do you think we shouldn't be doing this in core at all? (determining something that could function as a session id, or player id)

@liamillard The "apollosy" way to handle would be for the for variable to be named something like "analyticsMeta" that then gets shoveled `{ ...analyticsMeta }` into each call :)

packages/apollos-ui-media-player/src/Container.tsx
vinnyjth: @liamillard `WebVideo` doesn't expose _any_ events right now (and likely can't) so I think you're okay here :)

packages/apollos-ui-media-player/src/Container.tsx
```diff
@@ -151,6 +166,35 @@ const Container: React.FunctionComponent<ContainerProps> = ({
     ControlsComponent,
   });
 
+  // we need a session id for some analytics events. The session ID should be
+  // created on mount, and different for each video player
+  const { current: sessionId } = React.useRef<string>(Date.now().toString());
+  const analyticsMeta = { sessionId };
+
+  // ---------
+  // setup onEnd effect on unmount
```
vinnyjth, line 175: 💥  Super smart to include this :)

liamillard:
@conrad-vanl @vinnyjth @redreceipt I'm bringing this one back, since we'll need it to hook up analytics in the KC mobile app. I _think_ I responded to all feedback, but happy to make other changes!

redreceipt:
leaving this one for @vinnyjth or @conrad-vanl as their opinions and feedback on this section of code is much more valuable

vinnyjth (approved):
This looks good to me 👍

PR #2435, opened 2022-03-03
chore: add actionsheet as prayer pkg dependency

redreceipt:


vinnyjth:
@redreceipt do we want to move this into peer deps? We don't need to, but it will give us a bit more flex on the client side.

redreceipt:
Motivation for this PR was me upgrading just the UI prayer pkg to pick up a bug fix and actionsheet was missing. Thinking about it though, and tell me what you think, since it technically requires the provider from UIKit , maybe this PR should be adding UIKit peer del that has the actionsheet provider?

vinnyjth:
Yah - That seems more accurate. Or peer dep on both to be honest.

PR #2438, opened 2022-03-04
feat: Updated Button url logic

nlewis84:
This PR will add functionality to our Button Features in core.

COMPLETE_NODE action will be honored
Otherwise, it will open Auth'd in the In App browser

https://user-images.githubusercontent.com/72768221/157957760-9f3cef74-6152-476e-8a69-51803ed2a08e.mp4


redreceipt:
I don't love this because it attaches the token to every link regardless of what it is. Which just looks goofy and like it's a bug.

PR #2441, opened 2022-03-11
chore: removes incorrect navigation title

redreceipt:
currently navigating to a feed uses the previous action's title. this isn't right. We can use feature titles to display what's in the feed

Group 1


packages/apollos-ui-connected/src/FeaturesFeedTabConnected/FeaturesFeedTabConnected.js
```diff
@@ -77,9 +77,10 @@ export const createFeatureFeedTab = ({
       <TabStack.Screen
         component={FeaturesFeedConnected}
         name="FeatureFeed"
-        options={({ route }) => ({
-          title: route?.params?.title || 'Feed',
-        })}
+        options={{
+          title: '',
```
nlewis84, line 81: Does this title have to be an empty string, or could we just set headerLargeTitle to false?

redreceipt: if you omit, it sets to "FeatureFeed" (name of the navigated component)

nlewis84 (approved):
Looks good

PR #2442, opened 2022-03-14
chore: added devDependencies for v3

nlewis84:
@redreceipt These are the packages that @tristanguckenberger and I had to upgrade in order to get working builds of Oak Cliff and City First.


redreceipt:
the problem was the app wasn't booting?

nlewis84:
Yes, it wouldn't boot up.

redreceipt:
the only thing this PR does is add babel plugins. the current template app boots fine. seems like it could have been something else causing problems.

nlewis84:
Hmmm, this is the only thing that I changed on Oak Cliff and it went from not working to working. Same for Tristan. I actually went through all of the error messages one by one and ended up with the same packages getting updated as Tristan got seperately.

redreceipt (approved):
interesting. part of me thinks those plugins were getting defined via API pkg somewhere and when we pulled those out, they went missing. adding doesn't hurt.

PR #2443, opened 2022-03-15
fix: uses correct youtube URL regex to detect player

redreceipt:


vinnyjth (approved):
well done! nice to have snaps

PR #2446, opened 2022-03-22
fix: keyboard overlay on auth.

camrun91:
I found that changing to the base KeyboardAvoidingView kept this from happening in v2.42 I found that the same was true here unable to test android as it will not run past the initial screen.  There are several other places that the FlexedKeyboardAcoidingView is used: The auth screen is the only screen we have seen the issue with the keyboard overlaying the content.


camrun91:
@redreceipt if we are good with this then we will need to push this to v2 as well.

nlewis84:
Would the Search modal be another place they would see this?

redreceipt:
could you describe the problem in the description, not really sure what this fixes

camrun91:
> could you describe the problem in the description, not really sure what this fixes

Yeah I will copy from Basecamp and tag you there also. 

When adding a new account and you open the keyboard the name/email content is not visible as it get pushed higher on the screen. One of our phones we couldn’t get to the content unless the keyboard is closed (but then you can’t enter content) and one of our phones the content section was at the top and only could get to parts of it by scrolling the block. However, the rest of the page is blank so the block shouldn’t be so high.

camrun91:
@redreceipt here is a screen record of the bug. I will do one of the fix in just a second. 

https://user-images.githubusercontent.com/16566029/159724844-770038c8-dde7-454d-8936-772bd402a121.mp4

camrun91:
closing to put these changes to v2

PR #2447, opened 2022-03-22
fix: grabs correct keychain for ios deploys

redreceipt:
- fix: creates keychain for deploys on
- keychain


templates/mobile/ios/Podfile.lock
```diff
@@ -701,14 +701,14 @@ SPEC CHECKSUMS:
   Flipper-RSocket: d9d9ade67cbecf6ac10730304bf5607266dd2541
   FlipperKit: d8d346844eca5d9120c17d441a2f38596e8ed2b9
   fmt: ff9d55029c625d3757ed641535fd4a75fedc7ce9
-  glog: 5337263514dd6f09803962437687240c5dc39aa4
+  glog: 85ecdd10ee8d8ec362ef519a6a45ff9aa27b2e85
```
nlewis84, line 704: Should this file have changes?

redreceipt: We have no way of knowing which is correct because we haven't done deploys in awhile. But that's what happens when I run yarn install in the template. There's no harm right now.

nlewis84 (approved):

PR #2448, opened 2022-03-23
chore: better deploy error messages

redreceipt:


packages/apollos-cli/scripts/deploy-ios.sh
```diff
@@ -12,8 +12,8 @@ VERSION=$(fastlane run get_version_number xcodeproj:ios target:$TARGET |
 
 if [ "$TRACK" = "internal" ]; then
   fastlane run setup_ci
-  KEYCHAIN=$(security list-keychains -d user | sed -E "s/.*Keychains\/(.*)-.*/\1/")
-  fastlane match appstore -a "$APP_ID,$APP_ID.OneSignalNotificationServiceExtension" keychain_name:"$KEYCHAIN" --readonly
+  fastlane run unlock_keychain path:fastlane_tmp_keychain
```
nlewis84, line 15: Does the temp keychain not have a password and that is why we don't have to pass that parameter?

redreceipt: That's a good question :-) I think it's already unlocked and all this does is added to the search list. There's really no way to test it outside of a CI system. I know these PR's are repetitive and annoying, I'm annoyed too.

PR #2449, opened 2022-03-23
fix: V2 Auth screen bug

camrun91:
Here is the bug that we had found: The keybord causes the content to be pushed up and become scrollable. This is worse on some devices. 

https://user-images.githubusercontent.com/16566029/159742153-31c53502-c951-43a6-932e-37b9d25b701d.mp4

and here is it after these changes:

https://user-images.githubusercontent.com/16566029/159742365-af996771-63cd-4628-93da-3af87fa60240.mp4


redreceipt:
looks good, I'm gonna wait for snaps to change and quick a look after that. I have a feeling some will change.

camrun91:
@redreceipt any idea how to get these checks to re-trgger?

redreceipt:
Yea that's weird. Not sure how it snuck through.

redreceipt:
Fix the lint and tests. It's unfortunate there are unrelated failures, I think someone merged straight to the branch instead of doing a PR. Buts likely just snap changes and syntax errors so the best course of action would be to correct before merging.

camrun91:
OK I will get the lint stuff fixed and look at the tests.

redreceipt:
Those snaps don't look right. The override name is being passed into the style prop now?

packages/apollos-ui-auth/src/styles.js
```diff
@@ -24,10 +24,9 @@ const FlexedSafeAreaView = styled(
   'ui-auth.FlexedSafeAreaView'
 )(SafeAreaView);
 
-const FlexedKeyboardAvoidingView = styled(
-  { flex: 1 },
-  'ui-auth.FlexedKeyboardAvoidingView'
-)(KeyboardAvoidingView);
+const FlexedKeyboardAvoidingView = styled('ui-auth.FlexedKeyboardAvoidingView')(
```
vinnyjth, line 27: You might need to import `named`
```suggestion
const FlexedKeyboardAvoidingView = named('ui-auth.FlexedKeyboardAvoidingView')(
```

vinnyjth: Also maybe kill this component and import `KeyboardAvoidingView` directly, since it doesn't actually flex anymore :)

camrun91: Went ahead and removed this component.

vinnyjth (approved):

camrun91:
@redreceipt or @vinnyjth I have the lint fix ready to push here. However, I am not following where/how the tests are failing.

redreceipt:
looks like more snaps

redreceipt:
Make sure to yarn between switching branches before updating snaps. A lot of dependencies change versions.

camrun91:
Good call @redreceipt on the Yarn I did not even think about that. Both of these should pass now.

PR #2450, opened 2022-03-24
chore: better error reporting for the CLI

nlewis84:
@redreceipt This change gave me an error message I could work with to debug. Open to suggestions if you would want to do this another way.

|e.stdout|just e|
|---|---|
|Screen Shot 2022-03-24 at 8 41 31 AM|Screen Shot 2022-03-24 at 8 41 17 AM|


redreceipt:
Fixed here https://github.com/ApollosProject/apollos-apps/pull/2448

PR #2458, opened 2022-03-29
fix: give button features authenticated link support

redreceipt:


redreceipt:
@vinnyjth got time to look at this one today? it's blocking woodmen valley launch. should make rock-auth apps work but will be broken for new auth apps. although that might be ok because technically it's already broken?

packages/apollos-ui-connected/src/features/ButtonFeature.js
```diff
@@ -41,23 +41,39 @@ const ButtonFeature = ({ action }) => {
     }
   );
 
-  const handlePress = () => {
+  const handlePress = (openUrl) => {
```
vinnyjth, line 44: Looks great! 

We can merge this as is, but what do you think about importing this map instead? 

I would love to try and consolidate all the different uses of our action schema, so they have the same handler function 

https://github.com/ApollosProject/apollos-apps/blob/93bfe31a5e4470c62c6d232eea722f0956bc8851/packages/apollos-ui-connected/src/FeaturesFeedConnected/FeaturesFeedConnected.js#L12

vinnyjth (approved):

packages/apollos-ui-connected/src/features/ButtonFeature.js
redreceipt: automerge got em

vinnyjth: should have checked for that, we can do a fast follow PR or capture it in basecamp somewhere

PR #2460, opened 2022-03-29
fix: work with smaller screens.

camrun91:


camrun91:
This should fix the screen sizing issues seen here: https://github.com/Differential/woodmenvalley-app/pull/22#issuecomment-1082037515

redreceipt (approved):

nlewis84:
Looks good to me. You will need to update snapshots. If you haven't done that before, the fastest way is to `cd` into `ui-auth` and `ui-connected` and run `yarn test -u`

camrun91:
OK cool I missed the ui-connected updating those now.

PR #2462, opened 2022-03-31
feat: update Onboarding screens for easier customization

nlewis84:
This PR adds more layers of customization to the Onboarding slides. Here is an example `theme.js` file using these customizations to give a client a completely overhauled onboarding experience.
Screen Shot 2022-03-31 at 3 09 26 PM


nlewis84:
@redreceipt @conrad-vanl @vinnyjth We can discuss here on the PR instead of over a call. In essence, I took several static properties and set them up to be potentially customized. If you think this would be a core addition, I would particularly like feed back on the prop naming in `slides.js` [here](https://github.com/ApollosProject/apollos-apps/pull/2462/files#diff-6a930cb47aaa73504263237ce7f99b21ab587978e9f45886d664e0fceec1779dR34).

redreceipt:
I like it. `greeting` is good but I'd try to be more descriptive on `size` and `items`. Help the developer know what size and what items that is referring to.

nlewis84:
Added optional chaining to items to allow customization without items. Also, changed `size` to `appIconSize` and `items` to `listItems`

packages/apollos-ui-onboarding/src/LandingSwiper/AnimatedAppIcon.js
```diff
@@ -16,7 +16,7 @@ const RotatorContainer = styled({
   position: 'absolute',
 })(Animated.View);
 
-const Rotator = ({ index }) => {
+const Rotator = ({ index, size = 134 }) => {
```
redreceipt, line 19: i'd give this one a better name too

packages/apollos-ui-onboarding/src/LandingSwiper/AnimatedAppIcon.js
```diff
@@ -44,23 +44,28 @@ const Rotator = ({ index }) => {
...
   );
 };
 
 Rotator.propTypes = {
   index: PropTypes.number,
+  size: PropTypes.number,
 };
 
-const AnimatedAppIcon = () => {
+const AnimatedAppIcon = ({ size }) => {
```
redreceipt, line 57: and this one

packages/apollos-ui-onboarding/src/LandingSwiper/AnimatedAppIcon.js
redreceipt: well actually nevermind, this is fine. the component is simple enough to understand

packages/apollos-ui-onboarding/src/LandingSwiper/AnimatedAppIcon.js
redreceipt: this too. nevermind.

redreceipt (approved):

PR #2464, opened 2022-04-05
fix: Stop query thrashing by restoring old default fetch policy

vinnyjth:
This code was originally part of our app, but we removed it in an effort to reduce template footprint. 

I don't have an exact model for why this code fixes the "query thrashing" that we  see, butI believe it's due to our frequently self-referential queries and our frequent usage of `cache-and-network`. If a query runs, that lets a component load, that fires a query that contains data that is shared with the original query, the original query will re-run. By default, it will preserve the original fetch-policy (and hit the network again). With this patch, it will instead hit the cache and avoid the network bouncing. That's my theory, anyway. 

Before:

After:


redreceipt:
Simply cutting our network requests in half seems like such a bandaid. Shouldn't we fix something we've been talking about awhile which is the loading states? If we have data, it shouldn't show a loading state. Even if it does go back out to the network, the user won't know the difference.

vinnyjth:
@redreceipt I think both are issues. And I would argue that just fixing loading states is just as much a bandaid since the query thrashing (even if the user doesn't see a loading state popping back and forth) will make the app feel sluggish. 

I also don't think this is a bandaid? This restores the default apollo behavior prior to version 3. Even Crossroads uses it. 

https://github.com/Differential/crossroads-anywhere/blob/master/react-native/app/client.js#L244-L255****

nlewis84:
Does this address the underlying issue that I noticed which was, why are we querying for tabs when we are trying to load a content item?

redreceipt:
It just makes me wonder why it's `not` the default behavior anymore and what we're doing that forces us to go backwards.

vinnyjth:
> Does this address the underlying issue that I noticed which was, why are we querying for tabs when we are trying to load a content item?

I don't know :) but I can find out.

vinnyjth:
> It just makes me wonder why it's `not` the default behavior anymore and what we're doing that forces us to go backwards.

I would love if you could answer that for me, because I haven't been able to parse out why this change was made. I think the old behavior was less consistent (apollo made assumptions about how you would want your fetch policy configured) and the new behavior is more precise (but delivers a trickier out of the box experience)

redreceipt:
I agree it stops the bleeding though. @vinnyjth what do you think about @nlewis84 making the change on the app repos themselves and let me take a look over cool down to see I can't come up with something more comprehensive?

vinnyjth:
@redreceipt I think this fix is worth including in core. We included it as part of the transition to Apollos 3.0. We remove it, because we couldn't fully articulate why it was needed and couldn't create a test suite for it. We are able to fix outstanding bugs by reintroducing it.

It's not a hack, it's not particularly special or unusual (again, Crossroads uses it). It doesn't seem dangerou or correlated with other known issues.

vinnyjth:
Gerard's commit message has the best articulation of the issue/solution I've seen yet

https://github.com/Differential/crossroads-anywhere/pull/5906/commits/bb4845abdc3f6ea22211b26251f96c468557bfd5

vinnyjth:
@redreceipt This comment has the most helpful details I've read yet on why things changed. 

https://github.com/apollographql/apollo-client/issues/6760#issuecomment-668188727

It looks like what I said above is roughly correct, that the definition of "cache-and-network" changed to mean literally check cache, then make a network call **every time**. Previously, the Apollos Cache would fallback to "cache-only" after the first request cycle.

redreceipt:
I was under the impression `cache-and-network` always meant that

vinnyjth:
> I was under the impression `cache-and-network` always meant that

Apparently it used to only mean that for the first time a network request was made. I believe the complicating factor here is that the Apollo Components "re query" every time their associated cache changes. 

So if you query for an item, and then query for it's media, previously the original query would go fetch new data from its' cache. Without this fix, it goes back to the network again (even though nothing has technically changed)

redreceipt:
Also, I get why we are calling `cache-first` after `cache-and-network` but why after `network-only`? cache should never play a factor in those queries?

vinnyjth:
We could go either way with `network-only`. I think it's safer to maintain the existing implementation (pre-2.0). 

If we wanted to eliminate this configuration, I think that would take an evaluation of the queries in our app and determine which ones need new or different fetch policies

PR #2465, opened 2022-04-05
fix: Fully hide journal Navstack on android

camrun91:
in cleaning up the bottom sheet modal I found that on android there was a navigation action happening when you tap close to the profile image. 

https://user-images.githubusercontent.com/16566029/161835110-f13e57dd-89f0-428e-ac59-17c3bb2c4cc7.mp4

this is with the proposed fix. 

https://user-images.githubusercontent.com/16566029/161835226-b293ac4f-16bf-46bb-9b71-247b2b274ea7.mp4

@redreceipt this will also need to be pushed to v2


redreceipt (approved):
Good find!

redreceipt:
I'll cherry pick and release. I actually had to turn off auto releases for v2 because I created an infinite loop 🤦🏼‍♂️ check out version 2.46.198 😅

PR #2469, opened 2022-04-06
chore: another attempt to get deploys working on the CI

redreceipt:


nlewis84 (approved):
I did have good luck passing keychain_password as an empty string. Let's see how it goes!

PR #2472, opened 2022-04-08
fix: action list item padding and action list header

### LINKED ISSUE #2494: adjust padding around action list
Need to work on general padding and UI design on the action list. With only one item in the list and no titles, it should look like this

Screen Shot 2022-04-27 at 3 41 18 PM

tristanguckenberger:
closes #2494 

### The purpose of this PR is to
- adjust bottom padding for single action list items (without modifying the padding of action lists with multiple items)
- hide the action list header if the action list only has a single item

If there's a better way to do this, please let me know!

### BEFORE:

IMG_0046

### AFTER:

IMG_0048


redreceipt:
This may be good for @conrad-vanl to comment on. I'm unsure of the implications of adding positional padding to such a foundational component

PR #2475, opened 2022-04-11
fix: android maps crashing

camrun91:
This fixes the issue that @nlewis84 was seeing. The .getNode() was coming back as undefined. I tested the change and made sure it worked in IOS and android.


nlewis84:
Nice, just update tests and this should be good.

`yarn test -u`

nlewis84 (approved):

PR #2476, opened 2022-04-11
fix: set default size for avatars in cloud if there are none to show

nlewis84:
This PR fixes a bug found in the AvatarCloud used in Prayer onboarding. If there are no prayers to show, then there are no 'leaves', so we needed to bake in some defaults for the case of 0 avatars to display.

## The Issue
https://user-images.githubusercontent.com/72768221/162798812-ec5b82d8-0f1b-4a41-99a5-9d38945b0add.mp4

## The Solution
https://user-images.githubusercontent.com/72768221/162798821-773a2cfa-a0a7-40f2-8918-18036358f632.mp4


vinnyjth (approved):

conrad-vanl (approved):

nlewis84:
@vinnyjth There were no snap updates for this. Should I dig in on why testing is failing here in unrelated packages?

nlewis84:
nvm @vinnyjth. I found a lonesome snap in a different package that needed some attention.

PR #2479, opened 2022-04-21
fix: adds a default style prop to handle title overflow

tristanguckenberger:
The purpose of this PR is to fix featureTitles text overflow. Below is a quick demo using woodmen.

https://user-images.githubusercontent.com/19194337/164528827-cc07abcf-8b29-4f28-aeea-e613ec16ab58.mov


redreceipt (commented):
Is there a reason you didn't opt for a standard react native style sheet on the container? Or even our `styled` HOC? I think it's performs better.

tristanguckenberger:
No, defaultProps was the just first thing I thought of but I like the idea of using styled a lot better.

redreceipt:
I like it. fix the linter and tests and this ones good to go

vinnyjth (approved):

redreceipt:
@tristanguckenberger fix these snaps and you're golden here

redreceipt (approved):

PR #2480, opened 2022-04-21
feature: adds autoplay prop to ApollosPlayerContainer

tristanguckenberger:
### **The Purpose of this PR** 

Adds the autoplay prop to ApollosPlayerContainer. This allows us to control autoplay from apollos client theme overrides. 

### **autoplay => true**

https://user-images.githubusercontent.com/19194337/164532553-f301bfc2-6f35-456b-b678-bfa65a633fc6.mov

### **autoplay => false**

https://user-images.githubusercontent.com/19194337/164532785-83227a41-d584-4750-ab95-fbd8392d7522.mov

### **autoplay => undefined**

https://user-images.githubusercontent.com/19194337/164532809-11bdbc0f-bd8c-4dc8-a7b3-4b7425dba338.mov


packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -136,6 +136,7 @@ const NodeSingleConnectedWithMedia = ({
         }}
         audioOnly={!hasLivestream && !hasVideo && hasAudio}
         videos={data?.node?.videos}
+        autoplay={props?.autoplay}
```
redreceipt, line 139: pull it out of props and just use `autoplay` here so we're not sending it to other components that are using prop spreading

tristanguckenberger: @redreceipt Is there any reason why we shouldn't just remove autoplay here entirely and then, in [`Container.tsx`](https://github.com/ApollosProject/apollos-apps/blob/12e3bd4e4541ab051823108576a617db346cc650/packages/apollos-ui-media-player/src/Container.tsx#L238) export like so:

```
import { named } from '@apollosproject/ui-kit/src/theme';
...
export default named('ui-media-player.ApollosPlayerContainer')(Container);
```

I tested this, it works and it would allow us to directly override Container.tsx's props which is what autoplay actually belongs to, not NodeSingleConnectedWithMedia.

With this change we'd no longer need to add the autoplay prop to `ApollosPlayerContainer` inside of `NodeSingleConnectedWithMedia` and we wouldn't need to worry about setting and pulling autoplay off of NodeSingleConnectedWithMedia's props. This would also give us the ability to override other Container props like `audioOnly`.


The override could change to this in theme overrides:

```
'ui-media-player.ApollosPlayerContainer': () => () => ({
    autoplay: true,
}),
```

redreceipt: I like that WAY better, good thinking

redreceipt (approved):

PR #2481, opened 2022-04-21
chore: adds codeowners

redreceipt:


redreceipt:
@vinnyjth I put you on there in vacation situations. but we could also just force merge a PR as owners in a pinch if you prefer that over the extra noise of getting pinged?

redreceipt:
actually yea I'm gonna take you off and we can prove out the need for two owners. I like the accountability and speed that comes with knowing it's one person's responsibility.

vinnyjth:
👍

vinnyjth (approved):

PR #2485, opened 2022-04-22
chore: moves in default error handler

redreceipt:
a necessary step before making fragment types dynamic


redreceipt:
I still need to test and verify the cache behaves properly

PR #2487, opened 2022-04-22
chore: use react native community eslint rules

redreceipt:
gets rid of our eslint pkg, not something we should think about, instead just uses defaults from react-native-community


redreceipt:
I've learned we have been very lazy with our hook dependency arrays

redreceipt:
@vinnyjth what do you think about this one?

PR #2491, opened 2022-04-26
feat: named Scripture Feature and ability to show or hide share button

nlewis84:
This PR named the Scripture Feature and adds an override prop showShareButton. We should be able to easily override this component to disable/enable the display of this button, but I was running into apollo cache errors when trying to do this via conventional overrides. 

|Before|After|
|---|---|
|Screen Shot 2022-04-26 at 4 19 41 PM|Screen Shot 2022-04-26 at 4 15 54 PM|


redreceipt:
Seems like a lot for just adding a prop to hide a component

nlewis84:
@redreceipt fixed it. That should look better.

packages/apollos-ui-connected/src/features/ScriptureFeature.js
```diff
@@ -1,36 +1,45 @@
...
   contentId,
-  nodeId,
   featureId,
   isCard,
+  isLoading,
+  nodeId,
+  scriptures,
+  sharing: { message } = {},
+  showShareButton,
+  title,
```
nlewis84, line 18: Put these in alpha...I did this when testing out the changes and ended up just keeping it.

redreceipt (commented):
Use default props to set it to true

packages/apollos-ui-connected/src/features/ScriptureFeature.js
```diff
@@ -72,6 +81,8 @@ ScriptureFeature.propTypes = {
   title: PropTypes.string,
   nodeId: PropTypes.string,
   featureId: PropTypes.string,
+  // This prop is used purely for direct overriding of the display of the share button
```
redreceipt, line 84:
```suggestion
```

redreceipt (commented):
Niiice, once tests are fixed I'll approve

redreceipt (approved):

PR #2500, opened 2022-05-10
feature: Content progression

camrun91:
This PR implements the Playstate history and video playback starting at the last watched time stamp.
here is a screen record. 

https://www.loom.com/share/1e293b08a72a40628a8fa008ed0ff615


packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -136,6 +162,11 @@ const NodeSingleConnectedWithMedia = ({
         }}
         audioOnly={!hasLivestream && !hasVideo && hasAudio}
         videos={data?.node?.videos}
+        // modify these to call the mutation to write the interactions.
+        onPlay={(timestamp) => handleOnPlay(timestamp, data?.node?.videos[0])}
```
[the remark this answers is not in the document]
vinnyjth: I agree @redreceipt here :) since these calls are contextually related to the Video, we probably should let the Video component be responsible for making these calls.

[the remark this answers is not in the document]
camrun91: My understanding on how these handlers are working was that they are just returning the shared state of the Playhead.  Let me try this and see if pushing it down a level would work.

packages/apollos-ui-connected/src/NodeSingleConnected/useInteractWithNode.js
```diff
@@ -0,0 +1,61 @@
...
+      about conforming to Apollos API schema for [InteractionDataField]
+  */
+  const mutation = useCallback(
+    (mutOpts = { variables: {} }) => {
+      return _mutation({
+        ...mutOpts,
+        variables: transformVariables(mutOpts.variables),
+      });
+    },
+    [_mutation]
```
vinnyjth, line 55: Should options be listed here, just to be safe? (It might not be necessary since the reference to `_mutation` will likely change if `options` changes.) 

```suggestion
    [_mutation, options]
```

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -82,11 +82,37 @@ const NodeSingleConnectedWithMedia = ({
...
+        data,
+      },
+    });
+  }
   const { data } = useQuery(GET_MEDIA, {
     variables: { nodeId },
     fetchPolicy: 'cache-and-network',
   });
 
+  function handleOnPlay(playhead, video) {
```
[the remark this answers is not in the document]
camrun91: @redreceipt I did not see where we were doing any analytics in the player pkg. I could be missing it though.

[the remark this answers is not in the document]
redreceipt: https://github.com/ApollosProject/apollos-apps/blob/9ddf9391b8a4f2a1d035abe2cc53f34701788718/packages/apollos-ui-media-player/src/Container.tsx#L181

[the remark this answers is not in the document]
calebpanza: @redreceipt the way that the video player is set up right now is that it prefers prop drilling and callbacks. That method is set up in order to pass analytic information _back up_ to the parent renderer

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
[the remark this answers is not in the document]
calebpanza: The main problem I see with that is that the video player is currently set up to be contextual to the _video_, not the _user_. I think we could rewrite/consolidate these interactions in the future, but I would hesitate to break the current prop-drilling/callback convention that's already been established in the package. Thoughts?

camrun91:
@redreceipt @vinnyjth @calebpanza I talked to @liamillard since she was the one that set these callbacks up. They are using these callbacks in the KC mobile app. She did have a request for a few more things to be returned from them. I can look into returning those. 

However, I do not know if changing these to log the user progress in the video player package would impact KC app's usage or not.

liamillard:
@camrun91 The KC mobile app hasn't launched yet, so if we need to make changes to how these overrides are implemented in KC we can totally do that.

I don't think moving the callbacks up a level is bad, as long as they're still overridable in a theme. Does that make sense?

camrun91:
@redreceipt @vinnyjth @calebpanza just to make sure we are all on the same page here. I need to move the "new" `interactWithVideo` hook to live in the video play and the default action be to log the interaction, then return or pass up the values that @liamillard needs correct?

calebpanza:
> move the "new" interactWithVideo hook to live in the video play

I think both the current code and the rework mentioned in the comment above both resolve the immediate need of gathering analytics. 

With how the Video Player is written, passing in a callback function (current implementation) makes most sense to me personally. But again, both approaches resolve the problem. Whatever makes sense to everyone else is good with me 😁

vinnyjth:
> My only request is video components do all the video things. As little passing of props around as possible.

:+1:

I think this is what I was trying to say above. Try and avoid `NodeSingkeConnected` needing to figure out how to call interactions/analytics for the video player.

camrun91:
@redreceipt @vinnyjth I am fine with doing that my only hesitation is that The way that the video player package is currently we are not doing any mutations or anything from it. Do we want to introduce that into the video player package? and how much extra work/effort is that going to be?

redreceipt:
I don’t see why it should do GQL stuff. Every other high level UI pkg does. Makes more sense for the media player to do media player operations than the content component doing media player operations.

camrun91:
@redreceipt @vinnyjth @conrad-vanl the latest commit should be a good start on the ApollosPlayerConnected component. I will do some clean-up and push in the changes that Lia requested. But please provide feedback that you guys see.

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
```diff
@@ -0,0 +1,109 @@
...
+  const [interactWithVideo] = useMutation(INTERACT_WITH_NODE);
+
+  const track = useTrack();
+
+  const { data } = useQuery(GET_MEDIA, {
+    variables: { nodeId },
+    fetchPolicy: 'cache-and-network',
+  });
+
+  function handleOnPlay({ id, elapsedTime, title }) {
```
redreceipt, line 28: `useCallback` these puppies

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
```diff
@@ -0,0 +1,109 @@
...
+      },
+    });
+  }
+
+  return (
+    <ApollosPlayerContainer
+      source={mediaSource}
+      coverImage={data.node?.coverImage?.sources}
+      presentationProps={{
+        title: data.node.title,
```
redreceipt, line 83: Optional chain this one too?

packages/apollos-ui-connected/src/index.js
```diff
@@ -6,6 +6,7 @@ export AddCommentFeatureConnected, {
   GET_ADD_COMMENT_FEATURE,
   ADD_COMMENT,
 } from './AddCommentFeatureConnected';
+export ApollosPlayerConnected from './ApollosPlayerConnected';
```
redreceipt, line 9: Why are you exporting this?

packages/apollos-ui-media-player/src/Container.tsx
```diff
@@ -27,6 +28,8 @@ interface ContainerProps extends IPlayerMedia, FullScreenSlidingPlayerProps {
   onPlay?: Function;
   onPause?: Function;
   onEnd?: Function;
+  playheadStart?: number;
+  videos?: Array<IVideo>;
```
redreceipt, line 32: Nice!

templates/mobile/package.json
```diff
@@ -35,6 +35,9 @@
       "./jest.setup.js"
     ]
   },
+  "resolutions": {
+    "react-devtools-core": "4.24.3"
```
redreceipt, line 39: What's this for?

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
```diff
@@ -0,0 +1,109 @@
...
+      audioOnly={audioOnly}
+      videos={data?.node?.videos}
+      // modify these to call the mutation to write the interactions.
+      onPlay={handleOnPlay}
+      onPause={handleOnPause}
+      onEnd={handleOnEnd}
+      playheadStart={data?.node?.videos[0]?.userProgress?.playhead}
+      {...props}
+    >
+      <Component nodeId={nodeId} ImageWrapperComponent={Noop} {...props} />
```
redreceipt, line 94: Could this Component be defined here instead of passing it down from above?

redreceipt (commented):
Great work. Just a few questions.

packages/apollos-ui-connected/src/index.js
camrun91: I was making the assumption that it was "Standard" if it is not then I will remove it.

templates/mobile/package.json
camrun91: I can remove that I had to put that in to get flipper/react debug tools to function. I can remove that as well.

redreceipt: I've never used the dev tools in flipper, maybe you can show me some cool stuff? why does this need to be in resolutions vs dev deps? this is admittedly my own npm ignorance

camrun91: It is because the version that is "packaged" with flipper is not in sync with the version that is packaged with Reactnative. My understanding is that resolution forces the same version to be used.

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
camrun91: it could but it is defined and used in `nodeSingleConnected` in the instance that there is no media. if there is media it is then passed in here. So I think it would have to be defined 2 times essentially.

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
camrun91: yep got this fixed

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
camrun91: done

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
redreceipt: ah didn't see that. I still think we can refactor and clean it up but we can do that later.

camrun91: @redreceipt This should be the only thing left on this PR. What are your thoughts on it being needed by the parent component.?

camrun91:
@redreceipt is there something I need to do to fix that codecov report?

redreceipt:
nope! I guess tests are good.

redreceipt (approved):
excellent work

PR #2503, opened 2022-05-12
fix: only set KeyboardAvoidingView behavior for ios

tristanguckenberger:
### **Purpose**
This PR fixes a known issue with KeyboardAvoidingView and android. This bug affects both Oak Cliff and City First on android (other apollos apps are likely affected too) but the screen shots below just show Oak Cliff.

### **Before**
Screenshot_20220511-114600_Oak Cliff Bible Fellowship

### **After**
Screenshot_20220512-142705_Oak Cliff Bible Fellowship_Original


camrun91:
@tristanguckenberger I fixed these in v2 branch I think. you may make sure I did not miss any there. 
https://github.com/ApollosProject/apollos-apps/commit/e652dd6ab5a2e8f8a4e8e3e623fda3ef2909edb2

nlewis84:
Oak Cliff and City First are both v3 apps. @camrun91 did those change make it v3?

camrun91:
I don't think all the fixes got put into v3

tristanguckenberger:
I believe the problem here is that adding a style, i.e. `flex: 1`, to `FlexedKeyboardAvoidingView` conflicts with `KeyboardAvoidingView`'s `behavior` prop, specifically for android. I applied the changes from the v2 commit to make sure they work, they do, but there's a subtle difference ya'll should know about:

### **V2 Fix**
In the fix for V2, when applied to City First, if we remove flex from `FlexedKeyboardAvoidingView` or just use the original `KeyboardAvoidingView`, we then prevent the component from using the full available space above the keyboard resulting in the button pushed up against the form up top like so:
Screenshot_20220512-163530_City First_Original

### **PR Fix**
In the commit for this PR, we're able to keep the flex and maintain some styling continuity across platforms. My preference is to keep the experience for android and ios as similar as possible and I think we can do that in this instance if we stick with the changes in this PR for V3 apps.
Screenshot_20220512-163834_City First_Original

Let me know what ya'll think and/or if I've overlooked anything.

camrun91:
OK the only other thing I would say was make sure to test the changes on as small of an android phone as you can. Some of these changes when I was testing did not work with small screens. If that doesn't happen then I think these changes are good.

redreceipt (approved):

tristanguckenberger:
I did some testing with the smallest android phone sim in android studio and I believe this is behaving properly, but @camrun91 do you see anything like the issues you experienced in the video below? 

Someone gut check me, but I think the expected functionality here is for the screen above the keyboard to fill up as much space as it can and then allowing a user to scroll any overflow if needed.

### **Tests with this PR's changes on a tiny device**

https://user-images.githubusercontent.com/19194337/168313061-f4b3fd1e-39d2-49ca-9b8f-1d4fb905a847.mov

### **Tests without this PR's changes on a tiny device**

https://user-images.githubusercontent.com/19194337/168312498-d7844a19-9f87-4f60-ae10-7214fc4b91a3.mov

camrun91:
@tristanguckenberger @nlewis84 I think on v2 I made a change so that the next button stayed at the bottom on android. that way there was more of the input area visible. I Will let @conrad-vanl give his input as I am not good with design-type things.

nlewis84:
@tristanguckenberger This looks good in my testing. I'm going to go ahead and merge.

PR #2504, opened 2022-05-16
fix: protected route bug

vinnyjth:


redreceipt (approved):
💯 i wonder if this is a new bug or we had it before

redreceipt:
yea looks like i introduced this one

PR #2506, opened 2022-05-17
fix: repair tags list on tabs

redreceipt:
I've chosen to make it more restrictive than it was before since these features were broken anyway. plus adding arbitrary props here isn't the right level of customization for us moving forward

_NOTE: the chip list isn't working great and will need to be fixed in a separate PR_

Screen Shot 2022-05-17 at 2 40 56 PM


packages/apollos-ui-connected/src/TabsConnected/Tabs.js
```diff
@@ -87,12 +87,13 @@ const Tabs = ({ tabs }) => {
...
   const tabNavs = useMemo(
     () =>
-      tabs.map(({ title, icon, feed }) => ({
+      tabs.map(({ title, icon, feed }, i) => ({
         title,
         icon,
         component: createTab({
           featureFeedId: feed.id,
           options: {
+            headerLargeTitle: i !== 0,
```
vinnyjth, line 96: Is this to avoid the header large title on the home tab?

packages/apollos-ui-connected/src/TabsConnected/Tabs.js
```diff
@@ -102,9 +103,9 @@ const Tabs = ({ tabs }) => {
             ...(ApollosConfig.TABS[title]?.showProfile
               ? { headerLeft: ProfileButton }
               : {}),
-            ...(ApollosConfig.TABS[title] ?? {}),
           },
           title,
+          showTags: ApollosConfig.TABS[title]?.showTags || false,
```
vinnyjth, line 108: can we spread in tab options here as well? 

or are you avoiding that in order to introduce a more focused surface area?

packages/apollos-ui-connected/src/TabsConnected/createTab.js
```diff
@@ -11,13 +11,13 @@ import FeaturesFeedConnected from '../FeaturesFeedConnected';
...
 
-const Tab = ({ useTagFilter, feedViewProps, additionalFeatures }) => {
+const Tab = () => {
   const navigation = useNavigation();
   const route = useRoute();
   const [filteredTags, setFilteredTags] = useState([]);
   return (
     <>
-      {useTagFilter ? (
+      {route.params.showTags ? (
```
vinnyjth, line 20: 👍

packages/apollos-ui-connected/src/TabsConnected/createTab.js
```diff
@@ -11,13 +11,13 @@ import FeaturesFeedConnected from '../FeaturesFeedConnected';
 import ContentFeed from '../ContentFeedConnected';
 import TagFilterConnected from '../TagFilterConnected';
 
-const Tab = ({ useTagFilter, feedViewProps, additionalFeatures }) => {
```
vinnyjth, line 14: Thanks for removing these :)

packages/apollos-ui-connected/src/TabsConnected/createTab.js
```diff
@@ -30,8 +30,6 @@ const Tab = ({ useTagFilter, feedViewProps, additionalFeatures }) => {
               openUrl={openUrl}
               featureFeedId={route.params.featureFeedId}
               navigation={navigation}
-              feedViewProps={feedViewProps}
-              additionalFeatures={additionalFeatures}
```
vinnyjth, line 34: Do you recall what this feature even did?

vinnyjth (approved):

packages/apollos-ui-connected/src/TabsConnected/Tabs.js
redreceipt: yep

packages/apollos-ui-connected/src/TabsConnected/Tabs.js
redreceipt: smaller surface area. I don't love overrides corresponding to react-navigation individual screen options and also our own custom variables.

packages/apollos-ui-connected/src/TabsConnected/createTab.js
redreceipt: I have no idea how we used feed view props. `additionalFeatures` was for custom frontend components that you could define in your custom config.yml. it was the wild west back then 😂

PR #2509, opened 2022-05-23
feat: Scripture route and scripture detection

camrun91:
### Motivation
This PR sets up the groundwork for scripture, guides, and sermon notes. we needed a way to show scripture in the "Feed" of a content item with out it being a block of text. This also sets up bible ref detection and creates links to the scripture modal.

#### Scripture Screen
This is set up as a new Connected component `ScriptureScreenConnected` It is expecting that a scripture reference be passed in via route params. It then uses the useQuery hook to pull the data from our API and render it. The fetch policy on this hook should allow for caching of the data. 

This is built to allow enabling of Journals on scripture in the future. 

#### Scripture feature changes
The scripture feature now only renders a touchable that will navigate to the new scripture screen. From what I can tell this should not be a "Breaking change" this should work how things currently are just render less text and render a touchable. 

#### HTMLView changes
This is where most of the work is happening in this PR. First the HTMLView now has an additional prop `detectVerse` it defaults to false to allow backward compatibility. This was needed because the `ScriptureHTMLView` uses the normal HTMLView and processing the scripture HTML was causing an issue rendering references in the text of the verse. 

If detectVerse is true then the scripture detection runs at the parser level. It will wrap any reference in a new <scripture>. That tag is what tells the default renderer to return the new `ScriptureComponent` This is just rendering a styled text element with an `onTouch` prop to navigate to the scripture screen.

Here is an example of it running on the sim. 
https://www.loom.com/share/633f388414514fac8c9fe443a2555e17


packages/apollos-ui-htmlview/src/HTMLView/index.js
```diff
@@ -53,8 +56,21 @@ class HTMLView extends PureComponent {
...
+    const { start, end } = range;
+    return (
+      `${this.bibleApiCall}${range.book} c${start.chapter}v${start.verse}s'${start.section}' to ` +
+      `c${end.chapter}v${end.verse}s'${end.section}'`
+    );
+  }
+
   parse(html = '') {
-    this.parser.write(html);
+    const replaced = html.replace(this.verseRequiringRegex, (...args) => {
```
camrun91, line 69: @conrad-vanl @vinnyjth @calebpanza here is where I am currently detecting scripture. The issue with doing it here is I can not easily (that I am seeing) pass through the component to render. If I do it in the renderer it makes it hard to pass in the actual reference.

vinnyjth: > If I do it in the renderer it makes it hard to pass in the actual reference.

Can you explain this further? I think the renderer is the correct place to do this. Specifically - https://github.com/ApollosProject/apollos-apps/blob/3b3a745b508184653563b45e8175842b620c498f/packages/apollos-ui-htmlview/src/HTMLView/defaultRenderer.js#L70-L91

camrun91: Basically that (unless I am seeing what it is doing wrong. will make the entire section of text clickable. so a 
`<p> we ask you read Gen 4 </p>` the whole text in P will be clickable not just the "Gen 4. 

Unless there is a way I am not thinking of to break that text up and render it two different ways....
wait that gave me an idea... 

any input you have go ahead and leave here. I am going to try the idea I just had.

vinnyjth: you can return a fragment here (or any JSX) - so roughly, you could do this 

1. Split text into array of scripture references and normal strings (using regex)
2. map over array replacing strings matching the prior regex with a JSX tag as needed, and everything else with `Text` tags
3. Return the array, or a fragment with all the individual elements. 

More ideas and more efficient implementations possible: 
https://stackoverflow.com/questions/30474506/replace-part-of-string-with-tag-in-jsx

camrun91:
still draft but getting the core of this functionality down. Clean up the style and stuff and then some feature improvements.

camrun91:
@calebpanza @vinnyjth @redreceipt This should be good for review now.

packages/apollos-ui-connected/src/ScriptureScreenConnected/getScripture.js
```diff
@@ -0,0 +1,11 @@
+import ApollosConfig from '@apollosproject/config';
+import { gql } from '@apollo/client';
+
+export default gql`
+  query scriptures($reference: String!) {
```
redreceipt, line 5:
```suggestion
  query GetScriptures($reference: String!) {
```

packages/apollos-ui-connected/src/ScriptureScreenConnected/index.js
```diff
@@ -0,0 +1,56 @@
...
+import ScriptureItem from '@apollosproject/ui-scripture';
+import GET_SCRIPTURE from './getScripture';
+
+const ScriptureScreenConnected = ({ route, isLoading, ...props }) => {
+  const { reference } = route.params;
+  const { data } = useQuery(GET_SCRIPTURE, {
+    variables: { reference },
+    fetchPolicy: 'cache-and-network',
+  });
+  const scriptures = data?.scriptures ?? [];
```
redreceipt, line 17: should the whole thing just be `null` is scripture array is empty? (loading state)

packages/apollos-ui-connected/src/ScriptureScreenConnected/index.js
```diff
@@ -0,0 +1,56 @@
...
+ScriptureScreenConnected.propTypes = {
+  route: PropTypes.shape({
+    params: PropTypes.shape({
+      reference: PropTypes.string,
+      html: PropTypes.string,
+      version: PropTypes.string,
+      copyright: PropTypes.string,
+    }),
+  }),
+  isLoading: PropTypes.bool,
```
redreceipt, line 53: not currently using this prop

packages/apollos-ui-connected/src/features/ScriptureFeature.js
```diff
@@ -42,26 +47,24 @@ const ScriptureFeature = ({
...
-          version={version}
-        />
-      ))}
-    </Wrapper>
-  );
+  const navigation = useNavigation();
+  return scriptures.map((scripture) => (
+    <Touchable
+      onPress={() => {
+        navigation.navigate('Scripture', { ...scripture, isLoading });
```
redreceipt, line 54: we only use the reference right? and `isLoading` here is doing nothing

packages/apollos-ui-connected/src/ScriptureScreenConnected/index.js
camrun91: This one, in particular, I pulled from how it was rendering in the features stuff. I can def change it to null

packages/apollos-ui-connected/src/index.js
```diff
@@ -89,6 +89,7 @@ export NodeSingleConnected, {
 } from './NodeSingleConnected';
 export RockAuthedWebBrowser from './RockAuthedWebBrowser';
 export RockAuthedWebView from './RockAuthedWebView';
+export ScriptureScreenConnected from './ScriptureScreenConnected';
```
redreceipt, line 92: any reason to export this?

redreceipt: ah yes nvm, answered my own question

packages/apollos-ui-htmlview/src/HTMLView/ScriptureComponent.js
```diff
@@ -0,0 +1,35 @@
...
+import { useTheme } from '@apollosproject/ui-kit';
+import { useNavigation } from '@react-navigation/native';
+import PropTypes from 'prop-types';
+
+const ScriptureComponent = ({ verse, children }) => {
+  const navigation = useNavigation();
+  const theme = useTheme();
+  return (
+    <Text
+      style={styles.buttonLink(theme)}
```
redreceipt, line 12: ❤️

packages/apollos-ui-htmlview/src/HTMLView/index.js
```diff
@@ -53,8 +60,17 @@ class HTMLView extends PureComponent {
     }
   }
 
+  // Set up the new scripture route here to open in the modal.
+
   parse(html = '') {
-    this.parser.write(html);
+    if (this.props.detectVerse) {
```
vinnyjth, line 66: @camrun91 can you elaborate on this again? 

This feels more like a concern of the parser itself, rather than something that should be happening at the level of the HTMLView itself.

packages/apollos-ui-htmlview/src/HTMLView/defaultRenderer.js
```diff
@@ -93,6 +97,12 @@ const defaultRenderer = (node, { children }, handlePressAnchor) => {
   }
 
   switch (node.name) {
+    case 'scripture':
```
redreceipt, line 100: very cool

packages/apollos-ui-connected/src/ScriptureScreenConnected/index.js
camrun91: Also was not using the 2 or 3 other props there lol I got it cleaned up

packages/apollos-ui-connected/src/ScriptureScreenConnected/getScripture.js
camrun91: Done.

packages/apollos-ui-connected/src/ScriptureScreenConnected/index.js
```diff
@@ -0,0 +1,56 @@
...
+import PropTypes from 'prop-types';
+import { useQuery } from '@apollo/client';
+import { BottomSheetModalProvider } from '@gorhom/bottom-sheet';
+import { ModalView, PaddedView } from '@apollosproject/ui-kit';
+import { ScrollView } from 'react-native';
+import { SafeAreaView } from 'react-native-safe-area-context';
+import ScriptureItem from '@apollosproject/ui-scripture';
+import GET_SCRIPTURE from './getScripture';
+
+const ScriptureScreenConnected = ({ route, isLoading, ...props }) => {
```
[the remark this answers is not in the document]
redreceipt: what's the error? we should figure what prop is causing the error disappear and then deal with that.

packages/apollos-ui-htmlview/src/HTMLView/index.js
redreceipt: yea I agree. I'm a bit confused about the architecture and purpose of the different components

packages/apollos-ui-connected/src/ScriptureScreenConnected/index.js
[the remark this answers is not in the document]
camrun91: HAHA Well, it was an obvious fix... I got it working. I will push changes in a second.

packages/apollos-ui-htmlview/jest.setup.js
```diff
@@ -6,6 +6,16 @@ jest.mock('react-native-safe-area-context/', () => ({
   SafeAreaProvider: ({ children }) => children,
 }));
 
+jest.mock('@react-navigation/native', () => {
```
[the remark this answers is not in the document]
camrun91: The test was complaining about not knowing what the navigation object was until I added this. Do I need to import the utility here?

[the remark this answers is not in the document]
redreceipt: use it in the test. something like

```js
WithReactNavigator(
  <ComponentYoureTesting />
)
```

you can hunt for examples

packages/apollos-ui-htmlview/src/HTMLView/index.js
redreceipt: @vinnyjth are you referring to the server handling instead of react native?

camrun91: The Change I am making here is to let the parser parse the HTML and do an inline replace of the scripture reference with a wrapped `<scripture> {Reference} </scripture>` That then allows the `DefaultRenderer` to know that that specific part of HTML should be Rendered as a <ScriptureLink> (Going to be moving `ScriptureComponent` to UI-kit as `ScriptureLink `)

packages/apollos-ui-htmlview/jest.setup.js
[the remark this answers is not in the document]
camrun91: Got it. This will be fixed on the next push.

packages/apollos-ui-connected/src/ScriptureScreenConnected/index.js
```diff
@@ -0,0 +1,53 @@
...
+  const { reference } = route.params;
+  const { data } = useQuery(GET_SCRIPTURE, {
+    variables: { reference },
+    fetchPolicy: 'cache-and-network',
+  });
+  const scriptures = data?.scriptures ?? [];
+  return (
+    <ModalView {...props} onClose={() => props.navigation.goBack()}>
```
redreceipt, line 19: no prop spreading here either, please :)

camrun91: yep got that one already as well.

packages/apollos-ui-connected/src/ScriptureScreenConnected/index.js
```diff
@@ -8,20 +8,20 @@ import { SafeAreaView } from 'react-native-safe-area-context';
...
 
-const ScriptureScreenConnected = ({ route, isLoading, ...props }) => {
+const ScriptureScreenConnected = ({ route, navigation }) => {
   const { reference } = route.params;
   const { data } = useQuery(GET_SCRIPTURE, {
     variables: { reference },
     fetchPolicy: 'cache-and-network',
   });
-  const scriptures = data?.scriptures ?? [];
+  const scriptures = data?.scriptures ?? null;
```
redreceipt, line 17: no sorry, the array was perfect. I meant should this whole component render `null` on an empty array? like what does the loading state look like?

camrun91: haha ok, it is just the screen without any text. that does bring up a good point though. What should we do/show if our API is unable to get a reference? Like if it somehow tags John 100:3 that won't return any data. do we need an indicator of error?

packages/apollos-ui-htmlview/src/HTMLView/ScriptureComponent.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
```
redreceipt, line 1: you can delete this guy now, right?

camrun91: yep I think so.

packages/apollos-ui-connected/src/ScriptureScreenConnected/index.js
redreceipt: if the array is empty, just start with react natives activity indicator. that's good enough for MVP. that covers loading and error states.

packages/apollos-ui-htmlview/src/HTMLView/index.js
vinnyjth: What are the advantages of that approach vs replacing scripture with ScriptureLinks inline in the `parser`? 

I see that you're creating a form of intermediate representation here, but I'm not sure why it's needed or helpful :)

redreceipt:
deferring to @vinnyjth 's concerns here

camrun91:
@redreceipt the change you asked me to make to import the `withReactNavigator` for test looks like it broke something. when I added it the tests started failing and I can not tell how or why.

redreceipt:
i'll take a look

camrun91:
This looks like it should be good to go now as long as @redreceipt and @vinnyjth are good with the implementation.

PR #2510, opened 2022-05-23
fix: adjust content feed padding to account for header height on Android

nlewis84:
- fix: styles a background view to add padding equal to header height on Android
- fix: styles a background view to add padding
- chore: update snaps

This PR addresses a bug on Android where the height of the header is not being accounted for and so the header overlaps the content feed. 

|Before|After|
|---|---|
|Screenshot_20220523_153731|Screenshot_20220523_155912|
|Screenshot_20220523_153605|Screenshot_20220523_153543-1|

iOS unchanged
|User Settings|Feature Feed|
|---|---|
|Simulator Screen Shot - iPhone 13 - 2022-05-23 at 16 02 43|Simulator Screen Shot - iPhone 13 - 2022-05-23 at 16 02 32|


redreceipt:
Nice work. Fix the tests and I'll approve

nlewis84:
Sweet. Just working on figuring out why a ui-authentication test is failing.

redreceipt:
I'm seeing a failing snapshot in ui-connected according to the CI

nlewis84:
Screen Shot 2022-05-23 at 10 06 03 PM

redreceipt:
Do you need to yarn?

nlewis84:
@redreceipt so weird....I had to use `yarn test --updateSnapshot` in the ContentSingleFeaturesConnected folder in order for this snapshot to actually get changed. Usual process is `yarn test -u` in the package dir like `apollos-ui-connected`.

nlewis84:
Just noticed that all the nested screens (Likes, Personal Details, Change Password, Notification Settings) are effected by this bug as well. Turned off auto-merge so that I can get in there and fix them as well.

nlewis84:
More pics of the nested screens

|Before|After|
|---|---|
|Screenshot_1653402356|Screenshot_1653402005|
|Screenshot_1653402010|Screenshot_1653402360|
|Screenshot_1653402016|Screenshot_1653402364|
|Screenshot_1653402367|Screenshot_1653402020|

calebpanza:
It might be worth creating a Component or a Hook since the logic is reused in a handful of places?

nlewis84:
I don't think I've seen a utils directory anywhere. Any thoughts on where we could store something like that? I thought about adding a prop to BackgroundView that was androidHeaderPadding, then adding the logic to BackgroundView. @redreceipt may have thoughts on that.

redreceipt:
well if we were going to do that, it should probably go in `BackgroundView` but I didn't love the logic already (simply adding 32px of vertical space) but abstracting that up a level will make it a lot more pervasive. does react-navigation not have any helper spacing utilities we could use instead? it should know how big its navbar is.

nlewis84:
I'm not sure if React Navigation offers the ability to add margin to the nav header, but I'll look into it.

nlewis84:
The hook doesn't work up a level at the tab creation step because we are not yet inside a screen in a Stack.

packages/apollos-ui-kit/src/BackgroundView/index.js
```diff
@@ -1,15 +1,29 @@
 import PropTypes from 'prop-types';
 
-import { View } from 'react-native';
+import { Dimensions, Platform, View } from 'react-native';
 
 import styled from '../styled';
 
+const screenHeight = Dimensions.get('screen').height;
+const windowHeight = Dimensions.get('window').height;
+const navbarHeight = screenHeight - windowHeight;
```
redreceipt, line 9: Is there a case where this value isn't 0 and we wouldn't want to use it? Another to ask is could we scrap the prop and just make this default behavior if `navHeight > 0`?

packages/apollos-ui-kit/src/BackgroundView/index.js
```diff
@@ -1,15 +1,29 @@
...
+    theme,
+    androidHeaderHeight = false,
+    material = 'screen',
+    flexed = true,
+  }) => ({
     backgroundColor:
       theme?.colors?.background[material] || theme?.colors?.background?.screen,
     flex: flexed ? 1 : 0,
     height: flexed ? '100%' : undefined,
+    paddingTop: androidHeaderHeight
```
redreceipt, line 22: Let's go only set paddingTop if we're doing the nav bar thing. So don't set to 0 for iOS or not using nav bar height.

redreceipt (commented):
Couple thoughts

packages/apollos-ui-kit/src/BackgroundView/index.js
nlewis84: I just went on an exploration of this idea, and I did not come across any instance where it was > 0 and we did not want to use it.

redreceipt: I mean worst-case scenario, there's an extra 32 px of padding on a screen that doesn't use react-navigation, I say we go with this instead of introducing the prop

packages/apollos-ui-kit/src/BackgroundView/index.js
nlewis84: Would I need to spread out props, in order to not set paddingTop to anything?

packages/apollos-ui-kit/src/BackgroundView/index.js
nlewis84: I take it back...I found some in onboarding that are negatively affected by this.

packages/apollos-ui-kit/src/BackgroundView/index.js
redreceipt: Yea you'd either spread an empty object or one with just padding top in it

packages/apollos-ui-kit/src/BackgroundView/index.js
nlewis84: Turns out the nav bar height is calculated on app load. This is probably not a good look for us 😂
Screenshot_1653425099

redreceipt: Guess we need it. Maybe call it `avoidHeader` to be a clearer on what it does.

nlewis84:
Looks like the linter broke code coverage 😅

packages/apollos-ui-kit/src/BackgroundView/index.js
```diff
@@ -1,15 +1,24 @@
...
+const navbarHeight = screenHeight - windowHeight;
+
 const BackgroundView = styled(
-  ({ theme, material = 'screen', flexed = true }) => ({
+  ({ theme, avoidHeader = false, material = 'screen', flexed = true }) => ({
     backgroundColor:
       theme?.colors?.background[material] || theme?.colors?.background?.screen,
     flex: flexed ? 1 : 0,
     height: flexed ? '100%' : undefined,
+    paddingTop: avoidHeader
```
redreceipt, line 17: I still need you to spread this out instead of passing 0 here

packages/apollos-ui-kit/src/BackgroundView/index.js
```diff
@@ -1,15 +1,24 @@
...
+  ({ theme, avoidHeader = false, material = 'screen', flexed = true }) => ({
     backgroundColor:
       theme?.colors?.background[material] || theme?.colors?.background?.screen,
     flex: flexed ? 1 : 0,
     height: flexed ? '100%' : undefined,
+    ...(avoidHeader
+      ? Platform.OS === 'android'
+        ? { paddingTop: navbarHeight + theme.sizing.baseUnit * 2 }
+        : {}
+      : {}),
```
nlewis84, line 21: @redreceipt is this how you would do that? it works, but didn't know if you had in mind a different way.

redreceipt: Perfect

redreceipt:
very nice! I cleaned those up here so we can get a better look at what this PR is changing https://github.com/ApollosProject/apollos-apps/pull/2512

redreceipt (approved):
good work

PR #2520, opened 2022-06-03
feat: removed scripture component and add htmlFeature

camrun91:
This puts in place a very simple HTML Feature. it works with the scripture detection from earlier this week. 

Screen Shot 2022-06-03 at 12 30 48 PM


camrun91:
@conrad-vanl @redreceipt I am completely open to any styling changes you have on how that renders. Right now it renders the straight HTML in a padded view.

redreceipt:
I have no opinion on design. But could you give me a little test coverage :) a snap should be fine

redreceipt (approved):

camrun91:
hmmm ok, the test is passing locally but failing here...

calebpanza:
@camrun91 personally, I like that it just renders right on the screen. I wouldn't change a thing on it

PR #2523, opened 2022-06-08
fix: add resi livestream support

tristanguckenberger:
This purpose of this PR is to add livestream support for resi urls. We discovered recently that resi hls urls (resi links to .m3u8 files) do not work with react-native-webview on android. This fix derives a platform specific resi url from the backend's config.yml MEDIA_URLS field if a resi specific livestream is found. 

Related PR: [https://github.com/lcbc-digital/ephesus/pull/77](https://github.com/lcbc-digital/ephesus/pull/77)


redreceipt:
This one needs a better title. Also, what's its purpose in simple terms?

tristanguckenberger:
> This one needs a better title. Also, what's its purpose in simple terms?

Updated, lmk if that works. I updated the top bit to be a little more clear. But in super simple terms, this gets resi videos working for android during livestreams.

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -100,7 +100,14 @@ const NodeSingleConnectedWithMedia = ({
     data?.node?.liveStream?.media?.sources?.length;
 
   const mediaSource = useMemo(() => {
-    if (hasLivestream) return data?.node?.liveStream?.media?.sources[0];
+    if (hasLivestream) {
+      if (data?.node?.liveStream?.media?.sources?.length >= 2) {
```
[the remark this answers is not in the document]
tristanguckenberger: Thanks for the feedback. Updated! 

The latest commits now aim to get the platform specific link by determining the platform and then providing a .filter() the following required conditions:
- resi presence in the string
- presence/lack of the .m3u8 file extension

iOS => presence of resi && presence of .m3u8
Android => presence of resi && lack of .m3u8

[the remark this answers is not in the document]
vinnyjth: Love it!

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -99,8 +99,26 @@ const NodeSingleConnectedWithMedia = ({
     data?.node?.liveStream?.isLive &&
     data?.node?.liveStream?.media?.sources?.length;
 
+  const hasResiLivestream =
+    hasLivestream &&
+    data?.node?.liveStream?.media?.sources?.length >= 2 &&
+    (data?.node?.liveStream?.media?.sources[0]?.uri?.includes('resi') ||
```
vinnyjth, line 105: maybe

`data?.node?.liveStream?.media?.sources?.some(({ uri }) => uri.includes('resi))`

tristanguckenberger: Yeah, I like that better!

tristanguckenberger:
> Just so I'm clear, android can't play m3u8 files so this simply avoids that?

That's correct

tristanguckenberger:
> also, is there a reason this is going straight into v2 and not master first?

Yeah, this fix is for a v2 church so I thought it would go to v2.

redreceipt:
this is a bug in master too right? if so, move this to master branch PR

redreceipt:
> > > > Just so I'm clear, android can't play m3u8 files so this simply avoids that?
> > > 
> > > 
> > > That's correct
> > 
> > 
> > what format does work that we are expecting? and can iOS play that other format?
> 
> For android we're using a webplayer => https://control.resi.io/webplayer/video.html?id=(id here)
> 
> I tested just the webplayer link and it reaches the RNVideo component but fails to play.
> 
> But I think we could get it to work if we add the webplayers presence as a condition on this [line](https://github.com/ApollosProject/apollos-apps/blob/v2/packages/apollos-ui-media-player/src/useSourceComponents.tsx#:~:text=const%20isWebView%20%3D%20!source%3F.uri%20%26%26%20!!source%3F.html%3B) in `useSourceComponents.tsx`:
> 
> `const isWebView = (!source?.uri && !!source?.html) || (source?.uri?.includes('resi') && source?.uri?.includes('webplayer'))`
> 
> Basically enforcing webview for both android and ios if the source uri includes 'resi' and 'webplayer'.

we're getting somewhere but I'm very confused. Our server schema does not have a `source.html` field available so what is that code in media player doing?

this is where i'm headed, the logic here should be very simple: "if android, use livestream.html"
we don't need to care about resi, we know android can't play m3u8 files so just insure that never happens by only looking for html.

vinnyjth:
@redreceipt some context that might be missing 

> we know android can't play m3u8 files 
I think Android normally can, but the encoding that RESI uses on their m3u8 doesn't work out of the box with android

> we don't need to care about resi
I thought the same thing, but we actually have code in other parts of our app to handle resi specifically. Apparently it's actually a thing? 

https://github.com/ApollosProject/apollos-apps/commit/313008eddbfe52e0abad1eee4e073d2a780c25ab#diff-0500ed0b0cbe013d7e3afd6fe50dbf47c178e1a03a6033a81d305fc7bbde4821R327

redreceipt:
> I think Android normally can, but the encoding that RESI uses on their m3u8 doesn't work out of the box with android

that's helpful. 

> I thought the same thing, but we actually have code in other parts of our app to handle resi specifically. Apparently it's actually a thing?

but looking at our schema, there is no `source.html` field???

if we fix that, shouldn't this front-end logic simply be "if android and resi, ignore the m3u8" since we should already be handling html

vinnyjth:
>but looking at our schema, there is no source.html field???

@tristanguckenberger has a seperate PR on LCBC that adds a new source that's a webview link
https://github.com/lcbc-digital/ephesus/pull/77

tristanguckenberger:
> > but looking at our schema, there is no source.html field???
> 
> @tristanguckenberger has a seperate PR on LCBC that adds a new source that's a webview link [lcbc-digital/ephesus#77](https://github.com/lcbc-digital/ephesus/pull/77)

That's my bad! I thought I had added @redreceipt on there as well, just added.

tristanguckenberger:
> O wow. I understand now. So many thoughts...ok disregard my suggestions. We have bigger problems we can fix later. This should probably still go to master though since the bug is just as relevant.

Okay, is changing this to master as simple as clicking edit in top right of this page, by the title of the PR, and then changing this drop down to 'master'?
Screen Shot 2022-06-09 at 10 34 21 AM

What else is involved in getting this into the v2 branch?

redreceipt:
that should be it. after it's merged into master, we can cherry-pick it into v2.

tristanguckenberger:
There are some differences in [master](https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-ui-connected/src/NodeSingleConnected/index.js) vs [v2](https://github.com/ApollosProject/apollos-apps/blob/v2/packages/apollos-ui-connected/src/NodeSingleConnected/index.js). Notably Master wraps ApollosPlayerContainer in ApollosPlayerConnected, which doesn't exist in V2.

@redreceipt how would you fix this? Should this be two separate PRs, one for master and one for V2?

redreceipt:
yea probably two separate PRs

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -99,8 +99,27 @@ const NodeSingleConnectedWithMedia = ({
     data?.node?.liveStream?.isLive &&
     data?.node?.liveStream?.media?.sources?.length;
 
+  const hasResiLivestream =
+    hasLivestream &&
+    data?.node?.liveStream?.media?.sources?.length >= 2 &&
```
redreceipt, line 104: don't need this line right?

tristanguckenberger: We could probably get rid of 104, but for 103, correct me if this line of thinking is wrong, but I'm thinking we still want it to check if there is a livestream because we're looking for a specific instance of a livestream that also includes 'resi' in the string.

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -99,8 +99,27 @@ const NodeSingleConnectedWithMedia = ({
...
+    data?.node?.liveStream?.media?.sources?.some(({ uri }) =>
+      uri.includes('resi')
+    );
+
   const mediaSource = useMemo(() => {
-    if (hasLivestream) return data?.node?.liveStream?.media?.sources[0];
+    if (hasResiLivestream) {
+      return data?.node?.liveStream?.media?.sources?.filter((src) => {
+        if (Platform.OS === 'ios')
+          return src?.uri?.includes('resi') && src?.uri?.includes('.m3u8');
```
redreceipt, line 113: don't need these "includes(resi)" bits, we already have that covered with `hasResiLivestream`

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -99,8 +99,27 @@ const NodeSingleConnectedWithMedia = ({
...
+        if (Platform.OS === 'android')
+          return src?.uri?.includes('resi') && !src?.uri?.includes('.m3u8');
+        return null;
```
redreceipt, line 116: dont' need this `return null`

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
tristanguckenberger: Actually, I see what you mean.

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -99,8 +99,21 @@ const NodeSingleConnectedWithMedia = ({
...
     data?.node?.liveStream?.media?.sources?.length;
 
+  const hasResiLivestream = data?.node?.liveStream?.media?.sources?.some(
+    ({ uri }) => uri.includes('resi')
+  );
+
   const mediaSource = useMemo(() => {
-    if (hasLivestream) return data?.node?.liveStream?.media?.sources[0];
+    if (hasResiLivestream) {
+      return data?.node?.liveStream?.media?.sources?.filter((src) => {
```
redreceipt, line 108:
```suggestion
      const source = data?.node?.liveStream?.media?.sources?.filter((src) => {
```

redreceipt, line 108: ok so now, we should hijack `source` like we do down below so we don't have to alter media player stuff. something like:

```
if android: return {html: <URI without m3u8>}
else: return {uri: <URI with m3u8>}
```

tristanguckenberger: How would I do this inside a .filter? or are you suggesting to remove the filter?

redreceipt: the media player shouldn't care about resi, it should play whatever is passed to it. this is the component that should decide what to pass down.

try to rewrite my suggested algorithm above, knowing you may need to adjust or remove the `filter` statement

tristanguckenberger: Okay I took a crack at it, tested it and everything works still with the latest commit.

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -99,8 +99,32 @@ const NodeSingleConnectedWithMedia = ({
...
-    if (hasLivestream) return data?.node?.liveStream?.media?.sources[0];
+    if (hasResiLivestream) {
+      const source = data?.node?.liveStream?.media?.sources?.filter((src) => {
+        if (isAndroid) return !src?.uri?.includes('.m3u8');
+        return src?.uri?.includes('.m3u8');
+      })?.[0];
+
+      if (isAndroid) {
+        return {
+          html: `<div style="position:relative;overflow:hidden;padding-top:56.25%;"><iframe allow="autoplay; fullscreen" allowfullscreen="true" src="${source?.uri}" style="position:absolute;top:0;left:0;width:100%;height:100%;border:none;"></iframe></div>`,
```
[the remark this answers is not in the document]
redreceipt: I was under the impression if you passed `{html: " https://control.resi.io/webplayer/video.html?id=(id here)"}` the media player would work fine, is that not the case?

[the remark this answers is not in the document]
tristanguckenberger: It gets dropped into an html block [here](https://github.com/ApollosProject/apollos-apps/blob/dbd3dd2fd905900d1b04c049c1dc4458a94b5d0f/packages/apollos-ui-media-player/src/WebVideo.tsx#L42). If it isn't wrapped, it just prints out the url in the web view.

Here's what happens

https://user-images.githubusercontent.com/19194337/172973733-62ae1342-e005-4dfa-9f8d-ec31462773b4.mov

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -99,8 +99,30 @@ const NodeSingleConnectedWithMedia = ({
...
   const mediaSource = useMemo(() => {
-    if (hasLivestream) return data?.node?.liveStream?.media?.sources[0];
+    if (hasResiLivestream) {
+      const androidUrl = liveStreamSources?.find(
+        (src) => !src?.uri?.includes('.m3u8')
+      );
+      if (isAndroid && androidUrl)
+        return {
+          uri: androidUrl?.uri,
+        };
```
redreceipt, line 119:
```suggestion
      if (Platform.OS === 'android) return liveStreamSources?.find((src) => !src?.uri?.includes('.m3u8'));
```

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -99,8 +99,29 @@ const NodeSingleConnectedWithMedia = ({
...
   const mediaSource = useMemo(() => {
-    if (hasLivestream) return data?.node?.liveStream?.media?.sources[0];
+    if (hasResiLivestream) {
+      const androidUrl = data?.node?.liveStream?.media?.sources?.find(
+        (src) => !src?.uri?.includes('.m3u8')
+      );
+      if (Platform.OS === 'android' && androidUrl)
+        return {
+          uri: androidUrl?.uri,
+        };
```
redreceipt, line 116:
```suggestion
      if (Platform.OS === 'android') 
        return data?.node?.liveStream?.media?.sources?.find((src) =>
         !src?.uri?.includes('.m3u8')
       );
```

redreceipt (approved):

PR #2528, opened 2022-06-09
fix: android prayer closing on launch

vinnyjth:


vinnyjth:
Same as #2527  but for master

packages/apollos-ui-prayer/src/PrayerView/index.js
```diff
@@ -49,6 +49,8 @@ const SecondaryActionButton = withTheme(
   'ui-prayer.PrayerView.SecondaryActionButton'
 )(Button);
 
+const keyboardAvoidingBehavior = Platform.OS === 'ios' ? 'padding' : 'height';
```
redreceipt, line 52: it blows my mind this is necessary, feels like react-native should handle this one. I wanna say we should abstract this into some component in ui-kit but also I'm kinda tired of adding components that we will inevitably forget about.

vinnyjth: CleanShot 2022-06-09 at 11 17 43@2x

PR #2529, opened 2022-06-09
fix: Stop Videos on Church Websites from taking over

vinnyjth:
Videos on clients websites were taking over focus on the OpenID landing screen.


vinnyjth:
CleanShot 2022-06-09 at 15 12 37

redreceipt (approved):
two very curious props they have there

PR #2530, opened 2022-06-14
chore: remove change password (deprecated)

nlewis84:
This PR removes the Change Password component and also removes the screen from User Settings. This is no longer needed as we do not use passwords in our Auth flow any more.


packages/apollos-ui-connected/src/ContentSingleFeaturesConnected/__snapshots__/ContentSingleFeaturesConnected.tests.js.snap
```diff
@@ -252,7 +252,289 @@ exports[`ContentSingleFeaturesConnected should render 1`] = `
               "paddingVertical": 16,
             }
           }
-        />
+        >
```
redreceipt, line 255: this snap being affected is odd

redreceipt (approved):

packages/apollos-ui-connected/src/ContentSingleFeaturesConnected/__snapshots__/ContentSingleFeaturesConnected.tests.js.snap
nlewis84: Hmm, let me update the snaps again and see if it clears it up.

nlewis84: I'm not sure why this file got updated. The tests all pass with this change locally, but on the CI this causes tests to fail. I'm reverting this change to see if tests pass.

PR #2534, opened 2022-06-15
chore: change styled to withTheme on FollowList for better overrides

nlewis84:
This PR switches out styled for withTheme, so that you can more easily override non-style props such as title, bordered, pill, and type. Without this change, you can only override style attributes.


redreceipt:
if you're doing this, I'd prefer using a `StyleSheet` for styles and `named` for overrides. We need to be removing `styled` and `withTheme` calls.

nlewis84:
Okay, since I'm already in this file, I'll change out all the styles to the new pattern.

redreceipt:
I was gonna say we actually have a `ThemeConsumer` you can use for class components 😂 but I like this much better. 

Only question is where is the extra margin bottom coming from showing in the snaps?

redreceipt (approved):

PR #2536, opened 2022-06-15
chore: ui-authentication android refactor

nlewis84:
This PR addresses several QoL issues for Android users in the latest authentication flow. 

On the OpenId page, the iOS version of the app gets a header, while the Android version does not. Due to the fact that the absolutely positioned, we can add a BackgroundView with the new avoidHeader functionality to work around the fact that react-navigation does not automatically move content below the header (which causes and overlapping bug).

| Before | After |
|---|---|
|Screen Shot 2022-06-15 at 12 36 14 PM|Screen Shot 2022-06-15 at 12 19 45 PM|

On all pages of our authentication flow, the back button is oddly placed (due to transparent header). Getting rid of the header causes more issues with content falling under keyboards and generally making the UX worse on Android. Following the recommendations of Google, I have removed the back button on Android devices in favor of the hardware/software back button.

https://developer.android.com/guide/navigation/navigation-custom-back#:~:text=All%20Android%20devices%20provide%20a,button%20or%20a%20software%20button.
Screen Shot 2022-06-15 at 12 57 15 PM

| Before | After |
|---|---|
|Screen Shot 2022-06-15 at 12 36 48 PM|Screen Shot 2022-06-15 at 12 32 35 PM|


nlewis84:
I tested this in light and dark mode as well and it looks 👌

redreceipt (approved):
you're the man

PR #2537, opened 2022-06-15
fix: adds content padding to avoid actionsheet

redreceipt:


nlewis84:
We lose anything significant by dropping the props spreading? I guess we will find out and be able to clear it out in all the connected locations.

redreceipt:
I logged it, the only one that's coming down was the `paddingBottom` that wasn't doing anything

nlewis84 (approved):

PR #2540, opened 2022-06-20
fix: call logout after resetToAuth

tristanguckenberger:
The purpose of this PR is to prevent android crashing when a user logs out.

The following video shows this fix gets logout working again for both ios and android:

https://user-images.githubusercontent.com/19194337/174636854-e86d08f3-0abd-40f6-8285-5f447be9fc89.mov


redreceipt:
Why was this necessary? What was happening?

tristanguckenberger:
@redreceipt The app crashes on android when the user clicks logout from UserSettings, this is what was happening before: 

https://user-images.githubusercontent.com/19194337/174640753-b2d75fdc-fb44-46a7-9de9-2b136128c99b.mov

redreceipt:
Do we know why it was crashing on Android? Just wanna make sure we understand the fix.

tristanguckenberger:
I'm not entirely sure why it's happening, but something else I've noticed is that logout will also work again in the case where we remove the `await` from [this line](https://github.com/ApollosProject/apollos-apps/blob/ea9ec04df02a1d99e9ec31300fb2b7ddbeb60089/packages/apollos-ui-connected/src/UserSettingsConnected/UserSettings.js#L139). If you look at [this callback for logging out](https://github.com/ApollosProject/apollos-apps/blob/ea9ec04df02a1d99e9ec31300fb2b7ddbeb60089/packages/apollos-ui-authentication/src/Provider.js#L97-L107), it doesn't actually return anything, so I'm thinking the await here is unneeded and may actually be what's causing this.

vinnyjth:
I wish we could figure out what error was being thrown to cause the crash, but your fix seems fine to me.

vinnyjth (approved):

tristanguckenberger:
I can also get logout to work if the **_only_** change I do is making `useLogout` async on line 33 [here](https://github.com/ApollosProject/apollos-apps/blob/ea9ec04df02a1d99e9ec31300fb2b7ddbeb60089/packages/apollos-ui-authentication/src/Provider.js#L32-L33) OR if I make `useIsLoggedIn` above it async, either fixes this, no await needed on the returns. Seems odd...

Also, unrelated to the weird thing I just mentioned above but my android sim briefly flashed this error before the app crashed entirely, it was still present in when I went to reopen the app so I got a quick screen grab. This could be helpful too:
Screen Shot 2022-06-20 at 2 03 48 PM

redreceipt:
Good info Tristan. I like removing the await so both happen simultaneously. Especially the hook isn’t and doesn’t really need to be async.

redreceipt (approved):

PR #2541, opened 2022-06-22
fix: make back btn available for all platforms

tristanguckenberger:
The purpose of this PR is to make the back button available for all platforms. [Relevant basecamp QA post here](https://3.basecamp.com/3926363/buckets/22579289/todos/5040246327)

Previously the back button was only being rendered for ios due to a bug. After some testing with the back button enabled for android, I can confirm this is working.


redreceipt:
@nlewis84 just took this out based on Google best practices, is there a good reason to put it back?

nlewis84:
@tristanguckenberger Here's that PR and my notes: https://github.com/ApollosProject/apollos-apps/pull/2536

PR #2544, opened 2022-06-27
chore: change fetch policy so that prayers show up

nlewis84:
This PR changes the fetch policy on prayers to network-only so that prayer requests show up more quickly in app.


nlewis84:
@redreceipt Shoot me down if this is a bad idea.

redreceipt:
Not sure the performance hit is worth it. With cache and network, it should be correct on the following pull.

redreceipt:
yea we should figure out why it's not there on the next pull

nlewis84:
Pulled down the latest template files to grab the changes to client.js that @calebpanza was working on. Same trouble. I'll keep at it.

PR #2545, opened 2022-06-28
fix: Remove merged duplicates

tristanguckenberger:
This PR intends to handle a pagination issue affecting the v2 release of ui-connected ([reported in this ticket](https://3.basecamp.com/3926363/buckets/22579289/todos/5052456755)) where duplicate items were ending up in the merged result of the previous-results and the fetch-more-results. This unhandled duplicate was causing the issue shown in the 'before' video below. I've yet to identify why/how this duplicate is showing up in fetchMoreResult in the first place. So determining that could change the outcome of this PR assuming we discover or know of a better way to handle this.

I'm going to initially mark this as a draft due to jumpy-ness shown in the 'after' video when scrolling up. I've been trying to identify what's causing it but haven't had much luck so far and may need help.

### **Before**
https://user-images.githubusercontent.com/19194337/176213458-3b1b8b58-0630-4feb-8df0-ed898cb40355.mov

### **After**
https://user-images.githubusercontent.com/19194337/176214439-5b375748-87e1-4177-adc9-6c493450cb31.mov


packages/apollos-ui-connected/src/utils/fetchMoreResolver/index.js
```diff
@@ -26,12 +26,16 @@ export default function fetchMoreResolver({
...
         );
+        const uniqueSet = new Set();
         const newDataWithAdditionalItems = set(
           edgePath,
           [
             ...get(previousResult, edgePath, []),
             ...get(fetchMoreResult, edgePath, []),
-          ],
+          ].filter((item) => {
+            const key = item.node?.id;
```
redreceipt, line 36: I like the idea! But this will need to be more abstract, we probably use this fetch more resolver for more than just nodes with ids.

redreceipt (changes requested):

packages/apollos-ui-connected/src/utils/fetchMoreResolver/index.js
redreceipt: Line 30 is already trying to make a set, maybe there's a way to alter that call so it looks deeper?

tristanguckenberger: I've been playing around with abstraction but I'm having trouble wrapping my mind around removing duplicates from merged arrays where the path to their unique identifiers and/or their identifiers could vary by call. Am I over thinking this? Since this issue only occurs for item.node.id, should I just handle this for item.node.id and let anything else pass through? 

Or would it be better to modify where this resolver gets called so I can pass it a unique-id-key for use with a function that takes in a key as well as the array to de-duplicate? Or should we try to detect a key? 

Also can you elaborate by what you mean by altering the call on line 30 to look deeper?

Is the lodash set function comparable to vanilla js sets? My intention with the uniqueSet vanilla js set, was to only track the ids that had been already “seen” or passed over so I’d have a way to filter the duplicate item out.

redreceipt: I tend to think this solution is a bandaid to a deeper problem. do we know why calling for the next page in the search query is returning dupes? perhaps you could try to reproduce this with just the API and debug there?

PR #2546, opened 2022-06-28
feat: Progress Bar component

calebpanza:
## Motivation
In order to display a user's progress easily on screen, we have introduced a new ProgressBar component which will visual the users progress as a percentage of a horizontal container.

## API
```
<ProgressBar step={50} />
```

| Prop | Type | desc | notes |
| --- | --- | --- | --- |
| step | number | the amount of the bar that should be filled | _while any numeric value can be accepted, the component will clamp the input between 0 and 100_ |

## Notes
The ProgressBar will automatically fill the entire width of the parent container. For any use case that needs to limit the width, it will be the responsibility of a parent container to specify the width.

## Screenshots
<p align="center">

</p>


redreceipt:
this is beautiful

camrun91 (commented):
Looks great to me

camrun91 (approved):

redreceipt (approved):

PR #2550, opened 2022-06-29
chore: Pull theme from config

vinnyjth:


scripts/setup-app.sh
```diff
@@ -9,14 +9,19 @@ IOS_BUNDLE=$(curl -s --fail -H "x-api-key: $APOLLOS_API_KEY" -H "accept: text/pl
...
 rm -rf deploys/mobile
 cp -r templates/mobile deploys/mobile
+
 cd deploys/mobile || exit 1
 
 rm -rf fastlane
 mkdir fastlane
 
+rm theme.json
+echo $THEME > theme.json
```
redreceipt, line 23: This will probably need to be printf in case there's a carriage return in an override at some point.

scripts/setup-app.sh
```diff
@@ -9,14 +9,19 @@ IOS_BUNDLE=$(curl -s --fail -H "x-api-key: $APOLLOS_API_KEY" -H "accept: text/pl
...
 
 rm -rf deploys/mobile
 cp -r templates/mobile deploys/mobile
+
 cd deploys/mobile || exit 1
 
 rm -rf fastlane
 mkdir fastlane
 
+rm theme.json
```
redreceipt, line 22: Don't need this with >

vinnyjth, line 22:
```suggestion
```

scripts/setup-app.sh
vinnyjth, line 23:
```suggestion
printf $THEME > theme.json
```

redreceipt (approved):

PR #2552, opened 2022-06-30
chore: move auth and settings back into the template

redreceipt:
motivation for this PR was so we can use local images in the auth components instead of icons. turned out to be a lot of work.

I know @calebpanza I'm such a hypocrite 😔 the bulk of these changed files are moving `ui-authentication` and `UserSettings` back to the template and then getting tests working correctly. There's some external pkg that needs to be mocked in `ui-kit` or `ui-connected` that was breaking tests so instead of going through every import to find, I simply mocked those pkgs and the pieces we're using. Which turns out to actually produce much cleaner, faster tests.


templates/mobile/src/auth/Identity/VerificationConnected.test.js
```diff
@@ -0,0 +1,19 @@
+import renderer from 'react-test-renderer';
+import { LoginContext } from '../LoginProvider';
+import { AuthContext } from '../Provider';
+import VerificationConnected from './VerificationConnected';
+
+//jest.useFakeTimers();
+
```
redreceipt, line 7:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -31,11 +31,18 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
         with:
           node-version: 14
+      - uses: ruby/setup-ruby@v1
+        with:
+          working-directory: templates/mobile
```
redreceipt, line 39:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -31,11 +31,15 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
         with:
           node-version: 14
       - run: yarn
       - run: ./scripts/test-all.sh
+      - run: |
+          yarn --ignore-scripts
```
vinnyjth, line 40: why `yarn` twice?

vinnyjth: oh word, thanks!

PR #2553, opened 2022-07-01
chore: move in onboarding and settings

redreceipt:
this is actually what I needed to move the first time. this is using local images instead of the brand icon SVG

Screen Shot 2022-06-30 at 11 55 30 AM


templates/mobile/src/Providers.js
```diff
@@ -7,8 +7,8 @@ import {
   LiveProvider,
   ACCEPT_FOLLOW_REQUEST,
 } from '@apollosproject/ui-connected';
-import { checkOnboardingStatusAndNavigate } from '@apollosproject/ui-onboarding';
 
+import { checkOnboardingStatusAndNavigate } from 'onboarding';
```
vinnyjth, line 11: I typically that this pattern but I think it makes a lot of sense here. would be really frustrating to update all our imports otherwise if we had to use relative paths.

vinnyjth (approved):

templates/mobile/src/Providers.js
redreceipt: I agree

PR #2555, opened 2022-07-06
chore: adds icons to deploy process

redreceipt:
- chore: adds icons to deploy process
- stuff


deploys/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
```
vinnyjth, line 4: 👍  better than .gitkeep

scripts/generate-app-graphics.js
```diff
@@ -0,0 +1,90 @@
+const fs = require('fs');
```
vinnyjth, line 1: love it

scripts/setup-app.sh
```diff
@@ -5,11 +5,25 @@ fi
...
+WORDMARK=$(curl -s --fail -H "x-api-key: $APOLLOS_API_KEY" -H "accept: text/plain" "$API_URL/config/$CHURCH/APP.WORDMARK_URL")
+if [ -z "$WORDMARK" ]; then echo "APP.WORDMARK_URL invalid!" && exit 1; fi
 IOS_BUNDLE=$(curl -s --fail -H "x-api-key: $APOLLOS_API_KEY" -H "accept: text/plain" "$API_URL/config/$CHURCH/APP.APPLE_BUNDLE_ID")
+if [ -z "$IOS_BUNDLE" ]; then echo "APP.APPLE_BUNDLE_ID invalid!" && exit 1; fi
 ANDROID_BUNDLE=$(curl -s --fail -H "x-api-key: $APOLLOS_API_KEY" -H "accept: text/plain" "$API_URL/config/$CHURCH/APP.ANDROID_PKG_ID")
+if [ -z "$ANDROID_BUNDLE" ]; then echo "APP.ANDROID_PKG_ID invalid!" && exit 1; fi
 SERVER_URL=$(curl -s --fail -H "x-api-key: $APOLLOS_API_KEY" -H "accept: text/plain" "$API_URL/config/$CHURCH/APP.SERVER_URL")
+if [ -z "$SERVER_URL" ]; then echo "APP.SERVER_URL invalid!" && exit 1; fi
 GOOGLE_MAPS_KEY=$(curl -s --fail -H "x-api-key: $APOLLOS_API_KEY" -H "accept: text/plain" "$API_URL/config/$CHURCH/APP_GOOGLE_MAPS_API_KEY")
+if [ -z "$GOOGLE_MAPS_KEY" ]; then echo "APP.GOOGLE_MAPS_API_KEY invalid!" && exit 1; fi
```
vinnyjth, line 24: what are the chances we could frontload all these `ifs` at the start, let them all log, and then fail if any of them fires? 

I know its' a bit more work, but it prevents that really annoying "build -> watch log -> fix -> build -> watch log -> fix" cycle if you have a sequential series of variables that need to be set.

vinnyjth: absolutely love that you are adding these explicit errors

redreceipt: oh I see, so show all of them at once, good idea

vinnyjth: yah! so if any of them error, you can see what other ones would error before the script exits. so you have a reference for fixing them all at once

redreceipt: fixed

vinnyjth (approved):
I love it

PR #2563, opened 2022-07-07
feat: Fan Effect

calebpanza:
## Motivation
One of the interactions that we are using for the Daily Habits is to display your "habit cards" as fanned out when you complete a habit. While the current application is solely for that one screen, I wanted to introduce the component into the UI-Kit cause I think that we can do some really smart stuff later on if we "standardize" our animations or at lease centralize them.

## Component API
The component is simple. You import `FanEffect` and pass through each item you want "fanned" as a child. It will then get wrapped (automatically) in a `Rotator` component and rotate accordingly.

```
<FanEffect>
      <View style={styles.square} />
      <View style={styles.square} />
      <View style={styles.square} />
</FanEffect>
```

## Architectural Decisions
Right now, the component tree assumes the first item to be the "top" item. I _think_ that this might break XML or React conventions of handling a Z axis, so I would love to get some feedback on this. So if we're using the example above:

```
<FanEffect>
      <View style={styles.square} /> <!-- this is the top later -->
      <View style={styles.square} /> <!-- this is the middle later -->
      <View style={styles.square} /> <!-- this is the bottom later -->
</FanEffect>
```


packages/apollos-ui-kit/src/FanEffect/FanEffect.js
```diff
@@ -0,0 +1,84 @@
...
+    display: 'flex',
+  },
+  rotator: {
+    position: 'absolute',
+  },
+});
+
+export const Rotator = ({ children, index }) => {
+  const entering = () => {
+    'worklet';
```
redreceipt, line 33: Sorry, I'm ignorant to front end animation stuff. What's this for?

redreceipt (approved):

packages/apollos-ui-kit/src/FanEffect/FanEffect.js
redreceipt: Worklet

calebpanza: Yeah I'll be honest, I have no idea. I ripped this from the Animated App Icon and it was in there 🤷🏻‍♂️

Do you want to block this until I figure out what it's for or leave it?

conrad-vanl: https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/worklets/

redreceipt:
This caused me to add three new articles to my safari reading list

PR #2564, opened 2022-07-11
feat: Add sentry, Amplitude

vinnyjth:


templates/mobile/src/index.js
```diff
@@ -30,6 +30,14 @@ import Auth, { ProtectedRoute } from 'auth';
 import { Onboarding } from 'onboarding';
 import Settings from 'settings';
 import Tabs from 'tabs';
+import * as Sentry from "@sentry/react-native";
+
+Sentry.init({
+  dsn: "https://83621b940dcc49ada9fd28276ec679e1@o1184370.ingest.sentry.io/6565266",
```
redreceipt, line 36: Should this be hidden?

redreceipt:
Does this cover bugs and analytics?

vinnyjth:
@redreceipt sorry - this should still be a draft

templates/mobile/src/index.js
vinnyjth: @redreceipt I'm curious about your thoughts on this 

- Why shouldn't it live in code, not that we are in multitenant land? 
- If it doesn't live in code, where should it live? `default` in the Config table?

redreceipt: yea I think it can live in code if it doesn't need to be secret

redreceipt: if it does need to be secret, then yes, default just like the google maps API key

vinnyjth: I'm labeling it not secret :) 

https://forum.sentry.io/t/dsn-private-public/6297/2

templates/mobile/src/amplitude.js
```diff
@@ -0,0 +1,43 @@
+import { Amplitude } from '@amplitude/react-native';
+import { getVersion, getBundleId, getApplicationName } from 'react-native-device-info';
+
+const ampInstance = Amplitude.getInstance();
+ampInstance.init('2678bfa57b62e7a47f77505edd705c70');
+
+export const track = ({ eventName, properties = null }) => console.warn(eventName) || ampInstance.logEvent(eventName, properties);
```
vinnyjth, line 7:
```suggestion
export const track = ({ eventName, properties = null }) => ampInstance.logEvent(eventName, properties);
```

vinnyjth:
CleanShot 2022-07-12 at 10 47 18@2x

vinnyjth:
CleanShot 2022-07-12 at 10 47 46@2x

templates/mobile/android/sentry.properties
```diff
@@ -0,0 +1,4 @@
+defaults.url=https://sentry.io/
+defaults.org=differential-ka
+defaults.project=apollos-cluster-app
+auth.token=932242b637784c23ae6c4112c035429b5ce87c15646943f284c7a6629c981f1c
```
redreceipt, line 4: Should this be secret?

templates/mobile/ios/Podfile.lock
```diff
@@ -706,14 +730,14 @@ SPEC CHECKSUMS:
   Flipper-RSocket: d9d9ade67cbecf6ac10730304bf5607266dd2541
   FlipperKit: d8d346844eca5d9120c17d441a2f38596e8ed2b9
   fmt: ff9d55029c625d3757ed641535fd4a75fedc7ce9
-  glog: 85ecdd10ee8d8ec362ef519a6a45ff9aa27b2e85
+  glog: 5337263514dd6f09803962437687240c5dc39aa4
```
redreceipt, line 733: This is probably wrong

templates/mobile/ios/Podfile.lock
```diff
@@ -706,14 +730,14 @@ SPEC CHECKSUMS:
...
   hermes-engine: bf7577d12ac6ccf53ab8b5af3c6ccf0dd8458c5c
   libevent: 4049cae6c81cdb3654a443be001fb9bdceff7913
   OneSignalXCFramework: 10648e60ff3d335fbffa62d9b44b700b56d49b51
   OpenSSL-Universal: 1aa4f6a6ee7256b83db99ec1ccdaa80d10f9af9b
   Permission-Notifications: 9c6b5cc4f0e6599e9fc3395b77cebddc48f1be41
   Permission-PhotoLibrary: 8227a6ed9f6a971537afe63742d54f5f23a38fe2
-  RCT-Folly: 803a9cfd78114b2ec0f140cfa6fa2a6bafb2d685
+  RCT-Folly: a21c126816d8025b547704b777a2ba552f3d9fa9
```
redreceipt, line 740: And this

templates/mobile/src/amplitude.js
```diff
@@ -0,0 +1,43 @@
+import { Amplitude } from '@amplitude/react-native';
+import { getVersion, getBundleId, getApplicationName } from 'react-native-device-info';
+
+const ampInstance = Amplitude.getInstance();
+ampInstance.init('2678bfa57b62e7a47f77505edd705c70');
```
redreceipt, line 5: This should be secret

vinnyjth: I did some poking around, and general consensus is that it's okay to expose this API key. In a browser environment you would be able to snoop this.

templates/mobile/ios/Podfile.lock
vinnyjth: thanks as always

templates/mobile/android/sentry.properties
vinnyjth: It's the same as the one above, so no. 

https://github.com/getsentry/sentry-docs/blob/9d93a097bfb6fdb0b7cf283afc62ea24bc729535/src/docs/product/sentry-basics/dsn-explainer.mdx#dsn-utilization

vinnyjth: errr nvm this is used for debug symbol uploading. I'll look into it.

templates/mobile/ios/Podfile.lock
vinnyjth:
```suggestion
  RCT-Folly: 803a9cfd78114b2ec0f140cfa6fa2a6bafb2d685
```

templates/mobile/ios/Podfile.lock
vinnyjth:
```suggestion
  glog: 85ecdd10ee8d8ec362ef519a6a45ff9aa27b2e85
```

vinnyjth:
@redreceipt I caved and moved all variables to the DB.

redreceipt:
General curiosity, why the move to sentry over bugsnag?

vinnyjth:
@redreceipt a few reasons that all came together, 

- I have found Bugsnag's react-native support lackluster, hard to get the right details from error messages (could be a setup issue on our end) 
- We use Sentry for the Shovel, and we can't migrate to Bugsnag there. 
- Because we used it on Newspring, and according to Conrad, liked it :)

templates/mobile/src/amplitude.js
```diff
@@ -0,0 +1,44 @@
...
+import { getVersion, getBundleId, getApplicationName } from 'react-native-device-info';
+import ApollosConfig from '@apollosproject/config';
+
+const ampInstance = Amplitude.getInstance();
+ampInstance.init(ApollosConfig.AMPLITUDE_KEY);
+
+export const track = ({ eventName, properties = null }) => ampInstance.logEvent(eventName, properties);
+
+export const identify = () => {
+    const data = client.readQuery({
```
redreceipt, line 11: this was the linter giveaway. `client` and `gql` here are undefined

templates/mobile/src/amplitude.js
```diff
@@ -0,0 +1,44 @@
...
+      query: gql`
+        query currentCampusId {
```
redreceipt, line 13:
```suggestion
        query GetAnalyticsUser {
```

I'm imagining a debug scenario where no one is logged in yet and this query errors out and we have a hard time figuring out what's going on

templates/mobile/src/amplitude.js
vinnyjth: ah thanks! we swallow the error so I never saw it 😬

redreceipt (approved):

PR #2570, opened 2022-07-14
chore: update bottom sheet dependency

calebpanza:
## Motivation
The Bottom Sheet dependency that we use launched a major update to v4 which includes support for Reanimated v2. They also boast stability and performance improvements in their [release blog](https://gorhom.github.io/react-native-bottom-sheet/blog/bottom-sheet-v4).

As we continue to expand the exposure of the Comment List Feature and the visual language of the User Comment Input, it became clear that we will want this update in order to offer a better "bottom sheet experience" to users.

## Solution
I started by updating the Bottom Sheet dependency and lock file. From there, it was just a matter of confirming the update worked by testing and updating some of the properties on the `AddCommentInput` component.

## Gotcha's
This update seems to be doing some WILD stuff to snapshots. In order to use the `BottomSheetModal` component, we need to wrap our application in a `BottomSheetModalProvider` which gets included in the `Providers` at the UI Kit level. There were some styling changes introduces in this upgrade which means is that _every_ test that uses the UI Kit providers is now gonna fail.

There's really nothing that can be done about this other than just accept the changes, so I've gone ahead and hit up one big `yarn test -u` but still wanted to acknowledge it cause the files changed is... well.. significant. Lol.


calebpanza:
@redreceipt are you good with this PR?

redreceipt:
it's a big one so I'll want to test it as well. love the motivation. I'll make time to look it over.

calebpanza:
@redreceipt would the codecov errors just be an indication that there are new props available in this update that we're not covering in snapshots/tests?

redreceipt:
Something like that. But I think our codecov is messed up. Every PR is showing a 5% drop from 77 to 72 which isn't right

calebpanza:
@redreceipt does that mean that I'm safe to just merge this??

redreceipt:
yep!

PR #2571, opened 2022-07-14
fix: Comment Action Menu Touchability on Android

calebpanza:
## Motivation
I happened to be testing something that had to do with Comments and tried to edit a comment on an Android device. I was unable to tap the `ActionMenu` and went into attempting to fix it

## Solution
Basically, Android is not applying the padding to the Touchable, making the touch target so small that you can't interact with it. Instead of applying the padding to the Touchable directly, we're introducing a padded view _inside of_ the touchable and that seems to work

While working on the patch, I took some liberty to adjust/clean up some of the legacy styling that was on this component. Probably not 10000% necessary, but it felt appropriate to update some parts of it 🤷🏻‍♂️


packages/apollos-ui-kit/src/Comment/ActionMenu.js
```diff
@@ -12,22 +11,38 @@ const ActionMenuIcon = withTheme(({ theme: { colors, sizing } }) => ({
...
-  <ActionPosition>
-    <ActionTouchable {...props}>
-      <ActionMenuIcon />
-    </ActionTouchable>
-  </ActionPosition>
-);
+  // Generate our style sheet based on the current theme
+  // We're using the React.useMemo hook for optimization,
+  // the Styles object will be re-generated if the theme changes
+  const styles = useMemo(() => createStyles(theme), [theme]);
```
redreceipt, line 36: You sure stylesheets aren't memoized by default? I thought the was the point of defining outside the component. For the built in performance increases. These extra layers seem overkill.

calebpanza: Started using this approach due to an article that I read on integrating theming into your RN StyleSheets. Basically, the article suggested that defining a style (like `style.container`) as a function instead of an object bypasses the built in memoization cause the object is not defined at build time, but rather runtime when you call `style.container(theme)`.

So what this does is move the creation of the stylesheet to be at runtime when we have access to the `theme` object and then adds an _extra_ layer of performance security by memoizing that result since the function would run (and re-create the stylesheet) on every render

calebpanza: I'm not like 100% married to this approach. Just thought it was an interesting read that made some good points so I threw it in like this. Down to change it if you disagree, though. Syntax is just as easy either way. Lol

redreceipt: This is very interesting. It's made me want to dive in and really understand some things. I started using the functional stylesheet pattern on a hunch but I actually can't find any documentation confirming it should work.

redreceipt (approved):

PR #2572, opened 2022-07-15
fix: query for the correct likedCount so that it displays on User Settings

nlewis84:
This PR updates the query being used to display the number of liked content items on User Settings. I need to still update some snaps.

| Before | After |
|---|---|
|Simulator Screen Shot - iPhone 13 - 2022-07-15 at 08 47 47 | Simulator Screen Shot - iPhone 13 - 2022-07-15 at 10 26 54|


nlewis84:
Not expecting any snaps to change. Waiting to see what the CI gives me

packages/apollos-ui-connected/src/UserSettingsConnected/UserSettings.js
```diff
@@ -73,7 +73,11 @@ const UserSettings = () => {
       }
 
       likedContent {
-        totalCount
+        edges {
+          node {
+            likedCount
```
calebpanza, line 78: This is going to show the number of likes that a specific node has received. This isn't gonna show the total number of items that a user has liked...

calebpanza (changes requested):

packages/apollos-ui-connected/src/UserSettingsConnected/UserSettings.js
redreceipt: yea I agree, this doesn't seem right to me. we want the total for all content here, not just the first item

nlewis84:
Hmm I see what y'all are saying then there is a problem with how we are aggregating counts of liked content for a user because it is always zero.

PR #2578, opened 2022-07-22
fix: android default buttons

nlewis84:
- fix: android buttons look better now
- fix: also adjust borderColor based on Platform

This PR addresses an issue where transparent look great on iOS, but not Android. I have been making these changes to apps  over the last several months via overrides. This PR will just address the issue in Core.

## Screenshots
|FollowFeatureDark|FollowFeatureLight|PrayerViewDark|PrayerViewLight|
|---|---|---|---|
|Screenshot_1658497758|Screenshot_1658501323|Screenshot_1658497935|Screenshot_1658501302|
|FixedFollowFeatureDark|FixedFollowFeatureLight|FixedPrayerViewDark|FixedPrayerViewLight|
Screenshot_1658500773|Screenshot_1658500860|Screenshot_1658500782|Screenshot_1658500854|

PR #2580, opened 2022-07-25
chore: cherry pick of android button fix

nlewis84:
This is a cherry pick of a fix from master. https://github.com/ApollosProject/apollos-apps/pull/2578


vinnyjth:
looks like we need to update snaps

redreceipt (approved):

nlewis84:
@vinnyjth I think tests are broken on the CI. This is testing backend packages and not front end packages. Let me see if if I can figure this out.

nlewis84:
Ahh, yeah because it is matching all packages in the `packages` folder and we have front and backend packages there on this branch. I'm going to see if I can adjust that script to only test folders that have `ui` in the title

nlewis84:
@vinnyjth The script now only targets directories in the `packages` folder that start with `apollos-ui`. Let me know if you are down with this change.

PR #2583, opened 2022-07-26
fix: backgroundColor should match borderColor

nlewis84:
Fixing a mistake I made. Background Color should match Border Color for these Android buttons.


nlewis84:
Especially since my previous change to these colors didn't cause any snapshot to need to be regen'd

nlewis84:
CI passed. Not sure why I had snapshot updates locally. @vinnyjth this is good to go.

PR #2584, opened 2022-07-26
fix: backgroundColor should match borderColor

nlewis84:
Fixing a mistake I made. Background Color should match Border Color for these Android buttons.


nlewis84:
hmmmm test failed 🫤

PR #2589, opened 2022-07-28
fix: fix fallback for prayer request requestor

tristanguckenberger:
### **Purpose**
The purpose of this PR is to fix the 'request' fallback used in the case where neither `prayer.requestor?.nickName` or `prayer.requestor?.firstName` are present.

**Before**
Screen Shot 2022-07-28 at 14 03 55

**After**
Screen Shot 2022-07-28 at 14 03 20


packages/apollos-ui-prayer/src/screens/PrayerScreen/index.js
```diff
@@ -73,6 +73,11 @@ const PrayerScreen = ({
 
   const hasPrayed = data?.prayed?.success || prayer.isPrayed;
 
+  const requestorName =
```
vinnyjth, line 76: I think this logic is a bit off, if only `nickName` is defined, it's going to return `requestor?.firstName` instead of `requestor?.nickName`

What about something like 
```
const requestorName = [prayer.requestor?.nickName, prayer.requestor?.firstName, 'request'].filter(Boolean)[0]
```

tristanguckenberger: Yep you're right, totally overlooked that! This proposed solution returns an array of ["null", "request"] (seems a little weird that null is a string here...)so if we select the 0th index we select "null".

packages/apollos-ui-prayer/src/screens/PrayerScreen/index.js
```diff
@@ -73,6 +73,14 @@ const PrayerScreen = ({
 
   const hasPrayed = data?.prayed?.success || prayer.isPrayed;
 
+  const requestorName = [
+    prayer.requestor?.nickName,
+    prayer.requestor?.firstName,
+    'request',
+  ].filter((name) => {
+    return name !== null && name !== 'null' && name !== undefined;
```
vinnyjth, line 81:
```suggestion
    return name != null && name !== 'null'
```

`!= null` will catch both `null` and `undefined` :D

packages/apollos-ui-prayer/src/screens/PrayerScreen/index.js
vinnyjth: cool! good catch on the string `null`. It's a bummer that our server is sending `"null"` over the wire.

packages/apollos-ui-prayer/src/screens/PrayerScreen/index.js
tristanguckenberger: Oh nice, good to know!

vinnyjth (approved):

PR #2590, opened 2022-07-28
feat: Validate phone numbers better on the front end

vinnyjth:


templates/mobile/src/auth/IdentityConnect/phoneValidation.js
```diff
@@ -0,0 +1,14 @@
...
+
+export const test = (value) => {
+  if (value === null) {
+    return true;
+  }
+  try {
+    const number = new PhoneNumber(value, 'US');
+    return number.isValid();
+  } catch (e) {
+    console.log(e);
```
vinnyjth, line 11:
```suggestion
```

PR #2592, opened 2022-07-29
feat: Daily Habit Prayer

calebpanza:
Adds the UI for the Daily Prayer Habit


vinnyjth:
Looks like a linter error in: apollos-ui-connected/src/DailyHabitsConnected/PrayerExperience/PrayerSwipeScreen.js

packages/apollos-ui-connected/src/hooks/useCurrentPersonProfile.js
```diff
@@ -0,0 +1,31 @@
+import gql from 'graphql-tag';
+import { useQuery } from '@apollo/client';
+
+export const GET_CURRENT_USER_PROFILE = gql`
+  query getCurrentPersonProfile {
```
vinnyjth, line 5: We use an identical query here: 
https://github.com/ApollosProject/apollos-apps/blob/d179acc7d19369d874664c0eff965589b919b2d9/packages/apollos-ui-connected/src/UserAvatarConnected/getUserPhoto.js#L3

would it be worth consolidating?

templates/mobile/src/client/cache.js
```diff
@@ -144,6 +144,11 @@ const cache = new InMemoryCache({
         followedBy: relayStylePagination(),
       },
     },
+    Query: {
+      fields: {
+        prayerRequestsPaginated: relayStylePagination(),
```
vinnyjth, line 149: very cool

packages/apollos-ui-connected/src/hooks/useCurrentPersonProfile.js
calebpanza: Yeah I think so! My suggestion would be to consolidate and use the hook moving forward. It's a pattern I've used for quite a while and have become partial to.

But, of course, open to suggestions and other preferences, too! Just lmk

templates/mobile/src/client/cache.js
calebpanza: It's working like a charm! Really pleased with how this API turned out. I _think_ it would probably be worth pulling out the pagination to its own file at some point. Idk what the directory hierarchy could/should be, here?

packages/apollos-ui-connected/src/DailyHabitsConnected/DailyHabitNavigationStack/DailyHabitsNavigationStack.js
```diff
@@ -59,8 +65,11 @@ const DailyHabitsNavigationStack = () => {
 
       <Screen
         name={ScreenNames.PrayerInput}
-        component={PrayerInputScreen}
         sharedElements={sharedElements}
+        component={PrayerInputScreen}
+        initialParams={{
+          closeDailyHabits,
```
redreceipt, line 71: is there a reason this will ever need to be anything other than `goBack`? if not, you could define over there instead of passing through params

packages/apollos-ui-connected/src/DailyHabitsConnected/PrayerExperience/PrayerInputScreen.js
```diff
@@ -8,36 +8,32 @@
...
         leadingItem={<NavigationActions.Back />}
-        trailingItem={<NavigationActions.Done onPress={next} />}
+        trailingItem={<NavigationActions.Done onPress={onClose} />}
       />
-      <BackgroundView style={styles.container}>
-        <H4>Users Submit a Prayer</H4>
-      </BackgroundView>
+      <PrayerView
+        primaryActionText={'Share Prayer'}
+        onPressPrimary={() => null}
```
redreceipt, line 29: what is this function supposed to do?

packages/apollos-ui-connected/src/DailyHabitsConnected/PrayerExperience/PrayerInputScreen.js
```diff
@@ -8,36 +8,32 @@
...
+      >
+        <PrayerCard
+          profile={profile}
+          title="What would you like prayer for?"
+          onPrayerChangeText={() => null}
```
redreceipt, line 34: what is this function supposed to do?

packages/apollos-ui-connected/src/DailyHabitsConnected/DailyHabitNavigationStack/DailyHabitsNavigationStack.js
calebpanza: OH! Yes this needs to be done this way. In order to close the _entire_ Daily Habit (which is part of the main app navigator) I need to capture the navigation object _before_ rendering the second Navigator (ln:36)

I think a better option _could_ be to use a Context Provider and hook in order to make it so that we're not prop drilling as much?

packages/apollos-ui-connected/src/DailyHabitsConnected/PrayerExperience/PrayerInputScreen.js
calebpanza: Placeholder for what will be capturing/sending user input back to the server

packages/apollos-ui-connected/src/DailyHabitsConnected/PrayerExperience/PrayerInputScreen.js
calebpanza: Placeholder for what will be capturing/sending user input back to the server

redreceipt:
getting a 400 when trying to submit a prayer

https://user-images.githubusercontent.com/2659478/183226575-dfc1038f-4037-4f26-bdf1-82185700f9a0.mov

calebpanza:
@redreceipt the share portion wasn't actually connected, yet. I'm gonna spend a few minutes connecting that cause I see that the `ShareSheetConnected` got merged in now

calebpanza:
https://user-images.githubusercontent.com/45076058/183497194-57d4bce7-7bce-4253-a1a5-a3ddb1f517c4.mp4

calebpanza:
@redreceipt ready for another review on this guy for the **FULL** prayer flow

redreceipt:
@calebpanza I'm still getting the same 400 error when trying to submit a prayer as before

calebpanza:
That's so strange.. what API are you pointed to??

redreceipt:
@calebpanza I figured it out, apollos demo doesn't have a name (problem 1, is being fixed), we're still sending prayers to rock (potentially problem 2?)

Screen Shot 2022-08-09 at 12 08 00 PM

calebpanza:
Yeah so I hooked into the existing Prayer Posting endpoints. We didn't add any endpoint for this, so it's functionality that exists in today's production app(s). Did you want something adjusted to this flow given the encounter?

redreceipt (approved):
good work :)

PR #2595, opened 2022-08-01
chore: added a default 0 for following and followers

nlewis84:
Prevents the error below by having a default value, similar to likedCount. On v3.


nlewis84:
This is wrong....I'm noticing now after looking closer at it.

PR #2597, opened 2022-08-01
fix: wrap all touchables in a view

nlewis84:
This wraps all touchables in a view so that Android users can interact with them. Does not change the user experience for iOS users.


redreceipt (approved):

nlewis84:
Hmmm, these `connected` tests... 🧐

PR #2599, opened 2022-08-01
chore: Version Override

vinnyjth:
CleanShot 2022-08-01 at 14 20 04@2x

Allows us to (optionally) supply an initial version. Will only be used if it's higher than the current app store version (+1) so there's no reason to ever reset/unset it if we forget.


redreceipt:
I approved blindly, but what's the use case here 😂

vinnyjth:
> I approved blindly, but what's the use case here 😂

For deploying apps that have a larger version in testflight than what they have deployed in the App Store. 

I don't see an obvious way to get the version from TestFlight, so it seems easier to set an initial version :)

redreceipt:
Perhaps a better question, why would the version need to be lower?

vinnyjth:
See latest celebration build for an example: 

https://app.bitrise.io/build/100e20ef-15e4-4d91-8843-f1e8c766de4e

Latest App Version is 5.19, but the latest Test Flight version is 12.0.0

vinnyjth:
> Perhaps a better question, why would the version need to be lower?

I don't think it can be lower :)

vinnyjth:
@redreceipt Lol, fixed a typo in PR description ;)

redreceipt:
Fastlane may have a flag to overwrite the current edit version but I'm not positive

redreceipt:
Regardless, I like us having the ability to set the version anyway so good add

PR #2600, opened 2022-08-01
fix: Follow List Errors

calebpanza:
## Motivation
There were a couple of errors that were reported by @nlewis84 on the Celebration app. While I can't replicate those exact errors here, we did find some pretty obvious enhancements/errors. 

While we're not sure why we're seeing an error in 1 environment and not another, the solution presented here is solving the error on Celebration, so I figured it would be best to introduce it here.

### Button Imports
Import the Confirm, Hide and Request buttons directly from their files instead of the index to resolve import errors

### Cache Merge
Provides better default values for the pagination cache so that the UI can always expect _some_ value


redreceipt (approved):

vinnyjth:
thanks - just saw this crash the latest celebration build as well

PR #2604, opened 2022-08-02
chore: makes dark mode graphics optional

redreceipt:
its optional in code, this just mimics that in the setup


vinnyjth:
This is great but they aren't optional in code :/

vinnyjth:
Pardon my phone screenshot

image

I would love to make them optional but for all three apps I ported last week the "light" graphics were unacceptable in dark mode.

redreceipt:
> This is great but they aren't optional in code :/

you're right. what about my latest change?

redreceipt:
we'll leave them required

PR #2605, opened 2022-08-03
chore: switch wordmark and logo in app

redreceipt:


vinnyjth:
@redreceipt  I switched those intentionally - I believe the wordmark is supposed to be at the top of the tab screen and the logo is supposed to be on the landing page.

If you switch them, you will also need to update the retool dashboard to reflect that the aspect ratio required for each image is swapped. I can show you how to do that.

redreceipt:
the celebration mock shows it like this?
Screen Shot 2022-08-03 at 9 03 40 AM
Screen Shot 2022-08-03 at 9 03 46 AM

redreceipt:
and this is whats currently in master
IMG_9D35E4C89396-1

vinnyjth:
@redreceipt master is in a bad spot because the images were never update appropriately. 

While celebration mocks show a square for the icon, all other figma files I looked at had a horizontal wordmark. I confirmed this with @nlewis84 earlier last week: https://differential.slack.com/archives/C03BRNGDCPK/p1658945740904299

 

Here's is what Celebration looks like in this PR. 
CleanShot 2022-08-03 at 09 13 26@2x

And what it looks like in master:
IMG_79C312FAF446-1

vinnyjth:
It's not the same - here's crossing Landing and Tab
IMG_1330
IMG_1331

Here's what the app is currently expecting: 

**Logo:** 
Used in landing screen.
Aspect Ratio: 3.5

**Wordmark**
Used at top of tab screen
Aspect Ratio: 4

If we change from those values, we will need to either make the aspect ratio dynamic depending on the size of the image, or update Retool and cloudinary to crop to the appropriate new image size.

vinnyjth:
@redreceipt 
> that shows the wordmark and logo being the same? 

They are the same image in both, but different aspect ratios.

vinnyjth:
If I have the wrong idea here, I'm totally fine with that! But you gotta tell me what the correct scenario is :) I'm trying to match current designs I've seen in figma the bestI can.

PR #2608, opened 2022-08-03
chore: Test fixes v2

vinnyjth:


packages/apollos-ui-test-utils/src/index.js
```diff
@@ -11,21 +11,30 @@ import fragmentTypes from './fragmentTypes.json';
 
 async function renderWithApolloData(
   component,
-  existingTree,
-  { renderCount = 1 } = {}
+  existingTreeOrOpts,
```
redreceipt, line 14: this is weird to me, it can be a react component or options? and then the third param is also options?

theme.json
```diff
@@ -0,0 +1 @@
+{"colors":{"primary":"#3291e2","secondary":"#99a8b4","tertiary":"#bd9a5f"},"typography":{},"overrides":{}}
```
redreceipt, line 1: Did you mean for this to be here?

vinnyjth: nope
thanks

packages/apollos-ui-test-utils/src/index.js
vinnyjth: yah, I ended up ditching `existingTree` since we don't use it

redreceipt (approved):

PR #2609, opened 2022-08-03
chore: adds git status to CI

redreceipt:


redreceipt:
looks like it works

Screen Shot 2022-08-04 at 3 51 27 PM

vinnyjth (approved):

PR #2611, opened 2022-08-04
fix: adjust sed pattern

tristanguckenberger:
The purpose of this PR is to ensure we return the primary theme color instead of the tertiary theme color. I don't know much about regex or sed for that matter so take in this info with a mound of salt. 

@vinnyjth I tried the pattern you shared with me in slack and wasn't able to get it working. 

I did find [this stack overflow answer](https://stackoverflow.com/a/39752929) and it does the trick. Here's a [breakdown of the pattern](https://regex101.com/r/7S3Zn9/2).


tristanguckenberger:
> The purpose of this PR is to ensure we return the primary theme color instead of the tertiary theme color. I don't know much about regex or sed for that matter so take in this info with a mound of salt.
> 
> @vinnyjth I tried the pattern you shared with me in slack and wasn't able to get it working.
> 
> I did find [this stack overflow answer](https://stackoverflow.com/a/39752929) and it does the trick. Here's a [breakdown of the pattern](https://regex101.com/r/7S3Zn9/2).

Disregard 😅 I just needed to escape the double quotes, it's working! Tested and it's grabbing the right color, thanks for this @vinnyjth!

**(The current primary color, `#313e48` is highlighted in each window in the screenshot below)**
Screen Shot 2022-08-04 at 02 27 09


Also just spun up a local instance and sure enough, it works like charm!

**Splash Screen**
Screen Shot 2022-08-04 at 02 40 34

**Landing Screen**
Screen Shot 2022-08-04 at 02 41 01

scripts/setup-app.sh
```diff
@@ -48,7 +48,7 @@ cp -r templates/mobile deploys/mobile
 cd deploys/mobile || exit 1
 
 printf "%s" "$THEME" >theme.json
-PRIMARY=$(grep "primary" theme.json | sed -E "s/.*(#[[:alnum:]]+).*/\1/")
+PRIMARY=$(grep "primary" theme.json | sed -E "s/.*\"primary\":\s?\"(#[[:alnum:]]+).*/\1/")
```
redreceipt, line 51:
```suggestion
PRIMARY=$(grep "primary" theme.json | sed -E "s/.*\"primary\":[[:space:]]?\"(#[[:alnum:]]+).*/\1/")
```

Nice work! I think we may to use the Mac notation for spaces here. Sometimes it doesn't work with the normal symbols. Verify this though.

redreceipt (approved):

vinnyjth:
Well done!

PR #2613, opened 2022-08-04
feat: Daily Habit Feature Release

calebpanza:
## Motivation
This PR contains the final details required to launch the fully connected Daily Spiritual Habit Experience in its MVP form.

For a full product demo, please check out this [Loom](https://www.loom.com/share/8fdc0b7cf08f420281fa09dff26fa218)

## How to Test
You need to make sure that you have your Feature set up correctly in the dB. There is nothing inside of Staging nor production right now for this, but this configuration _should_ get you set up quickly

There is an API dependency on these changes: please make sure to use this [associated API branch](https://github.com/ApollosProject/apollos-cluster/pull/271) for it all to work as expected

1. Inside of the `feature` table, add a new feature:
```
type: **DailyHabitCardList**
apollos_type: **DailyHabitCardListFeature**
church_slug: **apollos_demo**
data: {
  "type": "DailyHabitCardList",
  "title": "Daily Spiritual Habits",
  "subtitle": "Boost your daily habit with a warm-up today.",
  "algorithms": [
    {
      "type": "HABIT_GRATITUDE"
    },
    {
      "type": "HABIT_PRAYER"
    },
    {
      "type": "HABIT_READING",
      "arguments": {
        "categoryIDs": [
          "5bf609c1-5689-421f-a27a-118d5b7006a7"
        ]
      }
    }
  ]
}
```

*You might need to update your `categoryIDs` inside of the `data`
**HABIT_READING requires that the Content Item includes a Scripture Feature inside of it. If you notice that it's not working as expected, make sure that the `categoryIDs` that are being chosen render Content Items that include at least 1 Scripture Feature

2. Add the new Feature to a Feature Feed (Home is the easiest)

Now that you're ready to test everything, go ahead and pop around on each of the different experiences. The "intended" use case would be to hit Gratitude then Prayer then Reading, but they aren't required to go in that order.

## What makes an Experience "Complete"?
**Gratitude:** posting a Gratitude
**Prayer:** praying for another user
**Read:** open up a Scripture Card on today's Daily Article


packages/apollos-ui-connected/src/FeaturesFeedConnected/FeaturesFeedConnected.js
```diff
@@ -127,6 +127,7 @@ class FeaturesFeedConnected extends PureComponent {
           {({ error, data, loading, refetch }) => {
             const features = get(data, 'node.features', []);
             this.refetchRef({ refetch, id: 'feed' });
+            console.log({ features });
```
vinnyjth, line 130:
```suggestion
```

redreceipt:
If I'm understanding correctly, there's an API PR that this one depends on to work correctly, right? If so, we should wait to merge that before marking this one ready to review.

calebpanza:
@redreceipt absolutely!

camrun91:
@redreceipt the pending API PR got merged so this should be able to be tested now from master.

redreceipt:
tagging @conrad-vanl to give his approval on the UI/UX

packages/apollos-ui-connected/src/features/DailyHabitFeature.js
```diff
@@ -0,0 +1,165 @@
...
+  Card,
+  H4,
+  H3,
+  BodySmall,
+  TouchableScale,
+  HorizontalTileFeed,
+} from '@apollosproject/ui-kit';
+import Icon from '@apollosproject/ui-kit/src/Icon';
+
+function useStyleSheet() {
```
conrad-vanl, line 26: Note for other PR reviewers / @redreceipt  and @vinnyjth : @calebpanza is starting a new standard here for constructing stylesheets. I don't have an opinion on it one way or the other yet...I kind of like it. But also 🤔 how we figure out standards like this vs having every coder adopt a slightly different practice.

packages/apollos-ui-connected/src/features/DailyHabitFeature.js
```diff
@@ -0,0 +1,165 @@
...
+    <Card style={styles.card}>
+      <View style={styles.cardContent}>
+        <View>
+          {!!chip && (
+            <View style={styles.chip}>
+              <Icon
+                name={chip?.iconName}
+                size={18}
+                fill={theme.colors.text.action}
+                weight="bold"
```
conrad-vanl, line 101: I think this should be solid? I actually don't see this represented in the prototype anywhere....we can have icons on habits? 🙌

packages/apollos-ui-connected/src/features/DailyHabitFeature.js
```diff
@@ -0,0 +1,165 @@
...
+              />
+              <H4 style={styles.chipText}>{chip?.labelText}</H4>
```
conrad-vanl, line 103: This is `H3` in Figma
Screen Shot 2022-08-16 at 8 02 42 AM

packages/apollos-ui-connected/src/features/DailyHabitFeature.js
```diff
@@ -0,0 +1,165 @@
...
+  const [styles] = useStyleSheet();
+
+  if (featureId && refetch && refetchRef) {
+    refetchRef({ refetch, id: featureId });
+  }
+
+  function renderItem({ item }) {
+    return (
+      <TouchableScale
+        hitSlop={{ bottom: 12, left: 24, right: 24, top: 16 }}
```
conrad-vanl, line 134: nice job getting hitslop in here!

packages/apollos-ui-connected/src/features/DailyHabitFeature.js
```diff
@@ -0,0 +1,165 @@
...
+        <HabitCard {...item} />
+      </TouchableScale>
+    );
+  }
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.header}>
+        {Boolean(feature?.title) && (
+          <H4 style={styles.secondaryText}>{feature.title}</H4>
```
conrad-vanl, line 146: We should probably abstract feature titles (this `H4` `H3` grouping) into its own component since every feature uses it.

We have a `FeatureTitles` component in Figma:

Screen Shot 2022-08-16 at 8 05 16 AM

packages/apollos-ui-connected/src/features/ScriptureFeature.js
```diff
@@ -72,6 +74,16 @@ const ScriptureFeature = ({
   showShareButton,
   title,
 }) => {
+  const [interactWithNode] = useMutation(INTERACT_WITH_NODE, {
+    variables: {
+      nodeId: featureId,
+      action: 'VIEW',
+    },
+    update: (cache) => {
+      cache.evict({ __typename: 'DailyHabitFeatureConnected' });
```
conrad-vanl, line 83: calling out this other usage of `cache.evict`...I'd like to understand why this is needed.

packages/apollos-ui-connected/src/hooks/useDailyHabitCardListFeature.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+`;
+
+export default function useDailyHabitCardListFeature(id) {
+  const query = useQuery(GET_DAILY_HABIT_CARD_LIST_FEATURE, {
+    variables: {
+      id,
+    },
+    fetchPolicy: 'network-only',
```
conrad-vanl, line 55: Why `network-only` ? This means we'll never get cached data here ... so every time this component mounts we have to roundtrip the server? Why can't we use `cache-and-network`?

conrad-vanl (commented):
@calebpanza first off, great work! I'm going to pick apart the UI as I think it's important that we start to get critical of UX. I don't think anything I see is merge blocking. However, before this PR is merged, I'd like to get all of the non-blocking TODOs migrated to Basecamp [here](https://3.basecamp.com/3926363/buckets/27088370/todos/5221910607) (cc @vinnyjth not sure how we should split this list up).

### General

- [ ] **Loading States**: I want to make sure we're showing loading states in situations that we have no data yet and we know we are waiting on network activity. The loading states should be something like:

1. No Cached Data, waiting on network activity: Show Loading State
2. Cached Data, waiting on network activity: Show Cached Data
3. No Data / Error, no network activity: Show empty/default state and/or error
4. Data, network refetching: Same as state 2 above5. 

In other words, we want to prevent a flash of default info before network activity has finished. I think the biggest example of this right now is in the "Be the first to journal" component that shows under your journal prompt. Every time this popover opens, it appears to initially mount with empty state ("Be the first to journal") then after loading the empty state disappears with the happy state (ie... "See what Alex & Others are saying"). We should really have some sort of loading state before showing an empty state, as this flash of content with no indication of network activity is confusing.

- [ ] **Modal Headers**: Weights look off / font sizes don't always match

| Video | Figma |
| ----- | ----- |
| Screen Shot 2022-08-16 at 7 48 52 AM | Screen Shot 2022-08-16 at 7 49 11 AM |
| Screen Shot 2022-08-16 at 7 49 50 AM | Screen Shot 2022-08-16 at 7 50 00 AM |
| Screen Shot 2022-08-16 at 7 57 02 AM | Screen Shot 2022-08-16 at 7 57 27 AM |


- [ ] **Missing Cancel Button** on Share Sheets. I don't want to rely on _only_ gestures for navigation flows

### Horizontal Habit Feed

- [ ] **Spacing is off**. Horizontal cards should be spaced 8 px apart. Title also look closer than designed.
- [ ] **Shadow is cut off**. Overflow of the shadow is hidden and the card shadows are cut off at the bottom.
- [ ] **Missing CTA Button** "Start" or "Continue" Daily Habit

| Video | Figma |
| ----- | ------ |
| Screen Shot 2022-08-16 at 7 32 49 AM  | Screen Shot 2022-08-16 at 7 32 27 AM |

###  Gratitude

- [ ] **Prompt Shuffling Logic**: The prompt shuffling shouldn't be random. It should iterate forward in the list in one direction. Random shuffle makes it hard for end-user to know how many prompts they have to choose from. I think this should switch to a simple algorithm, and in the future we could consider shuffling the array on load, but still have the prompt shuffling iterate through that array in a singular direction, one index at a time.

### Prayer

- [ ] **Prayer sorting logic**: Prayers shouldn't be sorted from oldest to newest naively, it should take into account:

1. Who you follow (prayers from people you follow should be sorted up)
2. Number of prayers (prayers that have been prayed fewer times should be sorted up)
3. Unique on Requestor (shouldn't show multiple prayers from the same requestor)

I understand this is a backend change, but documenting here so we don't lose it.

- [ ] **Prayer Prompt Design**: Doesn't match Figma

| Code | Figma |
| ----- | ----- |
| Screen Shot 2022-08-16 at 7 46 22 AM | Screen Shot 2022-08-16 at 7 46 42 AM |

### Daily Reading

- [ ] **Missing Checkmarks on Individual Scriptures**: The scripture references should check off as you read them.
Screen Shot 2022-08-16 at 7 53 02 AM

- [ ] **Small Icons should use Filled/Solid version** This pencil icon should use the solid variant b/c of its small size:

Screen Shot 2022-08-16 at 7 56 10 AM

vinnyjth:
@conrad-vanl Migrated your todos to a group in the F&I list for now
CleanShot 2022-08-16 at 09 26 20@2x

packages/apollos-ui-connected/src/DailyHabitsConnected/DailyHabitNavigationStack/DailyHabitsNavigationStack.js
```diff
@@ -28,10 +29,13 @@ import ShareScreenConnected from '../../ShareScreenConnected';
 
 const { Navigator, Screen } = createSharedElementStackNavigator();
 
-const DailyHabitsNavigationStack = () => {
+const DailyHabitsNavigationStack = ({ route }) => {
+  const client = useApolloClient();
   const { goBack } = useNavigation();
 
   function closeDailyHabits() {
+    client.cache.evict({ __typename: 'DailyHabitCardListFeature' });
```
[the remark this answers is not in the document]
calebpanza: Yeah so this was a bit of a "convenience" thing last minute and accomplishes 2 things:

1. I don't have to manually evict the cache on each of the Daily Habit completion points (which also means future habits will automatically update the Feature Feed)
2. If someone else participates in the Daily Habits while you're viewing the Daily Habits, it'll update the Social Row on the Horizontal Cards without manually writing refetch logic into the Daily Feature

packages/apollos-ui-connected/src/features/DailyHabitFeature.js
calebpanza: that was gonna be one of my cycle goals for next cycle: introduce a `useStyleSheet` hook into the UI kit theme dir. 

I really like the way that the API is landing, but there are some considerations that need to be made with how to use it globally and then also things like safe areas and what not

packages/apollos-ui-connected/src/features/ScriptureFeature.js
calebpanza: The Daily Habit Feature uses the CardListCard component to render cards. This means that I don't have an easy way to update the UI on the Daily Feature to show completion state without some kind of Provider and/or prop drilling a single CardListCard id into the component(s) that are triggering the completion state. By using `cache.evict`, it'll just guarantee a fresh network request be made on _all_ Daily Habit Features that are rendered in 1 or more Feature Feeds

_(although it does look like a typo in the typename there which is a good catch)_

packages/apollos-ui-connected/src/hooks/useDailyHabitCardListFeature.js
calebpanza: We could! This honestly might just be something that I put in there while debugging that I forgot to remove. I can update it cause I can't remember an explicit reason for adding this

packages/apollos-ui-connected/src/features/DailyHabitFeature.js
calebpanza: I'm gonna create a scope for UI Kit Improvements and I'll add this component to that list. (right now it's TableView, Button, Chip, Social Row, useStyleSheet and the Modal Navigation)

packages/apollos-ui-connected/src/DailyHabitsConnected/DailyHabitNavigationStack/DailyHabitsNavigationStack.js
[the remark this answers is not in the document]
conrad-vanl: @vinnyjth do you agree that calling `cache.evict` then calling the garbage collection method on the `cache` store (the next line of code) is a code smell?

packages/apollos-ui-connected/src/features/ScriptureFeature.js
conrad-vanl: I think an `optimisticUpdate` would handle this

packages/apollos-ui-connected/src/DailyHabitsConnected/DailyHabitNavigationStack/DailyHabitsNavigationStack.js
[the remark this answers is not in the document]
conrad-vanl: Shouldn't we use Apollos' cache update tools (https://www.apollographql.com/docs/react/data/mutations/#updating-local-data) instead?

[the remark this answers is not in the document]
conrad-vanl: `refetchQueries` might be a better way to do the same thing you're doing, if I understand correctly, but you can also update the cache directly and not require a network roundtrip at all (potentially...although the `refetchQueries` option is easier)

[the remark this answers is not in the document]
redreceipt: this convo doesn't feel resolved. for what it's worth, I don't like `cache.evict` either. seems like a smell at best. I like updating the cache manually if possible to avoid trips to the network. at a higher level, if it's hard to keep track of the variables needed to update the cache without passing them up and down through the tree, there's probably an architecture improvement that could be made. but I'll let @conrad-vanl make the final call here.

packages/apollos-ui-connected/src/features/ScriptureFeature.js
redreceipt: similar sentiment to the above comment.

packages/apollos-ui-connected/src/hooks/useDailyHabitCardListFeature.js
redreceipt: I'll wait for this to be pulled out.

calebpanza: Updated

vinnyjth:
Where are we at here @calebpanza? Sounds like @conrad-vanl is okay merging this given that we've cataloged some of the issues. 

This won't have any changes for existing apps that don't use the new features, right?

calebpanza:
@vinnyjth the only outstanding conversation is regrading the use of `cache.evict` and whether that needs to be addressed now or would be a refactor/improvement.

I think everything else has been determined to be an enhancement. And this work should not interfere with any of the existing features. It _is_ required in order to churches to start using the Daily Habit Feature.

conrad-vanl:
> @vinnyjth the only outstanding conversation is regrading the use of cache.evict and whether that needs to be addressed now or would be a refactor/improvement.

Yeah, I'd really like @vinnyjth's feedback on this one

packages/apollos-ui-connected/src/DailyHabitsConnected/DailyHabitNavigationStack/DailyHabitsNavigationStack.js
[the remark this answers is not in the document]
vinnyjth: @calebpanza let's drop a `TODO:` in here to make it clear that we should refactor this to use `refetchQueries` or `updateFragment` in the future. I agree that this solution is a bit nuclear, but I think it's an okay compromise.

conrad-vanl:
or @redreceipt's feedback not sure who will know better about apollographql client cache management stuff. It's a smell to be dealing with the garbage collector manually in my mind, but if y'all are ok with that lets get this shipped

vinnyjth:
> > @vinnyjth the only outstanding conversation is regrading the use of cache.evict and whether that needs to be addressed now or would be a refactor/improvement.
> 
> Yeah, I'd really like @vinnyjth's feedback on this one

I don't see `cache.evict` as much worse than `cacheWrite`, but I think there's a better way to do what @calebpanza wants to do. I think I can see why you went this route, @calebpanza, since doing a `refetchQueries` on feature in the feature feed takes a bit finesse. 

Let's drop a comment in reminding us that this isn't the best practice and move on.

calebpanza:
@vinnyjth @conrad-vanl it'll be a really easy rewrite during a cooldown in order to be more explicit about this. I'll add a comment now

vinnyjth (approved):

PR #2619, opened 2022-08-08
fix: set onboarding slide titles to correct color

redreceipt:
currently looks bad in dark mode

Screen Shot 2022-08-08 at 10 38 49 AM
Screen Shot 2022-08-08 at 10 38 30 AM


vinnyjth (approved):
thanks! this bugged me also

PR #2620, opened 2022-08-08
chore: cleans up webview feature

redreceipt:
essentially, I was trying to fix the height, realized it was back-end driven, but in the process found other things that weren't working right

Screen Shot 2022-08-08 at 11 28 48 AM


vinnyjth (approved):
rad! 

I made some of these same changes a couple weeks ago while solving a different issue, but ended up bailing on the whole pr

PR #2625, opened 2022-08-09
chore: conditionally load onboarding slides based on user info

nlewis84:
This PR builds the Onboarding Slides array based on the current user's information. I originally was getting clever with the shift method, but I think being more verbose means this all makes more sense.

|Status|Gif|
|---|---|
|Have First Name, Last Name, Gender, BirthDate|Simulator Screen Recording - iPhone 13 - 2022-08-09 at 10 08 16|
|Doesn't have First Name/Last Name|Simulator Screen Recording - iPhone 13 - 2022-08-09 at 10 09 29|
|Doesn't have Gender/BirthDate|Simulator Screen Recording - iPhone 13 - 2022-08-09 at 10 10 03|
|Needs full onboarding|Simulator Screen Recording - iPhone 13 - 2022-08-09 at 10 10 28|


redreceipt (approved):
❤️❤️❤️

vinnyjth:
Great screenshots @nlewis84  :D

templates/mobile/src/onboarding/Onboarding/index.js
```diff
@@ -36,18 +38,61 @@ function Onboarding(props) {
...
+  const needsNameOnboarding =
+    data?.currentUser?.profile?.firstName === null ||
+    data?.currentUser?.profile?.lastName === null;
+  const needsGenderBirthDateOnboarding =
+    data?.currentUser?.profile?.birthDate === null ||
+    data?.currentUser?.profile?.birthDate === null;
+
+  if (needsNameOnboarding && needsGenderBirthDateOnboarding) {
+    // user needs all onboarding
+    slides = props?.slides || [
```
redreceipt, line 60: If props.slides, you could bypass all this. But also, we don't really need props.slides anymore. I'll let you decide if you wanna make any changes.

nlewis84: I'll ditch it. 😉

vinnyjth:
@nlewis84 There are legit test failures you will need to look at :)

nlewis84:
AboutYou and AboutYouConnected tests are passing locally. 🤞

redreceipt:
why are we removing tests? seems like we're losing coverage here as well

nlewis84:
The test to see if it renders a birthdate and gender is obsolete now, because that slide will no longer show up if the current user has both those attributes. There appears to be an issue with tests where we are having to pass a date just to get it to match screenshots, which also means that we are passing a gender and a date (which is obsolete). One test I deleted because the snapshot kept being off by one day for no reason.

redreceipt (approved):
gotcha

PR #2626, opened 2022-08-09
fix: Update Scripture Card Touchable

calebpanza:
## Motivation
Cards should use the `TouchableScale` component instead of the `Touchable` (which uses a highlight effect). The Scripture Card was wrapped in the incorrect component and this PR resolves that issue

### Before
WRONG_TOUCHABLE

### After
https://user-images.githubusercontent.com/45076058/183715442-a5c824ea-c8bb-4fdf-b8bd-c19d7e7f779a.mp4


conrad-vanl (approved):

redreceipt (approved):
Niiice

PR #2627, opened 2022-08-09
fix: Scripture Journal SocialRow Icon Rendering

calebpanza:
## Motivation
In order to draw more clarity to what the icon is, the "sparkle" icon should be filled when displayed on the Social Row component.

I noticed an area in the conditional rendering that I thought could be improved so that the user state was being passed in to the render methods instead of just reading values from the parent scope, so I snuck that in here, too. 

| Before | After |
| --- | --- |
| IMG_19369FCD7415-1 | Simulator Screen Shot - iPhone 13 - 2022-08-09 at 13 31 40 |


conrad-vanl:
I would argue that the color of the icon should be the same tone as the text, but one "step" lower hierarchy.

So if the Text is text.secondary, the icon should be text.tertiary or gray3

camrun91:
> I would argue that the color of the icon should be the same tone as the text, but one "step" lower hierarchy.
> 
> So if the Text is text.secondary, the icon should be text.tertiary or gray3

I agree it looks good but seems to be a little darker or saturated than the text to the right I assume the change you suggested would lighten it some.

calebpanza:
@conrad-vanl do you think that a little splash of color might be a fun way to draw some attention to the feature and encourage engagement?

conrad-vanl:
> do you think that a little splash of color might be a fun way to draw some attention to the feature and encourage engagement?

It's not a practice that we use anywhere else in our design system right now. If we wanted to use color here, I'd say the text and potentially border should also be colored to match what we do elsewhere....I'd say make the icon the color I suggested above.

Screen Shot 2022-08-09 at 4 09 20 PM
Screen Shot 2022-08-09 at 4 09 27 PM
Screen Shot 2022-08-09 at 4 09 45 PM
Screen Shot 2022-08-09 at 4 10 07 PM
Screen Shot 2022-08-09 at 4 10 30 PM
Screen Shot 2022-08-09 at 4 10 44 PM

calebpanza:
Yeah the color isn't feeling right. I'm just gonna keep it neutral for now. Maybe try out some other things in Figma at some point, but I think that'll be the easiest way to get this puppy landed

calebpanza:
| Empty | Write Another | Join Others |
| --- | --- | --- |
| Simulator Screen Shot - iPhone 13 - 2022-08-09 at 20 24 27 | Simulator Screen Shot - iPhone 13 - 2022-08-09 at 20 24 33 | Simulator Screen Shot - iPhone 13 - 2022-08-09 at 20 25 25 |

redreceipt (approved):

PR #2628, opened 2022-08-09
fix: Add Journal Touch Targets

calebpanza:
## Motivation
When there is an empty state on a Scripture Journal, the touch target is not clear enough that you can add a journal by touching the Avatar.

## Solution
Update the empty state of the Journal Feed to include a CTA and a button. The existing touch target of the Avatar remains in the header, but this draws more visual attention to the User on what they can interact with and informs them that they will be the first to contribute to the conversation.

I'm open to suggestions on this change. One suggestion was brought up to not allow the drawer to be opened unless there are journals. My concern with that would only be consistency in the UX so the user doesn't have to "guess" if the drawer they're seeing can be opened or not. (but that's just an initial thought, open to critique!)

| Before | After |
| --- | --- |
| IMG_9811 | Simulator Screen Shot - iPhone 13 - 2022-08-09 at 14 26 20 |


camrun91:
@calebpanza what does it look like in dark mode? I like the click-through.

vinnyjth (approved):

conrad-vanl (changes requested):
So not to push back since it's already approved, I'm not a fan of there's now two actions on this screen that do the same thing (tapping the journal prompt at the top or pressing the button). I think this could be confusing to end-users as it can cause decision paralysis in knowing whether both actions do the same thing or not.

I would probably favor just having the text by itself ("No journals yet") vs a second CTA 🤔 "Be the first to add your thoughts"

Edit: Even better if we can prevent "swiping up" on this popover until there's journals to show

calebpanza:
@conrad-vanl I don't mind making that change at all! This was the "easier" of the 2 solutions, so I wanted to see how this felt implemented before going the extra mile with the sheet being non-openable. But the decision paralysis makes sense and I could see that being confusing. I'll spend some time tonight updating that.

calebpanza:
it's not perfect, but it's the best I've been able to figure out. There isn't really a hook into the index change from a child of the Bottom Modal (that I can tell). I have access to the Animated Value, so I'm checking to see if the index is 1 which would indicate that it's open. Attempts to close it in real time were causing the sheet to get "stuck". I can take another stab at it in the morning with a fresh head:

https://user-images.githubusercontent.com/45076058/183792664-6f044dd2-cbe7-40c6-b3bc-08d92a613131.mp4

vinnyjth:
@conrad-vanl Does this UX look good to you?

conrad-vanl:
Hmmm....based on the video it looks like the screen temporarily opens then closes? 

What if opening focused the text field when there's no journals?\


....just having a simple empty state might be best.

calebpanza:
Aight so I pushed a quick update with another suggested solution. There's a whole explanation/reason for this whole process that I recorded in this [Loom](https://www.loom.com/share/902a6503aaf8421fa05b32aa11e224c9). Looking forward to the feedback, here, but I think we're pretty close.

conrad-vanl (approved):

PR #2629, opened 2022-08-09
fix: Add Journal Placeholder Color

calebpanza:
## Motivation
The placeholder text when inputing a Journal was not getting set to a theme value, so it was not readable in dark mode. This PR resolves the system theme discrepancy.

| Before | After - Dark | After - Light |
| --- | --- | --- |
| IMG_C85655CCE4C2-1 | Simulator Screen Shot - iPhone 13 - 2022-08-09 at 14 38 02 | Simulator Screen Shot - iPhone 13 - 2022-08-09 at 14 38 48 |


packages/apollos-ui-connected/src/ScriptureScreenConnected/ScriptureJournals/UserInput.js
```diff
@@ -64,6 +64,7 @@ const UserInput = ({ onChangeText, value }) => {
           value={value}
           onChangeText={onChangeText}
           placeholder="What sticks out to me is..."
+          placeholderTextColor={theme.colors.text.tertiary}
```
conrad-vanl, line 67: Why did you choose `tertiary`?

The inputs and Figma uses `secondary` for placeholder text.

(and don't start telling me we should use artifacts here, I know that's a great reason for them I don't doubt)

calebpanza: Tbh, I just liked the way that it looked and I typically use a really muted color for Placeholders. I actually didn't check to see what our TextInput was already doing. I'll go ahead and make that update. It's a good callout.

redreceipt (approved):

PR #2630, opened 2022-08-09
fix: Update Community Share messaging

calebpanza:
## Motivation
The "Share with the Community" visually indicated some kind of user interaction in which you could enable/disable community sharing. This was confusing to the users as we are shipping an MVP in which all journals/prayers are publicly shared.

## Solution
Remove the TableView styling on the Community Share within the Share Sheet. Instead, replace it with a static line of text (per node type) that explains that posts will be visible to the community.

| Before | After |
| --- | --- |
| IMG_BE2397FA1456-1 | Simulator Screen Shot - iPhone 13 - 2022-08-09 at 16 34 57 |


packages/apollos-ui-connected/src/ShareScreenConnected/ShareSelection/CommunityShare.js
```diff
@@ -74,7 +80,7 @@ function typeSafeText(__type) {
     case 'ScriptureJournal':
       return {
         title: 'Share Journal',
-        summary: 'Your journal will show up on the community feed',
+        summary: 'Your journal will be visible by others',
```
conrad-vanl, line 83:
```suggestion
        summary: 'Your journal will be visible to others',
```

conrad-vanl (changes requested):

redreceipt (approved):

PR #2631, opened 2022-08-09
chore: adds mergify back

redreceipt:
with the size of the team, I think it's time to bring this back. I'll be aware with my approvals. this will save us time having to hit the 'update' button


vinnyjth:
CleanShot 2022-08-09 at 19 24 07@2x

redreceipt:
hmmm, I want it to simply obey our branch protection rules but it's saying I have to list a condition

redreceipt:
got it. we had this before. this way users can mark their own PR blocked if they don't want it to be auto merged for them.

vinnyjth (approved):

PR #2633, opened 2022-08-10
fix: separate intent filters

tristanguckenberger:
### **Purpose**
This PR ensures multi-tenant android apps are able to be located on a users device, post-install.

**Before**

https://user-images.githubusercontent.com/19194337/183943446-c08210f4-34e3-4c20-804a-9b3ad87fe335.mov

**After**

https://user-images.githubusercontent.com/19194337/183943576-3013dbc2-af22-4533-a365-526855e97bde.mov


redreceipt:
Good find! Can you confirm universal links still work?

redreceipt:
Ah ok i see what we did wrong. can you add these missing lines to the second intent to preserve linking https://reactnavigation.org//docs/deep-linking/

redreceipt (approved):
Nice work!

PR #2634, opened 2022-08-10
fix: add optional chaining

nlewis84:
This PR fixes a bug in the frontend that was causing a card to show up blank. This came up due to the rarely used Series in Progress/Content in Progress algos.

|The Error in Sentry|
|---|
|Screen Shot 2022-08-10 at 12 19 59 PM|
|The fix demo|
||


vinnyjth:
@nlewis84 Your solution looks good at a glance, but I have a feeling something fishy is going on one layer deeper. it seems strange that is `relatedNode` null to begin with. Do you happen to have the full response you are getting back from the server when `relatedNode` is null?

vinnyjth:
@nlewis84 Nevermind, I understand :D

vinnyjth:
@conrad-vanl Can I get your eyes on this UX? 

This is for the "series in progress" algorithm if you don't have any series in progress.

vinnyjth:
CleanShot 2022-08-10 at 14 07 23@2x

nlewis84:
Just thinking out loud here....why do we do this? Why don't we just not show the feature like we would on anything else that doesn't have any content? @conrad-vanl

vinnyjth:
@nlewis84 I think I can answer this - 

We got a lot of client feed saying "The upcoming feature isn't working / I can't find the upcoming feature / Please add the upcoming feature" :)

vinnyjth:
@nlewis84 Michael snuck in and merged this same code in another PR, so I'm going to close this :)

nlewis84:
Man, he is sneaky!

PR #2638, opened 2022-08-10
fix: avoids weird card cache bug

redreceipt:
fixes this

Screen Shot 2022-08-10 at 7 51 02 PM


vinnyjth:
@redreceipt Nathan has got the same fix here also :) 

https://github.com/ApollosProject/apollos-apps/pull/2634

redreceipt:
I've got it in another PR as well but it's blocking work and testing for all of us so figured it was simpler to get it in.

vinnyjth (approved):
let's merge whatever passes tests first

PR #2639, opened 2022-08-11
fix: use goBack in onPress

tristanguckenberger:
### **Purpose**
The purpose of this PR is to ensure that the AuthNavigator's `PlatformSpecificBackButton` works for iOS as expected when tapped by a user.

### **Testing**
**Setup**
1. Clone the repo and checkout the following branch: `add-goback-to-auth-navigator`
2. Pick any multi-tenant app for a fresh spin up and run the setup script (`setup-app.sh`)
3. Run the iOS app

**Steps to Test**
1. Login with OAuth
2. Enter your phone number, hit Next
3. From the "PIN" screen, hit back

**Expected Result from Test**
You should be taken back to the previous page.

**Before**

https://user-images.githubusercontent.com/19194337/184215140-a306b7e8-ce33-4411-8852-2ec9d28ec64e.mp4

**After**

https://user-images.githubusercontent.com/19194337/184215280-80fd7d95-2350-4727-9933-26af5c2d2f28.mov


redreceipt (commented):
Probably doesn't need to be a prom. Just get the go back function inside a component instead of passing it in

tristanguckenberger:
> Probably doesn't need to be a prop. Just get the go back function inside the component instead of passing it in

@redreceipt I'm not sure I follow and it's likely due to a gap in my knowledge. The component I'm passing `goBack` to (`HeaderBackButton`) is an import from `node_modules/@react-navigation/stack/lib/typescript/src/views/Header/HeaderBackButton.d.ts`. Is there a way for us to modify this file?

redreceipt:
Oh! I thought it was one of ours. You're good :)

redreceipt (approved):

PR #2641, opened 2022-08-17
fix: remove video select

tristanguckenberger:
### **Purpose**
The purpose of this PR is to ensure the video select feature only works in the case where videos have names. Conrad mentions in the relevant basecamp post [here](https://3.basecamp.com/3926363/buckets/27724884/todos/5175263026#:~:text=video%20select%20was%20intended%20to%20select%20between%20different%20types%20of%20videos%20as%20in%20%2D%20English%2C%20Spanish%2C%20or%20Modern%20Service%20and%20Traditional%20Service) (which I recommend looking at as well), that

> video select was intended to select between different types of videos as in - English, Spanish, or Modern Service and Traditional Service

whereas for some clients, specifically Crossings, video select is picking up 3 videos

Screen Shot 2022-08-17 at 10 53 19

which correspond to these in rock:

Screen Shot 2022-08-17 at 10 55 08

### **Solution**
**Since video select is picking up videos that don't match its intended purpose and since we are not handling null names, it seems appropriate to use the lack of a name as a way to disable videoselect when and if it is not used as intended, without disabling it entirely.**

_*NOTE: This solution below is based on the assumption that when used as intended (i.e. with specific language options or service types), names wont be null and the video select feature works. The solution I propose above also tries to avoid disabling the component entirely in case other churches are using video select with the intended types. Shoot me down here if I'm not making sense or if I need to dig deeper. Ultimately I'm looking for a way to disable videoselect for cases where it's not being passed the right type of videos.*_

### **Testing**
**Setup**
1. Clone the repo and checkout the following branch: `handle-video-select-null-names`
2. Pick any multi-tenant app (I'm using crossings_community_church) for a fresh spin up and run the setup script (`setup-app.sh`)
3. Run the iOS or Android app

**Steps to Test**
1. Login
2. Select any video item
3. You should no longer see the video select button if _any_ video.name in the videos array is null

### **Before**
Screen Shot 2022-08-17 at 12 27 18

### **After**
Screen Shot 2022-08-17 at 12 28 01


tristanguckenberger:
Also, I'm not having much luck finding where a video name gets set, and it's beginning to seem like this could be an unused or no longer used media player feature. If that's the case, and [given that the feature no longer fully works](https://3.basecamp.com/3926363/buckets/27724884/todos/5175263026#:~:text=The%20video%20select%20button%20is%20trying%20to%20set%20an%20onPress%20event%20but%20the%20react%2Dnative%20View%20component%20that%20it%27s%20built%20with%20doesn%27t%20provide%20(or%20no%20longer%20provides)%20onPress%20as%20a%20prop.%20We%20can%20use%20Pressable%20or%20TouchableOpacity%20in%20its%20place%20if%20we%20want%20this%20onPress%20functionality%20to%20work.), maybe we _should_ just disable this entirely?

redreceipt:
I’m good with killing it if it’s not working and we’re not using it

packages/apollos-ui-media-player/src/Controls/index.tsx
```diff
@@ -24,7 +24,6 @@ import Seeker from './Seeker';
 import AirPlayButton from '../AirPlayButton';
 import { PictureMode } from '../types';
 import CollapsedHeader from './CollapsedHeader';
-import VideoSelect from './VideoSelect';
```
redreceipt, line 27: you can go ahead and delete this file too

redreceipt (approved):

PR #2643, opened 2022-08-18
fix: correctly adds deep link intent

redreceipt:
even though multiple sources of documentation (RN, react navigation, Google Android docs) said what we were doing should have worked, it wasn't...🤷‍♂️. this makes it work

out


vinnyjth (approved):

vinnyjth:
I _think_ this makes sense to me based on what I've read. 

I think the path prefix might have been applying to the deep link scheme?

redreceipt:
> I think the path prefix might have been applying to the deep link scheme?

that's it! android merges all possible permutations of scheme, host, and pathPrefix (🤮 ) so it was leaking in

tristanguckenberger:
@redreceipt is this cmd right? It's a little hard to see in the gif but I think I have a version for Crossings:
`adb shell am start -W -a android.intent.action.VIEW -d Crossings://app-link/nav/give com.crossings.com`

I'm trying to test and I can't get it to launch into the give tab:

https://user-images.githubusercontent.com/19194337/185474822-97b67e0c-97cd-4134-8d9d-56ecdbeaaadc.mov

redreceipt:
Give tab won't work. Only connect tab and content items right now

PR #2644, opened 2022-08-22
feat: Community Single Navigation

calebpanza:
## Motivation
Community work for the cycle is still being shaped out a bit, but we still wanted to begin making code progress. So this PR introduces the navigation and the Screen Component for what will be our Community Single

## How to Test
I've updated the Production dB to introduce a Community Tab. It holds a single card that is deep linked to the Community Single screen. You'll be able to use that in order to see this work in action!

If you want to test locally, just trigger a command (however you so desire) to link to: `/community/:communityId`

### New Community Tab
Simulator Screen Shot - iPhone 13 - 2022-08-22 at 13 41 56

### Community Single in Action

https://user-images.githubusercontent.com/45076058/185987281-2e549e0b-5a81-4af3-8abf-0c7ff86b7fec.mp4

# NOTE :: this feature is not considered Production Ready, although it is protected from end-users since the route is hidden and only available through intentional integration


packages/apollos-ui-connected/src/CommunitySingleConnected/CommunitySingleConnected.js
```diff
@@ -0,0 +1,59 @@
...
+    justifyContent: 'center',
+  },
+});
+
+// Context
+const CommunitySingleContext = createContext({
+  id: null,
+});
+
+export const useCommunitySingle = () => {
```
redreceipt, line 31: why the hook? why not just useContext() in whatever component?

packages/apollos-ui-connected/src/CommunitySingleConnected/CommunitySingleConnected.js
```diff
@@ -0,0 +1,59 @@
...
+  return useContext(CommunitySingleContext);
+};
+
+const CommunitySingleConnected = ({ navigation, route }) => {
+  return (
+    <CommunitySingleContext.Provider
+      value={{
```
redreceipt, line 38: where is this being consumed?

packages/apollos-ui-connected/src/CommunitySingleConnected/CommunitySingleConnected.js
calebpanza: Personal preference. I would prefer to export the context as a hook and do 1 import instead of importing the Context and then consuming it with `useContext` on each component.

redreceipt: but the hook doesn't do anything? and by using `useContext` on the children, future developers don't have to wonder what it does

packages/apollos-ui-connected/src/CommunitySingleConnected/CommunitySingleConnected.js
calebpanza: There are a bunch of UI elements that are yet to be built out that will need to react based on the user's Membership Status in a community. I've found success in the past writing a Context Provider that could pass through a really simple `isMember` boolean, so I threw this in here now as a precursor for that work to come

packages/apollos-ui-connected/src/CommunitySingleConnected/CommunitySingleConnected.js
```diff
@@ -0,0 +1,59 @@
...
+        <Button title="Go Back" onPress={navigation.goBack} />
+      </View>
+    </CommunitySingleContext.Provider>
+  );
+};
+
+CommunitySingleConnected.propTypes = {
+  route: PropTypes.shape({
+    params: PropTypes.shape({
+      communityId: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
calebpanza: Ah good catch! Sorry for that. I've updated it

camrun91 (approved):
As long as @redreceipt is good with them these look good to me.

redreceipt (changes requested):
take out context provider. we can add back when we're actually using it

redreceipt (approved):

calebpanza:
it is removed!

PR #2645, opened 2022-08-22
chore: Migrate Post to UI Kit

calebpanza:
## Motivation
We have the social posts UI currently stored in the ShareSheet component inside of UI Connected. For the Community work we're doing, we're going to want a more "global" access to them.

## Solution
Took the core logic of the components and moved them to the UI Kit and imported those references inside of the UI Connected for Share Sheet


packages/apollos-ui-connected/src/ShareScreenConnected/NodePreview/PrayerRequestPreview.js
```diff
@@ -7,50 +7,15 @@
...
-          padding: theme.sizing.baseUnit,
-        },
-      }),
-      theme,
-    ],
-    [theme]
-  );
-}
-
 const PrayerRequestPreview = ({ text }) => {
```
redreceipt, line 16: this component is now five lines long and maybe doesn't need to be its own file?

packages/apollos-ui-connected/src/ShareScreenConnected/NodePreview/ScriptureJournalPreview.js
```diff
@@ -7,53 +7,16 @@
...
-          paddingBottom: theme.sizing.baseUnit * 0.5,
-        },
-      }),
-      theme,
-    ],
-    [theme]
-  );
-}
 
 const ScriptureJournalPreview = ({ text }) => {
```
redreceipt, line 16: same as this. and it looks the exact same as the prayer preview

calebpanza:
@redreceipt I see what you're saying and do agree. These 2 components just to happen to have the same API's and are very small. Ultimately, this is a stop gap until we rewrite these components with the UI Kit 2.0.

I just need this one small change to these files in order to unblock some cycle work to get Communities done which will then unblock me to be able to work on UI Kit 2.0.

Are we ok with the slight inefficiency knowing that it's intentional and only temporary?

redreceipt:
"Only temporary" is a slippery slope. As best I can tell, these components don't need to exist at all. Whatever is calling them can just write out their API in line. This will apply for UIKit 1 and 2 looks like. Why fix later when you can fix now. And such a small adjustment?

calebpanza:
Yeah that's a fair point. Went ahead and deleted those files so we should be good!

redreceipt (approved):

PR #2646, opened 2022-08-24
feat: Community Single

calebpanza:
## Motivation
This PR introduces the baseline styling for the Community Single. 

## Approach
The API is still in progress, so this PR introduces and explicit cache-loading that mocks the desired data structure. This should help 1. get UI out for review quicker and 2. give @camrun91 an easy way to see what the UI is expecting to make sure API design is in alignment.

Per conversations with @vinnyjth and @redreceipt in Basecamp, this PR localizes the Community Single hooks to the CommunitySingleConnected directory.

## Caveats
This PR does _not_ include the Social Feed. In order to keep the PR as small as possible, I've limited the PR to just the main content that describes a Community Single

Light Mode was not explicitly designed for in Figma, so this is a 1-1 inversion of the desired Dark Theme (I think it looks fine for the time being and should not keep us from merging the PR. @conrad-vanl can review and make revisions in Figma and we can push out subsequent `patch` PR's for micro-level changes)

This PR uses _only_ the existing UI Kit components and is not a perfect 1-1 recreation of Figma. This is because I will be working on UI Kit 2.0 in the next week or so. Once that work is completed, a `chore` PR will be pushed out that re-implements this screen as our first screen fully driven by UI Kit 2.0. At that time, it'll match the Figma mocks 1-1. This is just to make ample time in the cycle for UI Kit 2.0 _and_ not risk Community Cycle goals.

## Testing
Pull up the mobile app and connect to prod's dB. Load up the Community Tab and press into the Community. Your cache will auto-load with the test data and you'll be good to go!

In order to test loading and error states, manually update return values in the `useCommunitySingle` hook.

- [x] iOS test/style
- [x] iPad test/style
- [x] Android test/style
- [x] View a Community
- [x] Loading State
- [x] Error State
- [x] Join a community
- [x] Leave a community
- [x] Invite someone to the community

## Screenshots

| | Light Mode | Dark Mode |
| --- | --- | --- |
| Current User is not a Member | non_member_state | non_member_state |
| Current User is a Member | member_state | member_state |
| There are No Members in the Community | no_members | no_members |
| Loading State | loading_state | loading_state |
| Error State | error_state | error_state |

## Android Styling Considerations
There is a strange bug happening when using a Button with a transparent color where a gray box renders in the center of the button. To get around this (for now) I've gone ahead and made the "check button" a primary button on Android only

## iPad Styling Considerations
Instead of flexing the entire width, the "Invite"/"Join" button will have a max width of 350px


conrad-vanl:
Screen Shot 2022-08-25 at 11 49 27 AM

Would it be hard to get the user count up against the button, instead of spread out / space-between?

calebpanza:
> What's the deal with the green outline on this button?

I didn't notice that until after the screenshots were taken. It should be resolved in the simulator, though

> Would it be hard to get the user count up against the button, instead of spread out / space-between?

No not all! I can super easily make that switch

calebpanza:
@conrad-vanl here we go!
Simulator Screen Shot - iPad Air (5th generation) - 2022-08-25 at 12 22 11

calebpanza:
Double checked on that phantom border. We're all set now!

conrad-vanl:
Could you do an updated visual/video demo walking through the flows this PR introduces? Maybe for Visual PRs that could be a good deliverable so that someone like me can review design implementation without having to run it locally :)

packages/apollos-ui-connected/src/CommunitySingleConnected/components/Details.js
```diff
@@ -0,0 +1,134 @@
...
+  backgroundColor: theme.colors.text.primary,
+  borderColor: theme.colors.text.primary,
+}))(Button);
+
+const NeutralButtonLabel = styled(({ theme }) => ({
+  marginLeft: 4,
+  color: theme.colors.background.paper,
+}))(H4);
+
+const NeutralButtonIcon = withTheme(({ theme }) => ({
```
conrad-vanl, line 47: Is the intention to replace these with non-customized/styled components after Button is refactored?

packages/apollos-ui-connected/src/CommunitySingleConnected/components/Details.js
```diff
@@ -0,0 +1,134 @@
...
+                <NeutralButtonLabel>Join</NeutralButtonLabel>
+              </NeutralButton>
+            </NeutralButtonContainer>
+          )}
+          {currentUserIsMember && (
+            <IsMemberButton onPress={leaveCommunity} disabled={loading}>
+              <IsMemberIcon name="check" />
+            </IsMemberButton>
+          )}
+          {currentUserIsMember && (
```
conrad-vanl, line 108: I know this is common practice and web land, but I'm curious if it can still cause issues in react-native land. 

What are the possible values of `currentUserIsMember`? Whatever falsey value of `currentUserIsMember` will get returned and react-native will try to render it. If it could ever be an empty string for example, react-native will throw an error

packages/apollos-ui-connected/src/CommunitySingleConnected/components/TabNavigator/TabBar.js
```diff
@@ -0,0 +1,191 @@
+/**
+ * TabBar.js
```
conrad-vanl, line 2: Do you expect to pull this `TabNavigator` out of just Community?

conrad-vanl:
I'll let @vinnyjth do the primary code review but from my perspective so far the visuals look pretty good. I'd love to see a video walk through of the final state.

My only other thought re: code is what structures do we have setup that's planned to stay, and what structures are setup that's planned to be moved/refactored soon? I know @calebpanza wants to refactor a lot of the UI into `/mobile`, so I assume some of the styled buttons and such will be refactored once that's done. Another thought I had is it looked like everything is setup within `ui-connected` and new components introduced within just this `Community` folder structure. I wonder if it's intended for those things to stay there long term or get moved as well soon.

packages/apollos-ui-connected/src/CommunitySingleConnected/components/TabNavigator/TabBar.js
calebpanza: Yeah, I think so. I really like the way that it landed. Setting it up as its own file was just a personal organizational choice. When the component landed as well as it did, I got to thinking that it'll probably be worth breaking out into its own reusable item inside the `mobile/components` dir (after the restructure)

packages/apollos-ui-connected/src/CommunitySingleConnected/components/Details.js
calebpanza: Oh I see. Yeah I can either wrap it in a `Boolean()` or even just break it out into its own rendering function. That's not a bad call. The data provider will guarantee that it's always a boolean, but without Typescript I see how this does lead to a vulnerability.

packages/apollos-ui-connected/src/CommunitySingleConnected/components/Details.js
calebpanza: Yeah exactly. I used only (mainly) the HOC's and our UI components cause I was able to crank it out crazy fast that way

templates/mobile/package.json
```diff
@@ -7,6 +7,7 @@
   },
   "scripts": {
     "ios": "react-native run-ios",
+    "ipados": "react-native run-ios --simulator=\"iPad Air (5th generation)\"",
```
vinnyjth, line 10: win! thanks for adding this

packages/apollos-ui-connected/src/CommunitySingleConnected/hooks/useCommunitySingle.js
```diff
@@ -0,0 +1,183 @@
...
+            }
+          }
+        }
+        currentUserStatus
+      }
+    }
+  }
+`;
+
+const useCommunitySingle = (id) => {
```
vinnyjth, line 60: this approach of using the cache is a good compromise. nice work!

packages/apollos-ui-connected/src/CommunitySingleConnected/hooks/useCommunitySingle.js
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    const forceCacheInjection = false;
+    if ((data === undefined && !queryLoading) || forceCacheInjection) {
+      injectCache();
+    }
+  }, [data, queryLoading, client, id, variables]);
+
+  return {
+    loading: queryLoading || mutationLoading,
```
vinnyjth, line 165: should these be the same loading state? I imagine there's a future where you use the same hook to drive mutations and queries, and I feel it would be a bit odd if you showed a feed loading state when you join/leave.

maybe I'm wrong about that though

packages/apollos-ui-connected/src/CommunitySingleConnected/components/TabNavigator/TabBar.js
```diff
@@ -0,0 +1,191 @@
...
+  });
+
+  return (
+    <View>
+      <CoverImage communityId={communityId} />
+      <Details communityId={communityId} />
+      <Row>
+        {state?.routes?.map((route, index) => {
+          const { options } = descriptors[route.key];
+          const label =
```
vinnyjth, line 99: nested ternaries can be so hard to parse 😅 

what about an array operation instead? 
```
const label = [options.tabBarLabel, options.title, route.name].filter(Boolean)[0]
```

packages/apollos-ui-connected/src/CommunitySingleConnected/components/MembersPreview.js
```diff
@@ -0,0 +1,89 @@
...
+import { View } from 'react-native';
+import { Avatar, BodySmall, styled } from '@apollosproject/ui-kit';
+
+function intToString(num) {
+  num = num.toString().replace(/[^0-9.]/g, '');
+  if (num < 1000) {
+    return num;
+  }
+  let si = [
+    { v: 1e3, s: 'K' },
```
vinnyjth, line 23: neato!

vinnyjth (changes requested):
This looks great overall! 

I'm not too worried about leaving this in ui-connected. It all needs to move over to `mobile/src` eventually, but building out here won't make that transition that much harder.

packages/apollos-ui-connected/src/CommunitySingleConnected/hooks/useCommunitySingle.js
calebpanza: Yeah that's a good callout. The UI will take into consideration both if data exists _and_ if there is a loading state. Buttons, though, will be considered disabled anytime there is a loading state. So I was thinking of maybe passing in a `mutationLoading` or `queryLoading` at some point if I need to be able to respond specifically to one of those scenarios. In practice, though, I found myself using `isLoading={queryLoading || mutationLoading}` on each component and figured it was worth consolidation

packages/apollos-ui-connected/src/CommunitySingleConnected/components/TabNavigator/TabBar.js
calebpanza: tbh I ripped this straight outa the React Navigation Docs and only customized the visuals. I would be more than happy to make that adjustment, though. I think that suggested approach is really smart!

What's the benefit to that vs `const label = options.tabBarLabel ?? options.title ?? route.name`?

packages/apollos-ui-connected/src/CommunitySingleConnected/components/MembersPreview.js
calebpanza: #StackOverflow. Lol

calebpanza:
@conrad-vanl per your request, here is an updated [demo](https://www.loom.com/share/6c5c7a940876442ba305623e9c8a0b54) of the UI in action.

The first 2 minutes walk through the UI states that are built out, and then the last bit of it explains the dev approach and how I was planning to integrate UI Kit 2.0 not the mix.

packages/apollos-ui-connected/src/CommunitySingleConnected/components/Details.js
calebpanza: I went ahead and just updated the rendering to be a memoized function so it's a bit safer/more explicit.

packages/apollos-ui-connected/src/CommunitySingleConnected/components/TabNavigator/TabBar.js
conrad-vanl: TIL `??`

vinnyjth (approved):

redreceipt (approved):

conrad-vanl:
Watched the video. left a few minor comments on the loom but looking good!

PR #2647, opened 2022-08-27
chore: cleans up UpNext button

redreceipt:


packages/apollos-ui-connected/src/UpNextButtonConnected/index.js
```diff
@@ -1,135 +1,92 @@
...
         const finishedSeries = !upNextId;
-        const childContentItemsConnection = get(
-          data,
-          'node.childContentItemsConnection.edges',
-          []
-        );
+        const children = data?.node?.childContentItemsConnection?.edges || [];
         // We shouldn't show the button if the content doesn't support it.
-        // The button is considered "supported" if it's a series (has children) or has an up next Id.
+        // The button is considered "supported" if it's a series that has children
```
vinnyjth, line 40: I don't think all items with children should have an up next button. 

For example - a list of resources doesn't need an up next button. I think the `upNextId` only is present on certain content item types (or interfaces)

packages/apollos-ui-connected/src/UpNextButtonConnected/index.js
```diff
@@ -1,135 +1,92 @@
...
     }
-    return <Component loading />;
+    return <Button style={styles.upNextButton(theme)} loading />;
   }
   return (
     <Query
       query={GET_CONTENT_UP_NEXT}
       fetchPolicy="cache-and-network"
-      variables={{ nodeId: nodeId || contentId }}
+      variables={{ nodeId }}
```
vinnyjth, line 33: yessss

packages/apollos-ui-connected/src/UpNextButtonConnected/index.js
```diff
@@ -1,135 +1,92 @@
...
 
   // We want to avoid rendering the Query component if we don't have the ID.
   // Running the query without a contentId throws an error, and the query won't rerun.
-  if (!contentId && !nodeId) {
+  if (!nodeId) {
     if (!loadingEnabled) {
       return null;
     }
-    return <Component loading />;
+    return <Button style={styles.upNextButton(theme)} loading />;
```
vinnyjth, line 27: love that we can remove so many of these abstractions

packages/apollos-ui-connected/src/UpNextButtonConnected/index.js
```diff
@@ -1,135 +1,92 @@
...
+              navigation.push('ContentSingle', { itemId: upNextId })
             }
           >
-            {!finishedSeries ? continueText : doneText}
-          </Component>
+            {!finishedSeries ? (
+              <H4>{continueText}</H4>
+            ) : (
+              <>
+                <Icon size={theme.sizing.baseUnit * 1.25} name={'check'} />
```
vinnyjth, line 60: good simplification here 👍

vinnyjth (changes requested):

packages/apollos-ui-connected/src/UpNextButtonConnected/index.js
redreceipt: the problem is the server returns `upNextId=null` if the series is complete. that's why I took out the logic, it should show a disabled button. not a missing button.

packages/apollos-ui-connected/src/UpNextButtonConnected/index.js
```diff
@@ -1,135 +1,92 @@
...
-    name: 'check',
-    marginRight: baseUnit / 2,
-  }),
-  'ui-connected.UpNextButtonConnected.DoneTextIcon'
-)(Icon);
-
 const UpNextButtonConnected = ({
   continueText,
+  completeText,
   doneText,
```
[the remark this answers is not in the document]
redreceipt: gone

packages/apollos-ui-connected/src/UpNextButtonConnected/index.js
vinnyjth: that sounds reasonable to me - but I still don't think the solve is to show the button for every item, disregarding if it actually supports `upNext`

redreceipt: what content item has children where it shouldn't show up?

vinnyjth: >For example - a list of resources doesn't need an up next button. I think the upNextId only is present on certain content item types (or interfaces)

vinnyjth: CleanShot 2022-08-31 at 09 49 52@2x

`ProgressNode` is a distinct interface that not all content items that implement `ChildContentItemsConnection` implement

redreceipt: the up next button probably shouldn't be trying to retroactively make decisions on whether to show itself. this seems like a problem for one level up to solve.

vinnyjth: That was our approach for a while on these ContentNode connected components, we render them all and then they each decide if they should be show up. 

I'm more than happy if you have a different solve, but this does need to be solved :)

vinnyjth: > the problem is the server returns upNextId=null if the series is complete. that's why I took out the logic, it should show a disabled button. not a missing button.

This situation should be okay though, right? Since this statement will return false since `childContentItemsConnection.length` will be > 0

redreceipt: I changed it to key off `percentageComplete` because what it returns is more straightforward.

PR #2648, opened 2022-08-27
chore: moved test panel to its own component

redreceipt:
Screen Shot 2022-08-27 at 2 31 54 PM


templates/mobile/package.json
```diff
@@ -29,7 +29,7 @@
     ]
   },
   "resolutions": {
-    "react-devtools-core": "4.24.3"
+    "react-devtools-core": "4.19.1"
```
redreceipt, line 32:
```suggestion
    "react-devtools-core": "4.24.3"
```

vinnyjth (approved):

PR #2650, opened 2022-08-27
chore: make apollo client dynamic based on church

redreceipt:
this PR simply lays the groundwork. next, we'll need to get the theme from the server and load that when the client changes. lastly, we'll throw a church picker in the dev test panel.

Screen Shot 2022-08-27 at 7 09 26 PM


vinnyjth:
@redreceipt I'll review this after you resolve the conflicts with `master`

redreceipt:
@vinnyjth its ready

vinnyjth:
Looking great! One finding: 

Error Link should probably consume Church (will save headaches down the road) 
https://github.com/ApollosProject/apollos-apps/blob/b46d345812daf34f3f32d99ac75f1880a717d409/templates/mobile/src/auth/buildErrorLink.js#L17

redreceipt:
@vinnyjth ready

templates/mobile/src/index.js
```diff
@@ -104,109 +112,141 @@ const ThemedNavigationContainer = ({ children, navigationRef }) => {
 const { Navigator, Screen } = createNativeStackNavigator();
 
 const App = () => {
+  const [church, setChurch] = useState(
+    ApollosConfig.CHURCH_HEADER || 'apollos_demo'
```
vinnyjth, line 116: we aren't setting this when compiling the template into a specific church app. Should we be? Otherwise we'll be sending our API conflicting messages, 
```
url: city-first.production.apollos.app
x-church: apollos_demo
```

redreceipt: yea, good call. was gonna wait until it matters but I'll make that change. we'll now always be setting the server URL to the global production server. now the only purpose for the subdomains is universal linking.

redreceipt: OR we could disable this feature if the church header is set?

vinnyjth: I think always setting the `CHURCH_HEADER` is fine. Are you going to add that code to the setup-app step in this PR?

redreceipt: added

vinnyjth:
Innocuous error I noticed in `setup-app`. 

CleanShot 2022-09-08 at 14 11 56@2x

vinnyjth:
This looks good! But I did find a bug in master while testing this out. Details recorded here: https://3.basecamp.com/3926363/buckets/27088370/todos/5310570060

vinnyjth (approved):

PR #2652, opened 2022-08-31
fix: add CurrentVideo context and pass data to onEnd

tristanguckenberger:
### **Purpose**
The purpose of this PR is to prevent the app from crashing by ensuring handleOnEnd receives required data like id.

@vinnyjth Please let me know if there's a better way to do this... It seems like there should be, but I'm blanking.

### **Testing**
1. Check out fix-crash-onEnd
2. Start up any multitenant client (I used newspring)
3. Add console logs for id, elaspedTime, and title inside of [handleOnEnd](https://github.com/ApollosProject/apollos-apps/blob/68f9a5e6583a8bc7855f0d67ccab0c28f26b397f/packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js#L66)
4. With the app running open any video, scrub to a few seconds before the end and let it complete
5. You should see:
- the items get logged to your terminal
- the app no longer crashes

Below is a quick screencap showing that handleOnEnd receives the info it's looking for by logging the id, elapsedTime, and title to my terminal on the left.

https://user-images.githubusercontent.com/19194337/187772211-14066c51-e2d1-43f1-938d-4e3e315f3b17.mov


vinnyjth:
@tristanguckenberger Looking great my friend! Awesome that you dove into typescript land and got context hooked up. Not an easy task :D 

I think there might be a way to solve this without context that's a bit simpler and maybe a bit easier to manage down the road? It's a bit hard to explain in text so I've shot quick loom. Happy to pair on it if that's helpful :) 
https://www.loom.com/share/db22feb0ca9148c58c71bad9251f6703

packages/apollos-ui-media-player/src/Container.tsx
```diff
@@ -176,7 +176,14 @@ const Container: React.FunctionComponent<ContainerProps> = ({
   const { current: sessionId } = React.useRef<string>(Date.now().toString());
   const { title = null } = presentationProps ?? {};
   const id = videos[0]?.id ?? null;
-  const analyticsMeta = { sessionId, title, id };
+  const analyticsMeta = React.useMemo(() => {
+    return { sessionId, title, id };
+  }, [sessionId, title, id]);
+
+  const curriedOnEnd = React.useCallback(
```
vinnyjth, line 183: 💥

packages/apollos-ui-media-player/src/RNVideo.tsx
```diff
@@ -166,7 +166,10 @@ const RNVideoPresentation = ({
   }, []);
 
   const handleVideoEnded = () => {
-    onEnd();
+    const { elapsedTime } = playheadContext;
```
vinnyjth, line 169: can you get access to `totalDuration` here?

vinnyjth: it's not a requirement - but it will our analytics event additional data to work with 
CleanShot 2022-09-01 at 12 20 19@2x

tristanguckenberger: Yeah, we should be able to pull it from playheadContext along with elapsedTime, I'll add it!

packages/apollos-ui-media-player/src/Container.tsx
```diff
@@ -176,7 +176,14 @@ const Container: React.FunctionComponent<ContainerProps> = ({
   const { current: sessionId } = React.useRef<string>(Date.now().toString());
   const { title = null } = presentationProps ?? {};
   const id = videos[0]?.id ?? null;
-  const analyticsMeta = { sessionId, title, id };
+  const analyticsMeta = React.useMemo(() => {
```
vinnyjth, line 179: any reason you added `useMemo`?

packages/apollos-ui-media-player/src/RNVideo.tsx
```diff
@@ -166,7 +166,11 @@ const RNVideoPresentation = ({
   }, []);
 
   const handleVideoEnded = () => {
-    onEnd();
+    const { elapsedTime, totalDuration } = playheadContext;
+    onEnd({
+      elapsedTime,
+      totalDuration,
```
vinnyjth, line 172: 🙌

vinnyjth (commented):
Looking great! One question about `useMemo`, and then we can merge :)

packages/apollos-ui-media-player/src/Container.tsx
tristanguckenberger: It was a recommendation from eslint:
 
Screen Shot 2022-09-01 at 13 13 38

> The 'analyticsMeta' object makes the dependencies of useCallback Hook (at line 183) change on every render. To fix this, wrap the initialization of 'analyticsMeta' in its own useMemo() Hook.eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)

tristanguckenberger: I suppose we could add `// eslint-disable-line react-hooks/exhaustive-deps` on line 179 instead of wrapping it in useMemo?

vinnyjth: Nah, you're solve was correct! I was just curious :D

vinnyjth (approved):

PR #2653, opened 2022-09-01
feat: Add Storybook to Mobile Templates

calebpanza:
### Motivation
In preparation of the UI Kit 2.0, I've gone ahead and reinstalled Storybook into the main mobile template. This PR just re-introduces Storybook and subsequent PR's will include actual components

This Loom shows a walkthrough of the solution

### Solution
Storybook is installed at the root of the project inside of the `storybook` directory. Storybook needs to be manually enabled by the developer at build time. This time around, we will not be rendering the complete application inside of storybook. Instead, Storybook will be used exclusively for documenting/presenting UI Kit elements and Components. This decision was made in order to keep Storybook "simpler but better-er" and was also influenced by conflicting metro configurations between Storybook requirements and the apollos project.

When the app loads, the `metro.config.js` and the `index.js` at the root of the project will read the `projectVariant` value from `app.json`. When set to `"storybook"`, the correct metro config will be created and the Storybook UI will be registered for the application.

### Caveats
For the time being, you will have to decide at build time if you want to run the main application _or_ Storybook. Due to the nature of the current configuration (more on that in the Loom video) you will not be able to run both at the same time.

Storybook (in its current iteration) will be designed/intended for large screen devices, so running `yarn ipados` is advised.

Storybook can be run as a macOS application only when emulating an iPad app. There is no plan to introduce a native macOS app due to complexities in the project configuration and React Native support (at this time, anyway. We can look into more in the future)

### Testing
Open up the `app.json` file and set the `projectVariant ` to `"storybook"`. Run `yarn && yarn start` then open up a new tab and run `yarn ipados`

You should be greeted by the following screen:
Simulator Screen Shot - iPad Air (5th generation) - 2022-08-31 at 16 50 01


calebpanza:
@vinnyjth I'm leaving you as the primary reviewer on this PR only cause it introduces some new structural elements to the project.

@conrad-vanl I attempted to go all-the-way with a unique target for storybook that we could launch independently of the main app. Unfortunately there was not a very straight forward way to manage a custom `metro.config` based on specific targets/environment variables. I would really like to come back to this in the future, but for right now it's become a blocker and I really don't want it to get in the way of actual component work.

templates/mobile/app.json
```diff
@@ -1,4 +1,5 @@
 {
   "name": "apolloschurchapp",
-  "displayName": "Apollos Church"
+  "displayName": "Apollos Church",
+  "projectVariant": "app"
```
vinnyjth, line 4: neat! 

out of curiosity, is this a common approach?

templates/mobile/metro.config.js
```diff
@@ -1,14 +1,16 @@
 const {
   applyConfigForLinkedDependencies,
 } = require('@carimus/metro-symlinked-deps');
+const AppJson = require('./app.json');
```
vinnyjth, line 4: could you use .env files here if you `require('dotenv').config()`?

vinnyjth (commented):
Looks good overall! I'm a little skeptical of the `app.json` approach :) I'll push one more time for ideas on how to use the .env files, but if there really isn't a straightforward path this looks good.

templates/mobile/app.json
calebpanza: No, I don't think so. I saw it on a forum suggesting the approach when someone was attempting to manage this through Environment Variables and ran into similar issues that I did where they weren't coming through on the metro.config

templates/mobile/metro.config.js
calebpanza: I did not try that cause I saw that `dotenv` is no longer maintained? I'd be MORE than happy to handle this using `env`. That was always my preferred approach, but it just kept falling through

calebpanza:
@vinnyjth I agree fully with that hesitation. I can try out `dotenv`. I was just nervous to use it cause the package is no longer maintained.

The only other thing that I was thinking of was to introduce a new file called like `project.json` or something of the sorts. Basically, create a configuration file that is removed from the `app.json` so they can be managed independently. When CL/CI runs per-church, it might be easier to guarantee that we don't accidentally push out a Storybook app if the configuration is its own file. We would just need to add a rule during deployment that replaces the incoming `project.json` with a pre-templated one that has the `variant: "app"` property set

vinnyjth:
@calebpanza  🤔  :) 
CleanShot 2022-09-01 at 15 27 57@2x


I think it's still maintained. Let's try `dotenv` before we switch to using `app.json`. I don't want to introduce another mechanism of configuration unless we have to.

calebpanza:
...huh... that's so strange. Idk what I found, then, that wasn't maintained. Might have just misread or something. I'll make that adjustment, then, to use dotenv instead. I agree it's for sure the "better" approach

calebpanza:
@vinnyjth 😭😭😭

Add `PROJECT_VARIANT` to your env. Setting it to `storybook` will enable the Storybook App. Any other value will enable the main app

vinnyjth (approved):

vinnyjth:
nice work man!

PR #2655, opened 2022-09-01
feat: UI 2.0 theme

calebpanza:
### Motivation
The new UI Kit theme attempts to achieve the following goals:
1. Mirror Figma near 1-1
2. Draw clarity on which color values are statically defined vs. dynamically calculated based on system theme

### Solution
We take advantage of Styled Components abilities to 1. next theme providers and 2. use a function within a sub-theme in order to access the parent theme values

In the parent theme, we define static color values referred to as "BaseColors". These will later be accessible in the theme from 'theme.colors.base'. 

Immediately within the parent theme provider, we need to provide a second theme provider. This theme provider will render either the light or dark theme (based on the system theme). These sub themes intentionally do NOT hold hard color values. They serve to dynamically create the 'text', 'neutral', and 'fill' objects. 

In order to customize these themes, you can either replace the base colors OR just add an additional theme provider. The great part about using Styled Components is that they actively encourage stacking theme providers for specific scenarios. 

### Testing
Snapshots created for base theme and each the system themes (light/dark)

### Storybook
There's a new section in Storybook called "Foundations" which contains palettes for Base Colors and System Colors.

Switching your device between light and dark mode will automatically adjust the stories to use light/dark themes. 

Storybook is also set up to automatically add all of our theme providers to every story.


calebpanza:
Not sure who would be best to review this one, so I figured I'd tag everyone. Lol. But this is ready for review!

vinnyjth:
A few questions I couldn't spot answers for - 

- Does this solution currently use the custom theme stored in `theme.json`? 
- Does this change any of the code currently visible in the apps? Looks like all the changes are isolated to storybook, which is nice :)

calebpanza:
@vinnyjth yeah those are good questions. It looks like I forgot to attach the code walkthrough that answers that first one. Let me take a minute to find that.

As for code-touching: NO! Lol. So this was done to be completely in isolation from the current code-base, but can still be demo'd within Storybook

vinnyjth (approved):
Looks good! You should shoot a loom on some of the new conventions, like the `.style` files and the `Component.SubComponent` naming schemes :)

calebpanza:
Yeah that's a great idea! I ended up opting out of the `Component.SubComponent` convention and went for a bunch of exports instead. Just a style thing, I think. Can you think of any performance reasons to use the other?

PR #2656, opened 2022-09-06
feat: UI 2.0 typography

calebpanza:
### Motivation
This PR introduces the typography styles and components in the UI Kit 2.0. Some of the problems we tried to solve in this PR were:

1. Make type styles easy to access
2. Make Text Components simple to use and named on their intended use
3. Match Figma
4. Provide shorthands for common style overrides so that devs can spend less time overriding Type Components

### Solution
There are a bunch of files that each hold individual "typography categories" (headings, body, etc.). These contain the complete typographic style object needed to render the given style (think Figma text styles)

Along with those, we expose those styles using a base component for each one (H1, BodyText, etc.). Every text-based component is an extension of our `BaseText` component which allows for quick `color` and `textAlign` overrides with prop passing. This is in order to help devs avoid the 6+ lines of code needed every time you wanna change the color/alignment (which is very common).

_This [Loom](https://www.loom.com/share/a499c109b71c4d3d89addfbfebab2c1a) displays a walk through of the solution_

### Testing
`yarn test ui-kit` Snapshot and component testing for all typography styles and components sits at 100%. 🥳

Storybook documentation is written for all of our text styles, too! **Foundations/Typography** includes detailed documentation on our approach to typography, code examples, and best practices. **Foundations/Type Faces** has an interactive demo where you can type in a line of text and see the Type Components react accordingly.

There is also a "Typography" add-on if you open up the addons panel where you can set font colors and text alignment.

### Screenshots

| Typography Docs | Type Faces Demo |
| --- | --- |
| Simulator Screen Shot - iPad Air (5th generation) - 2022-09-06 at 16 59 54 | Simulator Screen Shot - iPad Air (5th generation) - 2022-09-06 at 17 00 00 |


camrun91:
@calebpanza looking over the documentation, it looks good. I am just going to drop conversation questions and such here as I find them so we can discuss them.

the color property is there an opportunity to add "security/sanity checks" here? Thinking if there is a way that we can force the use of a theme key (If my thinking is right I could pass in whatever color I wanted to to a bodyText component and it would render that color if it is supported by react native text.style 

If that is not right let me know.

calebpanza:
@camrun91 great minds! Lol. So it doesn't actually render any color you want. It'll do a lookup for the color within the theme (say text.tertiary). If it exists, it'll use the theme value. If it doesn't (text.highlight or #fff), it'll default to text.primary. Only theme colors are supported

camrun91:
OK cool then this one looks good to me.

camrun91 (approved):

vinnyjth (approved):

PR #2657, opened 2022-09-07
feat: UI 2.0 Spacing & Sizing

calebpanza:
## Motivation
In the new UI Kit, we've moved away from what we traditionally held as the concept of a `baseUnit` into an implicit grid of 4px. With this change, we wanted to the able to give the developers a set of tools that mimic the way that a designer is implementing their designs.

## Solution
We've introduced a new utility function called `unit` which will take an input (`multiplier`) and multiply by 4. It seems rather simple, but that's kind of the point. The main idea, here, is more to give an API for developers that attempts to match the intended use case of the designer.

There is a much deeper breakdown of the approach inside of both Storybook and this [Loom](https://www.loom.com/share/892d131613b14e99820a8c05e5c89816).

## Considerations
This does _not_ match what is in Figma, unfortunately. Figma does not natively support any concept of tokens/styles for spacing & sizing, so it will be the developer's responsibility to look at a design that has a padding of `12px` and interpret that into `unit(3)`

## Testing
Test coverage is at 100% for the `unit` function

Storybook has documentation on usage.


camrun91 (approved):
This one looks good to me.

vinnyjth (approved):

PR #2659, opened 2022-09-09
feat: UI 2.0 Button

calebpanza:
## Motivation
The motivation of this PR is to get the base-line features/functions of a button established. It's not a 100% production ready button that's completely caught up to Figma cause there was already enough stuff to test with what's here.

## Solution
Introduces a Button component inside of the UI Kit 2.0 that extends the functions/features of the `Pressable` component from React Native. We've implemented the basic functions/features of a button

1. set a text label
2. set style variant
3. trigger an action on press
4. set a loading state
5. set the button disabled
6. feedback on press

## Testing
100% test coverage for both functionality _and_ button state (mid/post animations). Some of the tests are written to really explicitly look for exact styles/animations since they were done quite intentionally

Storybook examples under `UI Kit/Button`

## Screenshots
| On Press | Loading | Disabled |
| --- | --- | --- |
| CleanShot 2022-09-09 at 12 32 50 | CleanShot 2022-09-09 at 12 34 25 | CleanShot 2022-09-09 at 12 33 39 |

## TODO
Icon support

Material background support


templates/mobile/metro.config.js
```diff
@@ -5,7 +5,7 @@ require('dotenv').config();
 
 module.exports = applyConfigForLinkedDependencies(
   {
-    resolverMainFields: ['react-native', 'browser', 'main'],
+    resolverMainFields: ['browser', 'main'],
```
vinnyjth, line 8: do we need `react-native` in this list?

calebpanza: Not anymore. I thought I had removed it, but must have forgotten to commit that change. I'll update that.

conrad-vanl (approved):

PR #2661, opened 2022-09-11
fix: cleans up auth error initialization

redreceipt:
1. cleans up some redundant asyncstorage smells
2. waits for cache to be cleared before rendering auth screen
3. adds auth error simulation

out


redreceipt:
@vinnyjth I worked on this all weekend trying to get `ProtectedRoute` to handle app initialization. It's tricky because it's driven off a context that hasn't been established at the auth error function declaration time. Maybe for another PR though. I don't love that Auth screen navigator handles cache initialization but this is close enough to what we have while fixing the error on short notice. I was able to clean a couple of other smells we don't need.

vinnyjth:
@redreceipt Looking great. I tested a bunch locally and I couldn't deduce (experimentally or logically) why we needed any net new code in the Auth navigator, so I removed it since it seemed like a code smell. Seems like you were on the same page there. 

I also found an unrelated error with refresh tokens - seems like they were erroring out b/c the errorLink didn't have any headers to pass down. Not sure why, but patched it by making sure we pass down an empty object if there are no headers.  

I also added a button to the auth link for simulating token refresh

vinnyjth:
Ah - I think I found why you added the extra code to the AuthNavigator - I'll see if I can find a cleaner way to test if the cache is finished initing.

vinnyjth:
@redreceipt Found a clean fix - need to set loggedOut = true before clearing the store.

vinnyjth (approved):

vinnyjth:
> @vinnyjth i think you took out the only part that matters? Now on an auth error, the cache will never be cleared

ahhhhhhh thanks - I forgot to check the async storage after auth error. I thought that `logout` was being called somewhere it wasn't. I'll peek again

vinnyjth:
Attached are gifs of the four states I tested for

1. Logged in user, stays logged in
2. User logging out, stays logged out
3. User who receives an error, stays logged out
4. Logging in

I tweaked the provider tree a wee bit, and I think I fixed all the issues. 

I took all the `apollo-client`  code out of auth-provider, and moved it into more specific sub-providers (that I think can be refactored out entirely at a later point). I also tweaked `ProtectedRoute` so there's no longer a "slide" animation when it mounts, it jumps straight to the initial screen.

templates/mobile/src/auth/AuthPushListener.js
```diff
@@ -0,0 +1,38 @@
...
+
+  useEffect(() => {
+    if (pushId && isLoggedIn) {
+      console.warn('Updating pushId', pushId);
+      updatePushId({ variables: { input: { pushProviderUserId: pushId } } });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [pushId, isLoggedIn]);
+
+  // Callback used for logging in.
```
vinnyjth, line 31:
```suggestion
```

templates/mobile/src/auth/AuthPushListener.js
```diff
@@ -0,0 +1,38 @@
...
+
+  // Callback used for logging out.
```
vinnyjth, line 33:
```suggestion
```

templates/mobile/src/auth/AuthPushListener.js
```diff
@@ -0,0 +1,36 @@
...
+  const { data: { pushId } = { data: { pushId: null } } } = useQuery(gql`
+    query {
+      pushId @client
+    }
+  `);
+  const [updatePushId] = useMutation(UPDATE_PUSH_ID);
+
+  useEffect(() => {
+    if (pushId && isLoggedIn) {
+      console.warn('Updating pushId', pushId);
```
vinnyjth, line 25:
```suggestion
```

templates/mobile/src/auth/AuthPushListener.js
```diff
@@ -0,0 +1,35 @@
...
+  const [updatePushId] = useMutation(UPDATE_PUSH_ID);
+
+  useEffect(() => {
+    if (pushId && isLoggedIn) {
+      updatePushId({ variables: { input: { pushProviderUserId: pushId } } });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [pushId, isLoggedIn]);
+
+
```
vinnyjth, line 30:
```suggestion
```

templates/mobile/src/auth/AuthPushListener.js
```diff
@@ -0,0 +1,35 @@
...
+
```
vinnyjth, line 31:
```suggestion
```

templates/mobile/src/auth/ProtectedRoute.js
```diff
@@ -2,25 +2,36 @@ import { useEffect } from 'react';
...
+          routes: [{ name: 'Tabs' }],
+        });
+      } else {
+        navigation.reset({
+          index: 0,
+          routes: [{ name: 'Auth' }],
+        });
+      }
+      SplashScreen.hide();
+    }, 0);
```
vinnyjth, line 33:
```suggestion
    });
```

PR #2665, opened 2022-09-13
feat: UI 2.0 Add Icon Support to Button

calebpanza:
## Motivation
On the initial development of the Button component, we had a placeholder property for both `leading` and `trailing` icons cause the Icon component wasn't available, yet. This PR introduces the Icon into the Button component.

## Testing
Test coverage is at 100%

Storybook written for it, too.

Simulator Screen Shot - iPad Air (5th generation) - 2022-09-13 at 16 26 37


templates/mobile/src/ui-kit/components/Button/ButtonTheme.js
```diff
@@ -27,6 +27,7 @@ const variants = {
       variant: 'filled',
       backgroundColor: theme.colors.text.action,
       textColor: theme.colors.text.primary,
```
vinnyjth, line 29: will you standardize on either a string or property access eventually? 

seems odd that some colors are passed via name, and some via referencing a config object.

vinnyjth (approved):

conrad-vanl (approved):
Pro/con of explicit "LeadingIconName" "TrailingIconName" props vs something like just "Leading" "Trailing" props where you pass in a component? For example, what if we needed to put an avatar in a button?

templates/mobile/src/ui-kit/components/Button/ButtonTheme.js
calebpanza: So that `ButtonTheme` is only intended to be used within the `Button` component itself. As a standard practice for publicly exposed components, when referring to a color (like typography) you'll pass in the object path: `color="text.primary"`.

The composition of the Button is made up of various theme colors dispersed between children (background and foreground elements) and so we consolidate that map into a limited number of `variant`s (held in the ButtonTheme file).

My eventual end goal (which I'll put a PR out for) is to export prop types and constants so that devs can more easily access these in a "type safe" manner. Haven't bit the bullet yet cause I'm still feeling it out. Plus, it begs the question of "so why not just use TypeScript?" which I don't have a good answer for, yet. 😆

tristanguckenberger (approved):

calebpanza:
@conrad-vanl right now the product work that we have scoped doesn't require anything more just the icons. I agree that it will likely (and should) become a requirement in the future, but I didn't want to creep on the current known scope of work.

PR #2673, opened 2022-09-21
feat: Community Stream Activity Feed

calebpanza:
## Motivation
This PR updates the Community Single to include an onScroll animation in preparation for there to be the addition of a FlatList for the Stream Activity Feed. While the feed is not complete, it is included in this PR as a first round and more PR's will come out to fully support the posting/viewing of each post-type.

## Considerations
There are a bunch of test activities inside of one of the Feeds inside of Stream, so this PR just hardcodes that ID to make testing easier. Once we're able to actually post from the front end, we'll update this to use the community I'm actually looking at.

## Testing
1. Make sure that you have the latest from the API (I _think_ you can use stage, but I did everything locally
2. Set up your dB Config to include `STREAM.API_KEY`, `STREAM.API_KEY_SECRET`, `STREAM.APP_ID`
3. Set up your mobile app env to include `STREAM_FEED_API_KEY` and `STREAM_FEED_APP_ID`
4. Launch the mobile app and navigate to the Community Single screen

| iOS | iPadOS | Android |
| --- | --- | --- |
| CleanShot 2022-09-22 at 09 46 24 | CleanShot 2022-09-22 at 09 49 17 | Screenshot_20220922-095400 |


vinnyjth:
How do I visit the Community Single screen? Connected to staging currently.

vinnyjth:
Animation is a little jittery on my iOS 13 Sim. 


https://user-images.githubusercontent.com/1637694/191798031-74fb8e5f-29d2-49fb-804e-acf0e5d958e9.mp4

calebpanza:
@vinnyjth yeah I'm noticing it on my machine now, too. I'm gonna remove it for the time being

calebpanza:
@vinnyjth I removed the "bounce" option so we should be good to go here

vinnyjth:
Not blocking - but wanted to catalog 

**Join button jumps in width after leaving group when there's less than three community members**


**Initial loading spinner feels like it's in the wrong spot**

vinnyjth (approved):

templates/mobile/src/stream/activity-feed/ActivityFeedProvider.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    setToken();
+  }, [isLoggedIn]);
+
+  if (!isLoggedIn || !streamConfig) {
+    return children;
+  }
+
+  console.log({ streamConfig });
```
vinnyjth, line 50:
```suggestion
```

vinnyjth:
@calebpanza I'm running into an error when signing out of the app, that I _think_ is related to this PR or another recent PR

https://user-images.githubusercontent.com/1637694/191820205-964d161e-767d-499c-ac0f-aef9781dcea2.mp4

vinnyjth (approved):

calebpanza:
@vinnyjth 

1. Button Jump: that's the expected behavior. The button fills the container but has a max width for tablets
2. Still have to fine-tune that component a bit.... It's Stream's FlatList. Thanks for catching that! I'll be putting those kinds of UI enhancements in another PR
3. That's really strange... only thing I can think of is that I might be accidentally be triggering a re-render of the app cause of the way the provider works. I might refactor our Navigation Stack to use `Group`'s so that we can just wrap the screens in the provider and not the entire app

I would love to do all of those enhancements in a "polishing" pr if you're ok with it?

vinnyjth:
> @vinnyjth
> 
> 1. Button Jump: that's the expected behavior. The button fills the container but has a max width for tablets
> 2. Still have to fine-tune that component a bit.... It's Stream's FlatList. Thanks for catching that! I'll be putting those kinds of UI enhancements in another PR
> 3. That's really strange... only thing I can think of is that I might be accidentally be triggering a re-render of the app cause of the way the provider works. I might refactor our Navigation Stack to use `Group`'s so that we can just wrap the screens in the provider and not the entire app
> 
> I would love to do all of those enhancements in a "polishing" pr if you're ok with it?

yep! polishing PR is okay.

calebpanza:
@camrun91 can you check this out when you get a chance?

camrun91:
I agree with @vinnyjth that the join button causes some odd animation with the invite button. but that can be fixed later.

camrun91:
For tracking's state @calebpanza it would be nice to have som way to know it is the end of the list. as it is now this feels like it is cutting something off to me. 


Screen Shot 2022-09-23 at 10 45 45 AM

camrun91 (approved):

PR #2674, opened 2022-09-22
fix: datepicker in dark mode

redreceipt:
- updates components to latest version
- moves date picker to new ui-kit
- converts to functional component

Screen Shot 2022-09-21 at 9 02 27 PM


vinnyjth (approved):
Tested on Android, looks good 👍

PR #2675, opened 2022-09-22
feat: first pass at ChipListFeature

nlewis84:
This PR reimplements the TagFilter into a ChipListFeature. 

- Title
- Subtitle
- List of chips that support title and icon
- Tap on a chip to be taken to a feed of content based on algorithm supplied

|Video in action|The Feed|
|---|---|
|||


vinnyjth:
@nlewis84 could you drop a chiplist into staging so I can play around? thanks!

nlewis84:
Done. Home Tab, 1st feature.

vinnyjth:
@nlewis84 I think your code here looks great! 

The remaining problems you noted, I think they are both on the API 

CleanShot 2022-09-22 at 14 35 51@2x

I would ditch the `title` on the dynamic `ContentFeed` you are creating, that will eliminate that awkward subheader. 

As for only a single card returning, we'll have to dig into that on the API :)

vinnyjth (approved):

packages/apollos-ui-connected/src/ChipListFeatureConnected/ChipListFeature/index.js
```diff
@@ -0,0 +1,74 @@
...
+  FeatureTitles,
+  H4,
+  Icon,
+  PaddedView,
+  styled,
+  withTheme,
+} from '@apollosproject/ui-kit';
+import { ScrollView } from 'react-native';
+
+const ChipsContainer = withTheme(
```
redreceipt, line 13: Don't use these HOCs, use stylesheets and the usetheme hook

packages/apollos-ui-connected/src/ChipListFeatureConnected/ChipListFeatureConnected.js
```diff
@@ -0,0 +1,52 @@
...
+    PropTypes.func,
+    PropTypes.object,
+  ]),
+  featureId: PropTypes.string.isRequired,
+  isLoading: PropTypes.bool,
+  refetchRef: PropTypes.func,
+};
+
+ChipListFeatureConnected.defaultProps = {
+  Component: ChipListFeature,
```
redreceipt, line 49: We don't need component props anymore

packages/apollos-ui-connected/src/ChipListFeatureConnected/ChipListFeature/index.js
vinnyjth: Won't we have to migrate these anyway when we port them to ui-kit 2.0?

packages/apollos-ui-connected/src/ChipListFeatureConnected/ChipListFeatureConnected.js
vinnyjth: > We don't need component props anymore

👍

packages/apollos-ui-connected/src/ChipListFeatureConnected/ChipListFeature/index.js
vinnyjth: Looks like ui-kit 2.0 needs all new paradigm for stylesheets. I don't see a good reason to avoid `useTheme` if we are in the context of the ui-kit / ui-connected package.

packages/apollos-ui-connected/src/ChipListFeatureConnected/ChipListFeatureConnected.js
nlewis84: fixed!

packages/apollos-ui-connected/src/ChipListFeatureConnected/ChipListFeature/index.js
vinnyjth: @redreceipt It's not that we're fixing it; it's that it's going to be rewritten to an entirely new paradigm (styled components). I don't think `stylesheets` is the final solution that @calebpanza settled on.

redreceipt: oh ok

vinnyjth:
@nlewis84 tests :)

packages/apollos-ui-connected/src/ChipListFeatureConnected/ChipListFeature/index.js
calebpanza: Yep! @vinnyjth is spot on. With our UI Kit stuff, the philosophy I'm taking is "quick and dirty to resolve the current problem cause it's getting rewritten anyway". Lol

not the sexiest philosophy, but it's working so far.

packages/apollos-ui-connected/src/ChipListFeatureConnected/ChipListFeature/index.js
```diff
@@ -0,0 +1,74 @@
...
+const ChipText = styled(({ theme }) => {
+  return {
+    textAlign: 'center',
+    alignItems: 'center',
+    justifyContent: 'center',
+    color: theme.colors.secondary,
+  };
+})(H4);
+
+const ChipButton = styled(({ theme }) => ({
```
redreceipt, line 32: theme isn't used here

packages/apollos-ui-connected/src/ChipListFeatureConnected/ChipListFeature/index.js
```diff
@@ -0,0 +1,74 @@
...
+          isLoading={isLoading}
+        />
+        <ChipsContainer horizontal>
+          {chips &&
+            chips.map((chip) => (
+              <ChipButton
+                key={chip.title}
+                onPress={() => onPressItem(chip)}
+                type={'tertiary'}
+                chipList
```
redreceipt, line 62: what is this prop doing?

redreceipt:
In your GIF above, "Test Tag" probably doesn't need to be duplicated. either the nav header or the feed title but not both (or something like that)

packages/apollos-ui-connected/src/ChipListFeatureConnected/ChipListFeature/index.js
redreceipt: nvm, answered my own question

vinnyjth:
> In your GIF above, "Test Tag" probably doesn't need to be duplicated. either the nav header or the feed title but not both (or something like that)

I believe that's been fixed in the most recent version of the backend code.

redreceipt (approved):

nlewis84:
@vinnyjth Argh! Test were passing until that merge where I took out prop-types. Let me see if I can figure out what's going on.

PR #2678, opened 2022-09-23
feat: Community New Post Notifier

calebpanza:
## Motivation
Replaces Stream's notifier with a custom one for Apollos

CleanShot 2022-09-23 at 14 05 10


camrun91:
cough...Dynamic Island...Cough... 

if that isn't "possible" then should it be a little bigger? it looks easy to not notice to me?

calebpanza:
lint and test passing locally, so I'm gonna merge this so we can work on getting a TestFlight build out

PR #2686, opened 2022-09-27
fix: Community Prayer Posts

calebpanza:
## Motivation
As a part of F&I, this PR includes the following fixes:

1. Prayers can be prayed for more than once (you have to navigate backwards and reload the Community in order to pray again)
2. Updated language for the bottom bar
3. A user must be a member of the Community in order to pray for a prayer
4. A user can pray for their own prayer

Simulator Screen Shot - iPhone 13 - 2022-09-27 at 11 40 09


camrun91 (approved):
looks good and seems to work as expected in sim.

packages/apollos-ui-connected/src/CommunitySingleConnected/components/activities/PrayerRequestActivity.js
```diff
@@ -0,0 +1,137 @@
...
+}))(View);
+
+const SocialBar = styled(({ theme }) => ({
+  flexDirection: 'row',
+  alignItems: 'center',
+  paddingBottom: theme.sizing.baseUnit,
+  paddingHorizontal: theme.sizing.baseUnit,
+}))(View);
+
+const Spacer = styled(({ theme }) => ({
```
conrad-vanl, line 40: Is this spacer repeated inside of each activity component? (Prayer, Gratitude, etc)?

Should this be rendered at a higher-level, like within the `renderItem` component for the `FlatList`, or wherever the component that chooses which activity component to render (Gratitude vs Journal vs Prayer, etc)? That way we minimize the risk of having different spacing logic floating around? (Really, the spacing seems to me to be a factor of how the feed is rendered, and not really a concern of this lower level component)

conrad-vanl (approved):
Like it!

We should make a Lottie animation at some point of these prayer hands doing something when you press pray

vinnyjth (approved):

packages/apollos-ui-connected/src/CommunitySingleConnected/components/activities/PrayerRequestActivity.js
calebpanza: No. Lol. Tbh it's an artifact of something else. I thought I would need conditional rendering of the button, so the spacer was there so I could conditionally render the spacer instead of padding. But intimately it's just a margin/padding. This will be handled by the `unit` method in a styled component when we do the UI Refactor (next week probably)

PR #2687, opened 2022-09-27
fix: onboarding slides don't go missing anymore

nlewis84:
This PR addresses a bug where onboarding slides would go missing and the nav dots at the bottom would get weird on us.

https://user-images.githubusercontent.com/72768221/192581168-7d39693d-3cd0-4219-8415-4249cfd1416c.mp4


templates/mobile/src/onboarding/Onboarding/index.js
```diff
@@ -51,45 +51,47 @@ function Onboarding(props) {
...
-    ];
-  } else {
-    // user has been onboarded
-    slides = [
-      FeaturesConnected,
-      LocationFinderConnected,
-      AskNotificationsConnected,
-      FollowConnected,
-    ];
+  if (slides.length === 0) {
```
vinnyjth, line 57: there's a way to do this that might be a bit more "reacty" :) 

it's something like this 


```
const [slides, setSlides] = useState()
useEffect(() => {
   if (!slides.length && !loading) {
     // determine correct slides
   }
   setSlides(computedSlides)
}, [needsNameOnboarding, needsGenderBirthDateOnboarding, loading])
```

your current approach is great, but it runs the risk of blowing up if a new prop is ever passed into `Onboarding` Then this problem would occur again, because `slides` would be set again but this time with new data about the user. 

Explicitly tracking `slides` in `state` will make sure that it's only set once, after the data loads for the first time :)

templates/mobile/src/onboarding/Onboarding/index.js
```diff
@@ -51,45 +51,47 @@ function Onboarding(props) {
     data?.currentUser?.profile?.firstName === null ||
     data?.currentUser?.profile?.lastName === null;
   const needsGenderBirthDateOnboarding =
-    data?.currentUser?.profile?.birthDate === null ||
+    data?.currentUser?.profile?.gender === null ||
```
vinnyjth, line 54: good catch :D

templates/mobile/src/onboarding/Onboarding/index.js
nlewis84: ooo I had something like this for a bit, but then I went the other direction.

nlewis84: I'm hitting some infinite looping due to needing to include slides in the dependency array for useEffect and it also getting set within the useEffect.

vinnyjth: ah that's fun. I think you can actually omit slides from the `useEffect` dependency array. This makes sense conceptually since you don't want the effect to respond to a change in `slides` :)

nlewis84: No luck with just omitting it. I'm trying to get clever with if statements inside the `useEffect` to avoid the loops.

nlewis84:
@vincent here's the video of this new way of doing it working.

https://user-images.githubusercontent.com/72768221/192632753-a79dbcf6-73af-4d2c-af4d-ae6b02901244.mp4

templates/mobile/src/onboarding/Onboarding/index.js
```diff
@@ -32,66 +33,85 @@ const FullscreenBackgroundView = styled({
...
+        } else {
+          // user has been onboarded
+          computedSlides = [
+            FeaturesConnected,
+            LocationFinderConnected,
+            AskNotificationsConnected,
+            FollowConnected,
+          ];
+        }
+        console.log(
```
vinnyjth, line 96: Make sure to delete this comment.log :)

templates/mobile/src/onboarding/Onboarding/index.js
```diff
@@ -32,66 +33,85 @@ const FullscreenBackgroundView = styled({
...
+  useEffect(() => {
+    let computedSlides = [];
 
-  const needsNameOnboarding =
-    data?.currentUser?.profile?.firstName === null ||
-    data?.currentUser?.profile?.lastName === null;
-  const needsGenderBirthDateOnboarding =
-    data?.currentUser?.profile?.gender === null ||
-    data?.currentUser?.profile?.birthDate === null;
+    const setSlidesFunction = async () => {
```
vinnyjth, line 49: wow! funky! I'm assuming this does the hard work of isolating the useEffect dependencies? good find :)

nlewis84: So data was flowing in async, and the useEffect fires off synchronously, so I needed to await data otherwise it would always think data was undefined, which meant you only ever got the 4 main slides.

nlewis84: Oh, and I tried to use `slides.length` instead of `slides.length === 0` but for some reason that was not working. I wasn't sure what was up with that, but the more explicit version works so 🤷‍♂️

nlewis84:
I think I may need to update some tests. Waiting for CI to pass.

nlewis84:
@vinnyjth Are we still unconcerned by tests failing in the front end like a couple of months ago?

templates/mobile/ios/Podfile.lock
```diff
@@ -764,14 +764,14 @@ SPEC CHECKSUMS:
   Flipper-RSocket: d9d9ade67cbecf6ac10730304bf5607266dd2541
   FlipperKit: d8d346844eca5d9120c17d441a2f38596e8ed2b9
   fmt: ff9d55029c625d3757ed641535fd4a75fedc7ce9
-  glog: 85ecdd10ee8d8ec362ef519a6a45ff9aa27b2e85
+  glog: 5337263514dd6f09803962437687240c5dc39aa4
```
nlewis84, line 767: @vinnyjth These hashes changes when I ran yarn. Keep or ditch?

templates/mobile/ios/Podfile.lock
```diff
@@ -764,14 +764,14 @@ SPEC CHECKSUMS:
...
   hermes-engine: bf7577d12ac6ccf53ab8b5af3c6ccf0dd8458c5c
   libevent: 4049cae6c81cdb3654a443be001fb9bdceff7913
   OneSignalXCFramework: 10648e60ff3d335fbffa62d9b44b700b56d49b51
   OpenSSL-Universal: 1aa4f6a6ee7256b83db99ec1ccdaa80d10f9af9b
   Permission-Notifications: 9c6b5cc4f0e6599e9fc3395b77cebddc48f1be41
   Permission-PhotoLibrary: 8227a6ed9f6a971537afe63742d54f5f23a38fe2
-  RCT-Folly: 803a9cfd78114b2ec0f140cfa6fa2a6bafb2d685
+  RCT-Folly: a21c126816d8025b547704b777a2ba552f3d9fa9
```
nlewis84, line 774: @vinnyjth Same here. Keep or ditch?

templates/mobile/ios/Podfile.lock
nlewis84: it failed a test, so I backed this change out.

templates/mobile/ios/Podfile.lock
nlewis84: it failed a test, so I backed this change out.

vinnyjth: yep! that's the right call :)

templates/mobile/src/onboarding/Onboarding/__snapshots__/Onboarding.test.js.snap
```diff
@@ -2,298 +2,4 @@
 
 exports[`the Onboarding component should render Onboarding 1`] = `null`;
 
-exports[`the Onboarding component should render Onboarding 2`] = `
```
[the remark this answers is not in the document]
nlewis84: ummmm I've been trying 😅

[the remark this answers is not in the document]
nlewis84: I tried throwing `{ renderCount: 2}` on the test and also tried setting `<Onboarding slides={[jest.fn(), jest.fn(), jest.fn(), jest.fn()]} data={mockedCurrentUserObject} />`. I'm looking into maybe needing to mock the useEffect now. I don't see any examples of us having to do that elsewhere in the code base.

vinnyjth (approved):
works great, thanks!

nlewis84:
I'll ping Michael tomorrow and see what his thoughts are on fixing the code coverage for this. Also, I don't know why tests are passing in the CI. They are not passing for me locally. 👀

redreceipt (changes requested):
Snap should render

nlewis84:
> Snap should render

@redreceipt Do you have any tips on adding test coverage for the code inside `useEffect`? Without adding a new package, I'm having trouble figuring it out.

templates/mobile/src/onboarding/Onboarding/index.js
```diff
@@ -35,61 +36,71 @@ export { ONBOARDING_VERSION };
...
-  } else {
-    // user has been onboarded
-    slides = [
-      FeaturesConnected,
-      LocationFinderConnected,
-      AskNotificationsConnected,
-      FollowConnected,
-    ];
+  if (loading || slides.length === 0) {
+    return null;
```
nlewis84, line 103: @vinnyjth I did try `{ renderCount: 3 }` and it didn't work. I think this code is the issue, because I'm having trouble setting state in the testing environment and slides will always be `falsy`.

templates/mobile/src/onboarding/Onboarding/index.js
```diff
@@ -35,61 +36,71 @@ export { ONBOARDING_VERSION };
...
   const navigation = useNavigation();
+  const [slides, setSlides] = useState([]);
 
   const userVersion = route?.params?.userVersion || props?.userVersion || 0;
   const { data, loading } = useQuery(WITH_USER_ID, {
     fetchPolicy: 'network-only',
   });
 
-  let slides = [];
+  useEffect(() => {
```
conrad-vanl, line 46: look up the documentation for `useEffect` and its return value. Whenever you do something async inside of `useEffect`, you need to make sure that async work exits if it runs after the component un-renders.

In other words, if the component un-renders and `setSlides` is called, I think in React Native you'll get an RSOD cuz you're trying to set state on a component that doesn't exist. You might be able to do that as simple as:

```js
useEffect(() => {
  let active = true;
  
  const setSlidesFunction = async () => {
    if (!active) return;
    // ...
  }
  // ...
  
  return () => { active = false; }
}, []);
```

templates/mobile/src/onboarding/Onboarding/index.js
conrad-vanl: Because you check for `slides.length` before data has been awaited?

templates/mobile/src/onboarding/Onboarding/index.js
```diff
@@ -35,61 +36,71 @@ export { ONBOARDING_VERSION };
...
+    let computedSlides = [];
 
-  if (loading) {
-    return null;
-  }
+    const setSlidesFunction = async () => {
+      if (slides.length === 0 && (await data)) {
```
conrad-vanl, line 50: Do you need the `slides.length` bit at all? With the way this `if` statement is structured, if `data` changes after it's initial value - then the slides won't be recomputed since this `if` is escaped if `slides` already has length.

So once `slides` has a length, no matter what happens to `data` the inside of this if statement never executes.

conrad-vanl: Also, I'm confused...how is `data` a promise? `data` is a promise?!

conrad-vanl (commented):
I left some disparate comments.

Overall, I don't understand why we need `useEffect`. Is it so that we can keep `slides` from changing after they've been set? It feels like that's a smell and we're not getting at the real reason the data is changing.

With that said, if you wanted to only compute slides once and keep it from changing, `useMemo` might be a better way to do it. Have `useMemo` be based on `loading` perhaps.

```js
let computedSlides = [];
const slides = useMemo(() => {
   // logic for figuring out what `slides` is:
  if (!loading && !computedSlides.length) {
     const needsNameOnboarding =
     // and so on....
  }
  return computedSlides;
}, [loading]); // you might get some compiler warnings about dependencies...ignore 'em
```

Not sure if this would work, but I think it would.

vinnyjth:
> Overall, I don't understand why we need useEffect. Is it so that we can keep slides from changing after they've been set? It feels like that's a smell and we're not getting at the real reason the data is changing.

This is exactly why. The onboarding slides update because the slides change currentUser, so the Apollo hook is smart (and in this case, irritating enough) to rerender and drop slides.

templates/mobile/src/onboarding/Onboarding/index.js
nlewis84: You need slides.length, because the first time through, you need to `setSlides` or onboarding will get skipped entirely. Yep, `data` is a promise. The big issue that we are fighting is that once a user fills out the First/Last Name slide, `data` is updated, which means that they no longer need the first slide, so slides were recalculating and dropping the first slide, but then that made all sorts of wonky behavior happen with what index of the slides we were actually on and how many dots there were at the bottom of onboarding. Having it structured this way means that slides are only calculated one time, the first time this component is rendered, index of slides doesn't get messed up when a user completes info on the About You or Gender/Bday pages, and the nav dots don't get messed up.

templates/mobile/src/onboarding/Onboarding/index.js
nlewis84: Interesting! I'll add a cleanup function to this.

nlewis84:
> With that said, if you wanted to only compute slides once and keep it from changing, `useMemo` might be a better way to do it. Have `useMemo` be based on `loading` perhaps.

I'll test out `useMemo` and report back. @vinnyjth is correct and I gave an in-depth answer in reply to one of your other comments @conrad-vanl

conrad-vanl:
So is the real thing that we simply want the query to stop running once we get our initial data set back? There is a `skip` option and we could be conditionally setting it, something like this stackoverlow answer: https://stackoverflow.com/a/60801406

I don't know if my `useMemo` advice above is right or any better, it could be worse. But simply just turning the query off, (by setting `skip=true`) after we've correctly constructed the slides array the first time might be the right solve.

> You need slides.length, because the first time through, you need to setSlides or onboarding will get skipped entirely.

I'm a little confused by this...are you saying that we're setting some value to slides initially so that onboarding doesn't get skipped? Should we just show a loading state or something else then?

vinnyjth:
@conrad-vanl Skip could work, but wouldn't we need `useState` for that? We would want to skip only after having fetched data once. Also I'm not sure how `skip` interacts with updates to a query unrelated to the original query call. In this instance - even if the `useQuery` doesn't refire directly, it's still getting new `data` because the internal cache is updating. I'm not sure if `skip` stops that propagation? 

`useMemo` sounds like a great route to me. @nlewis84 I have some time this morning to pair on it if you would like, we could try some cleanup and then see if we can get tests working.

conrad-vanl:
I'm thinking more about code cleanliness and clarity when someone goes back through and reads what's going on. having both `useMemo` and `skip` could be right as well..... right now it wasn't clear to me what we're trying to do here -> it looked like a code smell/mistake that we are preventing updating the component after we initially get data.

Ultimately turning the `userQuery` "off" feels like it's the most clear expression of what we want...

vinnyjth:
@conrad-vanl I've seen the pattern of `useEffect` / `useState` to capture data once and then avoid changing it more than once, I thought it was a accepted approach on how to do this :) 

in my mind the oddity is the use of await / inner-function

conrad-vanl:
We could also just use the `query` function instead of the hook - https://www.apollographql.com/docs/react/api/core/ApolloClient#ApolloClient.query

This would be async so we'd have to do it in a `useEffect` and `setState` ...

vinnyjth:
> Ultimately turning the userQuery "off" feels like it's the most clear expression of what we want...
> We could also just use the query function instead of the hook

These both feel strange to me - they make sense in component land, but they break down in stateless / functional land where a re-render could strike at anytime :) And the dependancy for how/when useQuery should be turned off is `useQuery` itself (which I think you address by mentioning  useEffect/setState)

vinnyjth:
I can see a version of this component that that uses `setState` for `slides` and determines slides based on `query` inside `useEffect` with no dependencies (so it only fires on mount). That makes sense to me.

nlewis84:
@vinnyjth I'll ping you in a bit!

conrad-vanl:
> in my mind the oddity is the use of await / inner-function

Maybe it's just the variable naming or something to clean this up....

```js
const setUpSlidesOnlyOnce = useCallback(async () => {
  await data;
  // set up slides based on data
}, [data]); // not sure if we should have listed dependencies

useEffect(() => {
  setUpSlidesOnlyOnce();
}, []);
```

vinnyjth:
y'all, this PR has excellent tests and works great. I vouch for it 👍 💯

redreceipt (approved):

PR #2688, opened 2022-09-27
feat: Add content table

tristanguckenberger:
### [Purpose](https://3.basecamp.com/3926363/buckets/27088370/todos/5221899447)
The purpose of this PR is to replace the ContentParentFeedConnected/ContentChildFeedConnected and swap out the horizontal feed for a vertical one like whats in the image below:

image

### Testing
1. Checkout add-content-table
2. Start up any multi-tenant client from deploys/mobile
3. Login and go to any series that has children and navigate by selecting the child/sibling items

### Demo
https://user-images.githubusercontent.com/19194337/192613421-cecbbe1d-4199-47a6-bf5f-e1f0d804ad1f.MP4


tristanguckenberger:
@vinnyjth I've gone back and forth here and ultimately _tried_ to cleanly consolidate everything to just being contained in `ui-connected/src/ContentTableFeedConnected/`. Is this too much consolidation? Should I re-break these out into smaller sized ui-kit components (e.g. ContentTable, ContentTableRow, ContentTableImage) like I had done before my recent refactors?

vinnyjth:
> I've gone back and forth here and ultimately tried to cleanly consolidate everything to just being contained in ui-connected/src/ContentTableFeedConnected/. Is this too much consolidation? Should I re-break these out into smaller sized ui-kit components (e.g. ContentTable, ContentTableRow, ContentTableImage) like I had done before my recent refactors?

I wouldn't spend too much time on the correct refactor, because we will need to re-write or refactor this component when we migrate to @calebpanza's new UI-kit :) what you are describing sounds great, as long as the data driven elements are separated reasonably from the ui elements.

calebpanza:
@tristanguckenberger this component actually inspired Conrad's refactor of the TableView component. I have it on my list as a high-priority component to port over to 2.0. if you'd like to pair up when I get to it, I'd be more than happy to do that with you!

vinnyjth:
@calebpanza that sounds amazing but note that we can't port this component to 2.0 without also porting ContentNodeConnected :( (correct me if I'm wrong)

calebpanza:
@vinnyjth that's ok! With the way that we're doing it, we would do like "which components are needed to port ContentNodeConnected?", prioritize and test those, _then_ port over the Connected components.

vinnyjth:
@calebpanza I was under the impression we couldn't do that without porting over the Connected Component because of how the package imports work. UI 2.0 packages live in `templates/mobile/..` and `ConnectedComponents` live in `ui-connected`, and `ui-connected` can't import code in `templates/mobile` right?

calebpanza:
Correct. BUT work done/needed to be done inside of UI Connected influences the priority of components that are brought into UI Kit 2.0. SO this PR should still get worked on and merged, but it just adds to the priority of getting the TableView component added.

And then once we're at a place to where the entire ContentConnected can be brought into the main `templates/mobile`, we can do that.

vinnyjth:
Cool! I'm all for that - so are you suggesting we merge this PR using old paradigms but prioritize moving over ContentConnected as a result?

calebpanza:
If there is necessary functionality that is time sensitive in this PR, then merge it. If this is just an experiment/new feature that isn't needed for some time, then we could hold off. Either way is fine by me. I've been doing a mix of the 2 approaches with Community and it's been working out well so far.

vinnyjth:
> If there is necessary functionality that is time sensitive in this PR, then merge it. If this is just an experiment/new feature that isn't needed for some time, then we could hold off. Either way is fine by me. I've been doing a mix of the 2 approaches with Community and it's been working out well so far.

thanks for your insight :) we need this for Woodmen, so let's go ahead and ship it as-is an then you can help us all migrate ui-connected down the road :D

packages/apollos-ui-connected/src/NodeSingleConnected/index.js
```diff
@@ -30,8 +29,7 @@ const NodeSingleInner = ({ nodeId, ImageWrapperComponent }) => (
     />
     <ScriptureNodeConnected nodeId={nodeId} />
     <NodeFeaturesConnected nodeId={nodeId} />
-    <ContentParentFeedConnected nodeId={nodeId} />
```
redreceipt, line 33: 😍 

you could probably go ahead and delete the old components

tristanguckenberger: Done! 😁

redreceipt (approved):

packages/apollos-ui-connected/src/ContentTableFeedConnected/ContentTableFeedConnected.js
```diff
@@ -0,0 +1,81 @@
...
+  const loadingEnabled = loadingStateTypes.includes(nodeId.split(':')[0]);
+  const variables = { nodeId };
+  const query = GET_CONTENT_TABLE_FEED_ITEMS;
+  const navigation = useNavigation();
+
+  if (!nodeId && !loadingEnabled) {
+    return <></>;
+  }
+
+  const dataLoader = ({ data }) => {
```
conrad-vanl, line 20: When you have inline functions like this consider `useCallback` as a potential performance improvement. Right now, every render, these functions are re-created even if they don't need to be. This may cause no actual improvement in practice, but it can if the components that are rendered below (like `ContentTableFeed`) are `pure` or `memo`-ized.

That would look something like:

```js
const dataLoader = useCallback(({ data }) => { 
   // ...
}, [ ... ]);
```

Here's an article on it:
https://dev.to/ltsharma/performance-optimisation-react-native-with-hooks-a77

packages/apollos-ui-connected/src/ContentTableFeedConnected/ContentTableFeed/index.js
```diff
@@ -0,0 +1,90 @@
...
+  const { data, loading, error } = useQuery(query, {
+    variables,
+    fetchPolicy: 'cache-and-network',
+  });
+
+  if (error || isLoading || loading) {
+    return <></>;
+  }
+
+  const dataLoaderContent = dataLoader({ data });
```
conrad-vanl, line 56: Help me understand -> why do we have this `dataLoader` prop? It looks like this component (which is a child) loads the data, but then expects a prop to be passed in that transforms the data?

Why not either:

- Load the data in the parent component and just pass the transformed data into this component, and let this component be 100% visual, or

- Load the data and transform it in this component and not have query and dataloader props?

packages/apollos-ui-connected/src/ContentTableFeedConnected/ContentTableFeed/ContentTableFeedRow.js
```diff
@@ -0,0 +1,120 @@
...
+
+const StyledCell = styled(() => ({
+  height: 75,
+  maxHeight: 75,
+  backgroundColor: 'transparent',
+  paddingHorizontal: 8,
+}))(Cell);
+
+const StyledCellTextSmall = styled(({ theme }) => ({
+  fontSize: theme.helpers.rem(0.675),
```
conrad-vanl, line 25: You did good work here matching the styles of the prototype. Question -> The [component in Figma](https://www.figma.com/file/YHJLj8pdFxWG9npF2YmB3r/UI-Kit-2.0?node-id=174%3A298) just uses `H5` and `UITextSmall` (which doesn't appear to exist in our current UIKit, so we might just use `UIText`).

What do you think are the advantages of using `CellText` here vs just importing like `H5`?

packages/apollos-ui-connected/src/ContentTableFeedConnected/ContentTableFeedConnected.js
```diff
@@ -0,0 +1,81 @@
...
+      const content = edges.map((edge) => edge.node);
+      const currentIndex = content.findIndex((item) => item?.id === nodeId);
+
+      return { content, currentIndex, nodeId };
+    }
+
+    return null;
+  };
+
+  const renderItem = ({ item }) => (
```
conrad-vanl, line 43: Whenever I see a `render`-function in a component I want to wrap it in a `useCallback` or figure out how to move the `renderItem` function itself outside of the component entirely. React Native's documentation confirms with this: https://reactnative.dev/docs/optimizing-flatlist-configuration#avoid-anonymous-function-on-renderitem

In this case, `renderItem` does not depend on any variables from within the component context, so you could easily move this `const renderItem =` outside of the component function altogether (like in the example in that link above) and everything will still work!

conrad-vanl (approved):
Just going to offer some suggestions for improvement

packages/apollos-ui-connected/src/ContentTableFeedConnected/ContentTableFeed/index.js
tristanguckenberger: > Load the data in the parent component and just pass the transformed data into this component, and let this component be 100% visual, or

I  think I set `dataLoader` as a prop trying to clean up/simplify this file, but this makes more sense

packages/apollos-ui-connected/src/ContentTableFeedConnected/ContentTableFeedConnected.js
tristanguckenberger: Good tips! I moved renderItem to ContentTableFeed to further separate data and ui (and per your suggestion, out the the component func in that file).

conrad-vanl:
@vinnyjth once this gets merged....we need to get it deployed out to Woodmen Valley I think? Is this the only thing holding up their guides changes?

vinnyjth:
> @vinnyjth once this gets merged....we need to get it deployed out to Woodmen Valley I think? Is this the only thing holding up their guides changes?

yep!

conrad-vanl:
Sweet. Lets trigger a build on City Rise too -> this is also something they asked for and I think will be happy.

vinnyjth:
Tested and looking great! Nice work!

PR #2689, opened 2022-09-28
feat: Pressables

calebpanza:
## Motivation
We use a couple of different "styles" of buttons based on specific product needs. Right now, those are "Button", "Chip" and "Icon". In practice, we've run into a couple of problems with these components:
1. Naming them in code can be challenging. `Icon` is already an existing component and `Chip` can be used as both a static label and a button. To name them `ButtonIcon` could imply that the component is the Icon rendered _inside_ a Button, while `Chip` doesn't actually inform the functionality of the component in its name.
2. These components share a lot of references, but are entirely different in the way they are represented visually. The interactions are the same, but the UI is not

## Solution
All of these components are built on top of the React Native `Pressable` component. This means that the full API of the RN Pressable should be available to all 3 styles of buttons represented above. In order to help drive clarity in both function and form, we're renaming our Button components with a prefix of `Pressable`. In the case of Icon and Chip, this is especially useful as `PressableIcon` and `PressableChip` both represent the form and function of the component in its name, while also nodding to the underlying API which _should_ make onboarding/comprehension of the components much easier and faster.

I've also pulled out the `ButtonTheme` and dropped it into the `themes` directory. This is because we want to be able to share the button theming across multiple components, and having them sit in a centralized location should help when searching through the component tree. There is also a new `PressableRipple` that houses the ripple effect inside of a `blocks` directory that will be intended to house lower level building blocks shared by components

# PressableButton
Renamed the Button component. Yes, the name is a bit redundant BUT the point is to hammer home clarity. It also helps provide a distinction from the existing `Button` component that exists in the current UI Kit which should help with migrations

# PressableIcon
| Default | Loading | Disabled |
| --- | --- | --- |
| CleanShot 2022-09-28 at 13 50 49 | CleanShot 2022-09-28 at 13 50 49 | CleanShot 2022-09-28 at 14 04 43 |

# PressableChip
| Default | Loading | Disabled |
| --- | --- | --- |
| CleanShot 2022-09-28 at 15 54 36 | CleanShot 2022-09-28 at 15 55 15 | CleanShot 2022-09-28 at 15 55 58 |


conrad-vanl (approved):
Should `ButtonTheme` be called `PressableTheme`?

Looks good!

calebpanza:
Yeah that's not a bad idea! Do you mind if I rename the _page_ in Figma to be called "Pressable" to match? I know this naming is more code-related, so I understanding if you want it organized differently on Figma

vinnyjth (approved):

conrad-vanl:
> Yeah that's not a bad idea! Do you mind if I rename the page in Figma to be called "Pressable" to match? I know this naming is more code-related, so I understanding if you want it organized differently on Figma

Yeah I'd prefer naming to match, so I'm good with renaming in figma

PR #2695, opened 2022-09-30
chore: adds church picker

redreceipt:
demo below. theme colors aren't working and I started to debug but aren't we changing theming soon? also, the wordmark is unimplemented. but the strength of this is seeing new churches' tabs and data.
 
https://user-images.githubusercontent.com/2659478/195435689-a17e4684-3874-4404-af05-1eac4fb953d2.mp4


redreceipt:
dependent on github.com/apollosproject/apollos-cluster/pull/323

calebpanza:
Really cool idea! I'm excited to play around with this in TestFlight. 

What insurances are we gonna have, though, to make sure that this feature doesn't accidentally ship in one of our church apps?

With the UI Kit work that I'm doing, I introduced a new environment variable called `PROJECT_VARIANT`. Maybe we could expand on this so that our options are:

`storybook` - for rendering the Storybook app
`multitenant` or `demo` - for rendering the Apollos Demo app with the Church Picker
`church` or `client` - for rendering an app for a single church without the Church Picker

redreceipt:
if you define `CHURCH_HEADER` in the env, you can't pick a new one

calebpanza:
We talked about this in person, but I'm reiterating it here:
Instead of enabling a feature by _omission_ of a value, can we have the logic read `if CHURCH_HEADER=apollos_demo, enable the church picker`. That way we don't change the convention of requiring a value for CHURCH_HEADER, but instead we resolve the _functionality_ of the app based on the value of CHURCH_HEADER

redreceipt:
Ping for a review on this one

vinnyjth (changes requested):
Crash on running with no `x-church` variables set 

```
 LOG  undefined {"errors": [{"extensions": [Object], "message": "Context creation failed: \"undefined\" not a valid church slug!"}]}
 WARN  Possible Unhandled Promise Rejection (id: 1):
TypeError: Cannot read property '__schema' of undefined
TypeError: Cannot read property '__schema' of undefined
    at _callee$ (http://localhost:8081/src/client/cache.bundle?platform=ios&modulesOnly=true&app=com.differential.apollos&dev=true&minify=false&runModule=true&shallow=true:149:51)
    at call (native)
    at tryCatch (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:23995:23)
    at invoke (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:24165:32)
    at anonymous (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:24037:30)
    at call (native)
    at tryCatch (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:23995:23)
    at invoke (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:24067:30)
    at anonymous (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:24077:21)
    at tryCallOne (/Users/distiller/hermes/build_iphonesimulator/lib/InternalBytecode/InternalBytecode.js:53:16)
    at anonymous (/Users/distiller/hermes/build_iphonesimulator/lib/InternalBytecode/InternalBytecode.js:139:27)
    at apply (native)
    at anonymous (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:30494:26)
    at _callTimer (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:30390:17)
    at _callReactNativeMicrotasksPass (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:30429:17)
    at callReactNativeMicrotasks (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:30639:44)
    at __callReactNativeMicrotasks (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:3109:46)
    at anonymous (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:2887:45)
    at __guard (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:3092:15)
    at flushedQueue (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:2886:21)
    at invokeCallbackAndReturnFlushedQueue (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.differential.apollos:2879:33)
```
    
CleanShot 2022-12-01 at 16 11 32@2x

`.env.` file is empty.

redreceipt:
sweet, thanks for reviewing. not shocked it's broken. it's pretty old. I'll see what's up and get it fixed.

redreceipt:
Oh I see now, @calebpanza wanted to make sure this only makes into the demo app which I agree with, so I changed it to require the `CHURCH_HEADER` variable. But I think there are other ways we can lock that down. I really like that no `.env` is required for development. For now, we can base the picker off `process.env.NODE_ENV` like our other development tools and then figure out a way to deploy on the demo app in a follow up PR. The immediate need is easily being able to switch churches in development which this will solve.

vinnyjth:
Gotcha - so should I re-add my `church_header` env variable or are you going to add a workaround?

vinnyjth:
I honestly can't remember where we netted out on what belongs in the .env :)

redreceipt:
> I honestly can't remember where we netted out on what belongs in the .env :)

my point exactly. that's why it's better to not require it. I'm gonna post an update.

calebpanza:
@redreceipt @vinnyjth I still have every intention to run with this and apologize for the delay. I just haven't had time to come back to this with all the work on Daily Habits. Should be able to prioritize contributing to this by the EOY, though.

redreceipt:
@vinnyjth you shouldn't need an `.env` now

templates/mobile/src/settings/UserSettings.js
```diff
@@ -176,6 +178,14 @@ const UserSettings = () => {
           {`App Version: ${getVersion()} (${Config.VERSION || 'development'})`}
         </H6>
 
+        {/* church picker */}
+        {process.env.NODE_ENV === 'development' ? (
```
vinnyjth, line 182: I like that this was `apollos_demo` because it made these changes super valuable for demoing the prod app with other churches data. Could the logic here be `NODE_ENV==dev or apollos_demo`?

redreceipt: done. good call. that takes care of all concerns.

vinnyjth (dismissed):
Incredible to see this in motion. 

Approving with two notes: 

1. Nothing user/campus related seems to work. That's not unexpected. And you seem to get booted back to `apollos_demo` if you try to change your campus. 
2. You get an error if you select a church without a theme (like `fake`) 
CleanShot 2022-12-06 at 15 47 33@2x

redreceipt:
that's no good. I'd like to fix before merging. the second is easy. the first I'll need to dig in. can you give me an example to work through?

vinnyjth:
@redreceipt A good example is to try and change your campus. 

I expect this to be a big challenge because the currently logged in user isn't valid in whatever church slug you end up selecting. Your best path is probably to log the user out immediately upon switching churches.

vinnyjth:
I approved because this has huge value today, even with the oddities. I'm fine if you focus on fixing the first issue in a follow up PR.

vinnyjth:
Really close! 

1. Login to the app.
2. Select "Woodmen Valley" 
3. Login to the app again
4. Visit the "Profile" screen and try and change your campus
5. The app kicks you back out to login and you are back in Apollos Demo
- The "selector" is defaulted to Apollos Demo, I wonder if it's "on select" is firing or if it's somehow changing the state.

redreceipt:
doesn't seem to be the campus call, that one comes back fine with the correct headers, must be something else

Screenshot 2022-12-06 at 5 10 05 PM

vinnyjth:
@redreceipt I think it's more likely the Connect screen is rerendering or something when you open the Campus selector, triggering the state of the the church to be reset to 0 (or something)

vinnyjth:
@redreceipt maybe because it's a navigation action? The MapView being inside a Navigator modal?

redreceipt:
something like that. I'm investigating. hard to believe that would cause the whole app to re-render

redreceipt:
verified it doesn't happen in master so something is triggering a full re-render of the whole app, just not sure what

redreceipt:
figured it out. seems to be a UI crash with the `Picker` from `react-native-picker`. commented out and the map view shows up fine. with it, even without using, the whole app re-renders.

redreceipt:
gonna upgrade the package, we're a major version behind

redreceipt:
upgrading versions didn't help. we may to consider a different UI.

redreceipt:
I'm gonna try and fix systematically:
- upgrade picker
- (if that doesn't work) upgrade maps pkg
- (if that doesn't work) remove picker from our repo and figure out another UI for this PR

redreceipt:
> I'm gonna try and fix systematically:
> 
> * upgrade picker
> * (if that doesn't work) upgrade maps pkg
> * (if that doesn't work) remove picker from our repo and figure out another UI for this PR

this theory didn't prove out on master so now I'm confused

redreceipt:
@vinnyjth got it. even fixed the theme colors. icons and word marks aren't working but I think we should pull those as URLs and pass them through the theme too. I can do that in another PR. I dismissed your review so you could retest with the changes.

vinnyjth (approved):
Looking great! Nice work. The button at the bottom of the picker was a good UI solve.

PR #2708, opened 2022-10-15
chore: remove onboarding versioning

redreceipt:


conrad-vanl:
Why are we removing onboarding versioning?

vinnyjth:
We will need versioning to make sure existing users see the new onboarding slides as we add them. 

An example of this is the upcoming "User Tags" onboarding slide.

PR #2709, opened 2022-10-16
fix: auth enhancements

redreceipt:
- easier to tap log in button
- forced openID if available
- cant skip openID
- auto loads openID
- removes webview

https://user-images.githubusercontent.com/2659478/196012023-0f4650b6-cd6c-417e-85a1-d08640a19995.mp4


conrad-vanl:
Screen Shot 2022-10-17 at 7 59 51 AM

WDYT about replacing this screen with just a centered `<ActivityIndicator/>` and then if the openid flow gets exited or error'd we can display a simple error message with a link/button to open OpenID again?

templates/mobile/src/auth/LandingSwiper/slides.js
```diff
@@ -71,43 +49,57 @@ export const Intro = named('ui-onboarding.LandingSwiper.slides.Intro')(
...
+              flexDirection: 'row',
+              justifyContent: 'center',
+              alignItems: 'center',
+            }}
+          >
+            <H5 secondary centered>
+              Already have an account?{' '}
+            </H5>
+            <PaddedView>
+              <ButtonLink onPress={onPress}>Log in →</ButtonLink>
```
conrad-vanl, line 74: I don't think the hit area on this is going to be any bigger then it was before, the padding is added outside of the link. I think you need to either add padding directly to `ButtonLink` or use `hitSlop`

redreceipt:
@conrad-vanl done, done, and done

redreceipt:
I think I need a lesson on hitslop. adding it does nothing.

conrad-vanl:
> I think I need a lesson on hitslop. adding it does nothing. padding worked though.

Yeah, you may not be able to notice any difference unless you're testing on a real device.

Can you provide an updated screenshot with the padding implementation?

redreceipt:
> Can you provide an updated screenshot with the padding implementation?

cc @conrad-vanl you can't really tell from the video but I'm clicking below, to the right, and above the button


https://user-images.githubusercontent.com/2659478/197589593-06f3f77f-642f-45b2-a7ca-7fe79f9e8c7c.mp4

vinnyjth (changes requested):
Sorry it's taken me so long to review this 😅 

We're showing the "Would you like to use the church_name .... " prompt on top of the first onboarding screen. Clicking confirm opens up the webview. We should see if we can avoid showing this prompt if you've already completed the flow.

vinnyjth:
Maybe helpful for debugging - I see three of these warnings: 
CleanShot 2022-10-27 at 09 11 32@2x

redreceipt:
ok @vinnyjth this is ready for another review. it ballooned much more than I originally intended on my most recent commit but I think I took care of the issue you found as well as some other small auth bugs. Most notably, I removed the `AuthContext` provider and just made the login/logout functions direct utility imports. It was causing subtle re-render bugs every time we set a new login status. Plus they don't need to be in the context anyway. I would appreciate a glance over all the code one more time as well as some light sim testing over various auth functions.

vinnyjth:
@redreceipt I like the cleanup you were able to achieve, but token refresh no longer works :/

vinnyjth:
Nevermind - same issue happens in master I think ... that worries me a bit. I'm trying to make sure token expiration is working as expected

vinnyjth (approved):

templates/mobile/src/auth/LandingSwiper/index.js
```diff
@@ -24,12 +24,14 @@ const LandingSwiper = () => {
   });
   const slideComponents = useMemo(() => {
     // Default to Auth / OpenID landing screen if we have OpenID providers, otherwise direct to IdentityEntry.
-    const defaultAuthScreen = data?.openIdProviders?.length
-      ? 'Auth'
-      : 'IdentityEntryConnected';
+    // const defaultAuthScreen = data?.openIdProviders?.length
```
vinnyjth, line 27:
```suggestion
    const defaultAuthScreen = data?.openIdProviders?.length
```

templates/mobile/src/auth/LandingSwiper/index.js
```diff
@@ -24,12 +24,14 @@ const LandingSwiper = () => {
...
+      // ? 'Auth'
```
vinnyjth, line 28:
```suggestion
      ? 'Auth'
```

templates/mobile/src/auth/LandingSwiper/index.js
```diff
@@ -24,12 +24,14 @@ const LandingSwiper = () => {
...
+      // : 'IdentityEntryConnected';
```
vinnyjth, line 29:
```suggestion
      : 'IdentityEntryConnected';
```

templates/mobile/src/auth/LandingSwiper/index.js
```diff
@@ -27,9 +27,11 @@ const LandingSwiper = () => {
     const defaultAuthScreen = data?.openIdProviders?.length
       ? 'Auth'
       : 'IdentityEntryConnected';
+      const defaultAuthScreen = 'IdentityEntryConnected';
```
redreceipt, line 30:
```suggestion
```

vinnyjth:
@redreceipt can you resolve conflicts for me? thanks :D

vinnyjth:
CleanShot 2022-11-01 at 08 40 15@2x

🤔  is this a legit failure?

vinnyjth:
My fix didn't work @redreceipt :(

redreceipt:
yea I'm looking into it now. must be something else.

redreceipt:
I think I got. The method I was using to identify the component was flaky (there was a missing trailing space at the end of the text) so I moved it to use `testID`.

PR #2710, opened 2022-10-18
feat: adds badge links

redreceipt:
https://user-images.githubusercontent.com/2659478/196431150-66123987-eb01-4876-99a0-2ea3a4be6553.mp4


packages/apollos-ui-connected/src/BadgeListFeatureConnected/getBadgeListFeature.js
```diff
@@ -15,6 +15,15 @@ export default gql`
             source
           }
           isActive
+          action {
+            title
+            action
+            relatedNode {
```
vinnyjth, line 21: Let's use the `relatedNode` feature fragment here. There's no reason not to, and it will make sure we support other actions down the road (this feature will continue to be used outside the context of studio C) 

https://github.com/ApollosProject/apollos-apps/blob/fdf88c1e186f07df126fd3c314653f985e638f6b/packages/apollos-ui-fragments/src/features.js#L435-L451

https://github.com/ApollosProject/apollos-apps/blob/fdf88c1e186f07df126fd3c314653f985e638f6b/packages/apollos-ui-fragments/src/features.js#L247

redreceipt: done

vinnyjth (approved):

PR #2711, opened 2022-10-18
chore: move tv and web into app templates

redreceipt:


calebpanza:
@redreceipt we might wanna rename the `web` directory to `tv-web` since we are likely to have multiple web based products

redreceipt:
> @redreceipt we might wanna rename the `web` directory to `tv-web` since we are likely to have multiple web based products

can we do that when that day comes so we can pick the most appropriate names

redreceipt:
> @redreceipt did the `lib` folder get moved over? I'm seeing an error with `import client from '../lib/apolloClient';` in the `AppProvider.js`

@jordanwade not sure why you're seeing that error but it seems like it would flag the linter in the CI if there was a problem. here's the folder structure which lines up with the import statement

Screen Shot 2022-10-19 at 4 57 45 PM

jordanwade:
Cool, I got it to work. Something strange must have happened locally. 👍

jordanwade (approved):

calebpanza (approved):

PR #2712, opened 2022-10-19
chore: Community UI kit 2.0

calebpanza:
## Motivation
1. Refactor Community Single locally in the `mobile` directory
2. Use UI Kit 2.0 + Styled Components
3. Remove all references to `@apollos/ui-connected` and `@apollos/ui-kit` from this screen

While there are a LOT of code changes here, the majority of them are translations from UI Kit 1.0 over to 2.0. Along the way, there were some additional bugs found in 2.0 that were addressed.

Not all of the features displayed in Community are working perfectly. Some of them exist as placeholders in order to offer smaller PR's that physically enable their functionality. That being said, everything that could be done previous is working as expected. 

I also took the liberty of adding some Error States to the screen which was fun!

| Light Mode | Dark Mode |
| --- | --- |
| iphone_light | iphone_dark |

## Testing
I've broken out some more details inside of [this Loom](https://www.loom.com/share/2b1c4002773246bc854b3666c025968c) to explain how to test and what to look for. In order to avoid an insanely long and detailed and boring description, I would advise just checking that out!


camrun91:
@calebpanza the only feedback I have from what I can see is that makeing the loading state icon a little more obvious would be good I think because it is so high on the screen it could get lost or not seen. this is a very minor thing and should not be blocking at all.

camrun91 (approved):
you can take this review with a pretty large grain of salt but it looks to function and work for me. If there are anti patterns or anything used I can't speak to that lol

calebpanza:
@camrun91 I like that idea! I had totally forgotten about that. I'll open up a new PR that updates the loading indicator. Gonna have to do a bit of magic here, I think, in case we have images that are really bright (in which the light loading indicator would hide)

PR #2715, opened 2022-10-19
feat: Custom Icon Generation

calebpanza:
## Motivation
We are using Phosphor as our icon set, but they don't have _everything_ that we couple possibly need/want for our products. We _could_ contribute to Phosphor's open-sourced project, but we could end up waiting for quite some time before seeing our contributions within our code base.

The problem presents itself as: we want to be able to add custom icons while still adhering to the principle of "phosphor is the icon set that we're using for Apollos"

## Solution
The phosphor-react-native library is actually just running a code-generation script that takes the Phosphor SVGs and translates them into RN components. I've gone ahead and localized their script and made a few adjustments to fit our workflow better.

In order to add a custom icon to the Apollos Project, simply export SVG's for each of the 6 Phosphor icon weights and drop them into the `svgs` directory organized into their weights. From there, run `yarn generate-icons` from the root of the `apollos-apps` workspace and voila! You'll now see a new directory appear in `templates/mobile/ui-kit` called `icons` which contain all of our custom icons.

To see it in action and hear an explanation of how it all works in more depth, I've put together [this Loom](https://www.loom.com/share/d10afaec2ee041009967f574cd8b2576).

## Tradeoffs or Unknowns
Right now, the script has to be run manually at the root. I would love if there were a way to tie this into the `postinstall` command. It doesn't take long to run, so I think it should be fine to run it after each build.

I'm unfamiliar with our build process for multi-tenant, so I will need a bit of guidance from @redreceipt in order to get this "production ready"

The landing directory for the icons is `templates/mobile` which doesn't scale very well to TV. I'll want to elaborate on some of our "shared code" practices with @jordanwade in order to make sure that this approach scales across all of our platforms. Should they dump the files into 2 different directories? Should we house/store all of our SVG components in a central place to import from?


jordanwade:
@calebpanza after our meeting yesterday. I had another thought, we should still do our best to make our internal developer experience as easy as possible if we are going to create a separate private package to import our icons. 

Mostly, just make adding/update/deploying the icons package quick and easy. I would hate to trade the symlinking problem we currently have with just another symlinking problem.

calebpanza:
So @jordanwade you're in favor of publishing a separate package that gets imported or you think that a code-gen solution within this package is better suited?

jordanwade:
> So @jordanwade you're in favor of publishing a separate package that gets imported or you think that a code-gen solution within this package is better suited?

After giving it a bit more thought, I think I like the separate package better. My point about symlinking/publishing being a pain point won't come up much since it doesn't seem like we will be adding icons often.

calebpanza:
Sounds great!

PR #2721, opened 2022-10-21
feat: Community Tab Type

calebpanza:
## Motivation
Our Community Tab experience is unique enough that we want to be able to solve these product problems separately from our main Feature Feed tabs. Users should be able to view a list of the communities they are a part of and navigate to each one. They should also see their Stream Activity Feed that consolidates all of their Communities directly on the Community Tab

## Solution
The API returns a new field called `type` as a part of the `tabs` query. The `createTabs` method will check each tab `type` and use a switch statement to render the appropriate UI.

I've also taken the liberty of breaking out a bunch of components into reusable files stored in the `components` directory. This will help us move on new Stream Activity Feed enhancements quicker in the future.

## Where to go next
There are some minor refactors to the `tabs` directory that I'll go through at a later time in order to make this whole "tab type" a bit cleaner to look at (there's a lot of prop drilling introduced here)

The Search functionality will be connected at a later point in time.

## Testing
Until it's merged, you will need to run the Cluster locally using [this branch](https://github.com/ApollosProject/apollos-cluster/pull/348#pullrequestreview-1151529541).

You will need to be in at least 1 community. Most of us _should_ be in communities from previous testing, but if you're not you could go into the GraphQL web client and run this mutation.

```
mutation joinCommunity($communityId: ID!) {
    joinCommunity(communityId: $communityId) {
      id
      isAdmin
      profile {
        id
        firstName
        lastName
        photo {
          uri
        }
      }
    }
  }
```

From there, just open up the Community Tab and you're good to go!

## Screenshots
CleanShot 2022-10-21 at 16 52 49

---
I've also put together [this Loom](https://www.loom.com/share/9bcc3e7360994035840d93e66c92cc48) in order to help make testing a bit easier


camrun91:
@calebpanza one thing I am noticing is that the name of the group that the activity is coming from seems to not show on every card. and then it will show the same name again a little later down. It is doing that on the loom on the iPad.

camrun91:
as seen here when I post gratitude in a community and then go back to the communities tab. I get the new item but no Idea where it is coming from. 

Screen Shot 2022-10-24 at 10 57 59 AM

calebpanza:
@camrun91 this is because not all of the Activities are referencing node id's that are in our production dB. So when the query goes to fetch the "label", some of them fail. We shouldn't see this issue on real accounts for churches that are in production

calebpanza:
@camrun91 good catch! It think it's displaying the UTC time instead of local. I have an item marked down in Basecamp for that

templates/mobile/src/components/CurrentUserCommunityList/SearchCommunityListItem.js
```diff
@@ -0,0 +1,47 @@
...
+        <List.Item
+          title="Find Community"
+          subtitle="This feature is not complete, yet, and will be implemented shortly"
+          leadingAccessory={
+            <ImageContainer>
+              <Icon
+                name="magnifying-glass"
+                weight="duotone"
+                color="text.secondary"
+                size={32}
```
camrun91, line 33: I thought at one point we weren't doing hard-coded sizes but referencing sizes from somewhere else. I could be completely off tough.

conrad-vanl:
>  Idk how we're gonna resolve this moving forward cause every time we connect to our local dB we're gonna see this issue...

Should we have a separate Stream Environment for local development? Is the problem because we're mixing production and local Stream environments?

camrun91:
> > Idk how we're gonna resolve this moving forward cause every time we connect to our local dB we're gonna see this issue...
> 
> Should we have a separate Stream Environment for local development? Is the problem because we're mixing production and local Stream environments?

I am actually going to be doing a basecamp post later today on this and the staging server. Me doing the post came out of a conversation @vinnyjth and I had last week.

templates/mobile/src/components/CurrentUserCommunityList/SearchCommunityListItem.js
calebpanza: Not quite. We _use_ to do everything that way, but for this instance it's a really specific size that can just be referenced here.

PR #2724, opened 2022-10-21
Feat: tv integrate login redux

dsilva609:
- replaces current auth screen with auth manager


packages/apollos-tv-shared/components/Auth/AuthConfirm.js
```diff
@@ -60,11 +52,13 @@ const AuthConfirm = () => {
...
       },
     });
-    router.push('/home');
+    router.push('/welcome');
   };
 
   const { values, setFieldValue, handleSubmit } = useForm(async () => {
-    const passcode = values.password;
+    const passcode = values.passcode;
+    console.log(`PASSCODE: ${passcode}`);
```
jordanwade, line 60: Did you want to keep these console.logs here?

jordanwade, line 60: Did you want to keep these console.logs?

jordanwade:
I think we will end up using `requestLogin` for the phone log in flow too. Instead of `requestPin`. Not sure if you want to do that in this PR or not.

PR #2731, opened 2022-10-25
feat: UI kit 2.0 Placeholders

calebpanza:
## Motivation
When we're building data flows, we're ideally not really seeing loading states. In fact, I've recently seen an article with an interesting thought that loading states should be considered more "errors" on the connectivity of our applications than a "feature" or "function". While I would consider that to be a bit extreme, I think it opens up an interesting concept about loading states: our users should see loading states as _little_ as possible.

What I wanted to introduce was a _super_ lean way to approach loading states into the UI Kit 2.0 project that adheres to the principle of "we shouldn't put much effort into building/maintaining cause our users shouldn't hardly ever see them". And this is where we landed!

## Solution
Exported as a part of the `blocks` directory is a new component called `Placeholder` and is exports some simple shapes (Circle, RoundedSquare, Line). These can be used inside of a component in order to build out a rough "fat marker" sketch of the component's loading state.

The main goal of this is to be _suuuuuper_ lean with loading states. Ideally they won't be shown very often, so the DX of building them should be nearly effortless.

## Testing
Read through the code and the storybook. It further explains the intended use cases and general approach.


calebpanza:
@conrad-vanl since it's just the building blocks, there isn't a _ton_ to show, but this is the Storybook which shows a handful of examples

CleanShot 2022-10-26 at 13 09 04

calebpanza:
I used the `text.placeholder`. It's a little light, but I don't mind it

jordanwade:
@calebpanza this is a great start and I personally prefer these types of loading states. Do you know if there are any pros or cons of using CSS vs SVGs, e.g., scaling, resolution, animation, etc.?

templates/mobile/src/ui-kit/blocks/Placeholder/Placeholder.styles.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import styled from 'styled-components/native';
+import { unit } from '../../utils';
+
+export const Circle = styled.View(({ size, theme }) => ({
+  height: size,
+  width: size,
+  borderRadius: 999,
+  backgroundColor: theme.colors.text.placeholder,
```
jordanwade, line 14: Couple questions:
1. In your theme setup is this a new text color item? 
2. Have you tested it on dark and light mode?

jordanwade: Also, Depending on the context of where these are render you might want to consider being able to pass in a different color and have `theme.colors.text.placeholder` be a default if no color/fill is passed in through props.

calebpanza: > In your theme setup is this a new text color item?

This is a new theme color value introduced by @conrad-vanl in the UI Kit 2.0 theme refactor

> Have you tested it on dark and light mode?

Yep! There are light mode/dark mode versions of the theme color

> you might want to consider being able to pass in a different color

If you _really_ needed to, you should be able to just wrap this in the `styled` HOC and override it.

calebpanza:
> Do you know if there are any pros or cons of using CSS vs SVGs

@jordanwade the main benefit I've been seeing inside of Figma is the ability to stack shapes on top of each other to automatically get visual hierarchy. In the screenshot below, I build out this loading screening using Figma to demonstrate what stacking these loading shapes on top of each other can create

image

calebpanza:
Still 100% open to suggestions on this! Just wanted to provide more insight as to the "why" behind the current implementation.

jordanwade (approved):

PR #2729, opened 2022-10-25
fix: account for translucent Android status bar

redreceipt:
Group 1

PR #2732, opened 2022-10-25
fix: removes extra space from various screens

redreceipt:
the `avoidHeader` prop solution was misguided. the root cause is fixed here: #2729

Group 1


calebpanza:
This is awesome. Thanks for catching this!

calebpanza (approved):

conrad-vanl:
This was re-tested on iOS right?

I feel like that prop was added for different reasons but I really don't recall...

redreceipt:
@conrad-vanl the prop was for android only

Screen Shot 2022-10-25 at 9 19 38 PM

PR #2734, opened 2022-10-26
feat: Communities Screen

calebpanza:
## Motivation
Users should be able to see a list of Communities so that they can join them

## Solution
A new screen called `Communities` that houses a single paginated list of Communities. As this screen requires some cards, I took this as an opportunity to implement the UI 2.0 Content Card in a really basic form.

## Screenshots
| Full List | Error with Data | Error with No Data |
| --- | --- | --- |
| CleanShot 2022-10-25 at 21 10 54 | CleanShot 2022-10-25 at 20 56 30 | CleanShot 2022-10-25 at 20 53 46 |


conrad-vanl:
Both of those error states are a new convention then what we've done elsewhere. I think these are "good enough" to launch with, but @calebpanza could you try and get a version of these in Figma at some point (I know you talked about that) so that we can iterate on them and figure out how we want to handle error states app-wide in a uikit-2.0 world?

PR #2740, opened 2022-10-27
chore: various updates and clean ups

redreceipt:
this also makes the `yarn app` command as described in the README from working without an `.env` file for a smooth experience


templates/mobile/Gemfile
```diff
@@ -3,5 +3,5 @@ source 'https://rubygems.org'
 # You may use http://rbenv.org/ or https://rvm.io/ to install and use this version
 ruby '~> 2.7'
 
-gem 'cocoapods', '1.11.2'
+gem 'cocoapods', '1.11.3'
```
calebpanza, line 6: Was this an intentional change? (just checking cause mine will auto-update sometimes and it's broken things in the pas when it does)

templates/mobile/Gemfile.lock
```diff
@@ -3,7 +3,7 @@ GEM
   specs:
     CFPropertyList (3.0.4)
       rexml
-    activesupport (6.1.4.1)
+    activesupport (6.1.7)
```
calebpanza, line 6: Are these intentional changes? (just checking cause mine will auto-update sometimes and it's broken things in the pas when it does)

templates/mobile/src/index.js
```diff
@@ -25,8 +25,6 @@ import {
   PersonFollowingConnected,
   SearchScreenConnected,
   ScriptureScreenConnected,
-  PrayerRequestInputConnected,
```
calebpanza, line 28: 🙇🏻‍♂️🙇🏻‍♂️🙇🏻‍♂️

templates/mobile/Gemfile
redreceipt: yea it was intentional, I'm just upgrading to the latest version

templates/mobile/Gemfile.lock
redreceipt: probably due to the updated version of Cocoapods

calebpanza:
So far it looks good to me! I can do a more explicit test with the sim first thing in the morning! It'll be the first thing on my list.

calebpanza (approved):
Everything's running well for me! iPhone and iPad load up and storybook loads up just fine.

vinnyjth (approved):

PR #2748, opened 2022-10-31
chore: adds tvOS deployments

redreceipt:
Screenshot 2022-10-31 at 9 07 48 PM


redreceipt:
this is getting hung up with fastlane trying to do release builds. it uses this command which we can use to shortcut debug: `xcodebuild -workspace ios/ApollosChurch.xcworkspace -scheme ApollosChurch-tvOS -destination 'generic/platform=tvOS' archive`

calebpanza (commented):
@redreceipt I was able to successfully install the tv app on my Apple TV, although no navigation works. Is that expected at this point?

jordanwade:
@calebpanza do you mean the remote isn't working? That would be strange regression.

calebpanza:
@jordanwade the remote technically works, although I am only able to interact with the Search button. Navigating any direction on the remote does nothing.

It _could_ be that this is just expected atm cause we don't have final content and such. but I just wanted to call it out in case this was working on the simulator in which it's not working on a physical device.

redreceipt:
@calebpanza if that is the case, it's beyond the scope of this PR and should be fixed elsewhere. All this does is get automated deploys working.

jordanwade:
@calebpanza thanks I'll capture that in an F&I TODO

calebpanza (approved):

PR #2749, opened 2022-10-31
chore: Cancel outdated runs

vinnyjth:
Uses [this action](https://github.com/marketplace/actions/skip-duplicate-actions#cancel-outdated-workflow-runs) to skip outdated runs.


calebpanza (approved):
🥳🥳🥳

PR #2753, opened 2022-11-02
fix(tv): color and padding on android inputs

jordanwade:
## ✏️ Description

Fixes background-color and padding issue on Android TV inputs.

## 🔬 To Test

1. Fire up TV app in Android sim, inputs should have a background that is different than the container and the input text should have padding on the left.

## 📸 Screenshots

image

## 📝 Checklist


dsilva609 (commented):
@jordanwade I did notice this on TVOS,  the spacing is pushed a little more than usual.

image

calebpanza (commented):
@jordanwade why limit this change to just Android? Why not keep the experience consistent across both platforms and set it across both?

jordanwade:
> @jordanwade why limit this change to just Android? Why not keep the experience consistent across both platforms and set it across both?

@calebpanza The goal is to keep the UI on all platforms as consistent as possible. But, both tvOS and Android TV have slight differences on how they handle sizing and default colors. I don't think `react-native-tvos` is as refined as `react-native`. But, we can continue to sharpen things in the UI as needed.

jordanwade:
@dsilva609 I pushed up another tweak, things should look a little better with the input spacing. Still sizing is aways a challenge between the platforms, but again we can continue to work on that as needed.

image

calebpanza (approved):

dsilva609 (approved):
LGTM 👍

PR #2758, opened 2022-11-03
migrate content node

redreceipt:


calebpanza:
Trying to track down why, but the StretchyView seems to not be working anymore

templates/mobile/src/content/ContentNodeConnected/index.js
```diff
@@ -114,17 +102,6 @@ const ContentNodeConnected = ({
 
 ContentNodeConnected.propTypes = {
```
calebpanza, line 103: Can you elaborate on the reason for changing/updating the Api structure on this? I'm noticing that this change is causing the stretchy view to not work properly and I'm concerned there might be other side effects.

Should we retain the current API as is and refactor another time?

redreceipt: Oh thanks for catching. Yea I removed them because I assumed they were just there for legacy prop overrides from theme file. I'll fix stretchy view.

redreceipt:
Ok I see you put it back, thanks

templates/mobile/src/content/ContentNodeConnected/index.js
calebpanza: What do you think about restoring all the original props for the time being? I'm familiar enough with the Stretchy View, but I'm not so familiar with the others which makes me a bit nervous. Idk the current uses and the implications of removing those.

redreceipt: The only component in the whole project that uses `NodeSingleConnected` and it only uses the `ImageWrapperComponent` prop so we're safe to remove. The majority of component props in the project were there for custom prop overrides but we don't do that anymore.

Screenshot 2022-11-04 at 10 12 53 PM
Screenshot 2022-11-04 at 10 13 12 PM

calebpanza (approved):

PR #2759, opened 2022-11-03
deploys android and androidTV together

redreceipt:
Screenshot 2022-11-23 at 11 58 01 PM
Screenshot 2022-11-23 at 11 58 05 PM


redreceipt:
still need to test this locally

vinnyjth:
This is awesome 🙌 

Will this result in a TV app being installable for all Android apps by default?

redreceipt:
that's the goal

vinnyjth:
> that's the goal

Gotcha. We might want to reconsider that (we should be able to avoid shipping a TV app). 

The majority of churches won't have the correct channels setup in order to provide a TV experience, and we don't prohibit shipping updates to those churches until we set them up with TV channels.

redreceipt:
yea I should be able to set an on/off switch

redreceipt:
ok this is ready to test locally but due to the high stakes situation of Android version codes, I wanna make sure I get the algorithm right before trying 😂 @vinnyjth @calebpanza double check my code and algorithm here. And if you have a better suggestion, I'm open to it. The version code is hours since Unix epoch with a 1/2 in front depending on mobile/tv. And then all of that added to an optional offset. So for Newspring it would be:

```
  2022000000 (offset)
+    1000000 (1 for mobile)
+     463200 (hours since unix epoch)
------------
  2023463200
```

TV would be `2024463200`

I also made the iOS build number the hours since the epoch to match. But it's almost irrelevant.

calebpanza:
>   2022000000 (offset)
+    1000000 (1 for mobile)
+     463200 (hours since unix epoch)
------------
  2023463200

So when the year turns, we'll update this to 2023?

redreceipt:
> First off, this looks great! I feel like this is making the most out of an android situation. Lol.
> 
> I know there's only so much that we can do, but is there any cap to the build number that we should be aware of? The offsets are huge which is great, but are we at all concerned that we could possibly hit a "cap" that is hidden in the Android docs somewhere?

`210000000` is the cap 😑

redreceipt:
Not too concerned about hitting it because, if my math is right, the hours won't dip into the 7th digit for another 50 years

redreceipt:
The draw back is we have a one hour release timetable. Even moving to minutes put us too close to the cap.

templates/tv/ios/Podfile.lock
```diff
@@ -457,4 +457,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: b8198b053d33b10a2bf51f86e92e550936c6a83a
 
-COCOAPODS: 1.11.3
```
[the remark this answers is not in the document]
redreceipt: that's the version of cocoa pods listed the `Gemfile` not sure why it was wrong to begin with.

calebpanza:
> The draw back is we have a one hour release timetable. Even moving to minutes put us too close to the cap.

I don't _love_ that, but as long as it's well documented and known that it's a limitation, then that's fine. It's takes a good 30 minutes to push a release anyway.

Maybe this is something that @nlewis84 could add as a consideration into the Deployment Dashboard?

vinnyjth:
Ah! so the goal here is that churches that have both an Android and TV app would run a different script to deploy both apps?  Churches that lack TV would continue to run the same script?

redreceipt:
Exactly

vinnyjth:
This looks great to me - 

My expectation is that when deploying to TV and Mobile, we would end up with two bundles in Google Play Console for the same bundle ID. One would be valid for a certain set of devices (mobile) and one would be valid for a different set of devices (TV). We would promote both to production. Does that sound right? 

Want me to run a test on Bitrise before deploying this or should we yolo and go straight to `master` and test from there?

vinnyjth:
Also happy to have you on board with the `timestamps as build numbers` bandwagon ;)

redreceipt:
likely blocked until we can get this patch into tv somehow

redreceipt:
@dsilva609 may have a `build.gradle` patch to unblock this?

dsilva609:
> @dsilva609 may have a `build.gradle` patch to unblock this?

Opened https://github.com/ApollosProject/apollos-apps/pull/2764 to fix the build

jordanwade:
@redreceipt do you need any help getting this one ready to review again and merged?

redreceipt:
I was sort of waiting on icons to be right to merge this but I suppose they are unrelated and can be fixed separately. I'll open this one up this afternoon and let you know when it's ready to review. sorry everything is held up. the path for all this feels very linear.

redreceipt:
it worked!

templates/mobile/android/app/src/main/AndroidManifest.xml
```diff
@@ -6,6 +6,10 @@
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+    <supports-screens android:smallScreens="true"
```
vinnyjth, line 9: These might not be the right values :) 

Here's what Crossroads discovered: 

https://github.com/Differential/crossroads-anywhere/pull/6808

https://github.com/Differential/crossroads-tv/pull/44

vinnyjth (changes requested):

templates/mobile/android/app/src/main/AndroidManifest.xml
redreceipt: good find. the tv size thing always seemed a little thin to me. fixed.

calebpanza: @redreceipt idk if I'm missing it or not, but it looks like the TV Manifest files were updated but not the Mobile ones per the [PR reference to the Crossroads Mobile App](https://github.com/Differential/crossroads-anywhere/pull/6808/files)

redreceipt: I'm not sure what's going on with the crossroads repo but the mobile _does_ require a touchscreen, no?

redreceipt: it's TV where, touchscreen should be false. that's straight from the docs https://developer.android.com/training/tv/start/start#no-touchscreen

calebpanza: Yeah I'm not too sure about the touch screen bit. I had the same thought you mentioned. Maybe their app use to be allowed on the TV store?

As for the device sizes, I don't see any specified in our application? I'm looking at `templates/mobile/android/app/src/main/AndroidManifest.xml` both on `master` and this branch and I don't see any specification

redreceipt: I'm assuming specifying sizes aren't necessary with using the touchscreen and TV booleans. It's not mentioned in the developer docs and crossroads just sets all sizes to true which seems like an unnecessary smell.

calebpanza (approved):

vinnyjth (approved):

PR #2760, opened 2022-11-04
feat: DailyHabitsExperienceCard Component 

calebpanza:
## ✏️ Description
We are introducing a new card type that is **specific** to Daily Habits. This card will be rendered exclusively on the Daily Habits tab for the time being, so we're introducing this component with the name of: DailyHabitsExperienceCard.

For reference, you can see the [Figma Component](https://www.figma.com/file/YHJLj8pdFxWG9npF2YmB3r/UI-Kit-2.0?node-id=4401%3A68140) as well as the [Daily Habit Screen](https://www.figma.com/file/YHJLj8pdFxWG9npF2YmB3r/UI-Kit-2.0?node-id=4296%3A61742) for different use cases.

## 🔬 To Test
1. Set your `PROJECT_VARIANT` to `storybook`
2. `yarn && yarn start` then `yarn ios`
3. Check out the story

## 📸 Screenshots
| Light | Dark |
| --- | --- |
| CleanShot 2022-11-04 at 14 05 25 | CleanShot 2022-11-04 at 14 05 35 |

## 📝 Checklist


vinnyjth (approved):
these new cards compose so nicely 💯

conrad-vanl:
Screenshot 2022-11-07 at 8 59 23 AM

Calling out differences between this and what's in Figma:

- Title ordering (big title on top vs bottom)
- Text next to avatars vs not
- Arrow on the right side of the card

conrad-vanl:
I want to make sure that we're documenting/calling attention to these differences (and explaining why) as these PRs come across, otherwise we're going to miss out on potentially intentional decisions, either way (Things that @calebpanza intentionally changed when he implemented it for one reason or another, or things thta I intentionally had designed a certain way that didn't get implemneted that way)

conrad-vanl:
Personally, I like the information hierarchy on what's in Figma better (I've noticed other community cards that have the information hierarchy reversed as well, and I haven't brought that up yet) and i also feel like these cards are text-heavier feeling

calebpanza:
Oh I see what you're saying. That's my bad. I jammed on this with @jordanwade before implementing it and I think I missed an area in Figma to update. You can see our work that we did [here](https://www.figma.com/file/YHJLj8pdFxWG9npF2YmB3r/UI-Kit-2.0?node-id=4296%3A61742)

Here were the justifications for each of the differences listed:

### Title ordering (big title on top vs bottom)
The primary CTA is the summary, while the title is more to associate a "branded name" with the feature so that it's familiar when referenced throughout the app. Given that the CTA gives you the "goal", it felt better suited to have more attention drawn to it

### Text next to avatars vs not
As we are introducing the Social Engagement and Avatar Row components to users, we wanted to make it very clear why you were seeing these and added the text to it. This is something that we felt could be refactored out of the design in the future, but for launch felt it might help provide more context to users.

### Arrow on the right side of the card
The cards should still be interactive for users even after they've completed an experience (so they could come back and do it again). In order to provide a clear indication that the cards are still interactive, we added the arrow which will stay visible even after the "check mark" has been filled.

calebpanza:
cc @conrad-vanl

PR #2761, opened 2022-11-04
fix: export ApollosPlayerConnected; replace GET_MEDIA import with query prop

tristanguckenberger:
## ✏️ Description
While setting up Community Christian I noticed that NodeSingleConnectedWithMedia would break each time I navigated to a specific item with media.

With the move of some of the ui-connected pkg contents to templates it looks like we now need to:
- export ApollosPlayerConnected from apollos-ui-connected/src/index
- pass GET_MEDIA as a prop now that it resides in templates and is required in pkgs -> ui-connected -> ApollosPlayerConnected

## 🔬 To Test
1. Spin up a one of the multitenant apps and view a content item that has media
- you should see an error immediately (see screenshots)
2. Checkout branch fix-NodeSingleConnectedWithMedia and test out the same app again
- the item should load and video playback should work

## 📸 Screenshots
**Before**
Screen Shot 2022-11-04 at 15 44 26

**After**
Screen Shot 2022-11-04 at 15 47 01

## 📝 Checklist

- [] New and existing tests and linting pass locally with my changes


redreceipt (dismissed):
Good catch. Move GET MEDIA to ui connected and import it in both. We don't really pass around queries anywhere.

tristanguckenberger:
> Good catch. Move GET MEDIA to ui connected and import it in both. We don't really pass around queries anywhere.

@redreceipt If it's more suitable for our needs I can move GET MEDIA into its own separate folder. Otherwise, I moved GET MEDIA into ui-connected -> ApollosPlayerConnected

vinnyjth (approved):

PR #2763, opened 2022-11-07
patches react native android build failure

redreceipt:
https://github.com/facebook/react-native/issues/35210


redreceipt:
slight hiccup...

Screenshot 2022-11-07 at 11 23 46 AM

dsilva609 (changes requested):
this patch will probably need to be done for the tv template as well for it's android build

vinnyjth (approved):

PR #2774, opened 2022-11-10
upgrades TV to RN68

redreceipt:
This is necessary so we can build without running in Rosetta and do special things for Fastlane. The notable change is you'll notice in the `package.json` we no longer need to run with `arch x86_64` (Rosetta)


templates/tv/package.json
```diff
@@ -3,11 +3,10 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "android": "react-native run-android",
-    "tvos": "arch -x86_64 react-native run-ios  --simulator \"Apple TV 4K (2nd generation)\" --scheme \"tvappnative-tvOS\"",
+    "androidtv": "emulator -avd \"$(emulator -list-avds | grep TV)\" & react-native run-android",
```
jordanwade, line 6: "androidtv" breaks yarn tv:android script in the root package.json.

```suggestion
    "android": "emulator -avd \"$(emulator -list-avds | grep TV)\" & react-native run-android",
```

jordanwade (changes requested):

jordanwade:
I can't get Android to work. I'm see this error.
image

dsilva609:
> I can't get Android to work. I'm see this error. image

Yeah I'm seeing the same on my end

redreceipt:
Me too. Trying to figure what I changed between testing and committing 🤦‍♂️ what’s weird is `react-native-video` gives instructions for installation on their docs that we aren’t following in master.

redreceipt:
@dsilva609 @jordanwade got it

package.json
```diff
@@ -19,10 +19,10 @@
     "nuke:cache": "watchman watch-del-all",
     "postinstall": "lerna run build --parallel",
     "generate-stories": "lerna run generate-stories",
-    "app": "yarn && (cd templates/mobile && yarn && yarn ios)",
-    "start:tv": "cd templates/tv && yarn && yarn start",
```
dsilva609, line 23: why was the tv start script removed?

redreceipt: Don't need it. Metro should launch automatically.

redreceipt: https://user-images.githubusercontent.com/2659478/201186648-30e981a7-5ea6-4349-a2ee-78c6cac2b662.mov

dsilva609: Probably a VS Code quirk but for me, it opens up a new terminal window which isn't preferable since there's already a built in terminal in the app so I like to separate the commands to keep everything organized.

image

redreceipt: I put it back for you

dsilva609: Awesome. I'm not sure who else uses it that way but, it's been really helpful.

dsilva609 (approved):

PR #2776, opened 2022-11-10
use church slug as deep link host

redreceipt:
removes the need for an extra variable in the DB


vinnyjth (commented):
Are there currently any live apps that have a deep link host that is different from the church slug?

redreceipt:
Probably. The consequences would just be old notification links wouldn't work, right?

vinnyjth:
Correct - and OpenID login wouldn't work for their users (might not be any overlap here though)

Also one or two apps hardcode deep links. Also can't confirm any overlap with digging in.

I think we backed off on this consolidation earlier this year. Might be worth going back to that PR and peeking at our specific reasoning back then.

redreceipt:
> Correct - and OpenID login wouldn't work for their users (might not be any overlap here though)

I didn't think this was a problem since it's the app itself generating the callback

> Also one or two apps hardcode deep links. Also can't confirm any overlap with digging in.

Which ones? I didn't think it could be custom at this point.

> I think we backed off on this consolidation earlier this year. Might be worth going back to that PR and peeking at our specific reasoning back then.

I'll see if I can track it down.

redreceipt:
@vinnyjth I'm the one who introduced it on 8/18 but I don't know why :( https://github.com/ApollosProject/apollos-apps/pull/2642

camrun91:
@vinnyjth and @redreceipt aren't all the links being created server side now? for some reason I thought they were or some of them were.

vinnyjth:
Correct!So old apps receiving newly generated links won’t be able to handle the new linksOn Nov 11, 2022, at 11:08 AM, Cameron Llewellyn ***@***.***> wrote:﻿
@vinnyjth and @redreceipt aren't all the links being created server side now? for some reason I thought they were or some of them were.

redreceipt:
It could use the config variable if present to support apps still using it.

redreceipt:
A better way to do it might be to embed the callback URL in the mutation so there aren't two sources of truth.

vinnyjth:
I really like the idea of using the config variable optionally. It doesn’t add much bloat and we don’t have to expose UI for setting it moving forwardSent from my iPhoneOn Nov 11, 2022, at 11:46 AM, Michael Neeley ***@***.***> wrote:﻿
A better way to do it might be to embed the callback URL in the mutation so there aren't two sources of truth.

redreceipt:
I can do that

PR #2778, opened 2022-11-11
TYPESCRIPT!

redreceipt:
This allows us to write typescript components if we choose that can co-exist alongside JSX components making for a smooth adoption. As the team gets larger, typescript will help us refactor code without fear of new bugs. this also prepares the way for migrating the media player pkg


conrad-vanl:
Oh man, this PR review is going to be fun to watch

redreceipt:
A big motivation for this is all the bugs I've caused by migrating old pkgs into the template 😅 when imports and props have a static shape, bugs magically disappear.

packages/apollos-ui-kit/src/index.js
```diff
@@ -1,4 +1,4 @@
-export ActionCard from './ActionCard';
+export { deafult as ActionCard } from './ActionCard';
```
redreceipt, line 1:
```suggestion
export { default as ActionCard } from './ActionCard';
```

camrun91:
tenor-163766927

redreceipt:
> just kidding I guess this is a good thing lol

Haha yea I was of the same opinion until recently. The fully opt-in nature of it makes it less intimidating. And it really does catch things that are easy to miss.

calebpanza:
VERY excited for this! I'm gonna hold my review until this afternoon cause I wanna take some time to fully understand everything. Thanks for looking into this!!

calebpanza (commented):
Looks good to me and the app boots! I'm not seeing any typescript examples in this PR, so I'm guessing the introduction of actual typescript files will be in future PR's?

redreceipt:
ExportFeature is typescript

calebpanza:
Yeah that looks great to me! With the introduction of this, did you have any thoughts/ideas on how to best organize types throughout the project?

calebpanza (approved):

conrad-vanl:
Just to be clear, this can't be merged without approving review from @vinnyjth, this is a larger discussion that needs to be had about coding standards

conrad-vanl:
To be clear, the merging of this PR is a signal that we are _deciding_ to shift to Typescript, having both TS and JS in our project long-term would be an additional code debt that to me outweighs the benefits of having TS at all. So I would want to ensure that we have a migration plan to progressively migrate our codebase fully to typescript. Agree/disagree?

I would want to be critical on answering:

1. Why now? Is now the _right_ time for a shift like this? How does it help us achieve our goals faster, better, or...?
2. How does this effect our execution speed in the _short_ term? There's arguments for the long-term benefit, but it we're focused on a velocity ramp over the next 6-12 months, does Typescript support that or hinder that?
3. What are the biggest risks of this decision?
4. Have we talked with other people around Differential about this? Our larger company benefits when we have some alignment of coding standards, practices, etc. While our teams generally have autonomy to push forward on decisions like this, it is our responsibility to consider the cross-team impacts

conrad-vanl:
> typescript will help us refactor code without fear of new bugs. 

This is the only benefit I see listed in the PR so far. Type checking can prevent a lot of bugs of a certain type....do we even have quantifiable data on how much type checking would avoid the bugs that we've ran into?

conrad-vanl:
I guess the other question(s) I would want to answer:

- This PR pushes for a slow/progressive adoption for Typescript. If the team wants to move to Typescript, would we be better served by taking a bigger bet?
- What other team considerations, collaboration and conversations have been had about this decision? Or is this a PR out of left-field that no one was expecting? Are these types of PRs serving our team the best, or would we be better served by taking a more collaborative, shape-up (or RFC, etc) approach to introducing engineering practices like this?

conrad-vanl:
Screenshot 2022-11-15 at 9 05 23 AM

😬

nlewis84:
My perspective from working on a large scale customer-facing TS project is that it allowed us to be extremely safe in our work, at the cost of moving quickly. Everything about that experience was slower though, from the actual coding to pushing up new features. 

I'm up for anything, but I don't know if moving to TS necessarily aligns with "Done well today is better than perfect tomorrow."

calebpanza:
@conrad-vanl really great questions and I will just answer them from my personal view:

> Why now? Is now the right time for a shift like this? How does it help us achieve our goals faster, better, or...?

We have to start somewhere, right? And as we're refactoring and adding so many new things, postponing the introduction will only intensify the introduction. There's also been a lot of discussion around testing best practices and PR reviews. One of the goals with the TS integration is that some of these problems are solved with the the type safety.

> How does this effect our execution speed in the short term? There's arguments for the long-term benefit, but it we're focused on a velocity ramp over the next 6-12 months, does Typescript support that or hinder that?

With a drip-feed approach, this shouldn't be the case. It's an extra step of decision making when writing new code (TS or JS) but that should be a matter of addressing in each PR. For wildly urgent issues, continue on with JS, but for everything else, migrate as it's touched.

> What are the biggest risks of this decision?

Lack of standards and the team not all being on board with the approach. Type-safety should encourage more strict coding guidelines/flavors, but if they're not followed now, then this is all for naught and would end up just being code smell.

> Have we talked with other people around Differential about this? Our larger company benefits when we have some alignment of coding standards, practices, etc. While our teams generally have autonomy to push forward on decisions like this, it is our responsibility to consider the cross-team impacts

Large sentiment from those I've talked to is that the continued use of vanilla JS is a desire for convenience. Nobody argues the safety-net and logic around TS, but rather groans about the learning curve of new coding standards. As they are working on larger and larger projects, though, the conveniences are outweighed by the safety net provided by TS. The conversations I've had have been a large motivating factor for my personal curiosity in the matter

I could see the entire UI Kit 2.0 being written in Typescript _very_ easily. As we have Tim bin cooldown, I would touch up each of the existing components and commit to all future work on that being fully TS enabled.

redreceipt:
sorry this was set to automerge, @conrad-vanl @vinnyjth I can revert if we want to wait for a final decision?

conrad-vanl:
My vote is to revert and reconvene. This PR has big side affects on how we work as a team and I don't think we've seriously talked about it outside of this PR, which isn't enough.

vinnyjth:
Yes. Let's revert.

redreceipt:
reverted

PR #2779, opened 2022-11-11
removes apollos config pkg

redreceipt:
don't need it anymore


redreceipt:
It gave us the ability to dynamically load global JS overrides. But we don't do that anymore.

vinnyjth (approved):

redreceipt:
❤️

PR #2780, opened 2022-11-11
makes deep and universal link config variables optional

redreceipt:


camrun91:
@redreceipt what is the best way for me to test this?

camrun91:
The code looks good I would love to test first just to make sure.

redreceipt:
thanks for taking a look @camrun91 . this one is actually dependent on https://github.com/ApollosProject/apollos-cluster/pull/368 which @vinnyjth found problems I need to fix first. I'll mark this as a draft until that one is ready.

redreceipt:
timing isn't right here

PR #2782, opened 2022-11-14
fix apple deploys on CI

redreceipt:


scripts/deploy-ios.sh
```diff
@@ -1,16 +1,10 @@
-TRACK=$2
+TRACK=$1
 
 if [ -z "$CHURCH" ] || [ -z "$APOLLOS_API_KEY" ]; then
 	echo "CHURCH and APOLLOS_API_KEY required!"
 	exit 1
 fi
 
-if [ -n "$(git status --porcelain)" ]; then
```
vinnyjth, line 8: did you want to rip this out?

redreceipt: yea, git status doesn't matter anymore since deploys are time based

vinnyjth (approved):

PR #2783, opened 2022-11-15
TYPESCRIPT (again)!

redreceipt:
for context: https://github.com/ApollosProject/apollos-apps/pull/2778


vinnyjth (approved):

redreceipt:
@vinnyjth this puts typescript back :) I've already taken typescript away in a previous commit.

vinnyjth:
@redreceipt woops. thanks for noting that for me

redreceipt:
> Am I going to be able to start the app and storybook, and contribute a bug fix without running into any difficulties?

assuming this was to me, yes.

vinnyjth:
> > Am I going to be able to start the app and storybook, and contribute a bug fix without running into any difficulties?
> 
> assuming this was to me, yes.

Was thinking a bit more rhetorically 😂  That's what I'm looking for when I review this PR

camrun91:
Just putting my two cents here. Looking at it from a new dev starting point I would much rather come on and know ok this project is Typescript rather than having to deal with the project is type script that may bleed over into javascript. 

My other concern would be, does this cause a potential roadblock to how quickly we are currently able to get things done? If we are wanting to use TS to enforce code standards and stuff that is great, however, have we defined those standards? and do those standards need to be true for all Apollos Repos?

conrad-vanl:
> possible to gradually adopt Typescript

@vinnyjth did you consider pros/cons on gradually adopting vs making a higher-appetite bet on it?

conrad-vanl:
I'm afraid that if we make the decision to adopt typescript with no real goal in mind ("gradual") that it'll never be migrated/done, and we'll be in a worse state along the way?

conrad-vanl:
As opposed to our team putting effort against this while we're still trying to close out cycle goals, WDYT about visiting this discussion during cool-down and december cycle?

calebpanza:
@conrad-vanl the pitch that I was going to make to @vinnyjth in the gradual integration of Typescript was to take our existing UI Kit 2.0 components and migrate them immediately to TS. From there, it would be a matter of 1. all new work done should be done in TS and 2. determine legacy screens that are relevant to cycle work and migrate them to TS/UI Kit 2.0 as a part of cycle goals

My pitch for TS in the app has been a followup to the work done in migrating our UI Kit and in my mind it was always tied to that effort. Since we're already implementing a "gradual" approach to the new UI Kit effort, it felt like a great time to introduce TS (and the coding standards/best practices that are established in it) as a piggyback off that effort without risking new features each cycle.

vinnyjth:
Let's talk about this during cooldown. I don't have the capacity to review this properly before end of cycle.

calebpanza:
Are we planning to come back to this anytime soon??

redreceipt:
Hope so

templates/mobile/.eslintrc.js
```diff
@@ -1,9 +1,13 @@
 module.exports = {
   root: true,
   extends: ['@react-native-community', 'plugin:react/jsx-runtime'],
+  parser: '@typescript-eslint/parser',
   env: {
     'jest/globals': true,
     'detox/detox': true,
   },
-  plugins: ['detox'],
+  plugins: ['detox', '@typescript-eslint'],
```
redreceipt, line 9:
```suggestion
  plugins: ['@typescript-eslint'],
```

templates/mobile/.eslintrc.js
```diff
@@ -1,9 +1,13 @@
 module.exports = {
   root: true,
   extends: ['@react-native-community', 'plugin:react/jsx-runtime'],
+  parser: '@typescript-eslint/parser',
   env: {
     'jest/globals': true,
     'detox/detox': true,
```
redreceipt, line 7:
```suggestion
```

vinnyjth (approved):
Couldn't spot any issues building or editing select files in the codebase 👍

PR #2786, opened 2022-11-16
fix: Android Build Issues

calebpanza:
## ✏️ Description
Android was failing to build due to a broken import within the Android platform code for the `react-native-passkit-wallet` package. Upon further review of issues in that repo, it seems like this was a known issue that was mitigated by including the `jetify` command as a part of the `postinstall` script.

We already had the `jetifier` package installed in our dev dependencies, so this PR simply adds the `jetify` command to the end of the `postinstall` script.

## 🔬 To Test
1. When you open up Android Studio, you can see any broken imports and you should see one in the `RNPassKitModule` file
2. After running `yarn` with the update inside of this PR, that error should go away
3. You might need to Clean the project and/or re-open Android Studio
4. You know it's working when the reference `import androids.core.content.FileProvider;` is in that file

## 📸 Screenshots
| Pre-Patch | Post-Patch |
| --- | --- |
| CleanShot 2022-11-15 at 21 48 21 | CleanShot 2022-11-15 at 21 49 11 |

## 📝 Checklist


redreceipt:
jetifier repo says it's automatic???

https://github.com/mikehardy/jetifier#do-you-need-this

Screenshot 2022-11-15 at 11 20 01 PM

redreceipt:
OOOO I get it...`react-native-passkit-wallet` is only on RN 58 😑

redreceipt (approved):

PR #2788, opened 2022-11-16
automatic match certs regeneration

redreceipt:
this means we will never have expired certs again and we won't have to generate the first time manually. It's been tested locally.


vinnyjth:
Running a test build here: https://app.bitrise.io/build/9006258e-dfb5-4450-b74c-b36bdddd7544

vinnyjth:
@redreceipt build succeeded (awesome!) but noticed these issues. 

Bundle wasn't valid
CleanShot 2022-11-16 at 15 03 05@2x

Cake and Mortor provisioning profiles were deleted from App Store Connect. Unfortunately that's not neighborly behavior and we shouldn't do that. 
CleanShot 2022-11-16 at 15 04 50@2x

redreceipt:
shoot. I tested on LCBC where they only had our bundle IDs so I assumed it was working correctly. The first error is surprising to me and I don't really understand it because I deployed LCBC to TestFlight successfully.

vinnyjth:
> The first error is surprising to me and I don't really understand it because I deployed LCBC to TestFlight successfully.

yes - it was surprising to me as well. it looked like the deploy was fine and matched my expectations from what I saw in Bitrise

redreceipt:
timing isn't right for this one

PR #2796, opened 2022-11-17
fixes tvos version number getter

redreceipt:
The current lookup tool we're using doesn't have tvOS information. This is also slightly more maintainable.

## iOS

Screenshot 2022-11-17 at 4 21 44 PM

## tvOS

Screenshot 2022-11-17 at 4 29 41 PM


redreceipt:
sorry not quite ready yet...

redreceipt:
ok @vinnyjth now it's ready

vinnyjth (approved):
Neat that we can get the version from `app_store_build_number`!

PR #2798, opened 2022-11-20
feat: add search screen feed connected

tristanguckenberger:
## ✏️ Description
WIP for Component that combines search, tags, and feature feeds.
Screen Shot 2022-11-19 at 21 06 19

For this to work the tab label must be set to "Search-Explore":
Screen Shot 2022-11-19 at 21 03 52

## 🔬 To Test
1.
2.

## 📝 Checklist


conrad-vanl:
I know that @calebpanza was working on some sort of "Tab Type" logic, maybe we could apply this here so that we don't have to hard-code the label.

I think progamming what chips/categories show up in the list might be an interesting challenge. Typically we could do the tags sorted by most number of items or most recently published in....however I'm sure churches will want a way to allow/denylist the list of tags we use.

@nlewis84 created a Chip List feature (not sure I remember what it's called for sure) a while back that I think we could use. My expectation for these chip lists is to be rather rudimentary - they don't actually filter or search the page in place, instead tapping on any of the tags opens up a new screen with just a feed of that tag's content. Similar to how the crossroads anywhere app's explore tag works.

vinnyjth:
I'll need a bit of help understanding the purpose of this page :) What's the advantage of this custom component vs using the Chip List and setting the search config on the tab to true? @conrad-vanl do you have the extra context here?

templates/mobile/src/tabs/Tabs.js
```diff
@@ -99,6 +99,7 @@ const Tabs = ({ tabs }) => {
                 ...(config.profile ? { headerLeft: ProfileButton } : {}),
               },
               title,
+              showTags: config.tagList,
```
calebpanza, line 102: Is this guaranteed to be a boolean or would it be safer to force cast this like `Boolean(config.tagList)` just in case?

vinnyjth:
Talked with @conrad-vanl and @tristanguckenberger - We'll take another look at this during cooldown, for now Community Christian is going to use the default search bar.

PR #2802, opened 2022-11-21
makes activity feed optional

redreceipt:
requiring these variables for development work raises the barrier of entry significantly for developers as we now have to know how to pull encrypted config variables from the DB. Without these variables, there are no required variables and the development experience is smooth and simple.


conrad-vanl:
Is there a way these variables could be pulled in an easier way? These variables are per-tenant right?

It seems like the Apollos Demo app, with the ability to switch tenants dynamically, will have to solve this?

I think this could be a reasonable short-term solve, but perhaps there's a bigger issue here we need to consider?

redreceipt:
Yes that's a good point. We really can't have any per-tenant variables as environment variables. I think the bigger solve is pulling out `react-native-activity-feed` on the front end. It's deprecated and they recommend only using server side SDKs. Only reason I'm doing this is because I've noticed in certain situations, if you don't define these variables, the app locks up and becomes unresponsive and now there's no hope of that being improved. 

Screenshot 2022-11-21 at 5 25 59 PM

redreceipt:
not necessary since the client side stream stuff is deprecated and will need to be pulled out completely

PR #2804, opened 2022-11-22
adds git hash as app version

redreceipt:
showing build number is no longer helpful since it's time based. this will tell support and developers what code exists in the app


templates/mobile/src/settings/__snapshots__/UserSettings.test.js.snap
```diff
@@ -119,7 +119,7 @@ exports[`UserSettings component renders UserSettings when logged in 1`] = `
           }
         }
       >
-        App Version: 1.0.0.1
+        App Version: 1.0.0 (development)
```
calebpanza, line 122: Could we include a snap of the expected response when this is not a `development` build?

vinnyjth (approved):

vinnyjth:
love this!

templates/mobile/src/settings/__snapshots__/UserSettings.test.js.snap
redreceipt: done

PR #2806, opened 2022-11-23
removes old tab view

redreceipt:
This removes the unused `TabView` and upgrades our `HorizontalTileFeed` to use the `react-native.FlatList` over a very old one from `react-native-gesture-handler`

https://user-images.githubusercontent.com/2659478/203456204-39bc33a5-390e-4c8e-bd6b-0d5208452af9.mov


vinnyjth:
Need to clone down and test but this looks good to me. 

I imagine we used the `gesture-handler` flat list for some old animation related functionality as well.

redreceipt:
> I imagine we used the `gesture-handler` flat list for some old animation related functionality as well.

Possibly, but the comment I removed only said it was necessary when having a `FlatList` inside of a `TabView` on Android

conrad-vanl (approved):

PR #2807, opened 2022-11-23
adds android e2e tests

redreceipt:
will need to create a virtual device through Android Studio called "Pixel" first. test with `yarn test:android`.


redreceipt:
currently blocked by `react-native-sha256` minSDK version

Screenshot 2022-11-22 at 11 35 56 PM

PR here: https://github.com/itinance/react-native-sha256/pull/60

redreceipt:
also may have to use `avdmanager` to create an emulator on the CI: https://developer.android.com/studio/command-line/avdmanager

redreceipt:
timing isn't right

PR #2808, opened 2022-11-23
Chore: Analytics refactor

conrad-vanl:
## ✏️ Description

Significantly refactors/simplifies the ui-analytics package and implements built-in support for a global Segment setup.

### Backstory
In evaluating our data needs, I've learned that:

- We need to track much better data for our own purposes. We have no cross-church analytics and no way to look at app performance in the macro
- This not only makes product decisions harder, but leaves all of our teams (marketing, sales, customer success, etc) without access to valuable shared data
- Furthermore, many of our clients are starting to desire more data sent to more platforms: Tableau, GA, Rock, etc. Setting up additional data destinations is difficult with our current setup
- Something like Astronomer is not optimized for click-stream data (although it works great in a reverse ETL setup like we have it now)

Also, `identify` is never being called on production, so even our current analytics tools is not getting that information.

### Customer Data Platform
I am still evaluating Customer Data Platform solutions. Segment, mParticle, Rudderstack, Salesforce, etc. Segment appears to be a clear choice for now:

- Price model nets it out cheaper then some competitors with lots of room to grow
- High number of integrations
- Ability to programmatic control destinations and filtering per-destination, meaning we could use Segment to funnel customer-specific data to customer-specific tools (such as use Segment to funnel Rock data for a single church into their Rock)

### ~~This removes our server-side analytics tracking~~
~~So far, this PR removes tracking to our server-side analytics. I'm not even sure if any church is using this outside of Crossroads (which, unfortunately, they use Segment....and the greatest drawback of Segment is they don't seem to have a Segment to Segment solution). However, these track events are accounting for a reasonable amount of our web traffic on the API, so removing these hits to our API might improve performance overall.~~

### ~~I made the TV Changes blind~~
~~I tested mobile, but I have not booted TV to see if these changes work without error. I believe they should, but will need more 👀 on this.~~

## 🔬 To Test
1. Add Segment key to your .env: `SEGMENT_KEY=rwkfWr0HpIW4eSUZqI30BmcMAdwsVrW0` (I don't anticipate using this key in production)
2. Boot up the sim, click around
3. Login to Segment and check [the live event stream](https://app.segment.com/differential-conrad/sources/apollos/debugger) (password in 1password)

## 📸 Screenshots

Screenshot 2022-11-22 at 11 50 12 PM

## ⚠️ To-do Before Merge

- [ ] Test that amplitude and onesignal events are still coming through correctly
- [ ] Test that interactions (like completing a content item) is going into the API correctly

## 📝 Checklist


conrad-vanl:
@vinnyjth / @jordanwade / @redreceipt not going to mark this ready for review, but this is something I've talked to @vinnyjth some about and I think it's to a point where I'd like y'alls feedback on:

- How I'm keeping ui-analytics and sharing between mobile and tv
- That I'm removing server-side tracking
- Other ideas before I continue investigating CDP options
- Anything else? should I remove some of the backward compatibilities and touch more code?

Also, this is nice. Most of the deletions are from snapshots but it is removing a lot of low-value code:
Screenshot 2022-11-22 at 11 57 27 PM

conrad-vanl:
PS...the main component/file to look at is `packages/apollos-ui-analytics/src/Provider.js`. All other changes were made to support that file. That's the new "Core" analytics provider that this PR proposes we use on react-native based projects.

jordanwade:
@conrad-vanl I tried to test this out on TV. There are a few simple import issues. But, the biggest problem is the `apollos-config` requires AsyncStorage. If I remember correctly, TVs can't rely on that since cache is blown away regularly and it has a pretty strict limit on its size. Instead, in the TV app we use a package called `react-native-default-preference` to act as a local storage for things like auth tokens.

Do you have any ideas on how to work around that dependence?

vinnyjth:
@jordanwade @redreceipt has a PR that gets rid of `ApollosConfig` :) 

I'm not sure if that also removes the AsyncStorage dependency also.

redreceipt:
it doesn't touch asyncstorage

conrad-vanl:
@vinnyjth I put Core Interactions (our graphql interactions) back into this package. Instead of relying on our dated code, I was able to leverage @jordanwade's version that they had in TV.

I tested that it works by completing the "Read Scripture" step of the daily reading.

Going to test up on TV then this _might_ be ready for merge.

conrad-vanl:
Had to rip out Segment on TV as it appears `segment-react-native` doesn't work on atleast tvOS. 

That's ok, once we validate segment's long term feasibility we can utilize Segment's JS library on the tv side.

calebpanza:
@conrad-vanl catching up on this PR. Is the intention for this to be additive so we can run both packages (Amplitude and Segment) alongside each other?

conrad-vanl:
@calebpanza yes, this PR is meant to maintain backwards compatible support for current clients that use Amplitude (same with OneSignal). 

Going forward, we'll likely use Segment to forward events to clients' Amplitude apps, but this can be launched without breaking anything for existing clients.

conrad-vanl:
@calebpanza i also just invited you to segment so you can test events flowing in locally.

Just set your segment key env var (look at the code to see what it should be 😉  )

The source event debugger sometimes has a delay (I thought it wasn't working for a while then POOF events started flowing in -> https://app.segment.com/apollos/sources/apollos/debugger )

packages/apollos-tv-shared/providers/AppProvider.js
```diff
@@ -12,15 +13,17 @@ import LiveStreamsProvider from './LiveStreamsProvider';
...
-      <BreakpointProvider>
-        <AuthProvider>
-          <LiveStreamsProvider>
-            <ThemeProvider customTheme={props.customTheme}>
-              {props.children}
-            </ThemeProvider>
-          </LiveStreamsProvider>
-        </AuthProvider>
-      </BreakpointProvider>
+      <AnalyticsProvider>
```
calebpanza, line 16: Same question about whether the addition on TV is valid given the issues with TV & Segment

packages/apollos-ui-analytics/package.json
```diff
@@ -38,27 +39,24 @@
...
-    "react-native": "*"
-  },
-  "dependencies": {
-    "react-native-device-info": "^4.0.1"
+    "react-native": "*",
+    "react-native-config": "*",
+    "react-native-device-info": "^10.3.0"
   },
   "devDependencies": {
-    "@apollo/client": "3.6.9",
```
calebpanza, line 55: What this intentional? On the `CoreInteractions.js`, you're importing from `@apollo/client`

packages/apollos-ui-analytics/src/CoreInteractions.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+`;
+
+const DEVICE_ID = DeviceInfo.getUniqueId();
+const MODEL = DeviceInfo.getModel();
+const PLATFORM = DeviceInfo.getSystemName() === 'tvOS' ? 'tvOS' : 'AndroidTV';
+const OS_VERSION = DeviceInfo.getSystemVersion();
+const APP_VERSION = DeviceInfo.getBuildNumber();
+
+export const useCoreTrack = (options = {}) => {
```
calebpanza, line 26: 📝 great chance to add some documentation

packages/apollos-ui-analytics/src/CoreInteractions.js
```diff
@@ -0,0 +1,89 @@
...
+    };
+  };
+
+  const [track] = useMutation(TRACK, options);
+
+  return (eventName, properties = {}) =>
+    track(mergeOptionsWithDefaults({ eventName, properties }));
+};
+
+export const useCoreIdentify = (options = {}) => {
```
calebpanza, line 59: 📝 great chance to add some documentation

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,146 @@
...
-    },
-  },
-});
+    }
+  }
+`;
 
-const Provider = ({
+export const AnalyticsProvider = ({
+  clients = defaultClients,
```
calebpanza, line 40: Just a random question. Is there any difference in using inline defaults vs. AnalyticsProvider.defaultProps?

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,146 @@
...
+
+  const clientsWithDefaults = useMemo(() => [{ track, identify }, ...clients], [
+    clients,
+    track,
+    identify,
+  ]);
+
+  const client = useMemo(() => {
+    return {
+      screen: (...args) =>
```
calebpanza, line 55: 📝 great chance to add some documentation

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,146 @@
...
+            // backwards compatibility: if no screen method is provided, use track
+            // This is mainly used to support Legacy Amplitude + CoreInteractions
+            // which does not provide a "screen" method.
+            if (!c?.screen && c.track) {
+              return c.track(...args);
+            }
+            return c?.screen && c.screen(...args);
+          })
+        ),
+      track: (eventName, properties, ...args) => {
```
calebpanza, line 67: 📝 great chance to add some documentation

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,146 @@
...
-export const useTrack = () => {
-  const ctx = useContext(AnalyticsContext);
-  return ctx.track;
+        return Promise.all(
+          clientsWithDefaults?.map(
+            (c) => c?.track && c.track(eventName, properties, ...args)
+          )
+        );
+      },
+      identify: (...args) =>
```
calebpanza, line 80: 📝 great chance to add some documentation. Are these segment-specific?

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,146 @@
...
+        Promise.all(
+          clientsWithDefaults?.map((c) => c?.identify && c.identify(...args))
+        ),
+      flush: (...args) =>
```
calebpanza, line 84: 📝 great chance to add some documentation. Are these segment-specific?

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,146 @@
...
+        Promise.all(
+          clientsWithDefaults?.map((c) => c?.flush && c.flush(...args))
+        ),
+      group: (...args) =>
```
calebpanza, line 88: 📝 great chance to add some documentation. Are these segment-specific?

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,146 @@
...
+        Promise.all(
+          clientsWithDefaults?.map((c) => c?.group && c.group(...args))
+        ),
+      alias: (...args) =>
```
calebpanza, line 92: 📝 great chance to add some documentation. Are these segment-specific?

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,146 @@
...
+        Promise.all(
+          clientsWithDefaults?.map((c) => c?.alias && c.alias(...args))
+        ),
+      reset: (...args) =>
```
calebpanza, line 96: 📝 great chance to add some documentation. Are these segment-specific?

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,146 @@
...
+  }, [clientsWithDefaults]);
+
+  // Auto track user
+  const { data } = useQuery(GET_ANALYTICS_USER, {
+    returnPartialData: true,
+    errorPolicy: 'ignore',
+  });
+  useEffect(() => {
+    if (!data?.currentUser?.profile?.id) {
+      return;
```
calebpanza, line 110: Should this return or does it need to clear the current identity?

Example:

- I log in --> Identity === Caleb
- I log out --> Identity === Caleb
- I log in as Conrad but something fails in the network request --> Identity === Caleb

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,146 @@
...
+    }
+    client.identify(data.currentUser.profile.id, {
+      ...data.currentUser.profile,
+      avatar: data.currentUser.profile.photo?.uri,
+    });
+  }, [client, data?.currentUser?.profile]);
+
+  // Reset when changing users
+  const lastId = useRef();
+  useEffect(() => {
```
calebpanza, line 120: The use of a separate effect seems a bit redundant and this effect could be captured in the comment above on line `108`. Save from another effect firing off

packages/apollos-ui-analytics/src/Provider.tests.js
```diff
@@ -0,0 +1,113 @@
...
+  },
+};
+
+const ProviderStack = ({ children }) => (
+  <MockedProvider mocks={[userMock]}>
+    <AnalyticsProvider clients={[testClient]}>{children}</AnalyticsProvider>
+  </MockedProvider>
+);
+
+describe('Provider', () => {
```
calebpanza, line 57: Tests are freaking out on my local machine. Says it'll take over 7 minutes to run 11 tests and this error keeps popping up

```  
●  Cannot log after tests are done. Did you forget to wait for something async in your test?
    Attempted to log "Warning: An update to AnalyticsProvider inside a test was not wrapped in act(...).

    When testing, code that causes React state updates should be wrapped into act(...):

    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */

    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
        in AnalyticsProvider (created by ProviderStack)
        in ApolloProvider (created by MockedProvider)
        in MockedProvider (created by ProviderStack)
        in ProviderStack".

      at console.error (../../node_modules/@jest/console/build/BufferedConsole.js:165:10)
      at printWarning (../../node_modules/react-test-renderer/cjs/react-test-renderer.development.js:110:34)
      at error (../../node_modules/react-test-renderer/cjs/react-test-renderer.development.js:88:9)
      at warnIfNotCurrentlyActingUpdatesInDEV (../../node_modules/react-test-renderer/cjs/react-test-renderer.development.js:11524:11)
      at dispatchAction (../../node_modules/react-test-renderer/cjs/react-test-renderer.development.js:5599:13)
      at ../../node_modules/@apollo/client/react/hooks/useMutation.js:76:21
      at tryCallOne (../../node_modules/react-native/node_modules/promise/lib/core.js:21:12)
      at ../../node_modules/react-native/node_modules/promise/lib/core.js:122:15
      at flush (../../node_modules/asap/raw.js:25:25)
```

Coverage also seems a bit low. Would love to help get coverage a bit higher due to the significance of the code.

packages/apollos-ui-analytics/src/withTrackOnPress.tests.js
```diff
@@ -1,56 +1,100 @@
...
+    let tree;
+    renderer.act(() => {
+      tree = renderer.create(
+        <ProviderStack>
+          <Button trackEventName="SomeEvent" onPress={jest.fn()} />
+        </ProviderStack>
+      );
+    });
+    await act(async () => await wait(0));
     expect(tree).toMatchSnapshot();
```
calebpanza, line 65: What is the snapshot proving in this situation? The test wants to confirm that the `trackEvenName` is being passed through properly, but the snapshot gives no indication that's the case. A better test might be to run the `onPress` on the button and confirm the proper event name is being passed through

packages/apollos-ui-analytics/src/withTrackOnPress.tests.js
```diff
@@ -1,56 +1,100 @@
...
+            trackEventName="SomeEvent"
+            trackProperties={{ ItemId: '123' }}
+            onPress={jest.fn()}
+          />
+        </ProviderStack>
+      );
+    });
+    await act(async () => await wait(0));
+
     expect(tree).toMatchSnapshot();
```
calebpanza, line 84: Same comment as above

packages/apollos-ui-analytics/src/withTrackOnPress.tests.js
```diff
@@ -1,56 +1,100 @@
...
+    let tree;
+    act(() => {
+      tree = renderer.create(
+        <ProviderStack>
+          <Button onPress={jest.fn()} />
+        </ProviderStack>
+      );
+    });
+    await act(async () => await wait(0));
     expect(tree).toMatchSnapshot();
```
calebpanza, line 98: Same comment as above

packages/apollos-ui-connected/src/DailyHabitsConnected/PrayerExperience/PrayerSwipeScreen.js
```diff
@@ -56,7 +56,7 @@ function useStyleSheet() {
 
 const PrayerSwipeScreen = ({}) => {
   const [styles] = useStyleSheet();
-  const { track } = useContext(AnalyticsContext);
+  const track = useTrack();
```
calebpanza, line 59: One of the comments in the code mentioned to prefer using the `useAnalytics` hook over the `useTrack`. Is this a case where we should update this section to use the new method?

packages/apollos-ui-prayer/src/PrayerExperienceConnected/index.js
```diff
@@ -27,7 +27,7 @@ const PrayerExperienceConnected = ({
     partialRefetch: true,
     returnPartialData: true,
   });
-  const { track } = useContext(AnalyticsContext);
+  const track = useTrack();
```
calebpanza, line 30: Same question as above. Should this be updated to `useAnalytics`?

packages/apollos-ui-prayer/src/screens/AddPrayerScreenConnected/index.js
```diff
@@ -48,7 +48,7 @@ const AddPrayerScreenConnected = ({
   const photo = userData?.currentUser?.profile?.photo;
   const profile = userData?.currentUser?.profile;
 
-  const { track } = useContext(AnalyticsContext);
+  const track = useTrack();
```
calebpanza, line 51: Same question as above. Should this be `useAnalytics`?

packages/apollos-ui-prayer/src/screens/PrayerScreen/index.js
```diff
@@ -36,7 +36,7 @@ const PrayerScreen = ({
   prayer,
   ...props
 }) => {
-  const { track } = useContext(AnalyticsContext);
+  const track = useTrack();
```
calebpanza, line 39: Same comment as above. Should this be `useAnalytics`?

templates/mobile/src/components/PrayerExperienceSwiper/PrayerExperienceSwiper.js
```diff
@@ -4,7 +4,7 @@
  * Displays a collection of prayer request that a user can swipe between and interact with.
  */
 
-import { AnalyticsContext } from '@apollosproject/ui-analytics';
+import { useTrack } from '@apollosproject/ui-analytics';
```
calebpanza, line 7: Is this suppose to be `useAnalytics`?

templates/mobile/src/analytics/segment.js
```diff
@@ -0,0 +1,16 @@
...
+import Config from 'react-native-config';
+
+let segmentClient = {
+  track: () => {},
+  identify: () => {},
+};
+
+if (Config.SEGMENT_KEY || process.env.NODE_ENV === 'production') {
+  segmentClient = createClient({
+    writeKey: Config.SEGMENT_KEY || 'rwkfWr0HpIW4eSUZqI30BmcMAdwsVrW0',
```
calebpanza, line 11: Probably shouldn't use a default value here. If no key is set, should probably just report to Sentry and crash the app or something

calebpanza (changes requested):
Sorry for the lengthy review! 😅 It's a _lot_ of pretty substantial changes and just wanna make sure it's all tidy. Excited for the updates and looking forward to gathering more data.

I'm more than happy to jump in and help resolve any of the changes here. @conrad-vanl let me know if you'd like for me to jump into any of these things.

conrad-vanl:
Screenshot 2023-01-20 at 2 10 04 PM
This escalated quickly. Thanks for the thoroughness! I'll dig in

packages/apollos-tv-shared/hooks/useAnalytics.js
```diff
@@ -1,110 +1 @@
...
-  };
-
-  const [identify] = useMutation(ANALYTICS_IDENTIFY, options);
-
-  return ({ eventName, traits }) =>
-    identify(mergeOptionsWithDefaults({ traits }));
-}
-
-export { useTrack, useIdentify };
+export { useTrack, useIdentify } from '@apollosproject/ui-analytics';
```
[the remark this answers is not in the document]
conrad-vanl: No, I don't think so, because the analytics package still provides a consistent interface across platforms. Could use it on Web too.

packages/apollos-tv-shared/package.json
```diff
@@ -9,7 +9,7 @@
     "test": "jest --no-watchman"
   },
   "peerDependencies": {
-    "@amplitude/react-native": "^2.6.0",
+    "@amplitude/react-native": "^2.17.0",
```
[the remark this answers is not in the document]
conrad-vanl: Yes

packages/apollos-tv-shared/providers/AppProvider.js
conrad-vanl: Going to resolve this comment and focus on the one above

packages/apollos-tv-shared/hooks/useAnalytics.js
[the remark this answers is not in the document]
conrad-vanl: Another way to think about it....

Before this PR TV and App were using a completely different analytics provider and framework. This PR makes it consistent and provides a hook-based mechanism that we can use across all Apollos Code (if we wanted to).

packages/apollos-ui-analytics/package.json
conrad-vanl: yeah there's a peer dependency listed but it's a mistake to not have it in dev as well

packages/apollos-ui-analytics/src/Provider.js
conrad-vanl: React community has moved to using this as the standard because it's more vanilla javascripty.

Other then that, no. It's considered a bad practice to reference `AnalyticsProvider.defaultProps` elsewhere, that would be the only difference

packages/apollos-ui-analytics/src/Provider.js
conrad-vanl: yeah i think this check keeps this useEffect from running if you haven't logged in yet

packages/apollos-ui-analytics/src/Provider.js
conrad-vanl: They're separate code flows though, I think splitting them out to separate `useEffects` increases readability? Clearer that they're unrelated code?

packages/apollos-ui-analytics/src/withTrackOnPress.tests.js
conrad-vanl: These were existing snapshots ported over, I did not touch them other then adding the `await act` stuff trying to fix the warnings.

`withTrackOnPress` is deprecated code that's only used in one place, I'm not sure it's worth fixing tests.

packages/apollos-ui-analytics/src/withTrackOnPress.tests.js
conrad-vanl: Resolving to clean up discussion list a little, we can discuss above

packages/apollos-ui-analytics/src/withTrackOnPress.tests.js
conrad-vanl: Resolving to clean up discussion list a little, we can discuss above

templates/mobile/src/analytics/segment.js
conrad-vanl: key is sharable / not considered a secret. We're using the same segment instance for all apps.

To me this is just like defaulting to `https://production.apollos.app` if a `DATA_URL` isn't provided.

vinnyjth: I'm with @calebpanza here. The default value should be in the Config DB `default` church slug. We've been bitten by having defaults in multiple places in the past :)

packages/apollos-ui-analytics/src/Provider.tests.js
conrad-vanl: Feel free to try and fix the tests freaking out. I spent hours on it.... if you start searching for this error message you'll eventually end up on a Github issue where one of the last comments is from @redreceipt with no resolution.

As for test coverage, here's the coverage report for the files focused on in this PR:
Screenshot 2023-01-20 at 2 20 34 PM

`Provider.js` is the main focus of texts that I wrote. Those two uncovered line #s that are listed are going to be really hard to capture in valuable tests. Lines 122-123 covers the code that handles reseting a user. I'm not sure how to fake this using Apollo Mocks. As we'd have to render the component once, change the mock, then re-render it. And this wouldn't show up in a snapshot test, so we'd have to just use mocked functions to show that `.reset()` and `.group()` gets called. What's the risk if this code doesn't properly fire? Well, interactions until the user logs in again would be tied to the user that was last logged in.

That second code group - 142-143, cover the deprecated `useTrack` function, which it sounds like we might just remove support for in this PR based on your other comments.

`CoreInteractions` was ported over almost verbatim from TV, which didn't include test coverage. 

The other components are existing components that are untouched by this PR, so I didn't work on test coverage either. I would rather remove those components from our code base then spend time writing tests for them, was just trying to keep this PR smaller. 😅

packages/apollos-ui-analytics/package.json
calebpanza: @conrad-vanl looks like the peer dependency got removed, too. So right now `apollo/client` is _not_ listed in any of the dependencies

packages/apollos-ui-analytics/src/Provider.js
calebpanza: In that case, might consider defining each of those using `useCallback` and then just having 1 `useEffect` that fires off each of the separate code flows.

packages/apollos-ui-analytics/src/Provider.tests.js
calebpanza: Ok cool! That makes sense. So for those files that are intended to be deprecated in the near-distant future, maybe we could just mark them with an intent to deprecate. That way we keep track of it and know to not spend tons of time on better coverage and such.

packages/apollos-ui-analytics/src/withTrackOnPress.tests.js
calebpanza: What would it take to remove it from that one place? Then we could just kill the file entirely which would probably help

calebpanza:
Lol. Hope it's not too much/wildly overwhelming! Really impressed by how much is covered with this PR and just wanna make sure to give it a proper test

packages/apollos-ui-analytics/package.json
conrad-vanl: oh i understand now

packages/apollos-ui-analytics/src/Provider.js
conrad-vanl: What's wrong with two `useEffect`s?

templates/mobile/src/analytics/segment.js
conrad-vanl: @vinnyjth That's fine, but how is this differen't then the way we're doing `APP_DATA_URL` [here](https://github.com/ApollosProject/apollos-apps/blob/master/templates/mobile/src/client/index.js#L26)? That's what I based this code on. I don't think we'd be setting `SEGMENT_KEY` ever except locally, and I do NOT want churches to be able to override and provide their own segment key. This is for our analytics purposes, not theirs'

packages/apollos-ui-connected/src/DailyHabitsConnected/PrayerExperience/PrayerSwipeScreen.js
conrad-vanl: There's currently pretty wide-spread usage of `useTrack` and I'd really like to try to avoid refactoring that out in this PR. I feel like it's already big enough. `useContext(...)` was worse and not as supported, and less widespread, so I atleast took those out.

conrad-vanl:
@calebpanza addressed _most_ of your comments. 

@vinnyjth since Caleb is out didn't know if you wanted to weigh in on the review at all here -> I know segment roll out is lower priority then other things but I'm trying to get a handle on our data which is blocked by this.

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,173 @@
...
+    if (lastId.current && lastId.current !== data?.currentUser?.profile?.id) {
+      client.reset(); // only call reset after a logout and log back in as a different user
+      client.group(church); // needs to be re-called after reset
+    }
+    lastId.current = data?.currentUser?.profile?.id;
+  }, [client, church, data?.currentUser?.profile?.id]);
+
+  // Auto-set group
+  useEffect(() => {
+    client.group(church); // also called on reset
```
vinnyjth, line 150: nice

packages/apollos-ui-prayer/src/PrayerExperienceConnected/index.js
vinnyjth: Answered above :) Minimizing amount of code changes.

packages/apollos-ui-prayer/src/screens/AddPrayerScreenConnected/index.js
vinnyjth: Answered above :) Minimizing amount of code changes.

packages/apollos-ui-prayer/src/screens/PrayerScreen/index.js
vinnyjth: Answered above :) Minimizing amount of code changes.

templates/mobile/src/analytics/onesignal.js
```diff
@@ -1,7 +1,7 @@
 import { snakeCase } from 'lodash';
 import OneSignal from 'react-native-onesignal';
 
-export default ({ eventName, properties }) => {
+export const track = (eventName, properties) => {
```
vinnyjth, line 4: I totally forgot we had these

templates/mobile/src/index.js
```diff
@@ -122,12 +123,37 @@ const linking = {
   },
 };
 
+const getActiveRoute = (state) => {
+  if (!state || typeof state.index !== 'number') {
+    return 'Unknown';
```
vinnyjth, line 128: So we don't risk a red screen error line 153.
```suggestion
    return { name: 'Unknown' };
```

templates/mobile/src/analytics/segment.js
```diff
@@ -0,0 +1,16 @@
...
+import Config from 'react-native-config';
+
+let segmentClient = {
+  track: () => {},
+  identify: () => {},
+};
+
+if (Config.SEGMENT_KEY) {
+  segmentClient = createClient({
+    writeKey: Config.SEGMENT_KEY,
```
vinnyjth, line 11: this feels like the right choice to me 👍

vinnyjth:
Seeing a lot of identifies after logging in

vinnyjth:
https://user-images.githubusercontent.com/1637694/214390491-6931f53c-cf0a-491b-903f-3981f08fea5b.mp4

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,173 @@
...
+  });
+  useEffect(() => {
+    if (!data?.currentUser?.profile?.id) {
+      return;
+    }
+    client.identify(data.currentUser.profile.id, {
+      ...data.currentUser.profile,
+      avatar: data.currentUser.profile.photo?.uri,
+    });
+  }, [client, data?.currentUser?.profile]);
```
vinnyjth, line 136: Removing `client` here stops an infinite identify loop from happening. Might be better options? 
```suggestion
  }, [data?.currentUser?.profile]);
```

packages/apollos-ui-analytics/src/CoreAnalaytics.js
```diff
@@ -0,0 +1,103 @@
...
+
+const IDENTIFY = gql`
+  mutation identifySelf($input: AnalyticsIdentifyInput!) {
+    identifySelf(input: $input) {
+      success
+    }
+  }
+`;
+
+const DEVICE_ID = DeviceInfo.getUniqueId();
```
vinnyjth, line 20: `getUniqueId` returns a promise - we'll have to refactor slightly

vinnyjth: CleanShot 2023-01-24 at 14 40 49@2x

scripts/setup-app.sh
```diff
@@ -25,6 +25,7 @@ GOOGLE_MAPS_KEY=$(curl -s --fail -H "x-api-key: $APOLLOS_API_KEY" -H "accept: te
 AMPLITUDE_KEY=$(curl -s --fail -H "x-api-key: $APOLLOS_API_KEY" -H "accept: text/plain" "$API_URL/config/$CHURCH/APP.AMPLITUDE_KEY")
 SENTRY_DSN=$(curl -s --fail -H "x-api-key: $APOLLOS_API_KEY" -H "accept: text/plain" "$API_URL/config/$CHURCH/APP.SENTRY_DSN")
 SENTRY_TOKEN=$(curl -s --fail -H "x-api-key: $APOLLOS_API_KEY" -H "accept: text/plain" "$API_URL/config/$CHURCH/APP.SENTRY_TOKEN")
+SEGMENT_KEY=$(curl -s --fail -H "x-api-key: $APOLLOS_API_KEY" -H "accept: text/plain" "$API_URL/config/$CHURCH/APP.SEGMENT_KEY")
```
vinnyjth, line 28: 💥

vinnyjth:
@conrad-vanl Fixed the identity and anonymous ID issues mentioned above, but you'll have to let me know if you care for my approach. 

Only other thing I think this PR needs is a clear way to pass on the church slug, I don't think we want to rely on the app name

vinnyjth:
Hm - the `group` might be enough but I don't see it in the amplitude debugger. Digging deeper.
CleanShot 2023-01-24 at 14 54 23@2x

conrad-vanl:
Yeah hoping we can rely on `group` for that, either way we'll want to be able to leverage [Destination Filters](https://segment.com/docs/connections/destinations/destination-filters/) by `church_slug`

vinnyjth:
I can see the group events flowing through in the segment logs - so maybe it's a syntax issue? 

CleanShot 2023-01-24 at 15 03 29@2x

packages/apollos-ui-analytics/src/CoreAnalaytics.js
```diff
@@ -0,0 +1,115 @@
...
+ * A hook that returns a function to track analytic events on cluster
+ * You should not call these directly, rather use the `useAnalytics` hook
+ * @example
+ * const track = useCoreTrack();
+ * track('Event Name', { property: 'value' });
+ */
+export const useCoreTrack = (options = {}) => {
+  const [anonymousId, setAnonymousId] = useState();
+
+  useEffect(() => {
```
vinnyjth, line 37: gut check my logic @conrad-vanl 

we'll need to do the same thing in `identify` as well

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,173 @@
...
-            notify: (error) => notifyFunctions.forEach((fn) => fn(error)),
-          }}
-        >
-          {children}
-        </AnalyticsContext.Provider>
-      );
-    }}
-  </ApolloConsumer>
-);
+  church,
```
vinnyjth, line 62: @conrad-vanl changed this to not have a default - it's handled above

packages/apollos-ui-analytics/src/Provider.js
```diff
@@ -1,168 +1,173 @@
...
+
+  // Reset when changing users
+  const lastId = useRef();
+  useEffect(() => {
+    if (lastId.current && lastId.current !== data?.currentUser?.profile?.id) {
+      client.reset(); // only call reset after a logout and log back in as a different user
+      client.group(church?.slug); // needs to be re-called after reset
+    }
+    lastId.current = data?.currentUser?.profile?.id;
+  }, [church, data?.currentUser?.profile?.id]);
```
vinnyjth, line 146: @conrad-vanl I'm noticing that `client` in the deps is causing this hook to fire on every page change. Do you have a better idea here? any ideas on how to make `client` mutate less?

conrad-vanl: 🤔

packages/apollos-ui-analytics/src/CoreAnalaytics.js
conrad-vanl: @vinnyjth do we need this ID? 
https://github.com/react-native-device-info/react-native-device-info#getuniqueid

Also, how is this being used? Do we really want https://github.com/react-native-device-info/react-native-device-info#getdeviceid instead?

conrad-vanl: Really I think we should turn `track` into a promise and await deviceId

packages/apollos-ui-analytics/src/Provider.js
conrad-vanl: I bet it has something to do with how `clientsWithDefaults` is getting constructed. This wasn't happening originally, and when I started implementing/fix on TV I changed this up quite a bit. I have an idea

packages/apollos-ui-analytics/src/CoreAnalaytics.js
vinnyjth: If we're trying to implement `anonymousId` I think we really need `uniqueId`. Refactoring as a promise sound great 👍

conrad-vanl: what do we need `anonymousId` though? 

We don't use it in the [Rock Interactions interface](https://github.com/ApollosProject/apollos-cluster/blob/master/src/data/analytics/interfaces/rock_interactions.js).

We do in Segment, but only crsds tv uses segment, and only on their TV app.

On Third Thought, WDYT about just removing core analytics?

I know it doesn't account for a lot of server time but it seems like dead code.

When we port Crsds TV over to Core TV, we'll just wire up their segment key on the frontend -> we can have two segment libraries installed at once so what we'll do is add support for a second segment key (or use our segment instance to forward segment events to theirs, if it can)

vinnyjth: Was thinking the same thing myself...  especially now that we're implementing segment. 

Can we confirm that the "VIEW" interaction is fired when viewing content items _without_ core analytics?

conrad-vanl: sure, that's an INTERACTION not an analytics event :) 

I was getting the two confused

vinnyjth: I think we might be relying on the analytics event creating an interaction as a side effect for that particular interaction though, would need to dig to confirm.

conrad-vanl: I don't think there's a VIEW interaction, just a COMPLETE

vinnyjth: yah here it is 

https://github.com/ApollosProject/apollos-apps/blob/283a5a3db7c26a93826c844ee1ae6bb812e9c3e7/templates/mobile/src/content/ContentSingleConnected/ContentSingle.js#L30-L46 which calls a method here https://github.com/ApollosProject/apollos-cluster/blob/master/src/data/analytics/interfaces/rock_interactions.js#L30 which... doesn't exist. 

yah safe to say we can remove it :) how are `VIEW` interactions being created?

conrad-vanl: I was about to say there's not `VIEW` interactions, but I do see them. only on `VideoMedia` though. Probably in the video player

vinnyjth: lol, didn't see the comments you added before I went digging. Looks like we create VIEW interactions on Scripture and Media. So yah - I vote we kill the core analytics functions.

calebpanza:
Trying to catch up on the progress made here. How can I help?

conrad-vanl:
Hit the green approve button? Lol. I think it's ready....

conrad-vanl:
@calebpanza @vinnyjth anything else you want to see on this?

calebpanza:
@conrad-vanl events are coming through on mobile. I don't see the "prayed for prayer" interaction coming through, though. Do we want a fast follow up for user interactions auditing or get it done here?

conrad-vanl:
Are you confusing interactions and analytics?

This does not touch interactions

calebpanza:
I guess to reword my question: are we intending to use Segment for interaction data as well? I'm noticing that interaction data is not being forwarded to Segment

conrad-vanl:
Correct. We are not currently forwarding interaction data to segment. This PR is intended to be a 1:1 replacement of current analytics without touching interacitons.

In most cases, we're sending both an analytic event and an interaction and so making this PR handle all interactions as well would result in double events being sent.

The intention would be to follow up with those things in subsequent PRs.

calebpanza:
ok perfect! Just wanted to confirm. I think I'm good with this, then

calebpanza (approved):

PR #2815, opened 2022-11-28
fixes content navigation

redreceipt:
this was my fault...

https://user-images.githubusercontent.com/2659478/204360713-5a83f490-bc03-4497-bd02-4a4a16cbee1a.mp4


vinnyjth (approved):

calebpanza:
@redreceipt I added a [`ScreenNames` enum](https://github.com/ApollosProject/apollos-apps/blob/master/templates/mobile/src/screens/ScreenNames.js) for these kinds of scenarios. While we're at it, might be worth adding the `ContentNavigator` key to this list.

calebpanza (approved):

redreceipt:
what does that enum get us that a unique string doesn't get us? we'd still have to go through and change the enum references all over the codebase, no?

redreceipt:
I take that back, it saves us from screen typos. wouldn't have helped with this particular bug but I can see how it would save us from others. useful if we're using typescript but an undefined screen enum name wouldn't show up until run time in plain ol javascript I think.

calebpanza:
Yeah it's just utilizing hard references vs. strings. Would have saved us in this case cause you would be able to update the Screen Value while keeping the reference the same and everything would have worked. But we can save that for whenever we get around to migrating Content Single to UI kit 2.0

redreceipt:
oh yea good call

PR #2819, opened 2022-12-01
fix: android header issues

nlewis84:
## ✏️ Description

This PR attempts to address some ongoing issues we are having with Android header avoidance. Right now we have content, featureFeeds, and searchConnected that is being hidden under the header. This adds back the 'avoidHeader' prop to solve the featureFeed issue, insetHeight to solve the content issue, and safeArea 'top' to solve the search issue. Because, these components are all formed slightly differently and in different context, it was necessary to use three different solves instead of just one 😢 . Open to suggestions if the reviewer sees a way to just use two (or even 1!) solve instead of 3.

## 🔬 To Test
1. Pull down this PR
2. Start up the Android sim and check: User Settings, Personal Details, Prayer, Search, Content Items, and Tab Feature Feeds.  

## 📸 Screenshots

|Location|Before|After|
|---|---|---|
|Home|Screenshot 2022-11-30 at 7 49 33 PM|Screenshot 2022-11-30 at 7 34 12 PM|
|Search|Screenshot 2022-11-30 at 7 49 51 PM|Screenshot 2022-11-30 at 7 34 36 PM|
|Content|Screenshot 2022-11-30 at 7 50 05 PM|Screenshot 2022-11-30 at 7 35 09 PM|
|User Settings|Screenshot 2022-11-30 at 7 50 17 PM|Screenshot 2022-11-30 at 7 35 18 PM|
|Personal Details|Screenshot 2022-11-30 at 7 50 27 PM|Screenshot 2022-11-30 at 7 35 27 PM|

## ⚠️ To-do Before Merge
None

## 📝 Checklist


templates/mobile/src/content/ContentSingleConnected/index.js
```diff
@@ -13,12 +14,14 @@ const ContentSingleNavigator = ({ route }) => (
...
-        headerStyle: { backgroundColor: 'transparent' },
+        headerStyle: {
+          backgroundColor: 'transparent',
+        },
         headerHideShadow: true,
         headerRight: ModalCloseButton,
         headerLeft: ModalBackButton,
         headerTitle: '',
-        headerTopInsetEnabled: false,
+        headerTopInsetEnabled: Platform.OS === 'android' ? true : false,
```
redreceipt, line 24: This is cool! You can't use this on each on the problem screens as a nav option instead of the other two solutions?

nlewis84: Let me investigate and I'll ping you with an answer!

nlewis84: @redreceipt Does not work on FeatureFeed or SearchScreenConnected.

nlewis84:
Ugh, Scripture Screen (modal) has overlapping issues as well on Android. I'm gonna make this draft again.

nlewis84:
Post Journal also.....

nlewis84:
Scripture issue is unrelated to this....I think I'll just drop a comment on it with a TODO. Basically the header size is too big for the container so it's top is getting cut off.

|Before adding a padding of 8|After adding the padding of 8|
|---|---|
|Screenshot_1669925998|Screenshot_1669925993|

nlewis84:
The StackNavigator for the ShareScreenConnected (Post Journal screen) was encountering the same android header issue (compounded because it also has a modal header to avoid). We were already using magical numbers here. I'm thinking I will make this less magical and use the deterministic logic that I built in the avoidHeader to calculate this prop value. 
Screenshot 2022-12-01 at 2 28 47 PM

nlewis84:
@redreceipt The more I mess with my original idea for avoidHeader, the more I like my logic. I plopped it into the ShareScreenConnected component and it worked beautifully!

I think the difference between something being hacky/smelly and great is whether you can import it from an npm package...I just need to package it up! 😂 

Why I think that logic worked in a different situation so well: It calculates the full screen height, then it calculates the current window height, then it subtracts the window from the screen. In the case of the FeatureFeed - This gives you the appropriate padding for just the Status Bar. In the case of the ShareScreenConnected - This gives you height for the Status Bar, the space between the Status Bar and the Modal, and the top (unaccessible) portion of the Modal.

Screenshot_1669927974

nlewis84:
Great question. When I try to rotate my sim the device rotates but the content does not 🤦‍♂️

vinnyjth:
@nlewis84 That might be true on a physical device as well! At least for Android

nlewis84:
@vinnyjth You are right! I just checked ios and android on real devices. We lock orientation on android.

redreceipt:
> I think the difference between something being hacky/smelly and great is whether you can import it from an npm package...I just need to package it up! 😂

Yea I don't mean to imply that it's bad code. A code smell to me is simply something that begs the question "this shouldn't be necessary. why is this necessary?" That's the question we can't answer and it's frustrating because the answer to that question could cause other problems down the road. But your prop is a perfectly good working solution for the problem at hand.

redreceipt:
I think leaving it on `BackgroundView` is fine

redreceipt:
but yes that sprawl is what I'm talking about. Likely there is a singular reason why it's necessary and if we could figure that out, I'm betting the solution is much smaller. But we don't have the time to investigate.

nlewis84:
Yeah, it's not just us either. This issue has comments from a few days ago. https://github.com/react-navigation/react-navigation/issues/10531

redreceipt:
Niiice. Yea maybe this will be solved when we upgrade to navigation@6 but who knows.

nlewis84:
This guy presented an interesting idea....set the header as translucent, then set the headerHeight manually as paddingTop. The hook he is using uses the same logic as our avoidHeader prop. Also, I'm leaning more that it is an issue with `react-native-reanimated` over `navigation`

https://github.com/software-mansion/react-native-reanimated/issues/2906#issuecomment-1082844611

redreceipt:
That wouldn’t surprise me and would be an easy experiment to prove out. Sent from my iPhoneOn Dec 1, 2022, at 5:01 PM, Nathan Lewis ***@***.***> wrote:﻿
This guy presented an interesting idea....set the header as translucent, then set the headerHeight manually as paddingTop. The hook he is using uses the same logic as our avoidHeader prop. Also, I'm leaning more that it is an issue with react-native-reanimated over navigation
software-mansion/react-native-reanimated#2906 (comment)

nlewis84:
His fix works as well for us. I'm going to experiment with placement of this logic wherever we have Navigators (setting `screenOptions`). Maybe it will touch less places that way.
Screenshot 2022-12-01 at 4 26 22 PM

nlewis84:
I'm going to have to close out the PR restart from scratch. I'm having to change too many things in the `deploys/mobile` folder just to see changes. So steps for me:

1. Close this PR
2. Pull down `dev`
3. Build the app
4. Save that `deploys/mobile` folder somewhere else on my computer (so I don't have to constantly rebuild a current state app
5. Begin working on a fix
6. If I need to reset my `deploys/mobile` folder, I'll just copy over the set aside version into my local repo.

redreceipt:
why do you need to use the deploys folder? just use the template to do the dev work

nlewis84:
I couldn't get `templates/mobile` to get past this screen on android: 

Didn't have this problem with the `deploys/mobile` folder.

nlewis84:
Putting this on hold and closing out the PR.

PR #2820, opened 2022-12-01
fix: Resolve Daily Habit Timezone Issues

calebpanza:
## ✏️ Description
The server needs to know my local timezone in order to accurately tell me if I've completed my Daily Habits with _my_ 24 hour day. This PR updates the front end to make adjustments based on a [backend update](https://github.com/ApollosProject/apollos-apps/pull/2820) that resolves issues resulting in inaccurate UI. Some of the bugs that these PR's resolve include:
- Completing an Experience _today_ marks _yesterday_ as complete
- Dates in the Streak were running from Sunday-Saturday instead of Monday-Sunday
- The id's coming back from the server resulted in inaccurate ID's which made cached results inaccurate

This PR introduces the [`react-native-localize`](https://github.com/zoontek/react-native-localize) package because we need to get the current user's timezone (America/New_York, America/Los_Angeles, Asia/Shanghai) in order to report that to the server and I couldn't find another way to do it.

## 🔬 To Test
1. Pull down the [Apollos Cluster branch](https://github.com/ApollosProject/apollos-apps/pull/2820) that has the new endpoints
2. Run the API locally
3. Start your app and open up Daily Habits
4. Complete 1 or more Experiences to progress through each stage
5. Change the timezone of your computer in order to see how different timezones effect historical data (New York --> Shanghai is a good example)

## ⚠️ To-do Before Merge

- [x] Ensure the [Apollos Cluster PR](https://github.com/ApollosProject/apollos-apps/pull/2820) is merged

## 📝 Checklist


templates/mobile/src/utils/useDailyHabits.js
```diff
@@ -137,6 +155,8 @@ function useDailyHabits() {
       'date'
     ) || [];
 
+  console.log(streak.map(({ date }) => date));
```
vinnyjth, line 158:
```suggestion
```

templates/mobile/src/utils/useDailyHabits.js
```diff
@@ -112,13 +109,34 @@ export const GET_DAILY_HABITS = gql`
...
+  return format(date, 'yyyy-MM-dd');
+}
+
 function useDailyHabits() {
+  const _startOfToday = startOfToday();
+  const _startOfWeek = startOfWeek(new Date(), { weekStartsOn });
+  const _endOfWeek = endOfWeek(new Date(), { weekStartsOn });
+  const timezone = getTimeZone();
+
+  console.log({
```
vinnyjth, line 122: remove this comment :)

templates/mobile/src/utils/useDailyHabits.js
```diff
@@ -112,13 +109,34 @@ export const GET_DAILY_HABITS = gql`
   }
 `;
 
+function formatWithTimeZone(date) {
```
vinnyjth, line 112: isn't this "format without time zone?" 

maybe better expressed as `formatLocalDate`?

vinnyjth (changes requested):
I'm mild skeptical about the need to pass the `dd-mm-yyyy` string instead of a timestamp, but I'll send that feedback in the api PR

calebpanza:
> I'm mild skeptical about the need to pass the dd-mm-yyyy string instead of a timestamp

Totally understand. It's a bit too much information for the date parsers to handle and we would end up using Regex on the server to just pull that specific bit out anyway. Check out the API logic cause it provides a LOT more context

dsilva609 (approved):

PR #2847, opened 2022-12-09
fixes multi-tenant theme colors

redreceipt:
## ✏️ Description

Correctly sets church theme on app initialization

## 🔬 To Test
1. Set `CHURCH_HEADER=<some church>` in your `.env` and boot the template

## 📸 Screenshots

Screenshot 2022-12-08 at 7 59 17 PM

## ⚠️ To-do Before Merge

N/A

## 📝 Checklist


templates/mobile/src/auth/ProtectedRoute.js
```diff
@@ -1,21 +1,35 @@
...
+      }
+    }
+  `);
+  const { theme } = data?.churches.find((church) => church.slug === slug) || {};
 
   useEffect(() => {
     // if we don't have the logged in state, do nothing
     // we're still showing the splash screen
-    if (isLoggedIn == null || !client) {
+    if (isLoggedIn == null || !theme) {
```
vinnyjth, line 27: This scares me a little bit :) 

I dislike tying app booting to the theme loading in. 

What if we defaulted theme to `APP.THEME`? We are still setting in in the `setup-app` script.

templates/mobile/src/auth/ProtectedRoute.js
```diff
@@ -1,21 +1,35 @@
...
       return;
     }
 
+    setChurch({ slug, theme: JSON.parse(theme) });
```
vinnyjth, line 31: 👍

templates/mobile/src/auth/ProtectedRoute.js
vinnyjth: Essentially I'm suggesting you 
- remove this check here
- keep the rest of this PR 
- on [this line](https://github.com/ApollosProject/apollos-apps/blob/4de733bb8371bb5e30e19fd9c2d6a212808ae180/templates/mobile/src/index.js#L149) default to `APP.THEME`

redreceipt: I know. I just see a clear future where we don't need a `theme.json` static file

vinnyjth: > they check for an internet connection and either say "sorry no connection!" or continue booting up.

I'm all for that, but if we're going to implement a "sorry no connection" screen, let's do that instead of delaying the splash screen from hiding. "App boots with no data" is an easier ticket to triage than "app never boots". 

And in better-case scenarios, I wouldn't want a slow connection or a slow response from the server to slow down the user seeing the splash screen hide. That "time to first paint" metric is really important to keep users engaged (https://web.dev/fcp/) 

I don't see the redundancy as bad at the moment; it's not holding us back in any way.

redreceipt: ok I'll revisit this idea later. I'll send another PR with the simple fix.

redreceipt:
timing isn't right

templates/mobile/src/auth/ProtectedRoute.js
vinnyjth: @redreceipt I can see where you're going - but I think having sane defaults on disk is a good strategy to mitigate risks :) You've shown how we can solve the problem of swapping themes when swapping churches, and I think we can backstop that approach with sane defaults for a better experience in production.

PR #2850, opened 2022-12-09
Upgrade react-navigation to v6

conrad-vanl:
## ✏️ Description

Fixes The Daily visual bugs such as dual-rendering, mainly by upgrading to the newest version of react-navigtaion.

## 🔬 To Test
1. Test in iOS - deep linking, tapping around, login, auth, etc
2. Test in Android - deep linking, tapping around, login, auth, etc

## 📝 Checklist


packages/apollos-ui-kit/src/ModalView/ModalButtons.js
```diff
@@ -18,24 +18,14 @@ const ModalButton = (props) => <StyledButtonIcon {...props} />;
...
-      navigation.navigate({
-        key: route.state.routes[route.state.index - 1]?.key,
-      });
-  }
-  if (!onPress) {
-    return null;
-  }
-  return <ModalButton name={'arrow-back'} onPress={onPress} {...props} />;
+  return (
+    <ModalButton name={'arrow-back'} onPress={navigation.goBack} {...props} />
```
conrad-vanl, line 22: One issue with this change is the back button always shows up, even if there's only one route in this stack. Like when you first open a content item. Before the back button would only show if you navigated to a child of that content item. Not sure if there's an easy fix / work around

conrad-vanl:
To anyone reviewing, here's the upgrade guide that was followed, may be helpful to understand any of the changes - https://reactnavigation.org/docs/upgrading-from-5.x

packages/apollos-ui-kit/src/ModalView/ModalButtons.js
calebpanza: I can spend a few minutes looking into this

templates/mobile/src/index.js
```diff
@@ -7,7 +7,6 @@ import {
   getStateFromPath,
 } from '@react-navigation/native';
 import { TransitionPresets } from '@react-navigation/stack';
-import { createNativeStackNavigator } from 'react-native-screens/native-stack';
 import 'react-native-gesture-handler'; // required for react-navigation
```
redreceipt, line 10: this line can be removed now, it's not needed in v6

packages/apollos-ui-connected/jest.setup.js
```diff
@@ -22,13 +22,20 @@ Animated.spring = (value, config) => ({
...
 });
 
-jest.mock('react-native-screens/native-stack', () => ({
+jest.mock(
+  'react-native-safe-area-context',
+  () =>
+    require.requireActual('react-native-safe-area-context/jest/mock').default
+);
+
+jest.mock('@react-navigation/native-stack', () => ({
```
redreceipt, line 31: seems odd we never had to mock this out but now we do

packages/apollos-ui-kit/src/ModalView/ModalButtons.js
```diff
@@ -18,24 +18,14 @@ const ModalButton = (props) => <StyledButtonIcon {...props} />;
...
+  );
 };
 
 export const ModalCloseButton = (props) => {
   const navigation = useNavigation();
-  const onPress = () => navigation.goBack();
+  const onPress = () => navigation.navigate('Tabs');
```
redreceipt, line 28: this won't be true for navigating back from content items in the liked feed in the user settings

templates/mobile/src/components/DailyHabitsTab/DailyHabitsTab.js
```diff
@@ -6,9 +6,10 @@
  * @see https://www.figma.com/file/YHJLj8pdFxWG9npF2YmB3r/UI-Kit-2.0?node-id=4454%3A68286&t=m7Ifzd5t3I97SmTy-4
  */
 
-import { useCallback, createContext, useContext } from 'react';
+import { useCallback, createContext, useContext, useMemo } from 'react';
 import { useFocusEffect, useNavigation } from '@react-navigation/native';
 import { useApolloClient } from '@apollo/client';
+import { useHeaderHeight } from '@react-navigation/elements';
```
redreceipt, line 12: `useHeaderHeight` is never used

packages/apollos-ui-connected/jest.setup.js
conrad-vanl: this is a new package we didn't have before.

I agree.... I ran into tons of weird gotchas with testing.

I think this becomes easier once we localize everything into the mobile project.

You're welcome to try removing some of these mocks and seeing if you can work through the errors in other ways.

templates/mobile/src/components/DailyHabitsTab/DailyHabitsTab.js
```diff
@@ -87,6 +88,13 @@ const DailyHabitsTab = ({}) => {
     } catch {}
   }, [scripture, today, navigation, cache]);
 
+  const contentContainerStyle = useMemo(
```
redreceipt, line 91: this should be an external stylesheet. using a memo here is probably cost negative since it's just returning an object. there's no compute intensive function for which the memo helps.

templates/mobile/__mocks__/@react-navigation/native-stack.js
```diff
@@ -0,0 +1,6 @@
+export const createNavigatorFactory = jest.fn();
+
+export const createNativeStackNavigator = () => ({
```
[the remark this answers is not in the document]
conrad-vanl: Similar to above, this is a new package we didn't have before, and we did have _some_ mocks for react-navigation before. I agree though, would be nice to get rid of them. We should be able to work through the errors without mocks, but I couldn't figure it out.

templates/mobile/src/components/DailyHabitsTab/DailyHabitsTab.js
conrad-vanl: 🤔 why is linter passing

templates/mobile/src/components/DailyHabitsTab/DailyHabitsTab.js
```diff
@@ -18,7 +19,7 @@ import {
 } from 'utils';
 import { ACTION_MAP } from 'features/FeaturesFeedConnected/FeaturesFeedConnected';
 
-import { RefreshControl, ScrollView, FlatList } from 'react-native';
+import { RefreshControl, ScrollView, Text } from 'react-native';
```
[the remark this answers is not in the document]
conrad-vanl: 🤔 why is linter passing

conrad-vanl:
@calebpanza I'm going to rely on you to implement @redreceipt's feedback to get this over the finish line. Just wanted to make that clear since I originally opened the PR

templates/mobile/__mocks__/@react-navigation/native-stack.js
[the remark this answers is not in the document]
redreceipt: I concede on mocking in `ui-connected` because it's going away so who cares. But why mock in templates when we think we won't shouldn't have to when it's localized. Are we trying to snap component trees that are too deep here? Said another way, maybe we're unit testing deeply nested component trees that include navigation when we don't have to be. And actually takes away from the effectiveness and maintainability of the unit test.

templates/mobile/src/components/DailyHabitsTab/DailyHabitsTab.js
redreceipt: because our rules allow it too. I'm just pointing it out in a code review

templates/mobile/src/components/DailyHabitsTab/DailyHabitsTab.js
redreceipt: that's true because the dependency list is empty. but memos with empty dependency lists should be external functions. that was my first clue ;)

calebpanza:
@conrad-vanl resolved the Modal Back button bug. Here's a video demo

https://user-images.githubusercontent.com/45076058/207147349-5c92aed0-9f91-4c2e-a5b2-a33b7faef51f.mp4

templates/mobile/__mocks__/@react-navigation/native-stack.js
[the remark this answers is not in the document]
calebpanza: @redreceipt I agree completely and unfortunately wasn't able to find a reasonable solution other than what you're seeing with these weird mocks. I feel like the issue here _might_ be in the way that this new Native component is rendering. Maybe the other one wasn't truly native or they had better jest setups installed in v5.

I feel like it's really a symptom of a greater issue with the current snapshots in which they're too broad to really prove anything about the components they're representing. Do you see this as release blocking, though? Could we document this as a future fix and come back to it later?

packages/apollos-ui-kit/src/ModalView/ModalButtons.js
calebpanza: Ah bummer... hmmm. Let me think about it a bit

templates/mobile/src/components/DailyHabitsTab/DailyHabitsTab.js
calebpanza: My bad. Thought I had removed it

templates/mobile/src/components/DailyHabitsTab/DailyHabitsTab.js
calebpanza: @redreceipt good catch! An artifact of some other testing I was previously doing with Header Heights. I'll resolve this now cause it really should be using Styled Components

templates/mobile/src/tabs/Tabs.js
```diff
@@ -88,20 +90,20 @@ const Tabs = ({ tabs }) => {
         component: createTab({
           type,
           featureFeedId: feed.id,
+          name: title,
           options: {
             headerLargeTitle: i !== 0,
-            ...(config.logo ? { headerCenter: HeaderWordmark } : {}),
+            ...(config.logo ? { headerTitle: () => <HeaderWordmark /> } : {}),
```
redreceipt, line 96: why was this changed to an anonymous function? the other two are still function definitions. And I think it's generally a no-no to have components as functions inside of other components because it will allocate memory for a new component on every render.

templates/mobile/src/tabs/createTab.js
```diff
@@ -45,24 +46,17 @@ const Tab = () => {
...
-  type,
-}) => {
+export const createTab = ({ featureFeedId, name, options, showTags, type }) => {
   const TabStack = createNativeStackNavigator();
+
   const TabNav = () => (
     <TabStack.Navigator
       screenOptions={{
-        headerHideShadow: true,
-        headerLargeTitle: true,
```
redreceipt, line 61: are we not using iOS large titles anymore or is it turned on by default in v6?

packages/apollos-ui-kit/src/ModalView/ModalButtons.js
calebpanza: Here we go! The changes are pushed now

https://user-images.githubusercontent.com/45076058/207154155-803b9b4a-db63-4cfe-8809-6295a705bbdd.mp4

templates/mobile/src/index.js
calebpanza: Done!

packages/apollos-ui-connected/src/DailyHabitsConnected/DailyHabitNavigationStack/DailyHabitsNavigationStack.js
```diff
@@ -6,7 +6,6 @@
  * Main Navigator for the DailyHabitsConnected
  */
 
-import React from 'react';
```
redreceipt, line 9: I believe `React` is still required inside our pkgs code because the version of RN they rely on still uses the old render logic. The eslint warning that always shows up is simply wrong I believe, probably due to the different eslint configs we have sprinkled around.

templates/mobile/src/tabs/Tabs.js
conrad-vanl: `headerCenter`/`headerTitle` works a lot differently in new version of react-navigation, and works differently then the other header props, it's kind of dumb. When it was just `HeaderWordmark` I was getting errors about "Object is not a function" -> I think. Or something like that. So....I made it a function. Might be worth trying again to confirm it's still needed, then just making it a named function at the top of the file or something..

templates/mobile/src/tabs/createTab.js
conrad-vanl: We are using them but I think they are set in a different spot now

packages/apollos-ui-connected/src/DailyHabitsConnected/DailyHabitNavigationStack/DailyHabitsNavigationStack.js
conrad-vanl: I didnt' get errors running the app or tests. Where would we show an error? I think my fix-lint on save in VScode automatically removed it.

templates/mobile/src/tabs/Tabs.js
calebpanza: https://reactnavigation.org/docs/elements/#headertitle

redreceipt:
wordmark on Android is off-center

Screenshot 2022-12-12 at 5 12 06 PM

redreceipt:
Also on Android, there's no title at all. I think it used to show up small where iOS showed up large.

Screenshot 2022-12-12 at 5 14 59 PM

redreceipt:
And the back button is there but it doesn't match the style of the close button which _I think_ this PR adjusts?

Screenshot 2022-12-12 at 5 16 02 PM

templates/mobile/index.js
```diff
@@ -1,6 +1,11 @@
 import ApollosConfig from '@apollosproject/config';
 import './theme';
-import { AppRegistry } from 'react-native';
+import { AppRegistry, Platform } from 'react-native';
+import { enableLayoutAnimations } from 'react-native-reanimated';
+
+// This is a hack to get around a bug in react-navigation
+// https://github.com/react-navigation/react-navigation/issues/10531
+enableLayoutAnimations(Platform.OS === 'ios');
```
redreceipt, line 8: is the purpose of this to fix the android header bug? it's hard to see fully on a sim but _I think_ it makes the iOS navigation animations weird. Could we only `enable(false)` on Android so the function isn't used on iOS at all and we just use whatever the default behavior is?

redreceipt:
Don't need done and back on profile screen

Screenshot 2022-12-12 at 5 20 47 PM

calebpanza:
> wordmark on Android is off-center

This is expected behavior since the navigate navigator changed the Title position on Android 13.

> Also on Android, there's no title at all. I think it used to show up small where iOS showed up large.

Because we're using the "large title" effect, the title will only show up on iOS. This is expected behavior with this upgrade

> And the back button is there but it doesn't match the style of the close button which I think this PR adjusts?

Interesting... I'll bet it's being auto-added by the OS. I'll look into that.

> Don't need done and back on profile screen

I think it's another artifact of my comment above. I'll check it out.

redreceipt:
Being off center is fine as long as the left most item lines up with the rest of the screen. In my screenshot above on the watch tab, there's too much space between the word mark and the left margin.

templates/mobile/index.js
conrad-vanl: @redreceipt's point may stand, a better way to do this with less chance of unattended consequences down the line may be:

`if (Platform.OS === 'android') enableLayoutAnimations(false);`

conrad-vanl:
> Also on Android, there's no title at all. I think it used to show up small where iOS showed up large.

It's only because there's a wordmark. It will show up on android if there's no wordmark. Pretty sure it works this way on master now too.... wordmark covers up the small item

> Being off center is fine as long as the left most item lines up with the rest of the screen.

@redreceipt this may just be how these android headers work on react-navigation. I'm not sure....we're not adding any padding/margin or anything.

@calebpanza do you know if react-navigation utilizes material 2 or 3 elements? 
https://m3.material.io/components/top-app-bar/overview
https://m2.material.io/components/app-bars-top

I would guess M2, and while yeah some of those screenshots use margins on top bars that are kinda whack.

Might be able to play with https://reactnavigation.org/docs/native-stack-navigator#headertitlealign

conrad-vanl:
> Don't need done and back on profile screen

I'd probably just hide the done button on android unless android subscribes to using "Done" buttons like iOS does. Done button is used there since it's a popover/modal on iOS and Done buttons are kind of a best practice

redreceipt:
> @redreceipt this may just be how these android headers work on react-navigation. I'm not sure....we're not adding any padding/margin or anything.

On tabs with the profile icon, it's all the way slid to the left and looks way better so I'm assuming there's no ghost padding for the word mark or titles either.

conrad-vanl:
> On tabs with the profile icon, it's all the way slid to the left and looks way better so I'm assuming there's no ghost padding for the word mark or titles either.

This is because that uses the "left" property.

The wordmark/title is in the "Title" property which gets rendered in the center / left aligned on android.

I think https://reactnavigation.org/docs/native-stack-navigator#headertitlealign would allow us to force center the wordmark/title even on android.

redreceipt:
I agree sticking with Android standard UI practices is best. But a “not-quite” left aligned title just seems off. Maybe the header left component need to be null or something for the title component to slide all the way over?Sent from my iPhoneOn Dec 13, 2022, at 9:02 AM, Conrad VanLandingham ***@***.***> wrote:﻿

On tabs with the profile icon, it's all the way slid to the left and looks way better so I'm assuming there's no ghost padding for the word mark or titles either.

This is because that uses the "left" property.
The wordmark/title is in the "Title" property which gets rendered in the center / left aligned on android.
I think https://reactnavigation.org/docs/native-stack-navigator#headertitlealign would allow us to force center the wordmark/title even on android.

calebpanza:
@redreceipt ah I see! Yes, I know exactly what you're talking about. I noticed it while @conrad-vanl and I were implementing the upgrade and decided that it wasn't a priority for that day, but I can resolve the issue now cause it def is a bit jarring.

Effectively, the photo isn't scaling down based on the size of the word mark, it's getting set explicitly. The image resize is set to `contain` which contains the image based on the center point of the container.

_Yellow is the full header container_

_Red is the image container_

_Green is the actual image_

CleanShot 2022-12-13 at 09 33 30


@conrad-vanl if the package is truly rendering native components, then on Android 12/13 it should be using M3 I would think. Either way, if we're going all in on native, then we should just let the UI render itself imo. If we want to start changes OS conventions, then we should just write our own navigation headers so they're consistent across all platforms.

conrad-vanl:
Ah, so the issue is that we're centering the image inside of a container that's wider then the image itself.

I'm wondering what the easiest/best solve will be. I don't think we can just remove `width` from the `<Image />` element. Actually! Since this is a local image and not a network image, I bet we can. Set height explicitly (say `height: 100%`) then don't set a width, and I _think_ that the image will automatically set its width.

If it proves difficult, I saw we center the title on Android only when we have a wordmark on the tab. Let text headers render natively.

calebpanza:
@conrad-vanl @redreceipt this is all set now!

| iOS | Android |
| --- | --- |
| Simulator Screen Shot - iPhone 14 Pro - 2022-12-13 at 11 23 35 | Screenshot_20221213-112122 |

redreceipt:
sweet. there's a couple other feedback items above: 
- only using `enableLayoutAnimations(false)` on Android
- back button rendering weird in content items
- back and done being redundant

redreceipt:
only thing left: 

207166828-aa706b98-8484-413d-85bd-c491d0c6c39b

calebpanza:
@redreceipt here we go

| 1st Item in Stack | 2nd Item in Stack |
| --- | --- |
| Screenshot_20221213-115935 | Screenshot_20221213-120002 |

redreceipt:
@calebpanza problem is android, here's the first item in the stack
Screenshot 2022-12-13 at 12 25 11 PM

redreceipt:
nice, I see your latest commit fixed it

redreceipt (approved):

calebpanza:
OH! Hahah. Sorry, I had committed the changes but didn't push them up and posted the comment pre-maturely. That's my bad! Lol

PR #2852, opened 2022-12-13
fix(search): set search query for android

dsilva609:
## ✏️ Description
- sets search query for android tv

## 🔬 To Test
1. Perform a search for content that doesn't exist
2. Result should show "There are not matches for {search term}"

## 📝 Checklist


jordanwade:
@dsilva609 do we need to add `data={state?.searchQuery === '' ? {} : contentItems}` to that too? I just noticed that is what we have in the `Search.js` file where we are using the `SearchProvider`.

dsilva609:
> @dsilva609 do we need to add `data={state?.searchQuery === '' ? {} : contentItems}` to that too? I just noticed that is what we have in the `Search.js` file where we are using the `SearchProvider`.

@jordanwade Good catch. Added that in

calebpanza (approved):

PR #2858, opened 2022-12-14
puts gesture handler back and upgrades to v2

redreceipt:
## ✏️ Description

Turns out it actually is still required with react-navigation. Plus always good to stay up to date. Also removes some custom Android native java code that isn't needed anymore on v2.

## 🔬 To Test
1. Boot app on iOS and Android and click around a bit
2.

## 📝 Checklist


calebpanza (approved):
This is looking good to me. I noticed an issue on the Search Feed, although I think we should just open up another PR to resolve cause I don't think this PR explicitly caused it.

PR #2866, opened 2022-12-15
feat: add recommended extensions

dsilva609:
## ✏️ Description
- adds a list of recommended extensions if you're using VS Code including
 - gitlens
 - github pull request
 - template string converter
 - markdown preview
 - code spell checker 

## 🔬 To Test
1. When you open up VS code and do not have some or all of the extensions installed, you will see a prompt to install them

## 📝 Checklist


redreceipt:
I don't love the idea of prescribing what tools or stack people should use. And I know that sentiment is unintentional but now it requires the team to be accountable for the tools we recommended. I don't know VSCode, is there not a way to set these on your machine outside the repo?

camrun91:
I would second what @redreceipt said. While I do use VS code and use some of the plugins/extensions mentioned, I don't necessarily love the idea of prompting people to install things they may not want to that are not required for them to work on the repo. 

I could see a pinned slack chat or basecamp post about what vscode plugins the team has found useful.

dsilva609:
> I don't love the idea of prescribing what tools or stack people should use. And I know that sentiment is unintentional but now it requires the team to be accountable for the tools we recommended. I don't know VSCode, is there not a way to set these on your machine outside the repo?

Yeah these are definitely just suggestions so I did keep the list pretty short and more targeted to the projects these will encompass. They can be ignored if not wanted as well.

calebpanza:
I'm not familiar with the entire list of plugins, but I think that suggesting plugins inside of the project itself is actually a really great idea. A Slack message or Basecamp post is easy to miss and not accessible to contractors (like Jack).

I wouldn't see this list as a permanent structure, either, so if someone doesn't like a plugin or it stop working, they just open up a PR removing/replacing it from the suggested list with a reasoning and it evolves over time.

Seeing that it's opt-in and inconsequential if you don't use VSCode, I don't see any harm with its introduction?

redreceipt:
I like a markdown file better kinda but the cultural thing that sticks out to me is assuming plugin suggestions are needed. Do we have a problem with folks not having good dev setups currently? Correct if I'm wrong but aren't the tools and plugins suggested based on RN dev and not necessarily Apollos? Shouldn't there be a level of expectation that in order to dev in the project, you need to have a basic understanding of RN dev and your environment? The heart behind it is an attempt to be helpful and I love that but I fear the opposite could be true if we attempt to do down the route of suggesting best tech tools to use. Are there examples of great OSS libraries out there that offer similar suggestions?

vinnyjth:
I think this is a great idea, and I might pare down the list. 

- They aren't required, just suggested. 
- They are a good starting point if you're unsure what tools you need to develop on the project. 
- They save you pain on simple issues - like the linter - when submitting your first PR. 
- Sure - you're taking responsibility for someone else's code - but we do that 2000x times with NPM and we manage to make it work.

.vscode/extensions.json
```diff
@@ -0,0 +1,16 @@
...
+  // See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+  // Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+  // List of extensions which should be recommended for users of this workspace.
+  "recommendations": [
+    "eamodio.gitlens",
+    "esbenp.prettier-vscode",
+    "github.vscode-pull-request-github",
+    "meganrogge.template-string-converter",
+    "shd101wyy.markdown-preview-enhanced",
```
vinnyjth, line 11: let's pull this one out - it's maybe a bit too adjacent :) 
```suggestion
```

vinnyjth (changes requested):
Let's takeout the markdown package for now :)

vinnyjth:
> Are there examples of great OSS libraries out there that offer similar suggestions?

NextJS is the first place I looked, and they have a set of recommended extensions. 
https://github.com/vercel/next.js/blob/canary/.vscode/extensions.json

.vscode/extensions.json
dsilva609: Yeah I'd agree with that one being a little too out there. It's is too specific to just writing out documentation.

vinnyjth:
> @redreceipt  The heart behind it is an attempt to be helpful and I love that but I fear the opposite could be true if we attempt to do down the route of suggesting best tech tools to use. 

I don't fully understand where you're coming from here. I'm not sure under what scenario recommending a set of extensions to get started would be unhelpful. If the tool is broken - that would certainly be true. Or if the tool required it's own set of learning to understand (like picking up `vim`) it would certainly be unhelpful. But I don't think that's the case here.

vinnyjth (approved):

redreceipt:
Yea...I don't disagree with being helpful for new devs, per se. Maybe my hesitation is just asking "is this the right place to put helpful suggestions?" Seems odd to me that every repo could have its own list of "helpful node and git plugins". Like if we think it's helpful here, is there a reason we shouldn't list the same plugins in ALL of our repos?

redreceipt:
regardless, I know @dsilva609 intention is to be helpful and there's certainly nothing wrong with that

calebpanza (approved):

jordanwade:
I think this is a great step in the right direction of starting to establish coding guidelines for VS code users. Especially, using prettier.

jordanwade (approved):

PR #2869, opened 2022-12-16
fix: Complete Daily Reading Interaction

calebpanza:
## ✏️ Description

As is stands today, the `COMPLETE` action gets triggered on the Content Single either when the user lands on the screen, or when they press a "complete" button.

For Daily Habits, our consumption of Content Single interactions is going to be a crucial piece of giving user's their correct "status". In order to compensate for the complexity, we're standardizing our interactions so that EVERY Content Item will store an `OPEN` interaction on load.

On the API, if a church has autocomplete set to true, the followup to an `OPEN` interaction will be `COMPLETE`. If the church has `autocomplete` set to `false`, there will still be a "complete" button that will store the `COMPLETE` interaction. 

On the API, whenever the `COMPLETE` action is triggered, we will do a check to see if the Content Item is today's Reading. If it is, there will follow a new interaction: `COMPELTE_DAILY_HABITS`

This front-end PR prepares for those API changes

## 🔬 To Test
1. Load up the app using Chase Oaks
2. Navigate to the Daily Reading
3. The checkmark should STAY
4. You should see a record in the dB inside the interactions table that says `OPEN`, `COMPLETE` and `COMPLETE_DAILY_HABITS`

## 📸 Screenshots
n/a

## ⚠️ To-do Before Merge
- [ ] There will need to be a Cluster update to enable this that @camrun91 will be working on

## 📝 Checklist


camrun91:
With this PR would this not mark the wrong item as completed for daily habits if a church changes the content item after the user loads the daily habits? 

Also this would keep the parent series that the daily habit is part of as being marked as in progress as that is triggered api side on the complete action as well. 

I dk if either of those are issues or not.

calebpanza:
@camrun91 for right now I don't think either of those are issue enough to keep the existing structure. The goal here is to take away as much friction as possible for a user to get their "completion".

I think all of the points you make above are really great to keep in mind once we have time to properly shape out the Daily Reading as its own standalone feature. But for now, this should be a fine enough solution.

templates/mobile/src/content/ContentSingleConnected/ContentSingle.js
```diff
@@ -25,6 +28,8 @@ const ContentSingle = (props) => {
     `,
     { variables: { nodeId } }
   );
+  const { scripture } = useDailyHabitsExperiences();
```
redreceipt, line 31: same as comment below. this is adding secret queries to all content items in the app, regardless of daily habits

calebpanza: If the church doesn't have a daily scripture set up, then nothing will trigger. 

Couple other notes:
- while we're currently piloting the feature with a handful of churches, Daily Habits is likely to become a defaulted tab on _all_ Apollos apps, so I don't see much risk in having this defaulted across all Content Items.
- as a fast followup, we will be shaping out a more explicit "Daily Reading" screen and experience that will live outside of the scope of the Content Single, so this is a stop-gap while we pilot the feature

templates/mobile/src/content/ContentSingleConnected/ContentSingle.js
```diff
@@ -1,5 +1,6 @@
-import { gql, useQuery } from '@apollo/client';
+import { useEffect } from 'react';
```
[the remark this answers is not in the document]
calebpanza: Oh my bad! Forgot to remove that

templates/mobile/src/content/ContentSingleConnected/ContentSingle.js
redreceipt: Ok I'm good with that

redreceipt:
Another high level question, why can't daily habits just function properly on complete whether it's by button or viewing depending on what the church wants?

redreceipt:
I am assuming churches with complete buttons would want daily habits to reflect button pushes.

calebpanza:
If a user opens up the Content Item from the Home/Discover/Watch tab, they should still get credit for their Daily Habits. Again, we need to flesh this out better to pull the Daily Reading into its own experience completely, but for right now our focus is on easy distribution of daily content and getting user's their Daily Completion status with as little friction as possible

redreceipt:
If they mean different things (one complete means "completed a challenge or item in series" and one means "completed a daily habit") then it feels like these should be two different interaction types. then you wouldn't have to conflate components on the front end to handle both situations

conrad-vanl:
Okay I have an idea. I agree with @redreceipt, and I don't like that this is adding a query to a very central view (I've already noticed that tab loading on all apps has slowed down a lot since we've made tabs persona-fied).

What if we did this:

Step One: Since Chase Oaks is the only app this is affecting right now, lets just add Completed buttons to their daily habit items. We can do this by filling in their Completed Button Text on their daily readings (we'll have to do this individually, we can do this for the next couple of days wroth, then tell Matt at Chase Oaks to do the rest) https://rock.chaseoaks.org/page/341?ContentChannelId=363

Step Two: On the frontend, change the default "COMPLETED" event that is fired when opening a content item to be "OPENED". Then, on the backend, add logic that triggers a additional "COMPLETED" event on that content item IF that content item does not have a completed button. The benefit here is that this actually fixes the special case with Chase Oaks, and we can turn it off, which means we can then remove these completed buttons on chase oaks. This also then allows us to use different logic for marking a content item as complete.

calebpanza:
@conrad-vanl talked through this with @camrun91 and we're gonna run with this.

tl;dr:

* Content Single will _always_ store the following interaction on Load: `OPEN`
* On the API, if a church has `autocomplete` set to `true`, the followup to an `OPEN` interaction will be `COMPLETE`
* On the API, the followup to a `COMPLETE` interaction will be to check if the Content Item is today's reading and will set `COMPLETE_DAILY_READING`
* On the front end, if `autocomplete` is set to `false`, there will be a "Complete" button rendered at the bottom of the Content Item

calebpanza:
I feel like we never truly landed on anything, here, so I'm gonna go ahead and close it as to not distract from relevant PR's

PR #2871, opened 2022-12-19
fix: Improvements to Daily Habit Streaks

calebpanza:
## ✏️ Description

This PR includes a couple of fixes to the Daily Habits Weekly Streak on the Daily Tab:
1. Update the range to be Sunday-Saturday (previously Mon-Sun)
2. Exclude future dates in the network request. This means that if we want to request Dec 1-7 but today is Dec 2, our network request will only ask for Dec 1 & 2

It's worth noting that a quick test on Postgres shows that requesting the full week takes about 336ms while requesting just 2 days (like in the example below) will take 269ms. Requesting just a single day took only 149ms. Not an insanely huge increase, but we're trying to squeeze out as much performance as we can get.

## 🔬 To Test
1. Load up Daily Habits

## 📸 Screenshots
| Screenshot | Description |
| --- | --- |
| Screenshot_20221219-145703 | First, note that the streak starts on Sunday and ends on Saturday. In this example, today is Monday, and so the network is only requesting Sunday and Monday |

## 📝 Checklist


templates/mobile/src/utils/useDailyHabitsStreak.js
```diff
@@ -6,14 +6,22 @@
...
+import drop from 'lodash/drop';
+import last from 'lodash/last';
 import orderBy from 'lodash/orderBy';
-import startOfWeek from 'date-fns/startOfWeek';
+
+import eachDayOfInterval from 'date-fns/eachDayOfInterval';
 import endOfWeek from 'date-fns/endOfWeek';
 import format from 'date-fns/format';
-import { getTimeZone } from 'react-native-localize';
+import startOfWeek from 'date-fns/startOfWeek';
```
redreceipt, line 17: why not combine the imports on all the `date-fns`?

templates/mobile/src/utils/useDailyHabitsStreak.js
```diff
@@ -60,29 +68,62 @@ function formatAsDateOnly(date) {
 }
 
 function useDailyHabitsStreak() {
-  const _startOfWeek = startOfWeek(new Date(), { weekStartsOn });
-  const _endOfWeek = endOfWeek(new Date(), { weekStartsOn });
+  // Local Variables
+  const _startOfWeek = startOfWeek(new Date());
```
redreceipt, line 72: maybe add a `const now = new Date()` and reuse instead of duplicating `newDate`. chances are slim, I know, but at midnight, duplication could cause problems

templates/mobile/src/utils/useDailyHabitsStreak.js
```diff
@@ -60,29 +68,62 @@ function formatAsDateOnly(date) {
...
   const timezone = getTimeZone();
 
   const query = useQuery(GET_DAILY_HABITS_STREAK, {
     fetchPolicy: 'cache-and-network',
     variables: {
       weekRange: {
-        startDate: formatAsDateOnly(_startOfWeek),
-        endDate: formatAsDateOnly(_endOfWeek),
+        startDate,
+        ...(startDate === endDate ? {} : { endDate }),
```
redreceipt, line 86: the `weekRange` variable can take a single date (if start and end are the same)?

templates/mobile/src/utils/useDailyHabitsStreak.js
```diff
@@ -60,29 +68,62 @@ function formatAsDateOnly(date) {
...
       'date'
     ) || [];
 
+  /**
+   * Once we have have a network response, we need to figure out how many days are between "today" and the end of the week
+   * We'll create an array of dates that represent future dates for this week and manually create the date objects
+   */
+  if (streak.length) {
+    const _endOfWeek = endOfWeek(new Date());
+    const lastInStreak = last(streak)?.date;
```
redreceipt, line 107: what does lodash.last do?

templates/mobile/src/utils/useDailyHabitsStreak.js
```diff
@@ -60,29 +68,62 @@ function formatAsDateOnly(date) {
...
+    const lastKnownDate = parseISO(lastInStreak);
+
+    const futureRange =
+      formatAsDateOnly(lastKnownDate) === formatAsDateOnly(_endOfWeek)
+        ? []
+        : eachDayOfInterval({ start: lastKnownDate, end: _endOfWeek });
+
+    streak = [
+      ...streak,
+      ...drop(
```
redreceipt, line 117: what does lodash.drop do?

templates/mobile/src/utils/useDailyHabitsStreak.js
```diff
@@ -60,29 +68,62 @@ function formatAsDateOnly(date) {
...
+        futureRange.map((date) => ({
+          id: `DailyHabit:FutureDate:${date.toISOString()}`,
```
redreceipt, line 119: this seems odd. are we building GQL objects from scratch here?

redreceipt:
I think some test coverage here would help green light the logical changes with high confidence. I'm not familiar enough with the context or the reasons for the changes to affirm otherwise.

templates/mobile/src/utils/useDailyHabitsStreak.js
calebpanza: Something that Drew mentioned at one point that date-fns and lodash don't tree shake and you save memory if you individually import. I can change it if this isn't relevant. It's just become a habit honestly

templates/mobile/src/utils/useDailyHabitsStreak.js
calebpanza: Oh wow that's a good catch. Yeah I'll update that

templates/mobile/src/utils/useDailyHabitsStreak.js
calebpanza: Yeah. So if you offer only a `startDate`, the query will return a single day. Right now the API throws an error if the start and end are both the same or if a non-date value is given for `endDate`. Probably a worthwhile patch on the API, but doing it this way got the front-end work done

templates/mobile/src/utils/useDailyHabitsStreak.js
calebpanza: Gets the last element in the array

templates/mobile/src/utils/useDailyHabitsStreak.js
calebpanza: Drops the first item in the array since it is "today" and that result is already provided by the query

templates/mobile/src/utils/useDailyHabitsStreak.js
redreceipt: oh nice. I just learned what tree shaking means.

looks like it was fixed in v2.17 https://github.com/date-fns/date-fns/issues/2207#issuecomment-782306914

templates/mobile/src/utils/useDailyHabitsStreak.js
calebpanza: In this case, the `id` from GQL is being used as the Key when rendering this data in a list. Since we're not fetching those future dates, we need to generate a unique id for them manually

calebpanza:
@redreceipt yeah that's not a bad idea. I've been slacking on unit testing GQL and integrated components which isn't good. Seems as good a time as any to start

templates/mobile/src/utils/useDailyHabitsStreak.js
calebpanza: That's really good to know! Makes imports MUCH simpler. Lol

calebpanza:
@redreceipt I went ahead and added a bunch of unit test. They explicitly mock all queries because the assumption here is that Apollo & React hooks are well tested. I don't need to test changes in network statuses, here, but rather test the logic that runs given the anticipated network states.

Comments galore to help make it really clear as to the intended outcome of the tests. And also resolved the tree-shaking matter.

templates/mobile/package.json
```diff
@@ -10,7 +10,7 @@
     "ipados": "react-native run-ios --simulator=\"iPad Air (5th generation)\"",
     "android": "if [ ! -f .env ]; then echo \"GOOGLE_MAPS_API_KEY=XXX\" > .env; fi && emulator -avd \"$(emulator -list-avds | grep Pixel)\" & react-native run-android",
     "lint": "eslint .",
-    "test": "jest",
+    "test": "TZ=UTC jest",
```
redreceipt, line 13: Why is this necessary?

templates/mobile/src/utils/__tests__/useDailyHabitsStreak.js
```diff
@@ -0,0 +1,192 @@
+import { useQuery } from '@apollo/client';
```
redreceipt, line 1: Can you add .test on the extension of this file so it matches. Also, I think that's the only way tests get picked up.

templates/mobile/package.json
calebpanza: If I don't have a consistent timezone, I can't accurately confirm date ranges. Running `yarn test` from the mobile directory without it would fail cause my local machine is America/New_York and it messes up the turn of days with where midnight lies

templates/mobile/src/utils/__tests__/useDailyHabitsStreak.js
calebpanza: Yeah sorry. Dumb oversight on my end.

redreceipt (approved):

PR #2875, opened 2022-12-20
simplifies PR template

redreceipt:
## ✏️ Description

The proposed template matches the shovel and cluster template. It's more concise and covers the basics necessary for the reviewer to know what's going on at a glance.

## 🔬 To Test

n/a

## 📸 Screenshots

n/a

## 📝 Checklist


calebpanza:
I'm having a hard time understanding the value add. If it helps, I can narrow down the things that have been particularly useful about the existing template so far:

1. Having a Table inside of the comments. I wouldn't say that the existing table template has to stay, but I have found it useful to just uncomment the table rather than creating it from scratch. Makes setting up PR's with lots of screenshots quicker
2. The "To-do before merging" section. Especially with the work that we've done on Daily Habits, there are often API branches that coincide with the work that we're doing on the frontend. This section offers a great gut-check for me personally

I understand that a more robust template requires more out of the individual setting up the PR, but isn't that the point? Even things that are trivial like checking off the boxes as the bottom of the review (imo) offer an intentional barrier to code-reviews that just remind us all that we're abiding by the same standards.

I'm not pushing/advocating that we keep the exact template that exists. If it's not useful, then it's not worth keeping. I just think that the proposed changes could swing too far to the opposite direction.

redreceipt:
That's strong enough for me :)

jordanwade:
My two cents, I believe this document should and can change to fit the entire teams needs. That might mean PR templates are slightly different for each code base. 

There are a few things that I think are really important to keep. First and always have clear Testing instructions. Second, the "To-do Before Merge" section is a life saver when you have other dependencies, it can be deleted if it isn't needed along with most other sections. Next, the screenshots section for any UI changes, I don't really care about the formatting. Lastly, a reminder or checkbox to document "new envs" or other setup changes.

I feel that as a PR author it is thoughtful to your reviewers/team members to take the time to do these things to speed up the review process. And in turn creates higher quality work that is less prone to mistakes. 

Also, when we add less experienced developers to the team, these types of things will help them understand what is expected from their code and help them avoid asking for reviews before PRs are ready.

Again, as long as the PR template accomplishes these things, I don't care how it is formatted.

PR #2886, opened 2022-12-22
chore: Update ButtonFeature.js

calebpanza:
## ✏️ Description
When debugging an issue in the Chase Oaks app with the "Complete" button on content items, I noticed a couple of things that component that could be cleaned up and figured I'd throw this together real quick. 

Update include:
1. Pulling the definition of the `ButtonWithAnalytics` out of the component
2. Pull out the query/mutation definition to be file-level definitions
3. Update the `refetchQueries` to explicitly include the variables and query (I've personally have had issues in the past with the reliability of just passing in the name of the query)
4. Memoize the `onPress` method
5. Couple file re-orgs and comments

## 🔬 To Test
1. Open up a Content Item that includes a Button Feature and make sure it works as expected
_I used the Chase Oaks app and tested the Content Item attached to their Daily Reading Experience on the Daily tab_

## 📸 Screenshots
n/a

## ⚠️ To-do Before Merge
n/a

## 📝 Checklist


templates/mobile/src/features/ButtonFeature.js
```diff
@@ -1,65 +1,76 @@
+import { useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { useMutation, useQuery, gql } from '@apollo/client';
 import { withTrackOnPress } from '@apollosproject/ui-analytics';
 import { Button, PaddedView, styled } from '@apollosproject/ui-kit';
 import { RockAuthedWebBrowser } from '@apollosproject/ui-connected';
 
+const ButtonWithAnalytics = styled(
```
redreceipt, line 8: any reason you're running this through the `styled` HOC? We wouldn't override something so generic anyway so it's really not doing anything here (except taking a performance hit).

redreceipt: Oh I see. You just moved it. Wanna remove that HOC while you're in there?

redreceipt: I'll approve regardless. still better than it was. removing that HOC would be icing on the cake.

redreceipt (approved):

templates/mobile/src/features/ButtonFeature.js
calebpanza: Oh sweet! I was gonna replace this with the new UI 2.0 Button but didn't cause of the named component in the HOC. If that's not needed, I'd prefer to try out the new `PressableButton` here, but maybe that's better in a separate PR?

redreceipt: I'm good with whatever you decide

calebpanza: I'm gonna take a stab at upgrading the component tonight. If it's more trouble than it's worth, I'll just scrap it and merge what's here

redreceipt: I'd do that in a separate PR so you can get another accurate review with different code

PR #2887, opened 2022-12-22
improves app deploy versioning

redreceipt:
## ✏️ Description

I've noticed a couple instances where in using our current method to override a version number, TV and mobile come out the same, sometimes jumping from v1.5.1 to v6.0.0. Also currently there's no way to set Android version numbers. I think it's appropriate the different OSs maintain different version numbers. It's also more correct to specify a version at deploy time since it will only be used once at most anyway. This will remove the need for the `INITIAL_VERSION` config variable.

Once this is validated, follow up work will be done to Bitrise and Retool so our Customer Success team can easily manage deploys and version numbers.

## 🔬 To Test
1. Deploy the demo app locally with `scripts/deploy-ios.sh internal <new version number>`

## 📸 Screenshots

|iOS|Android|
|---|---|
|Screenshot 2022-12-22 at 6 42 14 PM|Screenshot 2022-12-22 at 6 47 00 PM|

## 📝 Checklist


redreceipt:
cc @calebpanza @camrun91 @nlewis84 @tristanguckenberger this one will be important for everyone to be aware of. let me know if there's any questions on how it works.

calebpanza:
> I think it's appropriate the different OSs maintain different version numbers

We should probably also expand this option out to AndroidTV, FireStick and Roku. 

I'm a fan of keeping the app version numbers as consistent as possible, but I like the way that this flow puts that responsibility on the person deploying the apps. It'll keep us from being vulnerable to rare instances where we gotta patch 1 OS and they get offset from each other.

vinnyjth:
There are things about this approach I like - making it easy to push out new builds with a specific version in mind is nice - but I also like the forced consistency of persisting this data in the Config table. What do you think about expanding / revamping the `INITIAL_VERSION` approach, so it's more consistently applied and better named? 

My thinking is - if you are working on a new version of Newspring, you can bump the version once in the config and not have to worry about setting the correct version moving forward. 

it also opens the door to easily bumping the versions of all apps via a sql query, and then starting new builds for all apps (something we have talked about doing for a while now)

redreceipt:
> not have to worry about setting the correct version moving forward.

This PR has a set it once and forget it method too. But now there's not invalid data in the DB. We rarely need to mess with version numbers as it is for any church so this isn't something we'll have to think about _until_ deploy time once to reset for a church. Automated deploys are not a concern here because with either method, there's an interruption to the process to interject a new version number. And we'll always be transitioning apps off different systems like Subsplash or Rock or whatever so one number to rule them all is not really possible for us.

To make the current version work. We'd need to have in the DB:

```
NEWSPRING.DESIRED_IOS_VERSION
NEWSPRING.DESIRED_TVOS_VERSION
NEWSPRING.DESIRED_ANDROID_VERSION
NEWSPRING.DESIRED_FIRE_VERSION
NEWSPRING.DESIRED_ROKU_VERSION
```

And multiply that times however many churches we have. Plus after the first time, they are dead data. Instead of writing to a DB, it's just embedded in the API call to whatever service we use.

calebpanza:
@vinnyjth really interesting suggestion and something I've been thinking about this past few weeks. What about just simply storing an _Apollos Version_ in the dB that gets added on top of the client versions that were set when they entered Apollos.

**For example:** 

With the launch of Daily Habits, **Apollos** is now running v2.1.0 of its software.

So if NewSpring came into Apollos at **v6.0.0 for iOS** and **5.9.0 for Android** and Chase Oaks at **v4.4.10 for both**, we add those together (and reset the patch number in this case) so that:

NewSpring is now **8.1.0 for iOS**, **7.10.0 for Android**
Chase Oaks is now **6.5.0 for both**

---

With this approach, whoever is deploying the app can say "deploy major/minor/patch release of **Apollos**" from retool or whatever, and the system will just automatically apply the _Apollos_ software version on top of the church-platform existing version.

vinnyjth:
> To make the current version work. We'd need to have in the DB:

Can we do two versions? 

```
NEWSPRING.APP_VERSION
NEWSPRING.TV_VERSION
```

ideally we want to track iOS/Android version 1-1. If we ever need an emergency patch for both platforms, I think it's okay if the other platform skips a version. 


> And multiply that times however many churches we have. Plus after the first time, they are dead data. Instead of writing to a DB, it's just embedded in the API call to whatever service we use.

Having dead data after a build cycle feels better to me than putting the responsibility of knowing the correct version to build on all developers all the time.  If Caleb is working on shipping Chase Oaks, and then Nathan takes over, how we do we help guide Nathan on what version to specify in the build UI? 

Long term - we want to move to a fully automated build pipeline so pushing builds happens automatically on tagged commits (or similar mechanism). Do you see advantages of the approach you are suggesting in this PR given the future we hope to move towards?

redreceipt:
I'm trying to figure out why I like this. I think it comes down to DX. Currently when someone goes to cut a release, they have to "ask Michael or Vincent" how to set the version number. This forces us to make it an unavoidable part of releases. We could simply document the process but documentation is often a smell for a less-than-obvious procedure. We _should_ handle setting the config value in Retool regardless of the approach. This forces our hand to put it in the release UI somewhere. Hidden in a script flag is better than hidden in a config variable because both come through the script anyway. Now when we're doing ad-hoc releases locally, it's easier to set the version number.

vinnyjth (approved):
This is a step in the right direction and gives us more power and tools than we had before

PR #2888, opened 2022-12-23
fix: useDailyHabitsExperiences Default Values

calebpanza:
## ✏️ Description
The way that the `useDailyHabitsExperiences` hook was previously set up, it just returned whatever the server did (or didn't) return without any regard to nested properties within the objects for `gratitude` and `scripture`.

This PR introduces a new mapping structure to this hook in order to guarantee that there will _always_ be something returned for every property within each object. Some will return empty strings, others `null`. 

I also went through and added a bunch of unit tests to this hook as well in order to make sure that we catch errors like this more specifically moving forward.

## 🔬 To Test
1. Run the app and check out the Daily Experiences
2. `yarn test useDailyHabitsExperiences`

## 📸 Screenshots
n/a

## ⚠️ To-do Before Merge
n/a

## 📝 Checklist


redreceipt:
And in fact, it's actually less efficient looking at it because the gratitude components queries for prayer stuff and vice versa. Each component should only query for the data it cares about. That's how Apollo optimizes GQL the best way possible.

conrad-vanl:
It looks like all this PR does is add chaining to the already existing shared hook. So this PR probably does fix the issue that @calebpanza is setting out to fix. 

@redreceipt while it may be more fitting for a separate PR, could you give some more specific guidance on how to make more efficient?

calebpanza:
I would respectfully disagree. As we continue to expand on the offerings of Daily Habits throughout the app, small hooks like this allow for much leaner UI elements that don't need to worry about cache policies and data mapping, but rather explicitly rendering components. Having a separate file like this also allows for a much easier testing experience since we don't have to worry about component trees and virtual rendering

This pattern takes a lot of inspiration from the way that others at Di (Drew, Jordan, Ryan, etc) have approach projects in the past (like CF and Kingdom Resources) as well as the MVVM pattern adopted by Apple with the Swift/SwiftUI stack.

I do agree that `utils` feels like a funny place to house these kind of hooks and would love to introduce a root level directory called `hooks` for these kinds of queries and mutations.

redreceipt (approved):
It's preference and doesn't need to hold this up. This intends to add safety checks and does that.

calebpanza:
> And in fact, it's actually less efficient looking at it because the gratitude components query for prayer stuff and vice versa. 

@redreceipt in this case, that's actually by design. Lol. What I wanted to do is get data into the cache as quickly as possible. Sure, loading the first experience (Daily Habits Tab) might take a second longer to load, but since the data is now cached the Gratitude Experience doesn't have to worry about network load times in order to display data.

This isn't an approach that I would use everywhere, but given how small this query is, it was actually something that was specifically thought through. 

Now, I'm not gonna suggest this is a _perfect_ approach and will gladly break things out if this is proven to be detrimental to the product, but just wanted to call out that it wasn't an artifact, but rather an intentional decision.

PR #2889, opened 2023-01-04
fix: Update PrayerExperienceSwiper Keys

calebpanza:
## ✏️ Description

On the Prayer Swiper, we were placing component keys in the wrong place for both the Background Image and Profile Image. We are looping through the Prayer Requests and rendering components, but the `key` property was being set on the child and not the parent. This PR resolves that issue by moving the `key` props to the root level component rendered within the maps for both the Avatar and Background

## 🔬 To Test
1. Load up the prayer experience and make sure it still works

## 📸 Screenshots
Not much to show off since this is more of a best-practice fix

## ⚠️ To-do Before Merge
n/a

## 📝 Checklist


dsilva609 (approved):

vinnyjth:
@calebpanza This PR looks great, and ensures we are following best practices for React key usage, but it does not fix the prayer pagination issue. I wasn't sure if you intended this PR to be the fix, so just wanted to call that out. 


https://user-images.githubusercontent.com/1637694/210648274-e7986591-f518-437c-897d-b74af2ae1667.mp4

calebpanza:
@vinnyjth I gotcha! Yeah so this PR was _not_ intended to fix that. But rather I figured it was justified to get this up and out regardless of the prayer bug while I was trying to figure out replication steps. This was more an attempt at a "lean and mean" approach to PR's

calebpanza:
I should have commented here, as well, but I think I made that same comment inside of Basecamp

vinnyjth:
Awesome :D Just wanted to make sure.

PR #2892, opened 2023-01-04
chore: Relay Style Pagination Testing

calebpanza:
## ✏️ Description

The `relayStylePagination` cache policies are being used quite regularly at this point. There are over half a dozen instances of it being used. Overall, it's been serving us really well! But as it's being used more, it's opening us up to more and more risk if anything is broken or changed.

This PR looks to fill those gaps by:
1. Adding unit testing for the read policy
2. Adding unit testing for the merge policy
3. Adding unit testing for each of the utility functions
4. Moves the definition out of the `client.js` and into its own file
5. Expands the API of the policy generation to allow for additional cache keys to be defined without having to prop spread (dailyHabits)

## 🔬 To Test
1. run tests
2. run the app

## 📸 Screenshots
n/a

## ⚠️ To-do Before Merge
n/a

## 📝 Checklist


templates/mobile/src/client/cache.js
```diff
@@ -172,8 +62,7 @@ export const buildCache = async (serverURL, church) => {
           following: relayStylePagination(),
           followedBy: relayStylePagination(),
           communities: relayStylePagination(),
-          dailyHabits: {
-            ...relayStylePagination(),
+          dailyHabits: relayStylePagination({
```
redreceipt, line 65: this is strange to me. we're passing in an unused parameter to a function and then just returning right back. I think I liked how you were doing this before better.

calebpanza: Yeah that's fair. I was just trying to keep it so that every use of the `relayStylePagination` method was just called as a function and not break that convention up by requiring deconstruction for advanced cases. I guess that variable is kinda hidden, though, so that makes sense. I can revert that change.

redreceipt (approved):

PR #2893, opened 2023-01-04
fix(android): set background color for chip text in light mode

dsilva609:
## ✏️ Description
- fixes an issue on android where the background color of chip text doesn't match

## 🔬 To Test
1. Go to explore tab
2. Scroll down to the bottom with the campus tags
3. The text shouldn't have a white background inside the buttons
## 
📸 Screenshots
Before
image
After
image

## 📝 Checklist


templates/mobile/src/features/ChipListFeatureConnected/ChipListFeature/index.js
```diff
@@ -30,6 +30,7 @@ const ChipText = styled(({ theme }) => {
 
 const ChipButton = styled(() => ({
   borderRadius: 8,
+  backgroundColor: Appearance.getColorScheme() === 'light' ? 'lightgrey' : '',
```
calebpanza, line 33: I'm concerned that this is actually an error on the `Chip` component and not just the `ChipListFeature`. We also shouldn't use system colors in instances like this, but rather the Apollos theme colors

calebpanza (changes requested):

templates/mobile/src/features/ChipListFeatureConnected/ChipListFeature/index.js
dsilva609: Moved this over to `Chip` and it seems more consistent now.

image

calebpanza: This looks a lot better!

calebpanza:
Snaps are out of date

dsilva609:
> Looks good to me! I also noticed that the horizontal scrolling indicator is visible on the Chip List. Not required for this PR to get merged, but it would be nice to get that hidden on this component if you wanted to add that real quick

@calebpanza just pushed changes to hide that on the container

PR #2894, opened 2023-01-05
fix: Misc Fixes to usePrayerRequests

calebpanza:
## ✏️ Description

The following changes have been added in order to increase the stability and feature set of the `usePrayerRequests` hook including:
1. Spelling and grammar fixes in comments and variables
2. Complete test coverage
3. `prayerRequests` no longer returns `{ cursor: ..., node: ...}` and instead maps the `node` so the resulting array is just an array of prayer request objects
4. Updated the UI in respond to bullet number 3
5. Increase page size to 30
6. Forces `prayerRequests` to be unique

## 🔬 To Test
`yarn test usePrayerRequests`
Run the Daily Habits Experience to make sure that Prayer Requests are working

## 📸 Screenshots
n/a

## ⚠️ To-do Before Merge
n/a

## 📝 Checklist


templates/mobile/src/components/DailyHabitsTab/DailyHabitsTab.js
```diff
@@ -56,8 +56,12 @@ const DailyHabitsTab = ({}) => {
   const refetch = useCallback(() => {
     refetchDailyHabits();
     refetchStreak();
-    refetchPrayerReqeusts();
-  }, [refetchDailyHabits, refetchStreak, refetchPrayerReqeusts]);
+    // Insures that we refetch the first page of results
+    refetchPrayerRequests({
+      first: 10,
```
redreceipt, line 61: should this be 30? mildly uninformed question but seems like it fits

templates/mobile/src/utils/usePrayerRequests.js
```diff
@@ -64,7 +66,7 @@ export default function usePrayerRequests() {
 
   return {
     ...query,
-    prayerRequests,
+    prayerRequests: uniqBy(prayerRequests, 'id'),
```
redreceipt, line 69: shouldn't this be unique coming from the server? why does the app need to worry about sorting and filtering?

templates/mobile/src/components/DailyHabitsTab/DailyHabitsTab.js
calebpanza: AH. Yes. Yes it should

templates/mobile/src/utils/usePrayerRequests.js
calebpanza: Server-side errors/complexity in the current fetching could move a prayer request to a new position and end up incorrectly duplicating it across pages. There are longer-term solutions that are being talked about, but this change was asked for by @vinnyjth to catch it temporarily

templates/mobile/src/components/DailyHabitsTab/DailyHabitsTab.js
redreceipt: help me understand what the refetch is doing. why would we "refetch" the first page every time?

vinnyjth (commented):
I didn't review code, but I didn't experience any code glitches on my phone. 

Test praying for prayers (80+) swiping through prayers, and praying. All on iOS.

templates/mobile/src/utils/usePrayerRequests.js
redreceipt: This won't catch duplicates across pages. This only guarantees unique prayers on a given page. So if they are jumping around pages server-side, this won't do anything.

vinnyjth (approved):

templates/mobile/src/utils/usePrayerRequests.js
vinnyjth: @redreceipt It will b/c of how the hook combines with `fetchMore`. when `fetchMore` is run, the hook updates with a new (bigger) array of Prayers. 

Correct me if I'm wrong @calebpanza

I personally find that API confusing and hard to debug (not your problem @calebpanza, very much an Apollos thing!)

templates/mobile/src/components/DailyHabitsTab/DailyHabitsTab.js
calebpanza: Pull-to-refresh wasn't working. Instead, we opted to listen to the tab being focused being the refresh. And when the Daily tab becomes focused, we force a pull on all of our Daily Habits data to check for status/content updates

calebpanza: Pull-to-refresh needs to be restored, but this is got us to production as a less-but-better approach to the underlying problems

templates/mobile/src/utils/usePrayerRequests.js
calebpanza: This is playing in tandem with cache-merge policies. Yes, each page will be limited in it's results, but the cache is merging the existing results with the incoming, so this hook will build on top of the results and return _all_ prayer requests as a you request additional pages

redreceipt: OOO makes sense

PR #2895, opened 2023-01-05
feat(follow): update follow language

dsilva609:
## ✏️ Description
- Updates language to follow a user to better reflect flow

## 🔬 To Test
1. Go to individual prayer
2. Tap three vertical dot button
3. Should see `Request to follow [User]` in the menu

## 📸 Screenshots
image

## 📝 Checklist


dsilva609:
Updated it to say `Request to Follow [User]` but I'm not sure how much we should worry about sizing especially if there's much longer first name.

vinnyjth (approved):

calebpanza (changes requested):
@dsilva609 after seeing this in practice, I think that we'd be better suited using "Follow [user]" as the button text, but then updating the _alert_ to more accurately reflect the response from the server. If "FOLLOW" is sent back, "Great! You're following them" but if "REQUESTED" then "You've requested to follow them"

dsilva609:
> 

@calebpanza Swapped it back to the original button text and cleaned up the alert text to include the requester user name as well for more clarity.

calebpanza (approved):

PR #2896, opened 2023-01-06
Fix: content padding

camrun91:
## ✏️ Description
adding padding to show all content at the end of a content item. 

## 🔬 To Test
1. pull this repo
2. open a content item that has features and see all features

## 📸 Screenshots

Before change there are more items under journal prompt.
Screenshot 2023-01-06 at 2 35 06 PM

post change can see all features.
Screenshot 2023-01-06 at 2 31 37 PM


templates/mobile/src/content/NodeSingleConnected/index.js
```diff
@@ -22,16 +22,19 @@ import ContentNodeConnected from '../ContentNodeConnected';
 import GET_TITLE from './getTitle';
 
 const FlexedScrollView = styled({ flex: 1 })(Reanimated.ScrollView);
+const StyledView = styled(({ theme }) => ({
```
redreceipt, line 25: Don't use this styled HOC. Use a regular RN style sheet or y'all's new method. We found a performance hit with these old HOCs.

redreceipt (changes requested):

templates/mobile/src/content/NodeSingleConnected/index.js
calebpanza: Updated

calebpanza (approved):
This looks great and fixes the issues that we needed to resolve for Chase Oaks!

Just dropping this here as a reminder for myself to come back to this to think through a more robust solution to avoid large gaps as shown in the screenshot below.
CleanShot 2023-01-06 at 21 29 18@2x

PR #2897, opened 2023-01-06
Add user settings route

camrun91:
## ✏️ Description
This PR creates the route to allow direct navigation to the usersettings page. This will be used for new notifications for follow requests so people get sent to where they can see them. 

## 🔬 To Test
1. pull this pr
2. open safari in the sim. You can then put in the deep link with the route /usersettings this should open the app to the right page.


redreceipt (approved):

redreceipt:
@camrun91 would it be better to go all the way to the follow screen?

image

camrun91:
@redreceipt possibly I could not get that working and this was a quick fix so I could get the other stuff fixed for Chase Oaks.

if you have an idea how to make that work let me know.

redreceipt (approved):
This is fine for now, we'll probably want this one too. But long term we should point them directly at the follow screen.

PR #2900, opened 2023-01-10
feat(tv): removes progress value rounding for tv interaction

dsilva609:
## ✏️ Description
- progress value for saving watch state on tv would be rounded to the nearest 10 seconds
- this would cause unnecessary jumps in progress

## 🔬 To Test
1. Start a video
2. Let it play for more than 10 seconds
3. Pull up interactions for that video
4. progress should not always be rounded to nearest 10 seconds

## 📝 Checklist


calebpanza:
What's the harm in rounding to the nearest 10? If the interaction is only running every 10 seconds, wouldn't this just help to remove milliseconds from the data being passed into the server?

dsilva609:
> What's the harm in rounding to the nearest 10? If the interaction is only running every 10 seconds, wouldn't this just help to remove milliseconds from the data being passed into the server?

It would help with shaving off the milliseconds but it was also making it where having a timestamp in between 10 second intervals impossible. Say the video was paused at 15 sec, it would get saved at 20 sec. Should we worry about that upwards of 5 sec difference in watch states?

calebpanza:
That makes perfect sense. I didn't put it together that it was rounding _everything_

calebpanza (approved):

PR #2901, opened 2023-01-11
feat(video): add interactions for pausing, rewinding, and exiting videos

dsilva609:
## ✏️ Description
- adds interactions to handle watch states when a user
 - rewinds
 - pauses

## 🔬 To Test
1. start a video
2. pause (should create an interaction)
3. rewind (should create an interaction)

## 📝 Checklist


templates/tv/ios/Podfile.lock
```diff
@@ -434,8 +434,8 @@ SPEC CHECKSUMS:
   FBLazyVector: f78e9854b5205a3914cd64abd930eee3883dcf9b
   FBReactNativeSpec: 587b58901f40d6eeed47e1ec38f2f5d1076660a2
   fmt: 135c0c55547979f715b56dfa54037ececa96d07a
-  glog: 1c24d0cb9571f9d51d9f9344be9ddadea3e624de
-  RCT-Folly: ea6909cf09054ba87b406a21b14ec6799ce77686
+  glog: 5b768715ae7bf8149e9d269089c22585cee27eb7
```
calebpanza, line 437: Any idea why the Podfile would have changed in this?

calebpanza: never mind. Just saw your commit

packages/apollos-tv-shared/components/VideoPlayer/VideoPlayer.js
```diff
@@ -209,8 +220,12 @@ function VideoPlayer(props) {
   // ------------------------------------------
 
   // :: Exit/Unmount effect
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  useEffect(() => handleVideoExited, []);
+  useEffect(() => {
```
calebpanza, line 223: So I _think_ the way that this should be handled would be:

handleVideoExited ==> `useCallback` with `currentTime` as as dependency
useEffect has `handleVideoExited` as a dependency

Something about this flow atm doesn't quite make sense to me

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
```diff
@@ -25,13 +26,27 @@ const ApollosPlayerConnected = ({
...
 
+  const videoDuration = data?.node?.videos[0]?.duration?.toFixed(0);
+
+  const [currentPlayhead, setCurrentPlayhead] = useState(
+    data?.node?.videos[0]?.userProgress?.playhead
+  );
+
+  const isVideoComplete = useCallback(
+    (elapsedTime: number) => {
+      return elapsedTime?.toFixed(0) === videoDuration;
```
calebpanza, line 37: maybe `>=` just in case something gets mismatched for some reason. Makes it more exhaustive of a check

packages/apollos-tv-shared/components/VideoPlayer/VideoPlayer.js
dsilva609: goal here was to try to get the final interaction when the video unmounts, currently it triggers too often instead of just once

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
dsilva609: good catch 👍

packages/apollos-tv-shared/components/VideoPlayer/VideoPlayer.js
calebpanza: Oh I see. So basically the equivalent of "componentDidUnmount"? But wouldn't that trigger if the screen caused a rerender for some reason? The users didn't actually "exit" the video in that case and we could end up with messy data?

dsilva609: Yeah that's what I was trying to do. Looking into it more, this might be something that the parent component would have to handle like if the video content `isFocused` or not and just save the interaction at that point

calebpanza: We could listen to "on app background" and "on navigation" and it would accomplish the same thing, right?

calebpanza: We could also cut this from scope if it's too much. We'll always be covered within a 10 second window of where the user left off even if we don't store play head on exit and I'm fine with that as an option too

dsilva609: Yeah this specific watch state might be a little more involved than I initially thought. I'd be okay will cutting this one out if it does become too much.

dsilva609: removing it for now as it's not behaving as expected

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
```diff
@@ -25,13 +26,27 @@ const ApollosPlayerConnected = ({
...
+    (elapsedTime: number) => {
+      return elapsedTime?.toFixed(0) >= videoDuration;
+    },
+    [videoDuration]
+  );
+
   const handleOnPlay = useCallback(
     ({ id, elapsedTime, title }) => {
+      setCurrentPlayhead(elapsedTime);
       interactWithVideo({
```
calebpanza, line 45: At the definition of this hook, we should optimistically respond from the server and then manually update the local cache with the playhead information wherever we're querying for user progress (I think on the content item).

If you need examples of where to do this, I just recently implemented this on the [useCompleteNode hook](https://github.com/ApollosProject/apollos-apps/blob/master/templates/mobile/src/utils/useCompleteNode.js)

calebpanza (changes requested):
@dsilva609 Fast Forward and Rewind are _not_ working as expected on mobile.

1. Play video --> interaction
2. Press Fast Forward --> NO interaction
3. Press Rewind --> NO interaction
4. Pause --> interaction

I would expect an interaction to be written for steps 2 and 3

On mobile, there is no interaction being triggered for the 10 second interval

1. Play video --> interaction
2. Allow 10 seconds of the video to play --> NO interaction
3. Allow more seconds of the video to play --> NO interaction
4. Video completes --> interaction

Again, for steps 2 and 3 I would expect an interaction to be triggered

dsilva609:
@calebpanza added an `onProgress` event to the mobile video component. Seems like that will be able to start handling playhead changes.

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
dsilva609: Would this be to help handle showing the correct progress bar percentage?

calebpanza: This would help save us from having old data stored in the cache. It also keeps us from _having_ to re-query every time we want to know the user progress. Instead of `interactWithVideo` using `refetchQueries` in order to get the local cache state updated to reflect, we manually set it in good faith that our interaction was stored.

calebpanza: You make all network traffic effectively a background process at that point

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
```diff
@@ -86,6 +86,36 @@ const ApollosPlayerConnected = ({
...
+            nodeId: id,
+            action: 'VIEW',
+            data: { field: 'progress', value: elapsedTime },
+          },
+        });
+        return;
+      }
+
+      // Store the users' progress via an Interaction if enough seconds have elapsed
+      if ((elapsedTime - currentPlayhead).toFixed(0) >= 10) {
```
calebpanza, line 105: Let's get this `10` pulled out as a constant at the top of the file for easier access/visibility later on. Could even leave a comment that this is the default duration or something.

calebpanza (changes requested):

packages/apollos-ui-media-player/src/Container.tsx
```diff
@@ -200,6 +201,12 @@ const Container: React.FunctionComponent<ContainerProps> = ({
     [] // eslint-disable-line react-hooks/exhaustive-deps
   );
 
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
dsilva609: Can definitely be done. Noticed that there's a lot of `React.[something]` in this file

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
dsilva609: done

packages/apollos-ui-media-player/src/Container.tsx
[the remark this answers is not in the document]
dsilva609: done

packages/apollos-ui-media-player/src/Container.tsx
```diff
@@ -200,6 +201,12 @@ const Container: React.FunctionComponent<ContainerProps> = ({
...
+    const { elapsedTime, totalDuration } = playhead;
+    onProgress({ elapsedTime, totalDuration, ...analyticsMeta });
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
dsilva609: fixed and updated the other useEffects as well

packages/apollos-ui-media-player/src/Container.tsx
```diff
@@ -200,6 +201,12 @@ const Container: React.FunctionComponent<ContainerProps> = ({
     [] // eslint-disable-line react-hooks/exhaustive-deps
   );
 
+  React.useEffect(() => {
+    const { elapsedTime, totalDuration } = playhead;
+    onProgress({ elapsedTime, totalDuration, ...analyticsMeta });
```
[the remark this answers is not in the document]
dsilva609: @calebpanza  wrapped the `onProgress` call in a try catch and wrapped that in a 1 second throttle

packages/apollos-ui-media-player/src/Container.tsx
```diff
@@ -173,41 +179,52 @@ const Container: React.FunctionComponent<ContainerProps> = ({
...
+    if (autoplay) {
+      onPlay({ elapsedTime, totalDuration, ...analyticsMeta });
+    }
+    return () => {
+      return onEnd({ elapsedTime, totalDuration, ...analyticsMeta });
+    };
+  }, [analyticsMeta, autoplay, onEnd, onPlay, playhead]);
+
+  const throttledOnProgress = useCallback(
+    throttle((elapsedTime, totalDuration) => {
```
calebpanza, line 207: I see a warning here prompting an inline function to be used instead of passing `throttle` directly. Something about React not being able to handle the dependencies properly without. I think if you did `() => throttle(...)` that would do it. I also wonder if you even need the `useCallback` along with the throttle. Might be redundant and cause some issues?

calebpanza: I think we could probably get away with this being every `500` milliseconds

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
```diff
@@ -1,13 +1,17 @@
...
+import { ApollosPlayerContainer } from '@apollosproject/ui-media-player';
+import { throttle } from 'lodash';
+import PropTypes from 'prop-types';
+import React, { useCallback, useEffect, useState } from 'react';
+
 import GET_MEDIA from './getMedia';
+import INTERACT_WITH_NODE from './interactWithNode';
 
 const Noop = () => null;
+const INTERACTION_THROTTLE_SECONDS = 1000 * 1;
```
calebpanza, line 13: Could just name it "MILLISECONDS" and not have to worry about the multiplication

packages/apollos-ui-media-player/src/Container.tsx
```diff
@@ -1,49 +1,60 @@
...
-  React.useEffect(() => {
+  const ref = useRef(value);
+  useEffect(() => {
     ref.current = value;
   }, [value]);
   return ref.current;
 };
 
-const Container: React.FunctionComponent<ContainerProps> = ({
+const PROGRESS_THROTTLE_SECONDS = 1000 * 0.5;
```
calebpanza, line 55: Same thing here about naming

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
dsilva609: I can update that. My thought was to try to keep a consistent unit of time with all the constants

packages/apollos-ui-media-player/src/Container.tsx
dsilva609: Tried a few different ways if calling the throttle and it seems that only wrapping it a callback like this works at the moment. Having it inside the `useEffect` had it so it never triggered, having it just as `throttle(...)` strangely made it execute more often.

packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
calebpanza: Yeah I feel that. I wouldn't worry about that, though. The trottle method expects milliseconds, so let's just go with that

calebpanza (approved):

packages/apollos-ui-connected/src/ApollosPlayerConnected/useSetVideoProgress.js
```diff
@@ -0,0 +1,53 @@
+import { useRef } from 'react';
+import { useMutation } from '@apollo/client';
+import { throttle, isNumber, round } from 'lodash';
+
+import INTERACT_WITH_NODE from './interactWithNode';
+
+const INTERACTION_THROTTLE_MILLISECONDS = 1000;
+
+export default function useSaveVideoInteraction(videoId) {
```
dsilva609, line 7: leaving a reminder to rename this to match the file name

calebpanza: resolved!

calebpanza (approved):
@dsilva609 once you do another quick gut check, I'm good to get this merged!

packages/apollos-tv-shared/hooks/useVideoMediaProgress.js
```diff
@@ -0,0 +1,32 @@
...
+
+function useVideoMediaProgress(options = {}) {
+  const query = useQuery(GET_VIDEO_MEDIA_PROGRESS, {
+    fetchPolicy: 'cache-and-network',
+    ...options,
+  });
+
+  return {
+    userProgress: query?.data?.node?.userProgress || {
+      playhead: 0,
```
dsilva609, line 25: @calebpanza is this intended to be 0?

calebpanza: Only if there is no value given. In the case that the _entire_ `userProgress` object is missing from the resolved data, at least we have a valid object to avoid missing property references

PR #2904, opened 2023-01-13
locks pkg to node 18

redreceipt:
## ✏️ Description

`canvas` which we use to create icons and splash screen requires node <= 18

## 🔬 To Test
1. install node 19, yarn, watch it fail
2. install node 18, yarn, watch it pass

## 📸 Screenshots

n/a

## 📝 Checklist


calebpanza:
@redreceipt does this indicate that we should all be using node v18 locally? Mine is currently on like 16.x.x.

redreceipt:
I think 16 is fine there. I ran into another issue last night where I had to bump down to 16 to get something working but I can't tell if node 18 was the actual problem. If we deduce that 18 doesn't work, we can change it to 16. this PR just says "you absolutely can't use 19"

calebpanza:
Along with @dsilva609 why would we not add _additional_ constraints to this PR to just specify the same node version for everyone to be using? More constricting, sure, but feels like it would cast a much greater safety net for the whole team

redreceipt:
Mainly because I don’t know if 18 works or not so I don’t know what to set both to. Sent from my iPhoneOn Jan 13, 2023, at 12:31 PM, Caleb Panza ***@***.***> wrote:﻿
Along with @dsilva609 why would we not add additional constraints to this PR to just specify the same node version for everyone to be using? More constricting, sure, but feels like it would cast a much greater safety net for the whole team

calebpanza:
So why not just lock in at 16 since we know that works and then upgrade everyone to 18 at the same time?

Not meaning to by nudgey or nit-picky, just wanna make sure I fully understand the problems faced here

redreceipt:
Because saying 16 is required to install packages in the root when 18 works fine is incorrect.

calebpanza (approved):

PR #2906, opened 2023-01-16
Add explicit IDs to mutations

conrad-vanl:
## ✏️ Description

This adds explicit IDs to all mutations (where we can). The main benefit to this is with Stellate's [automatic cache invalidation](https://docs.stellate.co/docs/automatic-cache-invalidation-via-mutations).

## 🔬 To Test
1. Interact with nodes... see that IDs are being returned

## 📸 Screenshots

Proof that this works: https://www.loom.com/share/405667576e6949be9013dd8c0a9db2cc

## 📝 Checklist


calebpanza (commented):
I've recently had issue with the `interactWithNode` endpoint returning a node of type `Media`. Since `Media` is an abstract type, the server can't resolve it without more information, but our dB has a ton of rows of type `Media` instead of `VideoMedia` or `ImageMedia`. Because of this, I recently removed _most_ references of `node` from this mutation intentionally. If we need/want to enable this, we'll likely need to coordinate some API work before merging.

conrad-vanl:
@calebpanza not sure I quite follow, yeah this PR is really needed for stellate to work correctly. @vinnyjth might need to look at @calebpanza's comment.

conrad-vanl:
Screenshot 2023-01-22 at 12 25 44 PM

it does appear we have a lot of "Media" that uses the abstract type. I'm surprised this isn't causing other errors, as we should not be using abstract types. I imagine this is a shovel thing?

calebpanza:
@conrad-vanl yeah the exact issue you posted above is causing this error. Idk if that's a shovel thing or what, but when we resolve videos and/or images from places like content items, I'll bet we're explicitly specifying the type in the request which overrides the native type that we're seeing in the dB. Shouldn't be too rough of a migration, I would think. Each resource should have an extension on it that we could identify

vinnyjth:
It's a shovel thing :) one of our dags is writing this bad data. Can you plop a todo into the F&I list?

Thanks!

calebpanza:
Yeah that sounds good. I'll do that now. Do you think that's a this week kinda thing? If this PR gets in, we can launch Stellate as a part of the Community/Watch States work. If not, we won't include that in the scope of work

conrad-vanl:
@calebpanza are you saying there's a side affect right now that should prevent this PR from being merged? Or, merge this PR, but make sure to follow up by fixing our DAGs?

camrun91:
I think what he was seeing was that the API is unable to resolve the node type in order to resolve the id as the type is an abstract type.

conrad-vanl:
Yup @camrun91 there should be no items in our Database of `apollos_type = Media`. Needs to be an explicit Media Type.

vinnyjth:
Fixed the shovel media issue here: https://github.com/ApollosProject/apollos-shovel/pull/325

calebpanza:
@conrad-vanl I'm saying this should be considered blocked until the shovel gets merged _and_ migrations run on all existing items in the dB. Then this should be totally safe to merge

conrad-vanl:
https://github.com/ApollosProject/apollos-shovel/pull/325 has been merged.

Do we need to run backfills?

camrun91:
@conrad-vanl @vinnyjth and I just got the migration to run this should be good to go now I think

PR #2907, opened 2023-01-16
Create `ScriptureExperience` route and trigger `COMPLETE_DAILY_READING` on front end

conrad-vanl:
## ✏️ Description

This moves tracking the `COMPLETE_DAILY_READING` mutation to the front end, which would solve the button dillemmas we're having with Chase Oaks.

It also introduces a dedicated `ScriptureExperience` route and tees it up similar to Gratitude and Prayer. Long term, the UI/UX for the Daily Reading is going to change, and this gives us the component to go ahead and do it in.

The alternative to this approach might be to introduce a `VIEW` interaction that's triggered when any content is opened, and turn off autoComplete (essentially) across all apps. Then, key the `COMPLETE_DAILY_READING` side-affect off of `VIEW` instead of just `COMPLETE`.

The pro to having the frontend control the "COMPLETE" trigger is that as we adjust the ScriptureExperience UX we can move the trigger around in a non-breaking way for apps that haven't upgraded yet.

## 🔬 To Test
1. Open daily reading and see that it opens and navigates, and COMPLETE_DAILY_READING is logged in DB

## 📝 Checklist


calebpanza (approved):
Works for me. Until we get this rolled out to all churches _and_ in production _and_ an API build is pushed up, this is gonna cause duplicate records in the dB. There will also be a period of time in which non-upgraded users will no longer be able to complete the Scripture Experience

PR #2908, opened 2023-01-17
turns on amazon deploys

redreceipt:
## 🔬 To Test
1.
2.

## 📝 Checklist


redreceipt:
currently the plugin doesn't work right. we can either debug the fastlane plugin and fix it for them or create a new script that does the APK uploading without the plugin

PR #2910, opened 2023-01-18
chore: Enable Stream in Deployment Script

calebpanza:
## ✏️ Description

There's a configuration option within the code to allow us to specify whether Stream should be enabled _per deployment_. The problem is that the deploy script was never checking the dB for a value, so the value would _always_ be unset and never allow us to test Stream.

This also updates the Stream Provider to handle the string value coming in from the env more strictly.

## 🔬 To Test
1. Run the `setup-app.sh` script for `apollos_demo`
2. Confirm that `ENABLE_STREAM=true` is in the `.env`

## 📸 Screenshots
n/a

## ⚠️ To-do Before Merge
n/a

## 📝 Checklist


camrun91 (approved):

redreceipt:
Couldn't enable_stream be implied with the presence of a stream key and ID?

calebpanza:
The credentials are not client specific.

calebpanza:
Naw. We use 1 single Stream Configuration, but per-church enablement. This is also leftover from the "pull the plug" that we did a while back.

Once Communities is live, Stream will be automatically enabled for _all_ churches. We're just not there, yet

PR #2914, opened 2023-01-21
Switch to stellate

conrad-vanl:
## ✏️ Description

Stellate will serve traffic from https://cdn.apollos.app

Note: https://github.com/ApollosProject/apollos-apps/pull/2906 should really be merged first.
IT's not technically needed, but that PR will improve responsiveness on certain user interactions

## 🔬 To Test
1. Open app
2. Click around
3. 🎉

## 📝 Checklist


templates/mobile/src/client/index.js
```diff
@@ -22,9 +22,7 @@ const ClientProvider = ({ children, onAuthError, church }) => {
 
   useEffect(() => {
     const initialize = async () => {
-      const uri = (
-        Config.APP_DATA_URL || 'https://production.apollos.app'
-      ).replace(
+      const uri = (Config.APP_DATA_URL || 'https://cdn.apollos.app').replace(
```
camrun91, line 25: Are we wanting all new deploys to use this when we merge this? I was assuming that we would just set the `app_data_url` in the database? 

this does make it automatic and removes the need of a config var.

vinnyjth:
Fixed here: https://github.com/ApollosProject/apollos-shovel/pull/325

templates/mobile/src/client/index.js
conrad-vanl: apps don't currently support overriding `app_data_url`. 

So yeah, this PR is intended to do that.

redreceipt (approved):

PR #2918, opened 2023-01-24
feat(community): update gratitude input to match daily habit

dsilva609:
## ✏️ Description
- updates gratitude input in community to match the one in daily habits

## 🔬 To Test
1. Go to post a new gratitude in community
2. Should see updated input scren

## 📸 Screenshots

image
image

## 📝 Checklist


conrad-vanl:
Is there a reason we have separate Gratitude and Prayer inputs in community and daily? (I'm not saying it's a bad thing, just curious)

dsilva609:
> Is there a reason we have separate Gratitude and Prayer inputs in community and daily? (I'm not saying it's a bad thing, just curious)

@conrad-vanl the inputs on community were using the old design style so this change is meant to bring them up to the current one that's being used in daily

camrun91:
I think what @conrad-vanl is asking @dsilva609 is this: Is there a reason we are not using a shared component for these two screens or why the nav can't point to the same input?

conrad-vanl:
yeah, I'm asking @camrun91 's question

dsilva609:
> I think what @conrad-vanl is asking @dsilva609 is this: Is there a reason we are not using a shared component for these two screens or why the nav can't point to the same input?

I might be missing something here, but these are actually using the same shared input with the only difference is that Daily is using a full "experience" navigation style and in community it's just pulling out the single input section of the experience. We could make both community and daily have the full experience flow if we wanted though.

camrun91:
@conrad-vanl You know what the designs and stuff would call for.

calebpanza:
I can provide more context to this on Monday.

calebpanza (approved):

PR #2929, opened 2023-01-27
feat(community): add reply reaction to discussions

dsilva609:
## ✏️ Description

This PR enables User Replies to be **visible from the Community Feed**. When a Discussion Post has 1 or more replies added to it, a preview reply will display underneath the post along with a call to action requesting User Interaction.

# This PR does NOT
1. Allow a user to add a Reaction
2. Display a list of reactions
3. Allow the user to tap into a reaction

The reason for this is because there are 3 very specific UI cases that already have required quite a bit of code. In a Follow Up PR, we will enable the 3 things mentioned above

## 🔬 To Test
1. Open up the Apollos Demo App
2. Navigate to the Community Tab
3. Scroll through the feed to view the different reply interfaces

## 📸 Screenshots
**_!!! Important !!!_**
The following screenshots show all possible UI cases. Not all of these are currently live on the Community App. Instead, I manually manipulated the data in order to capture the screen shots. Please read the description for each to understand the logic behind each of these decisions

| Screenshot | Case | Description |
| --- | --- | --- |
| CleanShot 2023-01-30 at 17 27 20 | No Replies | When there are no replies on a post, we want to encourage the user to interact by offering a specifically call-to-action requisition their interaction |
| CleanShot 2023-01-30 at 17 38 17 | There is only 1 reply | When there is only 1 reply, the UI feels incomplete so only show a single avatar. The label doesn't make much sense when it says "See all 1 responses". We also want to encourage more interaction, and so we will hide the social bar and display the single reply along with a call-to-action encouraging the user to reply as well |
| CleanShot 2023-01-30 at 17 27 38 | There are 2 or more replies | When there are 2 or more replies, we will display a Social Engagement preview of the individuals who have submitted a reply with a label encouraging the user to view all of the replies to the post |

## ⚠️ To-do Before Merge
n/a

## 📝 Checklist


dsilva609:
@calebpanza Tested out this first phase and it's looking good 💪

calebpanza (approved):

PR #2946, opened 2023-02-01
chore: TV Auth Enablement Logic

calebpanza:
## ✏️ Description

Authentication is a core requirement for the TV product with the ability to override that field in the case of a Kid's Content app. The previous iteration of the code implemented auth as defaulted to "off" with overrides (in the env) required to enable it. In order to better reflect the business logic within the product, this PR reverses that logic so that Auth is _always_ enabled by default for all clients. This can be overridden by a Config Variable called `TV.DISABLE_AUTH` in the database

## 🔬 To Test
1. adjust the value of your local `env` and run the app
2. run the `setup-tv.sh` script for bayside and inspect the generated `env`. It should use `DISABLE_AUTH=false` as its value

## 📸 Screenshots
n/a

## ⚠️ To-do Before Merge
n/a

## 📝 Checklist


dsilva609 (approved):
@calebpanza  Looks good to me. Just leaving a note to make sure the config value gets added to Retool as well

PR #2951, opened 2023-02-02
prefer youtube media URLs

redreceipt:
## ✏️ Description

Chooses Youtube links first for video

## 🔬 To Test
1. Look at Bayside devotional videos, should be youtube

## 📸 Screenshots

Screenshot 2023-02-02 at 10 30 35 AM

## 📝 Checklist


packages/apollos-ui-connected/src/ApollosPlayerConnected/index.js
```diff
@@ -41,6 +41,7 @@ const ApollosPlayerConnected = ({
   nodeId,
   Component,
   audioOnly,
+  mediaSources,
   mediaSource,
```
redreceipt, line 45:
```suggestion
```

vinnyjth:
@redreceipt looking great - but there are some subtle bugs with existing code that will surface with these changes we might need to fix. 

These lines assume that the `videoId` and `videoDuration` come from the 0th video. We may need to apply the same selection strategy inside the `ApollosPlayerConnected` or (even better) consolidate some of this source/video selection to one of the components, maybe `ApollosPlayerConnected` shouldn't do any fetching on it's own? 

https://github.com/ApollosProject/apollos-apps/pull/2951/files#diff-dd8d85e421d3f1f939c3c28f8ebb6fb15ced20f0768a0c636d63559e0bd16892R51-R52

calebpanza:
This preference is gonna break tv... @conrad-vanl told me that we were in the process of removing all YouTube videos from Bayside. is that still happening?

redreceipt:
> This preference is gonna break tv... @conrad-vanl told me that we were in the process of removing all YouTube videos from Bayside. is that still happening?

TV will still use the native player

calebpanza:
That assumes we'll have both a native and YouTube file, though. Is that a safe assumption?

redreceipt:
@calebpanza I'll need to submit a PR to TV that filters out everything except HLS links

redreceipt:
ok @vinnyjth this is ready again. I removed the query from the `PlayerConnected` component (so it's not really connected anymore) and cleaned up the props to pass in all relevant data from the first time the query is run. All accompanying metadata comes from the preferred media now, YouTube or otherwise. I also cleaned up some unused props in the media player typescript components.

PR #2954, opened 2023-02-03
feat: Horizontal Media List Feature - Mobile

calebpanza:
# Please note that this PR is specifically for the Mobile App. TV will be a separate PR

## ✏️ Description

This PR introduces the `HorizontalMediaListFeature` into the `components` directory. Please note, this PR **does not** integrate the feature into the `FeatureFeed`. This PR does the following:

- Adds the `HorizontalMediaListFeature` to the `components` directory
- Renames the `MediaCard` to `MediaItem` to better reflect the design which is _not_ a card
- Adds a `useLivestreamIsActive` hook that accepts a `Livestream` object and locally manages reporting when the stream is live or not
- Adds a `LIVE` label to the `MediaItem`

## 📸 Screenshots and 🔬 Testing
**HorizontalMediaListFeature**
[Loom](https://www.loom.com/share/18be04b8814444a6988b4233ab947d4c)

To test this, set your `PROJECT_VARIANT` to `storybook` in your `env` and start the app. You can find a new story under `Components/HorizontalMediaListFeature` with a few demos.

**MediaItem**
[Loom](https://www.loom.com/share/3ab3b7f59b454e409422ba89185f3777)

Testing this is really just gonna be focused around making sure that imports are correct. Since the component has been renamed, that's the largest risk at play. To see the new `isLive` prop in action, the stories have also been updated accordingly

**useLivestreamIsActive**
[Loom](https://www.loom.com/share/98cbdce0710643c597a8c9866404b109)

To test this, read through the code and confirm that all potential edge-cases are taken into account. Tests are split up into a couple of different "groups", including a full simulation of a live event starting and ending. Read through the test file to make sure that the tests come to logical conclusions.

## ⚠️ To-do Before Merge
n/a

## 📝 Checklist


templates/mobile/src/components/HorizontalMediaListFeature/HorizontalMediaListFeature.js
```diff
@@ -0,0 +1,74 @@
...
+
+const HorizontalMediaListFeature = ({ title, subtitle, items }) => {
+  return (
+    <Styled>
+      {isNotEmpty(subtitle) ? <Subtitle>{subtitle}</Subtitle> : null}
+      {isNotEmpty(title) ? <Title>{title}</Title> : null}
+      <ItemList
+        hasTitle={isNotEmpty(title) || isNotEmpty(subtitle)}
+        data={items}
+        renderItem={renderItem}
```
redreceipt, line 54:
```suggestion
        renderItem={MediaItemWithLiveStreamManager}
```

templates/mobile/src/components/HorizontalMediaListFeature/HorizontalMediaListFeature.js
```diff
@@ -0,0 +1,74 @@
...
+  return !isEmpty(inp);
+}
+
+function renderItem({ item }) {
+  return <MediaItemWithLiveStreamManager item={item} />;
+}
+
+const MediaItemWithLiveStreamManager = ({ item }) => {
+  const isLive = useLivestreamIsActive(
+    item?.relatedNode?.__typename === 'Livestream' ? item?.relatedNode : null
```
redreceipt, line 31: could you just pass in the node and let the hook handle this logic? may need to extend it to other node types in the future

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
```diff
@@ -0,0 +1,62 @@
...
+      withTiming(0.65, { duration: ANIMATION_DURATION }),
+      -1,
+      true
+    );
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  return (
+    <LiveChipContainer entering={FadeIn} exiting={FadeOut}>
+      <DarkThemeProvider>
```
redreceipt, line 47: why do we have re-include this provider in the tree here?

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
```diff
@@ -0,0 +1,62 @@
...
+  useSharedValue,
+} from 'react-native-reanimated';
+import { DarkThemeProvider } from 'ui-kit/theme';
+import { unit } from 'ui-kit/utils';
+
+import Icon from '../Icon';
+import { SystemText } from '../typography';
+import { LiveChipContainer } from './MediaItem.styles';
+
+const paddingRight = unit(0.5);
```
redreceipt, line 23: why is this here outside of the component?

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
```diff
@@ -0,0 +1,62 @@
...
+    };
+  });
+
+  useEffect(() => {
+    opacity.value = withRepeat(
+      withTiming(0.65, { duration: ANIMATION_DURATION }),
+      -1,
+      true
+    );
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
redreceipt, line 42: why should this effect not fire if `opacity` changes (assuming it's some kind of memoized object)?

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
```diff
@@ -0,0 +1,62 @@
...
+  }, []);
+
+  return (
+    <LiveChipContainer entering={FadeIn} exiting={FadeOut}>
+      <DarkThemeProvider>
+        <Animated.View style={iconAnimatedStyles}>
```
redreceipt, line 48: does styled components not work with react-native-reanimated?

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
```diff
@@ -0,0 +1,62 @@
...
+        <SystemText small bold color="text.primary">
+          LIVE
+        </SystemText>
+      </DarkThemeProvider>
+    </LiveChipContainer>
+  );
+};
+
+LiveChip.propTypes = {};
+LiveChip.defaultProps = {};
```
redreceipt, line 60:
```suggestion
```

templates/mobile/src/ui-kit/components/MediaItem/MediaItem.js
```diff
@@ -12,22 +12,25 @@ import isEmpty from 'lodash/isEmpty';
...
 
-const MediaCard = ({
+const MediaItem = ({
   title,
   coverImage,
-  isInHorizontalList,
+  isInHorizontalList = false,
   label,
   progressBarProgress,
+  isLive = false, // todo : is there a better way to handle this that could apply styling for more chips? or should we leave `live` a separate entity?
```
redreceipt, line 33: wanna set these as default props instead?

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
calebpanza: In order to use the dark-mode text color, I have to nest another theme provider. This is similar to what we use to do with our ThemeMixin, but in a more "styled-component-y" way

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
calebpanza: So it doesn't get calculated on each render

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
calebpanza: Yeah it's causing an infinite loop if done that way. I originally tried that out cause of the linter warning.

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
calebpanza: They do, but the stylesheet created by re-animated needs to be generated using a hook they provide. In cases like these (where theme references aren't needed), I've found it easier to just negate styled-components and use straight up Reanimated.

templates/mobile/src/ui-kit/components/MediaItem/MediaItem.js
calebpanza: Per @conrad-vanl this is the more "React-y" way of handling defaults, now? I've always done it using `defaultProps` but updated this based on his comments.

templates/mobile/src/components/HorizontalMediaListFeature/HorizontalMediaListFeature.js
calebpanza: Not quite following. The only node type that will ever be "live" is the `Livestream` type, and we're passing in the entire `Livestream` node, here

templates/mobile/src/components/HorizontalMediaListFeature/HorizontalMediaListFeature.js
calebpanza: That's cool! I didn't realize you could pass the functional component reference in there. Updated!

templates/mobile/src/components/HorizontalMediaListFeature/HorizontalMediaListFeature.js
redreceipt: right, so instead of `HorizontalMediaListFeature` having to keep up with which node types are valid, the hook does. We only have to know about the shape of the relatedNode and act on it in one place, not two.

templates/mobile/src/ui-kit/components/MediaItem/MediaItem.js
redreceipt: I don't buy it. Why would there be a `defaultProps` field on a React component if it shouldn't be used? It's not deprecated.

redreceipt: @conrad-vanl explain yourself

calebpanza: I'll revert it if it gets this PR merged. Lol. I don't have a strong opinion about it

templates/mobile/src/components/HorizontalMediaListFeature/HorizontalMediaListFeature.js
calebpanza: OH! I see what you're saying, now. Ok yes. One moment

calebpanza: So sorry! That one took me a second

calebpanza: No other node types use the `start` and `durationInSeconds` so I think I can literally just kill that whole check

calebpanza:
@redreceipt lmk if this is good to go, now, and I'll merge and start working on TV

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
redreceipt: I'm confused. We use this unit function everywhere but we don't use global variables everywhere. Styles should only ever be computed once. That's the point of style sheets, do you don't declare global variables. Reanimated even recommends using a style sheets. We should too.

Screenshot 2023-02-07 at 4 49 19 PM

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
redreceipt: that tells me that opacity is changing often and we probably shouldn't ignore or else we'll have unexpected results here. that linter warning is there for a good reason.

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
redreceipt: We nest theme mixin when we want the theme to change, not to simply redeclare the theme. Why should we list the same context provider over and over. that defeats the purpose of react context.

templates/mobile/src/ui-kit/components/MediaItem/MediaItem.js
redreceipt: I do. I think we should use default props here.

templates/mobile/src/utils/useLivestreamIsActive.js
```diff
@@ -0,0 +1,78 @@
...
+  useEffect(() => {
+    if (!start || !durationInSeconds) {
+      return () => null;
+    }
+
+    /**
+     * Parsing ISO strings can be tricky and date-fns throws errors rather un-gracefully
+     * Instead of writing a bunch of conditions to take all parsing inconsistencies into account, we'll just use a try/catch to make sure we don't crash the whole app over the LIVE banner
+     */
+    try {
```
redreceipt, line 33: where are we expecting to get an error here? I don't like wrapping whole blocks of code with try/catch blocks because we miss unrelated errors down the line. We should identify where we think the error will come from, try to mitigate, and if we're still worried, only then wrap that single line in a try/catch

redreceipt: also in dev, we'd never know what was wrong or be alerted

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
calebpanza: This is the way that Styled Components recommends doing it. It's how we're handling light/dark switching and even with the Pressable(Button/Icon/Chip), each variant is its own "provider". I see what you're saying and wonder if it's a nuance in the language they chose to use, but their documentation recommends this approach: https://styled-components.com/docs/advanced#function-themes

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
calebpanza: You're right. I just added it back to the dependency list and it's working just fine, though. I guess early on there must have been something triggering a re-render or something but it's working just fine, now, without the linter exception

templates/mobile/src/components/HorizontalMediaListFeature/HorizontalMediaListFeature.js
calebpanza: Reverted this cause you can't render hooks outside of a functional _component_ and that renderItem method runs a function instead of rendering a component.

templates/mobile/src/utils/useLivestreamIsActive.js
calebpanza: fixed!

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
redreceipt: I just realized what `DarkThemeProvider` is. It's not really a provider in the sense that it's providing context. It's just overriding the theme. We're just forcing dark mode here. Shouldn't it be driven off the users device settings?

redreceipt (approved):

templates/mobile/src/ui-kit/components/MediaItem/LiveChip.js
calebpanza: This is something that I need to refine in the future, but not right now. From the system theme, the text color will switch back and forth, but I only want the nearly-white text color which lives exclusively in the dark theme.

conrad-vanl: said a different way is the chip itself is always "Dark Mode" regardless of device setting

PR #2958, opened 2023-02-04
only support HLS on TV

redreceipt:
## ✏️ Description

Only supports HLS video streams

## 🔬 To Test
1. Spin up any church TV app with videos and make sure they play
2.

## 📸 Screenshots

Screenshot 2023-02-03 at 8 10 40 PM

## 📝 Checklist


redreceipt:
@conrad-vanl I want to make sure to get your thumbs here. This assumes we will only ever support HLS video streams.

packages/apollos-tv-shared/components/VideoPlayer/VideoPlayer.js
```diff
@@ -51,7 +51,10 @@ function VideoPlayer(props) {
   const [progressTime, setProgressTime] = useState(0);
   const [paused, setPaused] = useState(false);
 
-  const videoMedia = props.parentNode?.videos[props.videoIndex];
+  // will find the first HLS video playlist provided
+  const videoMedia = props.parentNode?.videos.find((video) =>
```
calebpanza, line 55: I'm wondering if we should bake in something else that would catch if there are not videos and do something about it. If you look at the Content Single, it's determining whether or not the Video Player should render. So maybe the filter should happen at the parent level and allow the Video Player to assume the context of a valid video?

redreceipt: are you suggesting we don't even render the `VideoPlayer` if there are no videos?

calebpanza: Yeah pretty much. Either that or just not rendering the CTA that triggers the VideoPlayer

redreceipt: I could see that. But feels like it would live alongside this PR. Even in mobile, the video component figures out which video to play. And technically since this is TV, we should _never_ have content with no video attached so adding logic to selectively not show the video player seems futile.

calebpanza: Totally understand that. It's more of a catch-all in the case of faulty data. This way the app doesn't crash

calebpanza: Maybe less-but-better would be to log an error into Sentry and just not show the Video Player. Minimum code changes that way

redreceipt: `videoMedia` can already be null here and the app doesn't crash right? looks like that check is already in place.

calebpanza: Ah! I see it now.

calebpanza (approved):

PR #2961, opened 2023-02-06
Fix: change route to actually work.

camrun91:
## ✏️ Description
This PR just changes the route so that it works as expected when you click on the daily habit hero card for daily scripture. 

## 🔬 To Test
1. pull this branch 
2. run the app and open daily. 
3. use hero card to navigate to daily habits. 
4. see that all navigation works.


templates/mobile/src/components/DailyHabitsTab/HeroCard/InProgressCard.js
```diff
@@ -25,7 +25,7 @@ const InProgressCard = ({ title, subtitle }) => {
   function onPress() {
     switch (today?.nextExperience) {
       case 'scriptureExperience':
-        openScriptureExperience();
+        navigation.navigate(ScreenNames.ScriptureExperience);
```
vinnyjth, line 28: can please fix line 21 so it no longer references the unused var?

vinnyjth (commented):
Looks great! I think our linter isn't working right :)

templates/mobile/src/components/DailyHabitsTab/HeroCard/InProgressCard.js
camrun91: yep

vinnyjth (approved):

PR #2964, opened 2023-02-07
fix audio only.

camrun91:
## ✏️ Description

This PR swaps filter for find. Filter returns a shallow copy of the array find returns the first value in the array. 

## 🔬 To Test
1.pull the repo 
2.open an audio only item it will play. 

## 📸 Screenshots
fixed for woodmen
Screenshot 2023-02-07 at 3 05 18 PM


vinnyjth (approved):

camrun91:
@vinnyjth or @calebpanza it has been forever since I did tests on mobile. Does that look like a legit failure?

PR #2965, opened 2023-02-07
test/no-video-routing

dsilva609:
## 🔬 To Test
1. Go to a content series that has no videos in it (Item 5 in apollos_demo is a good one)
2. Press the hardware back button
3. It should send you back the the previous screen instead of exiting out of the app

## 📝 Checklist


vinnyjth:
@dsilva609 @calebpanza would it be better to solve this by not allowing you to navigate to items with no video? What's the value in visiting one of these items?

calebpanza:
@vinnyjth that's fine, but it still doesn't resolve the underlying issue and we'll have to figure this out at some point. Per our testing, the only way to get _any_ screen within this ContentSingle navigation stack to work properly is to render an input type (like TextInput) which means as we expand the feature offerings of TV, we're gonna run into the problem again

vinnyjth:
@calebpanza totally agree, would be great to know what's causing this! 

but even if we fix this we should really avoid navigating to a content with no video content :) hopefully we could avoid rendering it's card to begin with.

dsilva609:
Obsolete with newer pull request

PR #2972, opened 2023-02-09
fix(content-single): add height container for content single

dsilva609:
## ✏️ Description
- fixes issue where content single container was not the full height of the screen for android

## 🔬 To Test
1. Load up a video on Android tv
2. Should see bottom gradient on the screen

## 📸 Screenshots

image

## 📝 Checklist


packages/apollos-tv-shared/components/ContentSingle/ContentSingle.styles.js
```diff
@@ -16,6 +16,11 @@ import { system, utils } from '../../ui-kit';
 // safe to assume the values are fixed for TV devices.
 const window = Dimensions.get('window');
 
+export const Container = withTheme(styled.View`
+  height: ${window.height}px;
```
calebpanza, line 20: Can we not just flex: 1??

calebpanza (changes requested):

packages/apollos-tv-shared/components/ContentSingle/ContentSingle.styles.js
dsilva609: tried that and it still gives me the cut off looking image
image

calebpanza: And this is _only_ happening on Android? Does it change the iOS behavior?

Also what about height: 100%? Tryna avoid full-screen pixel values

dsilva609: Yeah only seems to change on android. Tvos looks to be ignoring the change and still works

calebpanza (approved):

PR #2979, opened 2023-02-16
Break out setup scripts

vinnyjth:
## ✏️ Description

Breaks out the setup scripts in a number of separate files. Same flow, but in a slightly different format. A couple of advantages.

- It's now much easier to debug step (you can run just the step you need)
- We can start to isolate Android/iOS specific steps. Saves us time (and money) on cloud-build platforms.
- Better separation of concerns encourages closer coupling of the why/how of build variables. 
- Enhanced ability to use Bitrise's caching without needing to rewrite or interupt the local flow. 

With this code merged, I'll be able to call each of the build steps independently on Bitrise and fine tune the build caching and omit stepson a per platform basis. 

Local setup/building should flow exactly as before but you'll see more logging. 

## 🔬 To Test
1. Run `./setup-app` as you would normally
2. Run `./build-ios` as you would normally 

## 📸 Screenshots
CleanShot 2023-02-16 at 09 51 18@2x

## 📝 Checklist


scripts/deployment/2_mobile_setup_deploy_folder.sh
```diff
@@ -0,0 +1,7 @@
+set -e
+
+echo "### 2. Mobile Setup Deploy Folder"
+echo "### Copying mobile template to deploys/mobile"
+
+rm -rf deploys/mobile
+rsync -av --progress templates/mobile deploys --exclude node_modules --exclude ios/Pods
```
vinnyjth, line 7: I migrated to `rsync` from `cp` because it's faster and can exclude folders we are deleting later on.

templates/mobile/ios/Podfile.lock
```diff
@@ -775,14 +775,14 @@ SPEC CHECKSUMS:
   Flipper-RSocket: d9d9ade67cbecf6ac10730304bf5607266dd2541
   FlipperKit: d8d346844eca5d9120c17d441a2f38596e8ed2b9
   fmt: ff9d55029c625d3757ed641535fd4a75fedc7ce9
-  glog: 85ecdd10ee8d8ec362ef519a6a45ff9aa27b2e85
+  glog: 5337263514dd6f09803962437687240c5dc39aa4
```
vinnyjth, line 778:
```suggestion
  glog: 85ecdd10ee8d8ec362ef519a6a45ff9aa27b2e85
```

templates/mobile/ios/Podfile.lock
```diff
@@ -775,14 +775,14 @@ SPEC CHECKSUMS:
...
   hermes-engine: bf7577d12ac6ccf53ab8b5af3c6ccf0dd8458c5c
   libevent: 4049cae6c81cdb3654a443be001fb9bdceff7913
   OneSignalXCFramework: 10648e60ff3d335fbffa62d9b44b700b56d49b51
   OpenSSL-Universal: 1aa4f6a6ee7256b83db99ec1ccdaa80d10f9af9b
   Permission-Notifications: 9c6b5cc4f0e6599e9fc3395b77cebddc48f1be41
   Permission-PhotoLibrary: 8227a6ed9f6a971537afe63742d54f5f23a38fe2
-  RCT-Folly: 803a9cfd78114b2ec0f140cfa6fa2a6bafb2d685
+  RCT-Folly: a21c126816d8025b547704b777a2ba552f3d9fa9
```
vinnyjth, line 785:
```suggestion
  RCT-Folly: 803a9cfd78114b2ec0f140cfa6fa2a6bafb2d685
```

scripts/deployment/2_mobile_setup_deploy_folder.sh
```diff
@@ -0,0 +1,7 @@
+set -e
```
redreceipt, line 1: can this step come before variables so we can put that file directly in the deploys folder? it doesn't belong and won't ever be used in templates.

scripts/deployment/2_mobile_setup_deploy_folder.sh
redreceipt: love that

scripts/deployment/2_mobile_setup_deploy_folder.sh
redreceipt: on second thought, I do think it should be separate because it's a different purpose but still think it should live inside deploys/mobile

vinnyjth: great feedback, I was thinking about this also 

one of my goals w/ these scripts is to bomb quicker if something goes wrong (hence all the `set -e`s). Not having valid a valid config is one of the first things that can go wrong, so I want to check for valid config before we do _anything_ else, (like copying over templates/mobile). I agree having the file in the root is a bit messy though, so here are some ideas - let me know which one you like most 

1. Create the file in templates/mobile, it will get copied over, delete it from `templates/mobile` after copy. 
2. Create the file in `/tmp`
3. Something else :)

redreceipt: I like creating the file in `tmp` out of those but then you have to figure out how to pass that tmp file in which I don't love. but honestly this script is so small and chance of failure is so small, I don't think it would be wrong to combine the two and just run this part first.

redreceipt: I mainly feel strongly about it because I think the templates folder should be purely for dev so if there's something in there with church specific variables, it can be confusing and cause problems later in dev

vinnyjth: Yah, I respect that a lot. 

I pushed up a version of bullet 1 from above 

> Create the file in templates/mobile, it will get copied over, delete it from templates/mobile after copy.

Let me know what you think, the script is setup in such a way that it will delete the templates/mobile/variables.env file even on a failure.

.gitignore
```diff
@@ -103,4 +103,6 @@ vendor
 **/.vscode/*
 !.vscode/extensions.json
 
-sentry.properties
\ No newline at end of file
+sentry.properties
+
+variables.env
```
vinnyjth, line 108: just in case

scripts/deployment/2_mobile_setup_deploy_folder.sh
redreceipt: why put it in templates in that case? could you put it in "deploys/temp" or something? even the root and then delete? avoiding the templates altogether seems appropriate

vinnyjth: I dropped it in templates b/c it get's pulled over automatically when we rsync

vinnyjth: @redreceipt That feels overly complex to me given the risk / reward of persisting the variables.env. 

What problem are we trying to solve? Devs getting confused about a potential variables file left over after an absolutely catastrophic script run (Right now you have a 5-second window to cmd+c in order to persist the file, and I think even cmd+c will trigger the file deletion code). 

What if we solved the ux another way by naming the `variables.env` file more clearly?

vinnyjth: In regard to the additional complexity - keeping track of a TEMP folder makes running the scripts one at a time much more complex in Bitrise / Github Actions b/c I have to transfer variables between actions. Not hard, but I really like the simplicity of the current setup.

redreceipt: I think I'm trying avoid future problems by not letting church variables leak into the templates folder. Right now this logic lives in three places. `setup.sh`, `1.sh`, and `2.sh`, and the `.gitignore`. if for some reason a dev comes along at any point and changes one of those files without fully understanding what this `variables.env` file is or the implications of changing the process, now we have a problem that's confusing at best, and insecure at worst. having a temp folder that handles deploys seems less complex and safer than our current process. It appropriately hides logic that devs should never see or care about.

redreceipt: you also won't have to hide the sentry.properties file. same exact conversation

redreceipt: you don't have to pass it in. I think you could even `export MOBILE_DEPLOY_DIR`

vinnyjth: The logic realistically lives in 1.sh and 2.sh. The .gitignore and setup.sh are protection against a user error :) 

I get your concern, though; I could reintroduce the lines of code in 1.sh that delete `variables.env` at the very beginning. 

> export MOBILE_DEPLOY_DIR

I don't think that works across Bitrise steps :/ you have to use their `envman` api. 
https://devcenter.bitrise.io/en/builds/environment-variables.html#setting-and-managing-env-vars-during-a-build

Using `temp` feels brittle and will make it more challenging to debug step by step b/c you'll have to identify the temp folder or work through the entire script step by step. Without temp you can pickup where the build failed (I see this as helpful when adding or changing the current steps)

redreceipt: why would we run scripts one at a time in bitrise?

vinnyjth: > why would we run scripts one at a time in bitrise?

a couple reasons 

- I don't want to run android steps on ios builds and vice versa 
- I want to handle the yarn/rm -rf node_modules a little differently (in a different order) so I can cache the node_modules

redreceipt: Ok back to my original suggestion, if you ran delete and copy deploys step first, you could put it there. I get you want to bomb out early, but the config step will only ever bomb out at most once. it won't be an exponential cost savings, it'll likely fail when we're setting a new app (after 2min instead of 1 min) and then we'll add variables and it will never fail again.

vinnyjth: okay I'm fine with that compromise

.gitignore
redreceipt: don't need any changes to this file now

vinnyjth: It doesn't hurt 🤷

redreceipt: and actually I'm not sure why we need to ignore sentry.properties here

vinnyjth: probs because people like me are klutz's and have a habit of dropping files all over the filesystem when testing / developing :)

redreceipt: it's the definition of dead code. it will never be used and when someone sees it, is confused why it's there.

vinnyjth: I think it's fine - we don't ever want to capture either file in git. I'll add a comment

redreceipt (approved):
❤️

calebpanza:
This is low key gonna be a huge win for local debugging, too! Thanks for all of this, guys! Really great stuff

PR #2982, opened 2023-02-17
fix(onboarding): add await and null checks

dsilva609:
## ✏️ Description
- fixes issue where verify step during auth would not progress to the next screen

## 🔬 To Test
1.
2.

## 📝 Checklist


calebpanza (approved):
So it looks like something with that request was timing just right that something was clearing from execution before running. I don't love the long term implications of this, but if it works, it works! I think this warrants a more robust handling of our Navigation Stack in order to push people to the right places at 1 level higher than what we're doing now

PR #2987, opened 2023-02-20
fix(routing): add back button for no content

dsilva609:
## ✏️ Description
- adds a back button to content that doesn't have valid media
- fixes routing issues for screens without media

## 🔬 To Test
1. Go to a content item that has no valid video media
2. You should see a button that will take you back to the previous screen

## 📸 Screenshots

image

## 📝 Checklist


calebpanza (changes requested):
While the primary button might be a bit overkill, can we update this to use it for the time being just for consistency. I'll make sure we get some designer eyes on this to make it less of an eye sore.

vinnyjth:
This PR is definitely a great use for usability, but should we go a step further and prevent navigation to items that don't have media as well?

calebpanza:
@vinnyjth I think that's a larger refactor that we absolutely should consider during next cycle when we're doing cleanup

calebpanza:
There's a product workflow that I don't want to entirely kill. Is there any chance that a church will want to have an item on TV available that doesn't actually have a video? I'm not sure, yet.

PR #2989, opened 2023-02-21
correctly fails on invalid version numbers

redreceipt:
## ✏️ Description

Another versioning PR... 🙄

## 🔬 To Test
1.
2.

## 📸 Screenshots

before and after

Screenshot 2023-02-21 at 10 49 37 AM

## 📝 Checklist


calebpanza:
Does it also need to on `android-and-androidtv`?

vinnyjth:
Love it - do you think we can move this logic to the `setup` steps? 

I would love to bomb out if a version number is incorrect quickly in the build process.

redreceipt:
That's an interesting thought. Not as it stands I don't think because currently iOS and android version numbers could be different. We wouldn't want iOS to bomb if it's just android that's screwed up. But there is probably a rearrangement we could make to catch certain problems early.

vinnyjth:
Gotcha - maybe we can save that for another PR then. On Bitrise we can only the iOS code on the iOS lane (and vice versa). 

I'll focus on moving the Bitrise config into Github.

vinnyjth (approved):

PR #2996, opened 2023-02-28
Segment: Make group calls more consistent

conrad-vanl:
## ✏️ Description

I'm seeing very few Group calls getting logged in segment
Also, not all destinations support Group

So, I'm trying to make the `.group` call triggered more often, as well as adding `church` to the `identify` call so that we have the church associated with each profile.

## 🔬 To Test
1. Boot app up, see that Group calls are getting logged to the terminal (analytics currently logs request to terminal)

## 📝 Checklist


calebpanza (approved):
Looks good to me! Def seeing an increase in the number of "group" calls coming through Segment

